PR #4, opened 2021-04-21
guide lines for contributing (wip)

christophmluscher:
This should be the start of some form of contributing guidelines

### DIFF
--- a/CONTRIBUTUNG.md
+++ b/CONTRIBUTUNG.md
@@ -2,6 +2,42 @@ Please always report problems by filling an issue here on GitHub. This also cove
 
 General rules:
 
-- If a class inherits directly/indirectly from the sisyphus class `Job` the class ends with "Job".
-- In a Sisyphus job class the output variables end in `out_*`
+    1. the code style suggestions of PyCharm should be followed
+
+Job general:
+
+    1. If a class inherits directly/indirectly from the sisyphus class `Job` the class ends with "Job".
+    2. In a Sisyphus job class the output variables end in `out_*`
+
+Job constructor order:
+
+    1. set inputs
+    2. set outputs (should be prefixed with `out_`)
+    3. define rqmt
+
+Job function order:
+
+    1. `__init__`
+    2. `tasks()`
+    3. task functions
+    4. helper functions
+    5. class functions
+    6. `__hash__`
+
+Rasr-Jobs:
+should always accept `crp`, `extra_config` + `extra_post_config`
+
+Import order:
+
+    1. standard library
+    2. external libraries (including sisyphus, yes I know it's not like that atm)
+    3. recipies
+
+Some general things:
+
+    1. Don't break a jobs hash
+    2. Really don't break a jobs hash
+    3. very strong preference to avoid config files / scripts in the recipe repo itself. Everything should be code
+    4. Jobs should be as deterministic as possible
+    5. `sh` function should definitely not be used
 
--- a/CONTRIBUTUNG.md
+++ b/CONTRIBUTUNG.md
@@ -8,6 +8,7 @@ Job general:
 
     1. If a class inherits directly/indirectly from the sisyphus class `Job` the class ends with "Job".
     2. In a Sisyphus job class the output variables end in `out_*`
+    3. If a Job has non-trivial requirements it should have a `self.rqmt`
 
 Job constructor order:
 
--- a/CONTRIBUTUNG.md
+++ b/CONTRIBUTUNG.md
@@ -1,0 +1,7 @@
+Please always report problems by filling an issue here on GitHub. This also covers reporting problems with the documentation (e.g. if sth is unclear).
+
+General rules:
+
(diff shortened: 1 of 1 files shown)


CONTRIBUTUNG.md
```diff
@@ -0,0 +1,44 @@
+Please always report problems by filling an issue here on GitHub. This also covers reporting problems with the documentation (e.g. if sth is unclear).
+
+General rules:
+
+    1. the code style suggestions of PyCharm should be followed
```
curufinwe, line 5: I would not formulate this in terms of PyCharm. I am not using PyCharm and I dont want to start. @albertz do you know if we can have the same code suggestions as a cmd-line tool?

JackTemaki: Can we not just add a check for "black" as a github PR hook?

CONTRIBUTUNG.md
```diff
@@ -0,0 +1,44 @@
...
+    5. class functions
+    6. `__hash__`
+
+Rasr-Jobs:
+should always accept `crp`, `extra_config` + `extra_post_config`
+
+Import order:
+
+    1. standard library
+    2. external libraries (including sisyphus, yes I know it's not like that atm)
```
JackTemaki, line 34: So Sisyphus not as first import anymore? I got quite used to it, but of course this is debatable.

CONTRIBUTUNG.md
```diff
@@ -0,0 +1,44 @@
...
+    3. recipies
+
+Some general things:
+
+    1. Don't break a jobs hash
+    2. Really don't break a jobs hash
+    3. very strong preference to avoid config files / scripts in the recipe repo itself. Everything should be code
+    4. Jobs should be as deterministic as possible
```
JackTemaki, line 42: I would formulate this as "jobs have to be deterministic, if this is not possible at all add a clear warning"

CONTRIBUTUNG.md
albertz: Black is just a formater, and by far does not cover all what PEP8 implies. (E.g. I'm sure it will not rename your variables.)

We actually use PyCharm as command line tool for automatic testing of code style in RETURNN, as part of CI, via [this script](https://github.com/rwth-i6/returnn/blob/master/tests/pycharm-inspect.py).

As already mentioned, part of that is `pycodestyle`, which can be used as a separate tool. (PyCharm by default ignores `E121,E123,E126,E226,E24,E704,W503,W504` which it probably checks in better ways - I haven't really looked into that.)

PyCharm still checks some more things, which are all reasonable. I don't think there are any other tools which can do the same. But we can of course check those also my human reviews. (All of the things are very reasonable anyway.)

There is also pylint. I tried to set it up once, i.e. to come up with reasonable defaults, but I gave up at some point because it misses many important things which are all covered by PyCharm, and also had some false positives, and otherwise some strange defaults.

CONTRIBUTUNG.md
albertz: Almost no jobs are 100% deterministic in practice. I'm sure when you run many things involving `float` at some point on different hardware, you will get different results.

And obviously, all our trainings are very much non-deterministic.

albertz: What I want to say is, this is a case which will very often happen (every training job), and we should also be prepared and be able to deal with this.

curufinwe: Float operations are also deterministic most of the time, or at least deterministic enough to not cause issues. Only in NN training we do so many of them to lead to different results.

CONTRIBUTUNG.md
curufinwe: I thought it's easier to treat sisyphus as a normal third-party library. It also reduces the number of import blocks from 4 to 3.

CONTRIBUTUNG.md
curufinwe: > Can we not just add a check for "black" as a github PR hook?

Right, black should always be part of the pipeline. The other PEP8 things are added on top of that.

albertz:
I remember that we got different results for gmm splitting depending on
amd64 vs x86 although we used float32 in both cases.


Eugen Beck ***@***.***> schrieb am Mi., 28. Apr. 2021, 13:53:

CONTRIBUTUNG.md
JackTemaki: I still would choose a stricter formulation. Of course hardware non-determinism can not be fixed in any way, but at least algorithmic non-determinism (random operations need a seed, no order of lines should change, number of output etc...) should not occur in any case.

albertz: I remember that we got different results for GMM splitting depending on amd64 vs x86 although we used float32 in both cases.

CONTRIBUTUNG.md
albertz: I also forgot MyPy, which can do type checks (but also just like PyCharm).

JackTemaki:
Can we just merge this now? I think we can continue the discussion if needed, but it would be good to atleast have a starting file.

PR #6, opened 2021-04-22
refactoring of returnn jobs

JackTemaki:
- refactored search and training
- added create_executable utility
- removed non-generic parameters or moved them to rasr_training

Note: this is not complete yet, and only serves as starting point for discussion

### DIFF
--- a/returnn/rasr_training.py
+++ b/returnn/rasr_training.py
@@ -5,22 +5,20 @@ from sisyphus import *
 Path = setup_path(__package__)
 
 import copy
-import json
-import os
-import pprint
-import stat
-import string
-import subprocess as sp
-import textwrap
 
 import recipe.i6_core.rasr as rasr
 import recipe.i6_core.mm as mm
 import recipe.i6_core.util as util
 
 from .training import ReturnnTrainingJob
+from .config import ReturnnConfig
 
 
 class ReturnnRasrTrainingJob(ReturnnTrainingJob):
+    """
+    Train a RETURNN model using rnn.py that uses ExternSpringDataset, and needs
+    to write RASR config and flow files.
+    """
     def __init__(
         self,
         train_crp,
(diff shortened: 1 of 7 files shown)


mmz33:
Removing `train_data` and `dev_data` from `ReturnnTrainingJob` is a good idea. I think these datasets should be added from the sis config. I am not sure about removing `num_classes`. I personally don't use this and also not sure if this should also be handled in the job. I guess also not.

albertz:
Without really looking into details, some random comments:

Should log files be part of the defined output? There is some potential information which we could take out from it, which is not available otherwise. And other jobs might even then depend on it. Some simple examples:
 * Speed information (training and also recognition, also RTF)
 * GPU memory usage
This can be relevant when we want to collect extensive statistics about training/recognition speed of various models vs their WER performance, and do that all automatically via a Sisyphus pipeline.

We sometimes use functions with `__` prefix (e.g. `__parse_python`). I think this is reserved for Python internal functions (`__init__`, `__str__`, etc), and should not be used by user code. I think PEP8 also states this. Also, in any case this is inconsistent.

(Maybe these comments should not get into this PR but into separate issues?)

JackTemaki:
> Should log files be part of the defined output? 

We can definitely consider adding this (Also adding outputs for existing files is always a safe change). The problem is that we then need might want to include "log_verbosity" in the hash, as this will have an effect on the usability of the log. Maybe we add a parameter "output_log", which does not need to be hashed itself, but then creates the output variable and adds log_verbosity to the hash.

Yes '__' should be avoided, as this is reserved for name mangling, and not for setting something as private.

returnn/training.py
```diff
@@ -419,8 +419,8 @@ def __init__(
         :param dict parameter_dict: provide external parameters to the rnn.py call
         :param int|str time_rqmt:
         :param int|str mem_rqmt:
-        :param tk.Path|str returnn_python_exe: the executable for running returnn
-        :param tk.Path |str returnn_root: the path to the returnn source folder
+        :param Path|str returnn_python_exe: file path to the executable for running returnn (python binary or .sh)
+        :param path|str returnn_root: file path to the RETURNN repository root folder
```
curufinwe, line 423: capital p?

returnn/rasr_training.py
```diff
@@ -100,9 +122,18 @@ def __init__(
         ) = ReturnnRasrTrainingJob.create_config(**kwargs)
 
         if self.alignment is not None:
-            self.class_labels = self.output_path("class.labels")
+            self.out_class_labels = self.output_path("class.labels")
 
     def create_files(self):
+
```
curufinwe, line 128: no newline, or is black doing that?

curufinwe (approved):
Looks good to me, only 2 very minor things. Fix + merge afterwards.

curufinwe:
Agree on the __ issue. Regarding log-file output: right now the Returnn output is an unstructured text-file with not even an implicit format. Thus I'd prefer not to add it to the jobs output. If we need specific runtime information from Returnn, it should be written into dedicated files with a clear (ideally documented) format.

albertz:
> Regarding log-file output: right now the Returnn output is an unstructured text-file with not even an implicit format. Thus I'd prefer not to add it to the jobs output. If we need specific runtime information from Returnn, it should be written into dedicated files with a clear (ideally documented) format.

The thing is, sometimes you only recognize afterwards that you want to use some information somehow (e.g. let's say the train loss output per mini-batch), and it would be too expensive to redo all your trainings. I bet many people extracted some information from log files. Usually we do that just manually.

Sure, it would be cleaner to store such things in some dedicated format. But this will rarely be used. And you might miss things.

But I also see your point. Not sure on it.

curufinwe:
OK, let's cross that bridge when someone has a need to do this, i.e. if one setup requires this information we output it in returnn into a separate file and add an output to the returnn job.

albertz:
> OK, let's cross that bridge when someone has a need to do this, i.e. if one setup requires this information we output it in returnn into a separate file and add an output to the returnn job.

Btw, regarding structured output, I was researching a bit into structured logging, which would solve this to some extent. I collected some stuff [here](https://github.com/albertz/wiki/blob/master/logging.md). Does anyone have experience with any of that?

curufinwe:
I looked at structlog and I like the idea of having a key for which you can output dictionaries. We then create two outputs, one for the console in string format and another binary one where we write pickled dicts (with some meta-data for the size of each entry).

curufinwe:
@JackTemaki just FYI: you can add a commit to the PR and then have github squash all commits in the PR into one for merging.

PR #9, opened 2021-04-28
rename filterbank keys

christophmluscher:
fixes erroneous keys and makes the keys more consistent with the other feature extraction jobs


curufinwe:
Why?

curufinwe (approved):

PR #12, opened 2021-04-29
update and add bliss<->txt conversion jobs

JackTemaki:
- segment_file and gzip option for CorpusToTxtJob
 - added matching CorpusReplaceOrthFromTxtJob

### DIFF
--- a/corpus/convert.py
+++ b/corpus/convert.py
@@ -1,13 +1,21 @@
-__all__ = ["CorpusToStmJob", "CorpusToTxtJob"]
+__all__ = ["CorpusToStmJob", "CorpusToTxtJob", "CorpusReplaceOrthFromTxtJob"]
 
-from recipe.i6_core.lib import corpus
+import gzip
+import itertools
 
 from sisyphus import *
+
+from recipe.i6_core.lib import corpus
+from recipe.i6_core.util import uopen
 
 Path = setup_path(__package__)
 
 
 class CorpusToStmJob(Job):
+    """
+    Convert a Bliss corpus into a .stm file
+    """
+
     def __init__(
         self,
         corpus_path,
@@ -17,6 +25,15 @@ class CorpusToStmJob(Job):
         name="",
         tag_mapping=(),
     ):
+        """
+
+        :param Path corpus_path: Bliss corpus
+        :param bool exclude_non_speech:
+        :param bool remove_punctuation:
+        :param bool fix_whitespace:
+        :param str name:
+        :param tuple[str, dict[str, str]] tag_mapping:
+        """
         self.set_vis_name("Extract STM from Corpus")
 
         self.corpus_path = corpus_path
@@ -25,7 +42,8 @@ class CorpusToStmJob(Job):
         self.fix_whitespace = fix_whitespace
         self.tag_mapping = tag_mapping
         self.name = name
-        self.stm_path = self.output_path("%scorpus.stm" % name)
+
+        self.out_stm_path = self.output_path("%scorpus.stm" % name)
 
     def tasks(self):
         yield Task("run", mini_task=True)
@@ -54,7 +72,7 @@ class CorpusToStmJob(Job):
                     if segment.rstrip() in tag_map:
                         tag_map[segment.rstrip()][i] = tag[0]
 
-        with open(self.stm_path.get_path(), "wt") as out:
+        with open(self.out_stm_path.get_path(), "wt") as out:
             for segment in c.segments():
                 speaker_name = (
                     segment.speaker().name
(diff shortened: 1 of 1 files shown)


corpus/convert.py
```diff
@@ -79,11 +81,22 @@ def run(self):
...
+        :param Path corpus_path: Bliss corpus
+        :param Path segment_file: segment file
+        :param bool gzip: gzip the output text file
+        """
         self.set_vis_name("Extract TXT from Corpus")
 
         self.corpus_path = corpus_path
-        self.txt_path = self.output_path("corpus.txt")
+        self.gzip = gzip
+        self.segment_file = segment_file
```
curufinwe, line 112: newline between member vars and outputs

corpus/convert.py
```diff
@@ -93,6 +106,59 @@ def run(self):
         c = corpus.Corpus()
         c.load(corpus_path)
 
-        with open(self.txt_path.get_path(), "wt") as out:
-            for segment in c.segments():
+        if self.segment_file:
+            with open(tk.uncached_path(self.segment_file), "rt") as f:
+                lines = [l for l in f.readlines() if len(l.strip()) > 0]
```
curufinwe, line 125: `lines = set(l for l in ...)` directly create the set from the iterator, no need for an intermediate list
also: why not strip the line that is added to the set?

corpus/convert.py
```diff
@@ -79,11 +81,22 @@ def run(self):
...
+        :param Path segment_file: segment file
+        :param bool gzip: gzip the output text file
+        """
         self.set_vis_name("Extract TXT from Corpus")
 
         self.corpus_path = corpus_path
-        self.txt_path = self.output_path("corpus.txt")
+        self.gzip = gzip
+        self.segment_file = segment_file
+        self.out_txt = self.output_path("corpus.%s" % (".txt.gz" if gzip else ".txt"))
```
curufinwe, line 113: creates two dots in the filename, better:
`self.out_txt = self.output_path("corpus.txt" + ".gz" if gzip else ""))`

corpus/convert.py
```diff
@@ -93,6 +106,59 @@ def run(self):
...
+        self.out_corpus = self.output_path("corpus.xml.gz")
+
+    def tasks(self):
+        yield Task('run', mini_task=True)
+
+    def run(self):
+        c = corpus.Corpus()
+        c.load(tk.uncached_path(self.corpus_path))
+
+        if tk.uncached_path(self.text_file).endswith(".gz"):
```
curufinwe, line 165: Use `uopen` https://github.com/rwth-i6/i6_core/blob/main/util.py#L144
Also: use `with` statement

corpus/convert.py
```diff
@@ -93,6 +106,59 @@ def run(self):
...
+            text = gzip.open(tk.uncached_path(self.text_file), "r")
+        else:
+            text = open(tk.uncached_path(self.text_file), "rt")
+
+        for recording in c.all_recordings():
```
curufinwe, line 170: You can directly iterate over all segments and then do something like:
```suggestion
        for segment, line in itertools.zip_longest(c.segments(), f)
           assert segment is not None
           assert line is not None
```

corpus/convert.py
```diff
@@ -93,6 +106,59 @@ def run(self):
...
+
+        out.close()
+
+
+class CorpusReplaceOrthFromTxtJob(Job):
+    """
+    Merge raw text back into a bliss corpus
+    """
+
+    def __init__(self, corpus, text_file):
```
curufinwe, line 147: This Job should work in tandem with the CorpusToTxtJob, i.e. support a list of segments to be passed into it as it relies on the line-number to match transcriptions to segments.

curufinwe (changes requested):

corpus/convert.py
```diff
@@ -93,6 +106,59 @@ def run(self):
...
+            out = open(self.out_txt.get_path(), "wt")
+
+        for segment in c.segments():
+            if not segments_whitelist or segment.fullname in segments_whitelist:
                 out.write(segment.orth + "\n")
+
+        out.close()
+
+
+class CorpusReplaceOrthFromTxtJob(Job):
```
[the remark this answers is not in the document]
JackTemaki: already is?

corpus/convert.py
JackTemaki: I just copied the job from an old setup, but yes I could add that functionality.

corpus/convert.py
JackTemaki: I wasn't aware of that util function, so yes that would be much nicer.

corpus/convert.py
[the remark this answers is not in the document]
curufinwe: sorry, looks like I only looked at one of the commits in the PR.

corpus/convert.py
curufinwe: :+1:

JackTemaki:
The problem is I do not have a pipeline to test this now, I am not sure if we should merge it just like this.

corpus/convert.py
```diff
@@ -93,6 +124,61 @@ def run(self):
...
+    def tasks(self):
+        yield Task("run", mini_task=True)
+
+    def run(self):
+        c = corpus.Corpus()
+        c.load(tk.uncached_path(self.corpus_path))
+
+        if self.segment_file:
+            with open(tk.uncached_path(self.segment_file), "rt") as f:
+                segments_whitelist = set([l.strip() for l in f.readlines() if len(l.strip()) > 0])
```
curufinwe, line 172: you are still creating an intermediate list, you can create the set from an iterator, no need for the list

corpus/convert.py
```diff
@@ -93,6 +124,61 @@ def run(self):
...
 
-        with open(self.txt_path.get_path(), "wt") as out:
-            for segment in c.segments():
+        if self.segment_file:
+            with open(tk.uncached_path(self.segment_file), "rt") as f:
+                segments_whitelist = set([l.strip() for l in f.readlines() if len(l.strip()) > 0])
+        else:
+            segments_whitelist = None
+
+        if self.gzip:
```
curufinwe, line 133: Actually you can also use uopen here, ".gz" is appended to path.

curufinwe (changes requested):
Please don't force push, just add a new commit. For the final merge I will squash them into a single commit. If you force push I don't see the diff w.r.t. the previous commit.

curufinwe (approved):

PR #14, opened 2021-04-29
Add VariableString class

JackTemaki:
- helper to embed tk.Path and tk.Variable
  objects into formatted strings
- allows for correct (non-absolute) hashing

This is a draft again, because I want to first get everyones opinion on this approach, and also on the name of the class.

@mmz33 reported a problem when he wanted to add output paths from jobs to the ExternSprintDataset config parameters that are part of a RETURNNConfig dict. If you do e.g. `config['xyz'] = "--my-parameter %s" % tk.uncached_path(some_path)` the hashing will be based on the absolute path, and as such will change when you move the setup.

The syntax now would be: `config['xyz'] = VariableString("--my-parameter %s", [some_path])`

### DIFF
--- a/util.py
+++ b/util.py
@@ -7,6 +7,7 @@ import xml.dom.minidom
 import xml.etree.ElementTree as ET
 
 from sisyphus import *
+from sisyphus.tools import sis_hash_helper
 
 Path = setup_path(__package__)
 Variable = tk.Variable
@@ -54,6 +55,40 @@ class MultiOutputPath(MultiPath):
             cached,
             gs.BASE_DIR,
         )
+
+
+class VariableString:
+    """
+    Enables to define strings containing Paths or Variables for e.g. ReturnnConfig or PipelineJob:
+    config[...] = VariableString("--some-parameter %s --another-parameter %i",
+                                 [sisyphus_path, sisyphus_variable])
+    """
+
+    def __init__(self, string, variables, allow_caching=False):
+        """
+
+        :param str string: any string with formatter symbols
+        :param tuple|list variables: Sisyphus Paths or Variables, or other object with ".get()" method
+        :param bool allow_caching: allows caching for tk.Path variables
+        """
+        self._string = string
+        self._variables = variables
+        self.allow_caching = allow_caching
+
+        assert isinstance(string, str)
+        assert isinstance(variables, list) or isinstance(variables, tuple)
+
+    def _sis_hash(self):
+        return sis_hash_helper((self._string, self._variables))
+
+    def get(self):
+        # use str for Path to get caching
+        return self._string % [
+            v.get_cached_path()
+            if isinstance(v, tk.Path) and self.allow_caching
+            else v.get()
+            for v in self._variables
+        ]
 
 
 def write_paths_to_file(file, paths):
@@ -252,4 +287,3 @@ def create_executable(filename, command):
         | stat.S_IXGRP
         | stat.S_IXOTH,
     )
-
--- a/returnn/config.py
+++ b/returnn/config.py
@@ -13,6 +13,8 @@ import pprint
 import string
 import textwrap
 
+from recipe.i6_core.util import VariableString
+
 
 def instanciate_vars(o):
     """
@@ -22,7 +24,7 @@ def instanciate_vars(o):
     :param Any o: nested structure that may contain Variable objects
     :return:
     """
-    if isinstance(o, Variable):
+    if isinstance(o, Variable) or isinstance(o, VariableString):
         o = o.get()
     elif isinstance(o, list):
         for k in range(len(o)):
--- a/text/processing.py
+++ b/text/processing.py
@@ -1,7 +1,10 @@
 __all__ = ["PipelineJob", "ConcatenateJob", "HeadJob", "TailJob"]
 
 import os
+
 from sisyphus import Job, Task, Path, global_settings as gs
+
+from recipe.i6_core.util import VariableString
 
 
 class PipelineJob(Job):
@@ -9,7 +12,7 @@ class PipelineJob(Job):
     Reads a text file and applies a list of piped shell commands
 
     :param Path text: text file (raw or gz) to be processed
-    :param list[str] pipeline: list of shell commands to form the pipeline
+    :param list[str|VariableString] pipeline: list of shell commands to form the pipeline
     :param bool zip_out: apply gzip to the output
     :param bool check_equal_length: the line count of the input and output should match
     :param bool mini_task: the pipeline should be run as mini_task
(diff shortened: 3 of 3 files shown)


util.py
```diff
@@ -56,6 +57,40 @@ def __init__(self, creator, path_template, hidden_paths, cached=False):
         )
 
 
+class VariableString:
+    """
+    Enables to define strings containing Paths or Variables for e.g. ReturnnConfig or PipelineJob:
+    config[...] = VariableString("--some-parameter %s --another-parameter %i",
+                                 [sisyphus_path, sisyphus_variable])
```
albertz, line 64: Why would this ever be needed?
You always should use sth like `["--some-parameter", sisyphus_path, "--another-parameter", sisyphus_variable]` instead, and that would not have this problem, right?

JackTemaki: This is not about a subprocess calls, but for any scenario where you want/need to pass a string to a job, but a part of that string depends on a sisyphus output. This should behave just like a tk.Variable/tk.Path.

albertz: But that's what I mean, or asked about. I cannot imagine any case where this should be needed.
Or rather: If there is any such case, I think this is bad, and should be cleaned up anyway.

mmz33:
I usually build SprintDataset using a helper class:
```
class SprintDataset(GenericDataset):
   
   .....
  
  def get_sprint_args(self):
    files = dict()
    files['config'] = tk.uncached_path(self.rasr_train_config)
    files['corpus'] = tk.uncached_path(self.corpus_xml)
    if self.segments_file:
      files['segments'] = tk.uncached_path(self.segments_file)
    files['features'] = self.features

    args = [
      "--config=" + files["config"],
      "--*.corpus.file=`cf {}`".format(files["corpus"]),
      "--*.corpus.segments.file=" + ("`cf {}`".format(files["segments"]) if 'segments' in files else ''),
      "--*.corpus.segment-order-shuffle=true",
      "--*.segment-order-sort-by-time-length=true",
      "--*.segment-order-sort-by-time-length-chunk-size=%i" % self.seg_order_sort_chunk_size,
      "--*.feature-cache-path=`cf {}`".format(files['features']),
      "--*.log-channel.file=" + self.log_channel,
      "--*.window-size=1",
    ]
    return args

  def as_returnn_opts(self):
    d = {
      'class': 'ExternSprintDataset',
      'sprintTrainerExecPath': self.trainer_exec,
      'sprintConfigStr': self.get_sprint_args(),
      'partitionEpoch': self.epoch_split
    }
    if self.segments_file:
      with open(tk.uncached_path(self.segments_file), 'r') as f:
        d['estimated_num_seqs'] = len(f.readlines()) // self.epoch_split
    # SprintDataset does not accept targets opts with defined class as OggZipDataset, thus they need to be removed
    for k in self.targets_data_opts:
      if 'class' in self.targets_data_opts[k]:
        del self.targets_data_opts[k]['class']
    d.update(self.targets_data_opts)
    d.update(self.other_opts)
    return d
```
       
The issue is that when building the args, I have to convert the `Path` object into `str`. But then the hash depends on the absolute `str` path rather than the `Path` object. This conversion to `str` should be only done inside a job and not here. This lead to issues later when I wanted to use another absolute path for the `Path` object but overwrite the hash with the old path. This won't work anymore.

albertz:
>  I have to convert the Path object into str using tk.uncached_path

I don't think you have to. That's what I mean. We can simply fix this. Sth like:
```
args = [
...
"--*.feature-cache-path=", files['features'],
"--use-cache-manager=true",
...
]
```

JackTemaki:
I am not really happy with the thought to eventually have to fix external software (in this case RETURNN or even RASR) just because there is a limitation in Sisyphus, but in this case this should work. (I am not sure how picky RASR is about the command line parameters)

If @mmz33 confirms this works, I will keep the code in my private repo if ever needed again.

albertz:
> I am not really happy with the thought to eventually have to fix external software (in this case RETURNN or even RASR) just because there is a limitation in Sisyphus, but in this case this should work.

Almost all command line tools use `--key`, `value` separately anyway. This is really the standard way since a long time.

All our tools can easily be fixed to support this in a similar way (and should be fixed anyway).

If there are really remaining rare cases, we can also easily write a wrapper script around it.

curufinwe:
I can see why such a wrapper could be handy sometimes, but my suggestion is to not do special support for it in CRNNConfig, just create an object that overloads __repr__ and implement the string conversion logic there.
Also: Please do not keep a private repo with different versions of core jobs. If you really really need a custom version of a job, put it in i6_private / i6_experiments and use that class.

JackTemaki:
I did not want to keep jobs private, just the VariableString code. Ok I will fix it to not use .get() but define `__str__` and `__repr__`, just like for tk.Path.

albertz:
> I can see why such a wrapper could be handy sometimes

But do we actually have use cases? (Cases where this is really the best solution, and not what I suggested, to rather rewrite this logic.)

Do we want to have code here for hypothetical (probably very rare) cases which might be useful at some point, but we do not have any use cases currently?

JackTemaki:
@patrick-wilken  pointed out a use case, the "laplace:<n>" notation for the dataset.

But I think this class is wrong here, with the change as Eugen suggested this is general enough to belong into the Sisyphus source code.

curufinwe:
Yes, these cases exist. E.g. in RASR configs you sometimes need to pass a prefix for the file-format (like bin:, xml:). The RASR config works with key val pairs, the problem is that the value here is a prefix + a path object.
Actually there is already such a wrapper: https://github.com/rwth-i6/i6_core/blob/main/rasr/config.py#L284
but it only works with paths, not variables (as the string value needs to be set at construction).

@JackTemaki maybe you could generalize it to use a string template and deal correctly with variables and put it in util while removing the string-wrapper class.

curufinwe:
> @patrick-wilken pointed out a use case, the "laplace:n" notation for the dataset.
> 
> But I think this class is wrong here, with the change as Eugen suggested this is general enough to belong into the Sisyphus source code.

sisyphus source code is also ok with me

albertz:
> pointed out a use case, the "laplace:n" notation for the dataset.

I'm not exactly sure I understand. But probably you should anyway use "laplace:.n" instead?

> E.g. in RASR configs you sometimes need to pass a prefix for the file-format (like bin:, xml:)

And this is also a case where I would argue this is bad, and we should fix this in a better way. (Like passing the file-format separately.)

mmz33:
> > I have to convert the Path object into str using tk.uncached_path
> 
> I don't think you have to. That's what I mean. We can simply fix this. Sth like:
> 
> ```
> args = [
> ...
> "--*.feature-cache-path=", files['features'],
> "--use-cache-manager=true",
> ...
> ]
> ```

This worked (without `=`). For caching, i guess this can be handled in sisyphus with Path(..., cached=true), and defining the cache function to return `cf ...`

PR #16, opened 2021-04-30
Librispeech Jobs

JackTemaki:
I wrote suggestions for how the download and bliss creator jobs may look like based on the code we already had.

There are some inconsistencies regarding the format with respect to the speakers, this needs to be discussed in the next meeting.

So please refrain from starting large discussions here for now, this is just for reference.

Comments on the code are welcome, but please keep in mind that this is mostly copy-pasted and needs to be updated based on the discussion results anyway.

A simple test case would look like this:
```python
  download_dev_clean = DownloadLibriSpeechCorpusJob("dev-clean")
  tk.register_output("dev-clean", download_dev_clean.out_corpus_folder)

  download_metadata = DownloadLibriSpeechMetadataJob()
  tk.register_output("librispeech_speakers.txt", download_metadata.out_speakers)

  create_dev_clean_bliss = LibriSpeechCreateBlissCorpusJob(
    corpus_folder=download_dev_clean.out_corpus_folder,
    speaker_metadata=download_metadata.out_speakers
  )

  tk.register_output("dev-clean.xml.gz", create_dev_clean_bliss.out_corpus)
```

### DIFF
--- a/datasets/librispeech.py
+++ b/datasets/librispeech.py
@@ -1,10 +1,11 @@
 __all__ = [
     "DownloadLibriSpeechCorpusJob",
     "DownloadLibriSpeechMetadataJob",
-    "LibriSpeechCreateBlissJob",
+    "LibriSpeechCreateBlissCorpusJob",
 ]
 
 import os
+import shutil
 import subprocess
 
 from sisyphus import *
@@ -43,9 +44,7 @@ class DownloadLibriSpeechCorpusJob(Job):
             "train-other-500",
         ]
 
-        # this is just needed to have the extraction target available
-        self._out_parent_folder = self.output_path(".", directory=True)
-
+        self._out_archive_folder = self.output_path("LibriSpeech")
         self.out_corpus_folder = self.output_path("LibriSpeech/%s" % self.corpus_key)
 
     def tasks(self):
@@ -68,20 +67,20 @@ class DownloadLibriSpeechCorpusJob(Job):
                     md5_out.write(line)
                     break
 
-        checksum_result = subprocess.check_output(
-            ["md5sum", "-c", "md5sum-%s.txt" % self.corpus_key]
-        )  #  type: bytes
-        assert b"OK" in checksum_result
-
+        subprocess.check_call(
+            ["md5sum", "--status", "-c", "md5sum-%s.txt" % self.corpus_key]
+        )
         subprocess.check_call(
             [
                 "tar",
                 "-xf",
                 "%s.tar.gz" % self.corpus_key,
                 "-C",
-                self._out_parent_folder.get_path(),
+                ".",
             ]
         )
+        shutil.move("LibriSpeech", self._out_archive_folder.get_path())
+        os.unlink("%s.tar.gz" % self.corpus_key)
 
 
 class DownloadLibriSpeechMetadataJob(DownloadLibriSpeechCorpusJob):
@@ -95,15 +94,14 @@ class DownloadLibriSpeechMetadataJob(DownloadLibriSpeechCorpusJob):
     def __init__(self):
         self.corpus_key = "raw-metadata"
 
-        # this is just needed to have the extraction target available
-        self._out_parent_folder = self.output_path(".", directory=True)
+        self._out_archive_folder = self.output_path("LibriSpeech")
 
         self.out_speakers = self.output_path("LibriSpeech/SPEAKERS.TXT")
         self.out_chapters = self.output_path("LibriSpeech/CHAPTERS.TXT")
         self.out_books = self.output_path("LibriSpeech/BOOKS.TXT")
 
 
-class LibriSpeechCreateBlissJob(Job):
+class LibriSpeechCreateBlissCorpusJob(Job):
     """
     Creates a Bliss corpus from a LibriSpeech corpus folder using the speaker information in addition
 
@@ -112,8 +110,6 @@ class LibriSpeechCreateBlissJob(Job):
 
     def __init__(self, corpus_folder, speaker_metadata):
         """
-        Generate a bliss xml from  the LibriSpeech corpus.
-
         :param Path corpus_folder: Path to a LibriSpeech corpus folder
         :param Path speaker_metadata: Path to SPEAKER.TXT file from the MetdataJob (out_speakers)
         """
@@ -129,11 +125,11 @@ class LibriSpeechCreateBlissJob(Job):
         yield Task("run", mini_task=True)
 
     def run(self):
-        self.get_speakers()
-        self.get_transcript()
+        self._get_speakers()
+        self._get_transcripts()
 
         c = corpus.Corpus()
-        c.name = os.path.basename(self.corpus_folder)
+        c.name = os.path.basename(tk.uncached_path(self.corpus_folder))
 
         for speaker_id, speaker_info in sorted(self._speakers.items()):
             speaker = corpus.Speaker()
@@ -153,7 +149,7 @@ class LibriSpeechCreateBlissJob(Job):
             segment = corpus.Segment()
             segment.name = name
             segment.start = 0
-            segment.end = "inf"
+            segment.end = float("inf")
             segment.orth = transcript["orth"].strip()
 
             recording.segments.append(segment)
@@ -161,22 +157,23 @@ class LibriSpeechCreateBlissJob(Job):
 
         c.dump(self.out_corpus.get_path())
 
-    def get_speakers(self):
+    def _get_speakers(self):
         """
         Extract the speakers from the SPEAKERS.TXT file
         """
         with open(tk.uncached_path(self.speaker_metadata), "r") as speakersfile:
             for line in speakersfile:
-                if line[0] != ";":
-                    procline = list(map(str.strip, line.split("|")))
-                    self._speakers[int(procline[0])] = [
-                        procline[1],
-                        procline[2],
-                        float(procline[3]),
-                        procline[4],
-                    ]
+                if line[0] == ";":
+                    continue
+                procline = list(map(str.strip, line.split("|")))
+                self._speakers[int(procline[0])] = [
+                    procline[1],
+                    procline[2],
+                    float(procline[3]),
+                    procline[4],
+                ]
 
-    def get_transcript(self):
+    def _get_transcripts(self):
         """
         Traverse the folder structure and search for the *.trans.txt files and read the content
         """
@@ -184,17 +181,18 @@ class LibriSpeechCreateBlissJob(Job):
             tk.uncached_path(self.corpus_folder), followlinks=True
         ):
             for file in files:
-                if file.endswith(".trans.txt"):
-                    with open(os.path.join(dirpath, file), "r") as transcription:
-                        for line in transcription:
-                            line_t = list(map(str.strip, line.split(" ", 1)))
-                            orth = line_t[1]
-                            procline = line_t[0].split("-")
-                            transcript = {
-                                "speaker_id": int(procline[0]),
-                                "chapter": int(procline[1]),
-                                "segment": int(procline[2]),
-                                "orth": orth,
-                                "path": dirpath,
-                            }
-                            self._transcripts.append(transcript)
+                if not file.endswith(".trans.txt"):
+                    continue
+                with open(os.path.join(dirpath, file), "r") as transcription:
+                    for line in transcription:
+                        line_t = list(map(str.strip, line.split(" ", 1)))
+                        orth = line_t[1]
+                        procline = line_t[0].split("-")
+                        transcript = {
+                            "speaker_id": int(procline[0]),
+                            "chapter": int(procline[1]),
+                            "segment": int(procline[2]),
+                            "orth": orth,
+                            "path": dirpath,
+                        }
+                        self._transcripts.append(transcript)
--- a/datasets/librispeech.py
+++ b/datasets/librispeech.py
@@ -1,0 +1,180 @@
+__all__ = ["DownloadLibriSpeechCorpusJob", "DownloadLibriSpeechMetadataJob"]
+
+import os
+import subprocess
+
+from sisyphus import *
+
+from recipe.i6_core.lib import corpus
+
+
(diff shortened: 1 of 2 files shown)


datasets/librispeech.py
```diff
@@ -0,0 +1,180 @@
...
+            segment = corpus.Segment()
+            segment.name = name
+            segment.start = 0
+            segment.end = "inf"
+            segment.orth = transcript['orth'].strip()
+
+            recording.segments.append(segment)
+            c.recordings.append(recording)
+
+        c.dump(tk.uncached_path(self.out_corpus))
```
christophmluscher, line 146: is there any difference between tk.uncached_path(self.out_corpus) and self.out_corpus.get_path()? do we have a convention for this? since out_ vars schould be Path objects and not str

datasets/librispeech.py
```diff
@@ -0,0 +1,180 @@
...
+    To get the corpus metadata, use
+    DownloadLibriSpeechMetadataJob
+
+    self.out_corpus_folder links to the root of the speaker_id/chapter/*
+    folder structure
+    """
+
+    def __init__(self, corpus_key):
+        """
+
```
christophmluscher, line 28: remove white space?

datasets/librispeech.py
```diff
@@ -0,0 +1,180 @@
...
+    def tasks(self):
+        yield Task('run', mini_task=True)
+
+    def run(self):
+        subprocess.check_call(["wget", "https://www.openslr.org/resources/12/md5sum.txt"])
+        subprocess.check_call(["wget", "https://www.openslr.org/resources/12/%s.tar.gz" % self.corpus_key])
+
+        with open("md5sum.txt", "rt") as md5_in:
+            with open("md5sum-%s.txt" % self.corpus_key, "wt") as md5_out:
```
christophmluscher, line 56: suggestion:
with open() as md5_in, open() as md5_out:

datasets/librispeech.py
JackTemaki: I think so far not, but then we should decide on one.

christophmluscher: @curufinwe your take?

curufinwe: For the out_ paths I'd go for `.get_path()` since we know it's a path obj and `.get_path` is a bit shorter.

christophmluscher (commented):
Naming convention for jobs which create bliss files, lexicon, etc?

<Corpus>Create<Bliss,Lexicon>Job

LibrispeechCreateBlissJob

what do you think?

JackTemaki:
I have no real opinion on this...

datasets/librispeech.py
```diff
@@ -0,0 +1,200 @@
...
+        with open("md5sum.txt", "rt") as md5_in, open(
+            "md5sum-%s.txt" % self.corpus_key, "wt"
+        ) as md5_out:
+            for line in md5_in:
+                split = line.strip().split(" ")
+                if split[-1].split(".")[0] == self.corpus_key:
+                    md5_out.write(line)
+                    break
+
+        checksum_result = subprocess.check_output(
```
curufinwe, line 71: call with --status and just check for return code

datasets/librispeech.py
```diff
@@ -0,0 +1,200 @@
...
+
+        subprocess.check_call(
+            [
+                "tar",
+                "-xf",
+                "%s.tar.gz" % self.corpus_key,
+                "-C",
+                self._out_parent_folder.get_path(),
+            ]
+        )
```
curufinwe, line 84: delete tar file at the end

datasets/librispeech.py
```diff
@@ -0,0 +1,200 @@
...
+            "dev-other",
+            "test-clean",
+            "test-other",
+            "train-clean-100",
+            "train-clean-360",
+            "train-other-500",
+        ]
+
+        # this is just needed to have the extraction target available
+        self._out_parent_folder = self.output_path(".", directory=True)
```
curufinwe, line 47: ugly, extract to work and move to output

datasets/librispeech.py
```diff
@@ -0,0 +1,200 @@
...
+        )  #  type: bytes
+        assert b"OK" in checksum_result
+
+        subprocess.check_call(
+            [
+                "tar",
+                "-xf",
+                "%s.tar.gz" % self.corpus_key,
+                "-C",
+                self._out_parent_folder.get_path(),
```
curufinwe, line 82: see above

datasets/librispeech.py
```diff
@@ -0,0 +1,200 @@
...
+
+    Defines outputs for "SPEAKERS.TXT, CHAPTERS.TXT and BOOKS.TXT"
+    """
+
+    # noinspection PyMissingConstructor
+    def __init__(self):
+        self.corpus_key = "raw-metadata"
+
+        # this is just needed to have the extraction target available
+        self._out_parent_folder = self.output_path(".", directory=True)
```
curufinwe, line 99: see above

datasets/librispeech.py
```diff
@@ -0,0 +1,200 @@
...
+
+        self.out_speakers = self.output_path("LibriSpeech/SPEAKERS.TXT")
+        self.out_chapters = self.output_path("LibriSpeech/CHAPTERS.TXT")
+        self.out_books = self.output_path("LibriSpeech/BOOKS.TXT")
+
+
+class LibriSpeechCreateBlissJob(Job):
```
curufinwe, line 106: *BlissCorpus

Bliss is also the format name of the lexicon

datasets/librispeech.py
```diff
@@ -0,0 +1,200 @@
...
+    """
+    Creates a Bliss corpus from a LibriSpeech corpus folder using the speaker information in addition
+
+    Outputs a single bliss .xml.gz file
```
curufinwe, line 110: see above

datasets/librispeech.py
```diff
@@ -0,0 +1,200 @@
...
+    """
+
+    def __init__(self, corpus_folder, speaker_metadata):
+        """
+        Generate a bliss xml from  the LibriSpeech corpus.
```
curufinwe, line 115: see above

datasets/librispeech.py
```diff
@@ -0,0 +1,200 @@
...
+                if line[0] != ";":
+                    procline = list(map(str.strip, line.split("|")))
+                    self._speakers[int(procline[0])] = [
+                        procline[1],
+                        procline[2],
+                        float(procline[3]),
+                        procline[4],
+                    ]
+
+    def get_transcript(self):
```
curufinwe, line 179: 1. get_transcript**s**
2. see above

datasets/librispeech.py
```diff
@@ -0,0 +1,200 @@
...
+            segment.start = 0
+            segment.end = "inf"
+            segment.orth = transcript["orth"].strip()
+
+            recording.segments.append(segment)
+            c.recordings.append(recording)
+
+        c.dump(self.out_corpus.get_path())
+
+    def get_speakers(self):
```
curufinwe, line 164: `_get_speakers` should be marked private as it modifies state and is not intended to be called from outside the class

datasets/librispeech.py
```diff
@@ -0,0 +1,200 @@
...
+
+        :param Path corpus_folder: Path to a LibriSpeech corpus folder
+        :param Path speaker_metadata: Path to SPEAKER.TXT file from the MetdataJob (out_speakers)
+        """
+        self.corpus_folder = corpus_folder
+        self.speaker_metadata = speaker_metadata
+
+        self.out_corpus = self.output_path("corpus.xml.gz")
+
+        self._speakers = {}  # dict(key: id, value: [sex, subset, min, name]
```
curufinwe, line 125: Why are speakers lists and transcripts dicts? It would be more consistent to use the same data-structure here.

datasets/librispeech.py
```diff
@@ -0,0 +1,200 @@
...
+
+        c.dump(self.out_corpus.get_path())
+
+    def get_speakers(self):
+        """
+        Extract the speakers from the SPEAKERS.TXT file
+        """
+        with open(tk.uncached_path(self.speaker_metadata), "r") as speakersfile:
+            for line in speakersfile:
+                if line[0] != ";":
```
curufinwe, line 170: you can save 1 level of indentation by checking for equality and then continuing

datasets/librispeech.py
```diff
@@ -0,0 +1,200 @@
...
+
+    def get_transcript(self):
+        """
+        Traverse the folder structure and search for the *.trans.txt files and read the content
+        """
+        for dirpath, dirs, files in os.walk(
+            tk.uncached_path(self.corpus_folder), followlinks=True
+        ):
+            for file in files:
+                if file.endswith(".trans.txt"):
```
curufinwe, line 187: see above

datasets/librispeech.py
```diff
@@ -0,0 +1,200 @@
...
+
+    def tasks(self):
+        yield Task("run", mini_task=True)
+
+    def run(self):
+        self.get_speakers()
+        self.get_transcript()
+
+        c = corpus.Corpus()
+        c.name = os.path.basename(self.corpus_folder)
```
curufinwe, line 136: tk.uncached_path?

datasets/librispeech.py
```diff
@@ -0,0 +1,200 @@
...
+            name = "{0}-{1}-{2:04d}".format(
+                transcript["speaker_id"], transcript["chapter"], transcript["segment"]
+            )
+            recording = corpus.Recording()
+            recording.name = name
+            recording.speaker_name = transcript["speaker_id"]
+            recording.audio = "{}/{}.flac".format(transcript["path"], name)
+
+            segment = corpus.Segment()
+            segment.name = name
```
curufinwe, line 154: replication necessary here? Why not leave the name empty (RASR will default to "1")

datasets/librispeech.py
```diff
@@ -0,0 +1,200 @@
...
+            segment.start = 0
+            segment.end = "inf"
```
curufinwe, line 156: better: `float("inf")`

lib/corpus.py
```diff
@@ -317,9 +321,11 @@ def dump(self, out, indentation=""):
         has_child_element = self.orth is not None or self.speaker_name is not None
         t = ' track="%d"' % self.track if self.track is not None else ""
         new_line = "\n" if has_child_element else ""
+        assert isinstance(self.end, float) or self.end == "inf"  # end can also be "inf"
```
curufinwe, line 324: if you set end to `float("inf")` you dont need special handling here

curufinwe (changes requested):

datasets/librispeech.py
JackTemaki: This is just how it was before. I am not sure if we want to change this, maybe we want to use some existing files that already use that labeling...

curufinwe: OK, better keep it that way then.

datasets/librispeech.py
JackTemaki: I did not write the original code, but I think the reason is that the speakers have specific key and need to be accessed with that, while the transcript "list" has no specific key attribute and just needs to be iterated. I do not really see a sense in changing that.

datasets/librispeech.py
```diff
@@ -43,9 +44,7 @@ def __init__(self, corpus_key):
             "train-other-500",
         ]
 
-        # this is just needed to have the extraction target available
-        self._out_parent_folder = self.output_path(".", directory=True)
-
+        self._out_archive_folder = self.output_path("LibriSpeech")
```
curufinwe, line 47: What I meant was not have this "hacky" output at all. See below for move operation.

datasets/librispeech.py
```diff
@@ -68,20 +67,20 @@ def run(self):
...
             [
                 "tar",
                 "-xf",
                 "%s.tar.gz" % self.corpus_key,
                 "-C",
-                self._out_parent_folder.get_path(),
+                ".",
             ]
         )
+        shutil.move("LibriSpeech", self._out_archive_folder.get_path())
```
curufinwe, line 82:
```suggestion
        shutil.move(os.path.join("LibriSpeech", self.corpus_key), self.out_corpus_folder.get_path())
```

datasets/librispeech.py
```diff
@@ -95,15 +94,14 @@ class DownloadLibriSpeechMetadataJob(DownloadLibriSpeechCorpusJob):
     def __init__(self):
         self.corpus_key = "raw-metadata"
 
-        # this is just needed to have the extraction target available
-        self._out_parent_folder = self.output_path(".", directory=True)
+        self._out_archive_folder = self.output_path("LibriSpeech")
```
curufinwe, line 97: see above

datasets/librispeech.py
```diff
@@ -153,48 +149,50 @@ def run(self):
...
-                    procline = list(map(str.strip, line.split("|")))
-                    self._speakers[int(procline[0])] = [
-                        procline[1],
-                        procline[2],
-                        float(procline[3]),
-                        procline[4],
-                    ]
-
-    def get_transcript(self):
+                if line[0] == ";":
```
curufinwe, line 166: indentation too large

curufinwe (changes requested):

datasets/librispeech.py
JackTemaki: This will not work, because there is also content in the "LibriSpeech" folder itself.

datasets/librispeech.py
JackTemaki: I do not see any error here...

datasets/librispeech.py
curufinwe: I'd suggest to update the DownloadMetadataJob and make it more specific.

datasets/librispeech.py
curufinwe: You're right. misread git diff.

curufinwe (commented):
Looks good, fix black issues and we can merge.

JackTemaki:
I just added a pre-commit hook that checks for black, so that it is impossible to miss that in the future.

PR #17, opened 2021-05-03
add black github action

JackTemaki:
I added a github action to check for the code formatting, I am not sure if we keep the default settings so I also added a `pyproject.toml` to configure black if needed. 

So open discussion:
- should we keep the weird default of 88 characters? or do we adjust this to the more common 120?
- should run black with a specific python version targeted? (If not, it will determine the version based on each file individually)

### DIFF
--- a/.github/workflows/black.yml
+++ b/.github/workflows/black.yml
@@ -1,0 +1,12 @@
+name: black
+on: [push, pull_request]
+jobs:
+  check-black-formatting:
+    runs-on: ubuntu-latest
+    steps:
+    - uses: actions/checkout@v2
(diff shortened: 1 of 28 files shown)


curufinwe:
Currently the code is formatted with the default 88 char limit and I dont see large problems.
I'd go for python 3.6 for now. Rational: https://github.com/rwth-i6/returnn/issues/487

JackTemaki:
Okay everything is ready now, I fixed the black version, applied it, and also added a requirements.txt so that the correct version can be easily installed.

curufinwe:
Messed up original commit, force pushed indended (rebased) version

PR #18, opened 2021-05-04
Bliss FFMPEG Job

JackTemaki:
- applies a generic ffmpeg filter string
  on each audio file
- can recover the duration of files if needed
- can be used as super class for more
  specific audio editing jobs
  (e.g. refactoring ChangeCorpusSpeed)

### DIFF
--- a/audio/ffmpeg.py
+++ b/audio/ffmpeg.py
@@ -114,15 +114,14 @@ class BlissFfmpegJob(Job):
         c = corpus.Corpus()
         c.load(tk.uncached_path(self.corpus_file))
 
-        # store index of last segment
-        for r in c.recordings:
-            audio_filename = self._get_output_filename(r)
-            r.audio = os.path.join(self.out_audio_folder.get_path(), audio_filename)
-
         from multiprocessing import pool
 
         p = pool.Pool(self.rqmt["cpu"])
         p.map(self._perform_ffmpeg, c.recordings)
+
+        for r in c.recordings:
+            audio_filename = self._get_output_filename(r)
+            r.audio = os.path.join(self.out_audio_folder.get_path(), audio_filename)
 
         if self.recover_duration:
             c.dump("temp_corpus.xml.gz")
--- a/audio/ffmpeg.py
+++ b/audio/ffmpeg.py
@@ -12,7 +12,7 @@ from recipe.i6_core.lib import corpus
 
 class BlissFfmpegJob(Job):
     """
-    Applies an FFMPEG audio filter to all recordigns of a bliss corpus.
+    Applies an FFMPEG audio filter to all recordings of a bliss corpus.
     This Job is extremely generic, as any valid audio option/filter string will work.
     Please consider using more specific jobs that use this Job as super class,
     see e.g. BlissChangeEncodingJob
--- a/audio/ffmpeg.py
+++ b/audio/ffmpeg.py
@@ -1,4 +1,4 @@
-__all__ = ["BlissFFMPEGJob"]
+__all__ = ["BlissFfmpegJob"]
 
 import copy
 import os
@@ -9,16 +9,16 @@ from sisyphus import *
 from recipe.i6_core.lib import corpus
 
 
-class BlissFFMPEGJob(Job):
+class BlissFfmpegJob(Job):
     """
-    Applies an FFMPEG audio filter to all recordigns of a bliss corpus.
-    This Job is extremely generic, as any valid audio option/filter string will work.
+    Applies an FFMPEG audio filter to all recordings of a bliss corpus.
+    This Job is only intended as a super-class for specific
+    Job implementations that need to make use of FFMPEG.
 
     WARNING:
         - This Job requires unique file names for the audio files
-        - Changing the duration of the audio files when you have multiple segments per audio,
-          as the segment information will be incorrect afterwards. Please use/write specific
-          jobs in that case (e.g. see corpus.data_augmentation.ChangeCorpusSpeed).
+        - Do not change the duration of the audio files when you have multiple segments per audio,
+          as the segment information will be incorrect afterwards.
 
     Typical applications:
 
@@ -63,35 +63,46 @@ class BlissFFMPEGJob(Job):
 
     The output stream that should be written into the audio is defined with `-map "[<output_stream>]"`
 
-    IMPORTANT! Do not forget to escape your quotation marks correctly for `-af` or `-filter_complex`
+    IMPORTANT! Do not forget to add and escape additional quotation marks correctly
+    for parameters to`-af` or `-filter_complex`
 
     """
 
     def __init__(
         self,
         corpus_file,
-        ffmpeg_option_string,
-        ffmpeg_binary=None,
+        ffmpeg_options=None,
         recover_duration=False,
         output_format=None,
+        ffmpeg_binary=None,
+        hash_binary=False,
     ):
         """
 
         :param Path corpus_file: bliss corpus
-        :param str|None ffmpeg_option_string: audio filter string, "-af" or "-filter_complex" with "-map" may be used
-        :param Path|str|None ffmpeg_binary: path to a ffmpeg binary
+        :param list(str)|None ffmpeg_options: list of additional ffmpeg parameters
         :param bool recover_duration: if the filter changes the duration of the audio, set to True
         :param str output_format: output file ending to determine container format (without dot)
+        :param Path|str|None ffmpeg_binary: path to a ffmpeg binary, uses system "ffmpeg" if None
+        :param bool hash_binary: In some cases it might be required to work with a specific ffmpeg version,
+                                 in which case the binary needs to be hashed
+
         """
+        if type(self) is BlissFfmpegJob:
+            raise Exception(
+                "Do not use the BlissFfmpegJob as is, "
+                "it is only intended to be a general class for specific applications"
+            )
+
         self.corpus_file = corpus_file
-        self.ffmpeg_option_string = ffmpeg_option_string
-        self.audio_folder = self.output_path("audio/", directory=True)
-        self.out = self.output_path("corpus.xml.gz")
-
-        self.ffmpeg_binary = ffmpeg_binary if ffmpeg_binary else "ffmpeg"
-
+        self.ffmpeg_options = ffmpeg_options
         self.recover_duration = recover_duration
         self.output_format = output_format
+        self.ffmpeg_binary = ffmpeg_binary if ffmpeg_binary else "ffmpeg"
+        self.hash_binary = hash_binary
+
+        self.out_audio_folder = self.output_path("audio/", directory=True)
+        self.out_corpus = self.output_path("corpus.xml.gz")
 
         self.rqmt = {"time": 4, "cpu": 4, "mem": 8}
 
@@ -128,7 +139,7 @@ class BlissFFMPEGJob(Job):
                 name, ext = os.path.splitext(audio_name)
                 audio_name = name + "." + self.output_format
 
-            nr.audio = os.path.join(tk.uncached_path(self.audio_folder), audio_name)
+            nr.audio = os.path.join(tk.uncached_path(self.out_audio_folder), audio_name)
             out_corpus.add_recording(nr)
 
         from multiprocessing import pool
@@ -139,7 +150,7 @@ class BlissFFMPEGJob(Job):
         if self.recover_duration:
             out_corpus.dump(tk.uncached_path("temp_corpus.xml.gz"))
         else:
-            out_corpus.dump(tk.uncached_path(self.out))
+            out_corpus.dump(tk.uncached_path(self.out_corpus))
 
     def run_recover_duration(self):
         """
@@ -163,7 +174,7 @@ class BlissFFMPEGJob(Job):
             )
             r.segments[0].end = new_duration
 
-        c.dump(tk.uncached_path(self.out))
+        c.dump(self.out_corpus.get_path())
 
     def _perform_ffmpeg(self, recording):
         """
@@ -178,7 +189,7 @@ class BlissFFMPEGJob(Job):
             name, ext = os.path.splitext(audio_name)
             audio_name = name + "." + self.output_format
 
-        target = tk.uncached_path(self.audio_folder) + "/" + audio_name
+        target = os.path.join(self.out_audio_folder.get_path(), audio_name)
         if not os.path.exists(target):
             print("try converting %s" % target)
             command_head = [
@@ -188,13 +199,18 @@ class BlissFFMPEGJob(Job):
                 "-i",
                 recording.audio,
             ]
-            command_tail = ["%s/%s" % (self.audio_folder, audio_name)]
-            if self.ffmpeg_option_string is None or len(self.ffmpeg_option_string) == 0:
+            command_tail = ["%s/%s" % (self.out_audio_folder, audio_name)]
+            if self.ffmpeg_options is None or len(self.ffmpeg_options) == 0:
                 command = command_head + command_tail
             else:
-                command = (
-                    command_head + self.ffmpeg_option_string.split(" ") + command_tail
-                )
+                command = command_head + self.ffmpeg_options + command_tail
             subprocess.check_call(command)
         else:
             print("skipped existing %s" % target)
+
+    @classmethod
+    def hash(cls, kwargs):
+        d = copy.copy(kwargs)
+        if not kwargs["hash_binary"]:
+            d.pop("ffmpeg_binary")
+        return super().hash(d)
--- a/audio/ffmpeg.py
+++ b/audio/ffmpeg.py
@@ -11,12 +11,13 @@ from recipe.i6_core.lib import corpus
 
 class BlissFfmpegJob(Job):
     """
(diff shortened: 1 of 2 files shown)


JackTemaki:
This is probably something that everybody should have a look at, because we would use this job to create the desired file format for all corpora as a starting point (so e.g. the conversion from .flac to .wav for LibriSpeech), but also use this job as a foundation for all other audio manipulation jobs, no matter if it is compression, data augmentation, amplitude normalization or any other desired audio operation.

audio/ffmpeg.py
```diff
@@ -0,0 +1,200 @@
...
+    The output stream that should be written into the audio is defined with `-map "[<output_stream>]"`
+
+    IMPORTANT! Do not forget to escape your quotation marks correctly for `-af` or `-filter_complex`
+
+    """
+
+    def __init__(
+        self,
+        corpus_file,
+        ffmpeg_option_string,
```
albertz, line 73: I don't think this should be a string but rather a list of arguments (i.e. list of strings).

This would be much cleaner. And also it could contain maybe references to other files (for adding noise, or whatever FFmpeg can do).

audio/ffmpeg.py
```diff
@@ -0,0 +1,200 @@
...
+        ffmpeg_binary=None,
```
albertz, line 74: Does it make sense that the ffmpeg binary is part of the hash? That implies that hashes between different systems are not really compatible (e.g. one system has `/usr/bin/ffmpeg` and the other has `/usr/local/bin/ffmpeg` or so). Also, it doesn't tell much (e.g. it doesn't tell what FFmpeg version this is).

One goal of the i6_core recipes is that we can also share data (datasets including preprocessed data such as features) across different systems, and this would break it.

audio/ffmpeg.py
```diff
@@ -0,0 +1,200 @@
...
+        c.dump(tk.uncached_path(self.out))
+
+    def _perform_ffmpeg(self, recording):
+        """
+        Build and call an FFMPEG command to apply on a recording
+
+        :param corpus.Recording recording:
+        :return:
+        """
+        audio_name = recording.audio.split("/")[-1]
```
albertz, line 175: I know some corpora which have a nested hierarchy of directories for the recordings (e.g. `A/1.mp3`, `B/1.mp3`, etc), so just picking the base name (`1.mp3`) would break this.

albertz:
More generic comments:

I see that it can be useful to preprocess the audio and then further work with it.
However, there are also cases where we might want to do it on-the-fly.
And there might be cases where we even might want to do it inside the network (i.e. in TensorFlow).

Ideally, it would be good to share the same code for all three cases. Because otherwise you never really know whether the implementation is really doing the same thing, and you cannot easily switch from one case to the other.

It is also easily possible to have the same code, by just having it all implemented in TensorFlow (or in RETURNN directly). We can then easily write a wrapper script around this to dump it for offline usage, or to have it as a dataset on-the-fly postprocessing function.

audio/ffmpeg.py
JackTemaki: This is why I already put a warning into the docstring

audio/ffmpeg.py
JackTemaki: I had to do it for myself, because different ffmpeg versions would produce different results for very specific filters and I needed to distinguish this in my pipeline. 
We could actually call "ffmpeg -version" in the hash function and then use this as hash. But then it might be difficult to install exactly the desired version on a system.

Another solution is to write specific jobs that inherit from this one, and allow only for limited parameters, which are then known to be stable across ffmpeg versions.

But no matter the approach, there will be difficulties/inconsistencies.

audio/ffmpeg.py
JackTemaki: This is why wrote this should be a super-class for more specific jobs, which then deal with the parameters more explicitly. Adding noise is a bad example, because this would not work with this job in any case.

audio/ffmpeg.py
albertz: Yes, but why not just fix it?

audio/ffmpeg.py
albertz: I don't understand. In any case, this should not be a string, but a list of arguments (list of strings).

JackTemaki:
Sharing the code for the three cases will be impossible. You can never guarantee that different libraries will have the same effect. This job is specifically for using FFMPEG, and will not be consistent with anything you do in python (well, I am sure you can somehow use the AV components but we did not do that anywhere yet besides for loading audio in RASR).

We can do what you suggested, but this is totally independent of this Job here.

For the actual pipeline, I suggest to write a sub-job that does only support changing the encoding with FFMPEG, as this will be the main use-case, and is based on this job.

audio/ffmpeg.py
albertz: I think in any case it should be made such that things can be shared (also across systems).

Also, as argued, just the path (`/usr/bin/ffmpeg`) doesn't really tell you anything. You update from Ubuntu 16 to Ubuntu 20, path stays the same, but you will get a different FFmpeg.

This problem of different behavior depending on version (and maybe also other things, such as Linux distribution, Linux kernel version, CPU type, whatever) is sth we anyway have. Maybe we should open a separate discussion (separate issue) on how to deal with this in general. I also don't have a good solution. Maybe you can just put some dummy string as argument, sth like `:param str comment:`, and you can write sth like "my custom FFmpeg which behaves different" to explicitly enforce a different hash.

In the usual case, we should assume that we get the same output though, even across different FFmpeg versions.

albertz:
> Sharing the code for the three cases will be impossible. You can never guarantee that different libraries will have the same effect.

Definitely not impossible. I already presented one possible easy solution.

> This job is specifically for using FFMPEG

Yes, this is why I wrote "more generic comments". I was questioning whether we should have this at all, and maybe not go for the solution I proposed. Or even if we take this for now, we probably should redo it later with my proposed solution (to make sure we get the same behavior and stay consistent).

audio/ffmpeg.py
albertz: To expand: One solution I commonly do in such cases: Find the common base path of all audio filenames. And then strip away this common base path.

JackTemaki:
> Definitely not impossible. 

Sure, if you also use a Tensorflow graph for offline processing or at least exactly the same library, yes this is possible.

>> chaning the encoding 
>
> Ah, if this is the only thing you actually want to do, I think I would just implement this, and not have this generic base job (for the arguments I presented before).

This is the main reason here, but I am also using this for silence filtering in my TTS-related setups. I would rather not re-implement this using completely different libraries (doing this on-the-fly is not an option anyway, for multiple reasons), but of course I can keep it outside the core repository.

albertz:
> > Definitely not impossible [to share same code for online, differentiable and offline processing].
> 
> Sure, if you also use a Tensorflow graph for offline processing or at least exactly the same library, yes this is possible.

Yes. What's the problem with this? Should be fairly straight-forward.

> ... but I am also using this for silence filtering in my TTS-related setups. I would rather not re-implement this using completely different libraries

That's what I'm saying. When people start using this base FFmpeg job now, it becomes more work in the future, to reimplement everything, once it is needed to do the same thing on-the-fly or differentiable.

> (doing this on-the-fly is not an option anyway, for multiple reasons)

Surely this is an option. I don't know your specific reasons now, and maybe you have some specific TTS related reasons. But in general, I think doing such things on-the-fly is very reasonable, and probably even to be preferred over the offline approach. For the same reason, it would also not really make sense to have SpecAugment being done offline. (Except maybe for an experiment to see how much you would loose by doing so.)

JackTemaki:
> Surely this is an option

Not a realistic one. I would need to implement the same "on-the-fly" processing in RASR, this will not happen. As multiple jobs are sharing this "processed" input, it would be wasted resources to do this on-the-fly every time. For data augmentation this is reasonable, but my silence preprocessing is not augmentation, but a fixed change of the corpus. But even for augmentation, nothing can easily be added to RASR. So as long as are using this (not with tensorflow training, there it is possible of course, but with GMM-based stuff), it needs to be offline.

Having the offline processing consistent with possible on the-fly processing is another issue, sure, but then you are limited to certain libraries (which does not have to be a problem, but might be one).

So in case of the FFMPEG "silenceremove" filter: how do I find a substitute that does the same and works in our on-the-fly scenario? There is librosa.effect.trim, but this works differently, and also does not remove in-between silence. So this means I would have to get into the libavfilter C API (pyffmpeg is discontinued, maybe there is another wrapper), and integrate this into the tensorflow Graph. I am sure I would spend hours on figuring out how to do that, which is not realistic compared to writing a single FFMPEG call that maybe takes 5 minutes in reading the documentation.

albertz:
> I would need to implement the same "on-the-fly" processing in RASR, this will not happen

No. You could just reuse the same code. That's my point. Write the code once in a way that it can be used for all cases.

> As multiple jobs are sharing this "processed" input, it would be wasted resources to do this on-the-fly every time.

Well, and on the other side, it is wasted disk space. Depends how much you augment. Maybe 1-2 copies, ok. Once you want to have 100 copies or so, it doesn't make sense anymore.

But anyway, this is not relevant. Sure there can be cases where you want to have it offline. I'm just arguing that it would be good to share the same code for all cases. To have it consistent, and to not require any need for reimplementation.

> Having the offline processing consistent with possible on the-fly processing is another issue, sure, but then you are limited to certain libraries (which does not have to be a problem, but might be one).

I don't really understand. Yes, it has to be in pure TensorFlow (either directly or within RETURNN). But I don't really see a problem with that. TF is made for such things.

> how do I find a substitute that does the same and works in our on-the-fly scenario?

You just implement it, or search for the code in Google?

> FFMPEG "silenceremove", librosa.effect.trim, libavfilter C API ... integrate this into the tensorflow Graph

Of course not. As I said, you would want to have shared code for all cases, including being differentiable. So it should just be simple pure TensorFlow code. You don't need any other library. TensorFlow can do all computations you need.

JackTemaki:
> You don't need any other library. TensorFlow can do all computations you need.

Okay now I see the problem in our discussion. From the perspective that one is actually able to (easily, with a realistic time scope) implement the desired functionally in Tensorflow, your arguments make sense, but:

> You just implement it, or search for the code in Google?

https://github.com/FFmpeg/FFmpeg/blob/master/libavfilter/af_silenceremove.c

I am not re-implenting that in Tensorflow (and RASR? At least here I could use the API directly) for a simple experiment. This would be a huge waste of time and effort, and I think it is quite unlikely that a threshold based sample removal function would be differentiable.

albertz:
> I am not re-implenting that in Tensorflow for a simple experiment.

Well, yes, maybe not for a simple experiment. But once you want to it more systematically, you need to.

Also, speaking about silence removal, I don't think this is so difficult?
E.g. something similar to [this](https://www.tensorflow.org/io/api_docs/python/tfio/experimental/audio/trim).

Anyway, this becomes off-topic here. My argument is just that it is all possible, and this is the solution which is ultimately better.

JackTemaki:
> E.g. something similar to this.

This is something different.

But anyway, I think the best solution is here to take away the possibility to run an arbitrary FFMPEG command as a job, and just add some general FFMPEG related functionality and write a specific encoding job based on this.

audio/ffmpeg.py
```diff
@@ -0,0 +1,200 @@
...
+import copy
+import os
+import subprocess
+
+from sisyphus import *
+
+from recipe.i6_core.lib import corpus
+
+
+class BlissFFMPEGJob(Job):
```
vieting, line 12: `BlissFfmpegJob` would be more consistent with `ReturnnTrainingJob` etc.

audio/ffmpeg.py
```diff
@@ -0,0 +1,200 @@
...
+    """
+    Applies an FFMPEG audio filter to all recordigns of a bliss corpus.
```
vieting, line 14:
```suggestion
    Applies an FFMPEG audio filter to all recordings of a bliss corpus.
```

audio/ffmpeg.py
```diff
@@ -0,0 +1,200 @@
...
+    This Job is extremely generic, as any valid audio option/filter string will work.
+
+    WARNING:
+        - This Job requires unique file names for the audio files
+        - Changing the duration of the audio files when you have multiple segments per audio,
+          as the segment information will be incorrect afterwards. Please use/write specific
+          jobs in that case (e.g. see corpus.data_augmentation.ChangeCorpusSpeed).
```
vieting, line 21: The first sentence doesn't make sense to me.

audio/ffmpeg.py
```diff
@@ -0,0 +1,200 @@
...
+        :param Path corpus_file: bliss corpus
+        :param str|None ffmpeg_option_string: audio filter string, "-af" or "-filter_complex" with "-map" may be used
+        :param Path|str|None ffmpeg_binary: path to a ffmpeg binary
+        :param bool recover_duration: if the filter changes the duration of the audio, set to True
+        :param str output_format: output file ending to determine container format (without dot)
+        """
+        self.corpus_file = corpus_file
+        self.ffmpeg_option_string = ffmpeg_option_string
+        self.audio_folder = self.output_path("audio/", directory=True)
+        self.out = self.output_path("corpus.xml.gz")
```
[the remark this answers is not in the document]
JackTemaki: Ah yes, this should be changed. I am not sure if `audio_folder` is needed as actual output, but I guess I can add this if ever needed.

audio/ffmpeg.py
```diff
@@ -0,0 +1,200 @@
...
+            command_tail = ["%s/%s" % (self.audio_folder, audio_name)]
+            if self.ffmpeg_option_string is None or len(self.ffmpeg_option_string) == 0:
+                command = command_head + command_tail
+            else:
+                command = (
+                    command_head + self.ffmpeg_option_string.split(" ") + command_tail
+                )
+            subprocess.check_call(command)
+        else:
+            print("skipped existing %s" % target)
```
curufinwe, line 200: `logging.info`

curufinwe (changes requested):

audio/ffmpeg.py
```diff
@@ -0,0 +1,200 @@
...
+        yield Task("run", rqmt=self.rqmt)
+        if self.recover_duration:
+            # recovering is not multi-threaded, so force cpu=1
+            recover_rqmt = copy.copy(self.rqmt)
+            recover_rqmt["cpu"] = 1
+            yield Task("run_recover_duration", rqmt=recover_rqmt)
+
+    def run(self):
+        in_corpus = corpus.Corpus()
+        out_corpus = corpus.Corpus()
```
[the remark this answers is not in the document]
JackTemaki: Oh yes, I think that would save a lot of lines.

audio/ffmpeg.py
```diff
@@ -0,0 +1,200 @@
...
+        """
+        audio_name = recording.audio.split("/")[-1]
+
+        if self.output_format is not None:
+            name, ext = os.path.splitext(audio_name)
+            audio_name = name + "." + self.output_format
+
+        target = tk.uncached_path(self.audio_folder) + "/" + audio_name
+        if not os.path.exists(target):
+            print("try converting %s" % target)
```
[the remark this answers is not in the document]
JackTemaki: We should add this to the guidelines

audio/ffmpeg.py
```diff
@@ -0,0 +1,200 @@
...
+
+        p = pool.Pool(self.rqmt["cpu"])
+        p.map(self._perform_ffmpeg, in_corpus.recordings)
+
+        if self.recover_duration:
+            out_corpus.dump(tk.uncached_path("temp_corpus.xml.gz"))
+        else:
+            out_corpus.dump(tk.uncached_path(self.out))
+
+    def run_recover_duration(self):
```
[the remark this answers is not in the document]
JackTemaki: I do not want to have a corpus file as output that contains invalid information. This is why I wanted this functionality to be part of the same job.

audio/ffmpeg.py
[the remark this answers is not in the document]
curufinwe: :ok_hand:

audio/ffmpeg.py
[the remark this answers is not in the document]
curufinwe: But then it should not be optional.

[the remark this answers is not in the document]
JackTemaki: I will make it true per default, and force this for the specific encoding job. But I would still leave this option to turn that off, to save computation time and I/O load when you are sure that nothing can go wrong.

audio/ffmpeg.py
```diff
@@ -0,0 +1,205 @@
...
+        self.ffmpeg_binary = ffmpeg_binary if ffmpeg_binary else "ffmpeg"
+        self.hash_binary = hash_binary
+
+        self.out_audio_folder = self.output_path("audio/", directory=True)
+        self.out_corpus = self.output_path("corpus.xml.gz")
+
+        self.rqmt = {"time": 4, "cpu": 4, "mem": 8}
+
+    def tasks(self):
+        yield Task("run", rqmt=self.rqmt)
```
curufinwe, line 106: The `run` task is resumable, but lacks the `resume="run"` parameter

curufinwe (changes requested):
One more minor thing.

curufinwe (approved):

PR #22, opened 2021-05-11
enable flat bliss corpus merge

JackTemaki:


corpus/transform.py
```diff
@@ -329,9 +332,14 @@ def run(self):
         for corpus_path in self.corpora:
             c = corpus.Corpus()
             c.load(str(corpus_path))
```
christophmluscher, line 334: why `str` and not `tk.uncached_path`?

corpus/transform.py
```diff
@@ -329,9 +332,14 @@ def run(self):
...
-
-        merged_corpus.dump(tk.uncached_path(self.merged_corpus))
+            if self.use_subcorpora:
+                merged_corpus.add_subcorpus(c)
+            else:
+                for rec in c.all_recordings():
+                    merged_corpus.add_recording(rec)
+                merged_corpus.speakers.update(c.speakers)
+
+        merged_corpus.dump(tk.uncached_path(self.out_merged_corpus))
```
christophmluscher, line 342: I think we decided on `self.out_merged_corpus.get_path()`

curufinwe (changes requested):

curufinwe (approved):

PR #23, opened 2021-05-12
add tone extraction func

christophmluscher:


meta/system.py
```diff
@@ -303,6 +303,26 @@ def plp_features(self, corpus, num_deriv=2, num_features=23, **kwargs):
...
+        self.feature_bundles[corpus]["tone"] = f.out_feature_bundle
+
+        feature_path = rasr.FlagDependentFlowAttribute(
+            "cache_mode",
+            {
+                "task_dependent": self.feature_caches[corpus]["tone"],
+                "bundle": self.feature_bundles[corpus]["tone"],
+            },
+        )
+        self.feature_flows[corpus]["tone"] = features.basic_cache_flow(feature_path)
```
JackTemaki, line 324: the `uncached_tone` features are missing

JackTemaki (changes requested):

meta/system.py
```diff
@@ -303,6 +303,26 @@ def plp_features(self, corpus, num_deriv=2, num_features=23, **kwargs):
             )
         self.feature_flows[corpus]["uncached_plp"] = f.feature_flow
 
+    def tone_features(self, corpus, prefix="", **kwargs):
+        timestamp_flow = self.feature_flows[corpus][
+            kwargs.pop("timestamp_flow", "energy")
```
curufinwe, line 308: Why is timestamp flow "hidden" in kwargs? Wouldn't it be better to have that as an explicit arg for `tone_features`?

curufinwe (changes requested):

meta/system.py
christophmluscher: since the ToneJob is not based on the FeatureExtractionJob, there is no `uncached_tone`
I don't think this can be added easily, since the tone job has multiple steps. That would take more refactoring and I dont want to do that now

curufinwe (approved):

PR #25, opened 2021-05-12
Returnn Search Job

mmz33:


returnn/search.py
```diff
@@ -14,6 +14,178 @@
...
+        device="gpu",
+        time_rqmt=4,
+        mem_rqmt=4,
+        cpu_rqmt=2,
+        returnn_python_exe=None,
+        returnn_root=None,
+    ):
+        """
+        :param dict[str] search_data: dataset used for search
+        :param dict[int, Checkpoint] model_checkpoints: Contains checkpoints pointing to TF models.
```
JackTemaki, line 39: It is not a good idea to pass a list of checkpoints, and then select only one by epoch, as the hash will change when you alter this list, even if the actually selected checkpoint is the same.

So we have to find variants for the inputs that are correct, but also work in any case. My suggestion would be to have the following two options:
 - model output folder + epoch
 - fixed `Checkpoint` object

returnn/search.py
```diff
@@ -14,6 +14,178 @@
...
+        self.rqmt = {
+            "gpu": 1 if device == "gpu" else 0,
+            "cpu": cpu_rqmt,
+            "mem": mem_rqmt,
+            "time": time_rqmt,
+        }
+
+    def tasks(self):
+        yield Task("create_files", mini_task=True)
+        yield Task("run", resume="run", rqmt=self.rqmt)
```
JackTemaki, line 84: Is search really resumable? I think this will crash....

returnn/search.py
```diff
@@ -14,6 +14,178 @@
...
+            else gs.RETURNN_PYTHON_EXE
+        )
+
+        self.returnn_root = (
+            returnn_root if returnn_root is not None else gs.RETURNN_ROOT
+        )
+
+        self.out_search_file = self.output_path("search_out")
+
+        self.returnn_config = self.create_returnn_config(
```
JackTemaki, line 69: This should not be in the middle of the output definitions I think...

returnn/search.py
```diff
@@ -14,6 +14,178 @@
...
+        yield Task("create_files", mini_task=True)
+        yield Task("run", resume="run", rqmt=self.rqmt)
+
+    def create_files(self):
+        config = self.returnn_config
+
+        config.config["load"] = self.model_checkpoint.ckpt_path
+        config.write(self.out_returnn_config_file.get_path())
+
+        with open("rnn.sh", "wt") as f:
```
JackTemaki, line 92: use `util.create_executable`

JackTemaki (changes requested):

returnn/search.py
mmz33: this won't work since `self.out_search_file` is used in `self.create_returnn_config`

returnn/search.py
mmz33: I thought this means that it will resubmit the job if it crashes or not?

returnn/search.py
mmz33: actually this should be of type `dict[int, List[Checkpoint]]`. Also, the checkpoints list is usually fixed and created in `ReturnnTrainingJob`. 
The problem here is that `epoch` can be a `tk.Variable` and so `.get()` should be only called from here. So I guess the first option would work. Actually this was what I had before. But at some point we agreed that it is better to use the `Checkpoint` class.

returnn/search.py
```diff
@@ -14,6 +14,178 @@
...
+                **self.search_data,
+            }
+        else:
+            config["search_data"] = self.search_data
+
+        post_config = {
+            "device": device,
+            "log": ["./returnn.log"],
+            "log_verbosity": log_verbosity,
+            "multiprocessing": True,
```
albertz, line 165:
```suggestion
```
(Not needed)

returnn/search.py
albertz: I don't quite understand. 

This should be a single `Checkpoint` object. Not a model output folder + epoch. And also not a dict of multiple checkpoints.

> The problem here is that epoch can be a tk.Variable

We should never do this. I think we can always replace such code. If there are cases where this is not easy/clear, maybe we should discuss separately, as this is sth more fundamental.

returnn/search.py
```diff
@@ -14,6 +14,178 @@
...
+        :param float|int mem_rqmt: job memory requirement in GB
+        :param float|int cpu_rqmt: job cpu requirement in GB
+        :param tk.Path|str|None returnn_python_exe: path to the RETURNN executable (python binary or launch script)
+        :param tk.Path|str|None returnn_root: path to the RETURNN src folder
+        """
+        self.search_data = search_data
+
+        self.model_checkpoint = model_checkpoints[
+            epoch.get() if isinstance(epoch, tk.Variable) else epoch
+        ]
```
albertz, line 55:
```suggestion
        self.model_checkpoint = model_checkpoint
```
(Via discussion above)

returnn/search.py
```diff
@@ -14,6 +14,178 @@
...
+        sp.check_call(call)
+
+    def create_returnn_config(self, returnn_config, log_verbosity, device):
+        """
+        Creates search RETURNN config
+        :param ReturnnConfig returnn_config: object representing RETURNN config
+        :param int log_verbosity: RETURNN log verbosity
+        :param str device: RETURNN device, cpu or gpu
+        :return: search ReturnnConfig
```
albertz, line 134:
```suggestion
        :rtype: ReturnnConfig
```
Btw, should we keep using the docstring to specify types?
Or rather use Python type annotations? (I don't find whether we discussed or decided on this.)

returnn/search.py
mmz33: Yes I would agree that this should be a single checkpoint. This is what I had before. But at some point, we wanted to select recognition epochs using a Job in Sisyphus config. This returns a `tk.Variable` representing the epoch (using `self.output_var`). That's why I do this now. The solution for this would be not to use `Checkpoint` but pass the model output folder + epoch. If we want to ignore this for now, then a single `Checkpoint` will work.

returnn/search.py
mmz33: yeah also not sure about this. i think it is not yet discussed.

returnn/search.py
albertz: > But at some point, we wanted to select recognition epochs using a Job in Sisyphus config. This returns a `tk.Variable` representing the epoch (using `self.output_var`).

I think this should be solved in some way as well (but independent from this PR). Maybe we should open a separate issue to discuss, if the solution is not obvious.

`ReturnnSearchJob` definitely should just accept a single `Checkpoint`.

Maybe we can also extend/modify `Checkpoint` itself, but that again should be a separate PR/discussion, not related to `ReturnnSearchJob`, which just uses `Checkpoint`.

JackTemaki: Having to pass "epoch" as `tk.Variable` is not necessary, as in this case there can (and should) be a separate job that "selects" a model. The only situation where you need to have this model_folder + epoch input is if you want to work with models that can be deleted during the training (e.g. intermediate checkpoints), and using the `update` functionality of a Job. This is a very specific case, and is usually not needed (You can tell RETURNN to keep some specific epochs I think, I am always fine with the pre-defined 80 and 160 so I never checked this).

mmz33: ok then i will just use single checkpoint

returnn/search.py
JackTemaki: But you can put it after it, right?

returnn/search.py
JackTemaki: Only if it is interrupted, e.g. out of time or memory. But then it will restart and crash, because the output file already exists. I am not sure if it is a good idea here to have automatic restarts with search jobs. At least with the RASR decoding jobs I was usually not happy if they restarted on their own and I did not notice, because usually this meant something in the configuration is wrong.

returnn/search.py
albertz: > Having to pass "epoch" as `tk.Variable` is not necessary, as in this case there can (and should) be a separate job that "selects" a model.

Yes, this is what I had in mind. Btw, the current `Checkpoint` class, will it actually have the same hash then? This `SelectEpochJob` (or however you would call it) likely would put a symlink or so, right? And `Checkpoint` would use the path of the symlink, which has a different hash than the checkpoint file itself? I don't know. Just asking. If this is a problem, as said, we should discuss in a separate issue, not here.

> The only situation where you need to have this model_folder + epoch input is if you want to work with models that can be deleted during the training (e.g. intermediate checkpoints), and using the `update` functionality of a Job.

I don't understand. Why can't you use `Checkpoint` for this case?

returnn/search.py
mmz33: ah ok i see.

returnn/search.py
mmz33: ah yeah to fix the order of output paths. ok

returnn/search.py
JackTemaki: > Why can't you use Checkpoint for this case?
It is probably even possible, I never looked at this again. But this is too specific to be considered here anyway.

> Btw, the current Checkpoint class, will it actually have the same hash then

No, this will not be the case. If e.g. epoch 247 would be determined to be the "best" epoch from a Job that does this, and you would start with epoch 247 "by hand", those would be two different recognition jobs. This is of course redundant, but this can not be resolved as this is not possible with the Graph concept defined by Sisyphus. It is not the "Content" (e.g. the physical checkpoint) that is hashed, but the ownership. So Sisyphus would create 2 decoding jobs, because "Training.<hash>/output/checkpoint.247" and "GetBestCheckpoint.<hash>/output/best_checkpoint" are different, even if "best_checkpoint" is just a symlink to checkpoint.247

albertz: >> Btw, the current Checkpoint class, will it actually have the same hash then
>
> No, this will not be the case.

We could fix this. E.g. Sisyphus could resolve symlinks (but only so far as they keep staying relative to the work dir) before it takes the hash. Or sth like that. They are surely many potential ways to fix this.

Or, maybe actually this is fine for the case of "best checkpoint", because when you rerun this part, and due to non-determinism, the best epoch changes, the "best checkpoint" hash will still be the same as before, which is maybe exactly what we want. Not sure.

But let's further discuss elsewhere, this is all off-topic for this PR here.

mmz33: ok i guess this can be resolved then.

returnn/search.py
```diff
@@ -14,6 +14,167 @@
...
+
+    def run(self):
+        call = [
+            tk.uncached_path(self.returnn_python_exe),
+            os.path.join(tk.uncached_path(self.returnn_root), "rnn.py"),
+            self.out_returnn_config_file.get_path(),
+        ]
+        sp.check_call(call)
+
+    def create_returnn_config(self, returnn_config, log_verbosity, device):
```
christophmluscher, line 119: I was just wondering if we want to merge the `create_returnn_config` functions from the training and search job? But probably this could be a different issue/pull request

returnn/search.py
```diff
@@ -14,6 +14,167 @@
...
+
+        cmd = [
+            tk.uncached_path(self.returnn_python_exe),
+            os.path.join(tk.uncached_path(self.returnn_root), "rnn.py"),
+            self.out_returnn_config_file.get_path(),
+        ]
+
+        util.create_executable("rnn.sh", cmd)
+
+        os.chmod(
```
christophmluscher, line 95: move this into `create_executable`?

christophmluscher: wait. is already in `create_executable`. so the `os.chmod` should be unnecessary

mmz33: oh yes.

returnn/search.py
mmz33: what do you mean by merge? they are quite different right?

JackTemaki: I think discussing unifying "common" code can wait until we have all Returnn jobs pushed (e.g. I will also submit a "forward") job). As long as the behavior of the jobs does not change, refactoring can be done at any time.

returnn/search.py
```diff
@@ -14,6 +14,156 @@
...
+        res = copy.deepcopy(returnn_config)
+        res.config = config
+        res.post_config = post_config
+
+        return res
+
+    @classmethod
+    def hash(cls, kwargs):
+        d = {
+            "returnn_config": kwargs["returnn_config"].hash(),
```
JackTemaki, line 158: We should fix the hashing behavior of ReturnnConfig completely so that calling this explicitly is not needed anymore, but this can be done in another PR as this affects other jobs as well.

JackTemaki (approved):
There are things to discuss, but this is more global and not directly related to this PR, so the rest seems fine to me.

christophmluscher (approved):

PR #26, opened 2021-05-12
Add rqmt dict for BlissToOggZip job

mmz33:
this can be used if the dataset is large.

### DIFF
--- a/returnn/oggzip.py
+++ b/returnn/oggzip.py
@@ -28,7 +28,6 @@ class BlissToOggZipJob(Job):
         no_conversion=False,
         returnn_python_exe=None,
         returnn_root=None,
-        rqmt=None,
     ):
         """
         use RETURNN to dump data into an ogg zip file
@@ -59,14 +58,15 @@ class BlissToOggZipJob(Job):
             returnn_root if returnn_root is not None else gs.RETURNN_ROOT
         )
 
-        self.rqmt = rqmt
-
         self.out_ogg_zip = self.output_path("out.ogg.zip")
+
+        self.rqmt = None
 
     def tasks(self):
         if self.rqmt:
             yield Task("run", rqmt=self.rqmt)
-        yield Task("run", mini_task=True)
+        else:
+            yield Task("run", mini_task=True)
 
     def run(self):
         args = [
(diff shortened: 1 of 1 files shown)


returnn/oggzip.py
```diff
@@ -40,6 +41,7 @@ def __init__(
         :param bool no_conversion: do not call the actual conversion, assume the audio files are already correct
         :param Path|str returnn_python_exe: file path to the executable for running returnn (python binary or .sh)
         :param Path|str returnn_root: file path to the RETURNN repository root folder
+        :param dict[str] rqmt: job's resources requirements dict
```
JackTemaki, line 44: type is `dict[str, Any]|None`, also note here that setting this will disable running the job as mini task

returnn/oggzip.py
```diff
@@ -57,9 +59,13 @@ def __init__(
             returnn_root if returnn_root is not None else gs.RETURNN_ROOT
         )
 
+        self.rqmt = rqmt
```
JackTemaki, line 62: The default style is to put self.rqmt after the output definitions at the end of the constructor.

JackTemaki (changes requested):

returnn/oggzip.py
```diff
@@ -28,6 +28,7 @@ def __init__(
         no_conversion=False,
         returnn_python_exe=None,
         returnn_root=None,
+        rqmt=None,
```
curufinwe, line 31: I'd not put the rqmt in the constructor. It's OK to have them set later. (This is the default for many jobs)

returnn/oggzip.py
```diff
@@ -57,9 +59,13 @@ def __init__(
...
         )
 
+        self.rqmt = rqmt
+
         self.out_ogg_zip = self.output_path("out.ogg.zip")
 
     def tasks(self):
+        if self.rqmt:
+            yield Task("run", rqmt=self.rqmt)
         yield Task("run", mini_task=True)
```
curufinwe, line 69: This should be in an else branch now.

curufinwe (changes requested):

JackTemaki (approved):

curufinwe (approved):

PR #28, opened 2021-05-19
Add CorpusToReturnnTextDictJob

JackTemaki:
This is a Job we use to convert the bliss test sets into the format to use the WER computation with the job from the RETURNN tools.

### DIFF
--- a/corpus/convert.py
+++ b/corpus/convert.py
@@ -212,16 +212,15 @@ class CorpusToTextDictJob(Job):
         segments = None
         if self.segments_file_path:
             with uopen(self.segments_file_path) as f:
-                segments = set(line.decode().strip() for line in f.readlines())
+                segments = set(line.decode().strip() for line in f)
 
         for segment in c.segments():
             orth = segment.orth.strip()
-            key = segments.fullname()
-            if segments:
-                if key not in segments:
-                    continue
+            key = segment.fullname()
+            if segments and key not in segments:
+                continue
             dictionary[key] = orth
 
         dictionary_string = pprint.pformat(dictionary, width=1000)
-        with open(self.out_dictionary.get_path(), "wt") as f:
+        with uopen(self.out_dictionary.get_path(), "wt") as f:
             f.write(dictionary_string)
--- a/corpus/convert.py
+++ b/corpus/convert.py
@@ -2,6 +2,7 @@ __all__ = ["CorpusToStmJob", "CorpusToTxtJob", "CorpusReplaceOrthFromTxtJob"]
 
 import gzip
 import itertools
+import pprint
 
 from sisyphus import *
 
@@ -182,3 +183,51 @@ class CorpusReplaceOrthFromTxtJob(Job):
                 segment.orth = line.strip()
 
         c.dump(tk.uncached_path(self.out_corpus))
+
+
+class CorpusToReturnnTextDictJob(Job):
+    """
+    Extract the Text from a Bliss corpus to fit the "{key: text}" structure of RETURNN
+    """
+
+    def __init__(self, corpus, segments=None):
+        """
+
+        :param corpus: bliss corpus file
+        :param segments: a segment file as optional whitelist
+        """
+        self.corpus_path = corpus
+        self.segments_file_path = segments
+
+        self.out_dictionary = self.output_path("text_dictionary.py")
+
+    def tasks(self):
+        yield Task("run", mini_task=True)
+
+    def run(self):
+        c = corpus.Corpus()
+        c.load(tk.uncached_path(self.corpus_path))
+
+        dictionary = {}
+
+        segments = None
+        if self.segments_file_path:
+            if tk.uncached_path(self.segments_file_path).endswith("gz"):
+                segment_file = gzip.open(
+                    tk.uncached_path(self.segments_file_path), "rb"
+                )
+            else:
+                segment_file = open(tk.uncached_path(self.segments_file_path), "rt")
+            segments = [line.decode().strip() for line in segment_file]
+
+        for segment in c.segments():
+            orth = segment.orth.strip()
+            key = segments.fullname()
+            if segments:
+                if key not in segments:
+                    continue
+            dictionary[key] = orth
+
+        dictionary_string = pprint.pformat(dictionary, width=1000)
+        with open(self.out_dictionary.get_path(), "wt") as f:
+            f.write(dictionary_string)
--- a/corpus/convert.py
+++ b/corpus/convert.py
@@ -192,8 +192,8 @@ class CorpusToTextDictJob(Job):
 
     def __init__(self, corpus, segments=None):
         """
-        :param Path: bliss corpus file
-        :param Path|None: a segment file as optional whitelist
+        :param Path corpus: bliss corpus file
+        :param Path|None segments: a segment file as optional whitelist
         """
         self.corpus_path = corpus
         self.segments_file_path = segments
@@ -211,7 +211,7 @@ class CorpusToTextDictJob(Job):
 
         segments = None
         if self.segments_file_path:
-            with uopen(tk.uncached_path(self.segments_file_path)) as f:
+            with uopen(self.segments_file_path) as f:
                 segments = set(line.decode().strip() for line in f.readlines())
 
         for segment in c.segments():
--- a/corpus/convert.py
+++ b/corpus/convert.py
@@ -185,16 +185,15 @@ class CorpusReplaceOrthFromTxtJob(Job):
         c.dump(tk.uncached_path(self.out_corpus))
 
 
-class CorpusToReturnnTextDictJob(Job):
+class CorpusToTextDictJob(Job):
     """
-    Extract the Text from a Bliss corpus to fit the "{key: text}" structure of RETURNN
+    Extract the Text from a Bliss corpus to fit a "{key: text}" structure (e.g. for RETURNN)
     """
 
     def __init__(self, corpus, segments=None):
         """
-
-        :param corpus: bliss corpus file
-        :param segments: a segment file as optional whitelist
+        :param Path: bliss corpus file
+        :param Path|None: a segment file as optional whitelist
         """
         self.corpus_path = corpus
         self.segments_file_path = segments
@@ -218,7 +217,7 @@ class CorpusToReturnnTextDictJob(Job):
                 )
             else:
                 segment_file = open(tk.uncached_path(self.segments_file_path), "rt")
-            segments = [line.decode().strip() for line in segment_file]
+            segments = set(line.decode().strip() for line in segment_file)
 
         for segment in c.segments():
             orth = segment.orth.strip()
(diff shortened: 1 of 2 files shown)


JackTemaki:
As a side note, maybe we want to remove the "Returnn" from the name, as this can also be used for other purposes...

corpus/convert.py
```diff
@@ -182,3 +183,51 @@ def run(self):
...
+
+class CorpusToReturnnTextDictJob(Job):
+    """
+    Extract the Text from a Bliss corpus to fit the "{key: text}" structure of RETURNN
+    """
+
+    def __init__(self, corpus, segments=None):
+        """
+
+        :param corpus: bliss corpus file
```
mmz33, line 196: type is `Path`

corpus/convert.py
```diff
@@ -182,3 +183,51 @@ def run(self):
...
+        self.segments_file_path = segments
+
+        self.out_dictionary = self.output_path("text_dictionary.py")
+
+    def tasks(self):
+        yield Task("run", mini_task=True)
+
+    def run(self):
+        c = corpus.Corpus()
+        c.load(tk.uncached_path(self.corpus_path))
```
mmz33, line 209: use `.get_path()`

corpus/convert.py
```diff
@@ -182,3 +183,51 @@ def run(self):
...
+class CorpusToReturnnTextDictJob(Job):
+    """
+    Extract the Text from a Bliss corpus to fit the "{key: text}" structure of RETURNN
+    """
+
+    def __init__(self, corpus, segments=None):
+        """
+
+        :param corpus: bliss corpus file
+        :param segments: a segment file as optional whitelist
```
mmz33, line 197: type is `Path`

corpus/convert.py
```diff
@@ -182,3 +183,51 @@ def run(self):
...
+
+    def run(self):
+        c = corpus.Corpus()
+        c.load(tk.uncached_path(self.corpus_path))
+
+        dictionary = {}
+
+        segments = None
+        if self.segments_file_path:
+            if tk.uncached_path(self.segments_file_path).endswith("gz"):
```
mmz33, line 215: use `.get_path()`

corpus/convert.py
```diff
@@ -182,3 +183,51 @@ def run(self):
...
+                segment_file = gzip.open(
+                    tk.uncached_path(self.segments_file_path), "rb"
```
mmz33, line 217: also here

corpus/convert.py
```diff
@@ -182,3 +183,51 @@ def run(self):
...
+                )
+            else:
+                segment_file = open(tk.uncached_path(self.segments_file_path), "rt")
```
mmz33, line 220: also here

corpus/convert.py
```diff
@@ -182,3 +183,51 @@ def run(self):
...
+            segments = [line.decode().strip() for line in segment_file]
```
mmz33, line 221: maybe this should be a set instead?

mmz33 (changes requested):

mmz33:
> As a side note, maybe we want to remove the "Returnn" from the name, as this can also be used for other purposes...

yeah i would agree with this

JackTemaki:
@mmz33 for inputs we usually still use `tk.uncached_path`, but we could also change that. I am not sure what the status is on this on the guidelines.

corpus/convert.py
```diff
@@ -182,3 +183,45 @@ def run(self):
...
+
+
+class CorpusToTextDictJob(Job):
+    """
+    Extract the Text from a Bliss corpus to fit a "{key: text}" structure (e.g. for RETURNN)
+    """
+
+    def __init__(self, corpus, segments=None):
+        """
+        :param Path: bliss corpus file
```
mmz33, line 195: parameter name missing

corpus/convert.py
```diff
@@ -182,3 +183,45 @@ def run(self):
...
+        :param Path|None: a segment file as optional whitelist
```
mmz33, line 196: parameter name missing

mmz33 (changes requested):

JackTemaki:
> > @mmz33 for inputs we usually still use `tk.uncached_path`, but we could also change that. I am not sure what the status is on this in the guidelines.
> 
> I think we should change that. this can lead to hash problems later. also JTP told me that he wanted to remove this feature anw at some point.

Inside job tasks this can never lead to hash problems, you are referring to using this outside. Nearly all current jobs use this, also `uopen` uses this internally, so this is completely fine.

mmz33:
> Inside job tasks this can never lead to hash problems, you are referring to using this outside. Nearly all current jobs use this, also `uopen` uses this internally, so this is completely fine.

well yes. but also now this means you allow inputs of type `str.` so you should update the parameters docs.

JackTemaki:
> > Inside job tasks this can never lead to hash problems, you are referring to using this outside. Nearly all current jobs use this, also `uopen` uses this internally, so this is completely fine.
> 
> well yes. but also now this means you allow inputs of type `str.` so you should update the parameters docs.

Just because it would work does not mean this is a good idea, so I would leave it this way.

corpus/convert.py
```diff
@@ -182,3 +183,45 @@ def run(self):
...
+    def run(self):
+        c = corpus.Corpus()
+        c.load(tk.uncached_path(self.corpus_path))
+
+        dictionary = {}
+
+        segments = None
+        if self.segments_file_path:
+            with uopen(self.segments_file_path) as f:
+                segments = set(line.decode().strip() for line in f.readlines())
```
christophmluscher, line 215: Isn't `for line in f` more memory efficient since it does not load the complete file at once? not 100% sure.. 
not sure if this is even really relevant.. more curiosity on my side :)

corpus/convert.py
```diff
@@ -182,3 +183,45 @@ def run(self):
...
+
+        for segment in c.segments():
+            orth = segment.orth.strip()
+            key = segments.fullname()
+            if segments:
+                if key not in segments:
```
christophmluscher, line 221: merge into one `if`?

corpus/convert.py
```diff
@@ -182,3 +183,45 @@ def run(self):
...
+
+class CorpusToTextDictJob(Job):
+    """
+    Extract the Text from a Bliss corpus to fit a "{key: text}" structure (e.g. for RETURNN)
+    """
+
+    def __init__(self, corpus, segments=None):
+        """
+        :param Path corpus: bliss corpus file
+        :param Path|None segments: a segment file as optional whitelist
```
christophmluscher, line 196: add maybe option for blacklist? like FilterCorpusBySegmentsJob (I think)

corpus/convert.py
```diff
@@ -182,3 +183,45 @@ def run(self):
...
+        for segment in c.segments():
+            orth = segment.orth.strip()
+            key = segments.fullname()
+            if segments:
+                if key not in segments:
+                    continue
+            dictionary[key] = orth
+
+        dictionary_string = pprint.pformat(dictionary, width=1000)
+        with open(self.out_dictionary.get_path(), "wt") as f:
```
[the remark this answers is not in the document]
mmz33: `uopen` uses `tk.uncached_path`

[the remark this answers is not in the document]
christophmluscher: hm I know what you mean.. out variables should use `get_path`
but I find `uopen` more readable..
not sure what the best policy here is..
out variables should use `get_path` since we know they are a sisyphus Path Object. But still use uopen when we directly use the file?

corpus/convert.py
JackTemaki: But then we would need to add it to all jobs that take segments for consistency, well of course we can do that...

JackTemaki: I created an issue for this, then we are reminded of this and can do it later on for all jobs.

christophmluscher: yes fine with me

corpus/convert.py
```diff
@@ -212,16 +212,15 @@ def run(self):
...
-                if key not in segments:
-                    continue
+            key = segment.fullname()
+            if segments and key not in segments:
+                continue
             dictionary[key] = orth
 
         dictionary_string = pprint.pformat(dictionary, width=1000)
-        with open(self.out_dictionary.get_path(), "wt") as f:
+        with uopen(self.out_dictionary.get_path(), "wt") as f:
```
christophmluscher, line 225: with uopen you would not need `get_path` right!? -> more readable

corpus/convert.py
```diff
@@ -182,3 +183,44 @@ def run(self):
...
+
+        segments = None
+        if self.segments_file_path:
+            with uopen(self.segments_file_path) as f:
+                segments = set(line.decode().strip() for line in f)
+
+        for segment in c.segments():
+            orth = segment.orth.strip()
+            key = segment.fullname()
+            if segments and key not in segments:
```
curufinwe, line 220: in RASR the whitelist not only refers to the full-name but can also just be the segment name. Better keep the support here the same. I.e. also test if `segment.name not in segments`

corpus/convert.py
curufinwe: agree

curufinwe (changes requested):

corpus/convert.py
```diff
@@ -2,6 +2,7 @@
```
christophmluscher, line 2: add CorpusToTextDictJob

christophmluscher (approved):

corpus/convert.py
```diff
@@ -36,7 +37,7 @@ def __init__(
         """
         self.set_vis_name("Extract STM from Corpus")
 
-        self.corpus_path = corpus_path
+        self.corpus_path = bliss_corpus
```
mmz33, line 40: also use `self.bliss_corpus` for all changes.

mmz33 (changes requested):

mmz33 (approved):

corpus/convert.py
```diff
@@ -27,7 +28,7 @@ def __init__(
     ):
         """
 
-        :param Path corpus_path: Bliss corpus
+        :param Path bliss_corpus: Bliss corpus
```
curufinwe, line 31: usually strings and Path objects are supported for file-paths.

corpus/convert.py
```diff
@@ -49,11 +50,10 @@ def tasks(self):
         yield Task("run", mini_task=True)
 
     def run(self):
-        corpus_path = tk.uncached_path(self.corpus_path)
         tag_map = {}
 
         c = corpus.Corpus()
-        c.load(corpus_path)
+        c.load(self.bliss_corpus.get_path())
```
curufinwe, line 56: see above

corpus/convert.py
```diff
@@ -181,4 +179,54 @@ def run(self):
...
+        if self.segments_file_path:
+            with uopen(self.segments_file_path) as f:
+                segments = set(line.decode().strip() for line in f)
+
+        for segment in c.segments():
+            orth = segment.orth.strip()
+            key = segment.fullname()
+            if segments:
+                if (
+                    not self.invert_match
```
curufinwe, line 221: if invert_match is true, no segment will be written to the corpus at all.

PR #29, opened 2021-05-20
add set stm files function

christophmluscher:
adds a function to create and set the stm files in the meta system class from the bliss corpus file

### DIFF
--- a/meta/system.py
+++ b/meta/system.py
@@ -166,6 +166,11 @@ class System:
         }
         self.scorer_hyp_arg[corpus] = "ctm"
 
+    def set_stm_files(self, corpus, **kwargs):
+        self.stm_files[corpus] = corpus_recipes.CorpusToStmJob(
+            self.crp[corpus].corpus_config.file, **kwargs
+        ).out_stm_path
+
     def store_allophones(self, source_corpus, target_corpus="base", **kwargs):
         self.jobs[target_corpus]["allophones"] = lexicon.StoreAllophonesJob(
             self.crp[source_corpus], **kwargs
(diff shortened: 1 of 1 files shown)


meta/system.py
```diff
@@ -166,6 +166,11 @@ def set_kaldi_scorer(self, corpus, mapping):
         }
         self.scorer_hyp_arg[corpus] = "ctm"
 
+    def set_stm_files(self, corpus, **kwargs):
```
JackTemaki, line 169: we should start to add docstring to all functions.

JackTemaki (changes requested):

meta/system.py
```diff
@@ -166,6 +166,15 @@ def set_kaldi_scorer(self, corpus, mapping):
...
+        """
+        create and set the stm files
+        :param Path|str corpus: Bliss corpus file path
```
JackTemaki, line 172: But this is wrong now or not?
This should be just the corpus "name" as string, right?

JackTemaki (changes requested):

meta/system.py
christophmluscher: ah yes you are right..

meta/system.py
```diff
@@ -166,6 +166,15 @@ def set_kaldi_scorer(self, corpus, mapping):
         }
         self.scorer_hyp_arg[corpus] = "ctm"
 
+    def set_stm_files(self, corpus, **kwargs):
```
curufinwe, line 169: From a set function I expect to have a parameter for the value that I want to set. A better name for what this function does is `create_stm_from_corpus`.

curufinwe (changes requested):

PR #31, opened 2021-05-25
Add rqmt dict for CompileTFGraph job

mmz33:
This is needed in case the local machine has old CPU arch. This can hang forever sometimes with the compiled TF python paths we use. I already encountered this on my machine. Also this happens on hiwis machines.

### DIFF
--- a/returnn/compile.py
+++ b/returnn/compile.py
@@ -60,8 +60,13 @@ class CompileTFGraphJob(Job):
         self.out_model_params = self.output_var("model_params.pickle", pickle=True)
         self.out_state_vars = self.output_var("state_vars.pickle", pickle=True)
 
+        self.rqmt = None
+
     def tasks(self):
-        yield Task("run", resume="run", mini_task=True)
+        if self.rqmt:
+            yield Task("run", resume="run", rqmt=self.rqmt)
+        else:
+            yield Task("run", resume="run", mini_task=True)
 
     def run(self):
         returnn_config_path = self.returnn_config


curufinwe (approved):

albertz:
Shouldn't we deal with this strange TF hang issues in a more generic way? This probably occurs for many other jobs as well.

mmz33:
> Shouldn't we deal with this strange TF hang issues in a more generic way? This probably occurs for many other jobs as well.

Hmm yeah. For example, I also thought about another job where this can also happen. For example, `ReturnnComputeWERJob` job. This also runs locally. i am not sure what is the best way to fix this. For this, I usually use another python path. actually for this job we can also do the same. but this is not a generic solution for sure. i am not sure what is the best solution here.

albertz:
> i am not sure what is the best way to fix this.

Maybe we should just fix the TF installation. I'm not sure it make sense to have all sort of workarounds for some strange cases.

We discussed already about the problem. This is very likely due to old GCC. For reference, here is the issue: https://github.com/tensorflow/tensorflow/issues/42035

JackTemaki:
> We discussed already about the problem. This is very likely due to old GCC.

As a comment, this only happens on machines "< Haswell", I tested it on 10-15 different machines, as I had to find some that work for my Hiwis.

curufinwe:
My guess is that it will go away once the chair upgrades to ubuntu 20.04 and we can recompile TF with a newer compiler. But until then this feature is an OK workaround.

albertz:
But we anyway also need a different workaround to make TF work on those machines (e.g. my desktop).

PR #33, opened 2021-05-28
refactor samples_flow

JackTemaki:


features/common.py
```diff
@@ -65,10 +80,10 @@ def samples_flow(
     if input_options is not None:
         input_opts.update(**input_options)
 
-    prescale = input_opts.pop("prescale", None)
+    input_node_type = "ffmpeg" if audio_format != "wav" else "wav"
```
curufinwe, line 83: no, at least sph is also supported. please check list of rasr audo nodes

curufinwe (changes requested):

christophmluscher (approved):
Do we want to catch the `scale_input` edge cases?
For example:
- `x 1`
- `* 0`
But this could be a different pull request

features/common.py
christophmluscher: https://www-i6.informatik.rwth-aachen.de/rwth-asr/manual/index.php/Audio_Nodes
- wav
- nist/sphere
- flac
- mpeg/mp3
- gsm
- htk
- phonedat
- oss

curufinwe:
I don't think it's necessary to do this. Maybe the `* 0` case should emit a warning, but not more.

features/common.py
JackTemaki: Okay. The question here is how much we want to trust the ffmpeg node in terms of performance and speed, as ffmpeg can open everything but `htk` and `phonedat` (at least I did not find those in the format list).

In the beginning I added format="ffmpeg" to the audio corpus instead of writing `.ogg`, but I did not like this...

So we can discuss for which of those formats we want to use the explicit nodes, because I was also using the ffmpeg node for e.g. flac one time (but this is quite slow in that case, because decoding flac is expensive).

curufinwe: The ffmpeg code is also optional (or at least to me it is more optional than the other libraries), I think it is better to use the "native" audio nodes if they are able to parse the format.

JackTemaki: Okay!

curufinwe (approved):

PR #34, opened 2021-05-31
Add CompileNativeOpJob

curufinwe:


returnn/compile.py
```diff
@@ -105,3 +107,64 @@ def hash(cls, kwargs):
...
+        self.returnn_python_exe = (
+            returnn_python_exe
+            if returnn_python_exe is not None
+            else gs.RETURNN_PYTHON_EXE
+        )
+        self.returnn_root = (
+            returnn_root if returnn_root is not None else gs.RETURNN_ROOT
+        )
+
+        self.out_op = self.output_path("op.so")
```
JackTemaki, line 138: I think here it might be better if the file is actually named like the op, so using `native_op` instead of `"op.so"`

returnn/compile.py
```diff
@@ -105,3 +107,64 @@ def hash(cls, kwargs):
...
+                tk.uncached_path(self.returnn_root), "tools/compile_native_op.py"
+            ),
+            "--native_op",
+            self.native_op,
+            "--output_file",
+            "compile.out",
+        ]
+        logging.info(args)
+
+        sp.run(args, check=True)
```
JackTemaki, line 162: I would use `util.create_executable` so that the command can be run manually.

mmz33 (approved):

JackTemaki (approved):

PR #37, opened 2021-05-31
Create LICENSE file

albertz:
Just to have it standard conform (as other repos have it).

MPL was already mentioned in the Readme, so that is what I choose here.
(Although I wonder a bit about the choice. Apache, MIT, simplified BSD are all much more common and otherwise very similar.)

### DIFF
--- a/LICENSE
+++ b/LICENSE
@@ -1,0 +1,373 @@
+Mozilla Public License Version 2.0
+==================================
+
+1. Definitions
+--------------
+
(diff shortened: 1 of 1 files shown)


curufinwe (approved):

curufinwe:
We chose the license for sisyphus because it requires users to publish their modifications of the software if they modify existing files, but does not force you to publish code that you write "around" the application. This creates an incentive to contribute back.

Issue #32, opened 2021-05-27
Removing tk.uncached_path


JackTemaki:
As there were also plans to remove `tk.uncached_path` completely (well, making it deprecated) from Sispyhus, I would suggest so start removing this from all locations and replacing it by `get_path()` and `get_cached_path()`, which is much cleaner. In most parts we do not want to have the dual usage of `str/tk.Path`, and if you want to pass external wrapping this as a Path before passing this to a Job is much cleaner.

If there are no objections within the next days, I will refactor that myself.

curufinwe:
Moving to Path only is ok for me. But if we change the caching behavior we should maybe switch to something where the caching is fully integrated also into the python code. I.e. instead of having the path object modify the string it should provide the location of the cached object. Even better: do this through an open wrapper. Something like:
```
class MyJob(job):
  def __init__(self, path):
    self.path = path

  def run(self):
    with self.path.open("rt"):  # applies the cache function and open the cached file
       pass
   with self.path.open("wt"): # also applies the cache function but also copies the file back to the original filesystem once it is closed
    pass
  some_library.process_read_only(self.path.get_cached_path())  # if some other library processes the file in read-only fashion use the cached path
  some_library.write_to(self.path.get_path())  # otherwise we path the "normal" path
```

JackTemaki:
We already have the definable `file_caching` function as wrapper, so adding some implementation here should be no problem, and does not require to change anything in Sisyphus. So even adding a subprocess call there to call `cf` might already do the correct thing (which can of course be replaced by a nicer python library that does this)

curufinwe:
The function only does some basic manipulation of the string, but ideally we switch this to something better integrated. Actually there are still some places where `cf` is hardcoded and some places where the cache_manager is called explicitly in python.

michelwi:
In the spirit of sisyphus, all inputs to a job should be `Path` objects (or parameters of a basic type) so a path to a tool or script (e.g. returnn) _should_ in fact be a `Path` object and `str` _should_ be disallowed except for string-type parameters. 
This would also check if the tool exists and only then run the jobs.

My personal ideal world solution for returnn (and RASR, SCTK, etc) would be to use a CloneGitRepositoryJob and set the output Path object as the input of the Job that uses returnn.
A CreatePythonVEnvJob can even create the necessary returnn_python_exe as a path object and install a curated set of required libraries.
This way we would also avoid having people modify their local version of returnn and never push back the changes they made.

> Do not break old setups.

[...]

PR #39, opened 2021-05-31
Remove hardcoded cachmanager uses

curufinwe:
This will break hashes. PR mainly for visibility of this breaking change.

### DIFF
--- a/cart/estimate.py
+++ b/cart/estimate.py
@@ -111,7 +111,7 @@ class AccumulateCartStatisticsJob(rasr.RasrCommand, Job):
             "alignment.flow"
         )
         config.acoustic_model_trainer.cart_trainer.example_file = (
-            "`cf -d cart.acc.xml.$(TASK).gz`"
+            "cart.acc.xml.$(TASK).gz"
         )
 
         alignment_flow.apply_config(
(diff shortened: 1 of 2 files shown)


JackTemaki:
Why does this break hashes?

curufinwe:
string in the config changes

christophmluscher (approved):

JackTemaki (approved):
It is good to have this fixed!

mmz33 (approved):

michelwi:
You removed `cf -d some_file` calls from the rasr config which changes the behavior:

With cf -d the file is created on some temporary (local) folder while rasr is running and is only copied to the given location on completion of the job.
Without, the file is opened in place on the fileserver.

Here in the case of cart this may not influence performance much, but for creating (big) lattices, this functionality might be desirable. Do we have an option to enable it in a non-hardcoded way?

curufinwe:
Not yet. But it was suggested in #32

albertz:
Instead of this `cf ...` syntax, doesn't Sprint have sth like `--*.enable_cf=True` or so? And also, is there a way to add options to Sprint (such as `enable_cf=True`) which are ignored for the hash?

curufinwe:
There is a compile time option to toggle on support for the cache manager. But there is no way to toggle this at runtime. In any case the current cache manager solution is only one way to do caching and if we "do it right" now we should look for a general interface.

Marvin84 (approved):

PR #41, opened 2021-06-01
Returnn Compatible BPE Jobs

mmz33:
Add BPE train and apply jobs that uses @albertz subword-nmt fork. This is compatible with RETURNN. I already used these jobs to train attention models for example. Now in the future, we probably want to have one job that uses `sentencepieces` software.

@curufinwe Should I remove the old train and apply BPE jobs? I am not sure if they are used.

### DIFF
--- a/bpe/apply.py
+++ b/bpe/apply.py
@@ -15,12 +15,16 @@ import i6_core.util as util
 
 
 class ApplyBPEModelToLexiconJob(Job):
-    def __init__(self, bpe_code, lexicon_path, vocabulary_path=None, subword_nmt_repo=None):
+    def __init__(
+        self, bpe_code, lexicon_path, vocabulary_path=None, subword_nmt_repo=None
+    ):
         self.bpe_code = bpe_code
         self.lexicon_path = lexicon_path
         self.vocabulary_path = vocabulary_path
 
-        self.subword_nmt_repo = subword_nmt_repo if subword_nmt_repo is not None else gs.SUBWORD_NMT_PATH
+        self.subword_nmt_repo = (
+            subword_nmt_repo if subword_nmt_repo is not None else gs.SUBWORD_NMT_PATH
+        )
 
         self.out_converted_lexicon = self.output_path("lexicon.xml.gz", cached=True)
 
@@ -109,7 +113,9 @@ class ApplyBPEToTextJob(Job):
         """
         self.text_file = text_file
         self.bpe_codes = bpe_codes
-        self.subword_nmt_repo = subword_nmt_repo if subword_nmt_repo is not None else gs.SUBWORD_NMT_PATH
+        self.subword_nmt_repo = (
+            subword_nmt_repo if subword_nmt_repo is not None else gs.SUBWORD_NMT_PATH
+        )
         self.bpe_vocab = bpe_vocab
 
         self.out_bpe_text = self.output_path("words_to_bpe.txt")
@@ -132,5 +138,5 @@ class ApplyBPEToTextJob(Job):
         if self.bpe_vocab:
             cmd += ["--vocabulary", tk.uncached_path(self.bpe_vocab)]
 
-        util.create_executable('apply_bpe.sh', cmd)
+        util.create_executable("apply_bpe.sh", cmd)
         sp.run(cmd, check=True)
--- a/bpe/apply.py
+++ b/bpe/apply.py
@@ -11,20 +11,18 @@ from sisyphus import *
 Path = setup_path(__package__)
 
 from i6_core.lib.lexicon import Lexicon
-from i6_core.tools.git import *
+import i6_core.util as util
 
 
 class ApplyBPEModelToLexiconJob(Job):
-    def __init__(self, bpe_code, lexicon_path, vocabulary_path=None):
+    def __init__(self, bpe_code, lexicon_path, vocabulary_path=None, subword_nmt_repo=None):
         self.bpe_code = bpe_code
         self.lexicon_path = lexicon_path
         self.vocabulary_path = vocabulary_path
 
-        self.subword_nmt_repo = CloneGitRepositoryJob(
-            "https://github.com/rsennrich/subword-nmt.git"
-        ).out_repository
+        self.subword_nmt_repo = subword_nmt_repo if subword_nmt_repo is not None else gs.SUBWORD_NMT_PATH
 
-        self.converted_lexicon = self.output_path("lexicon.xml.gz", cached=True)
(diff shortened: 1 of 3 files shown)


bpe/apply.py
```diff
@@ -95,3 +95,45 @@ def run(self):
...
+        """
+        :param Path|str text_file: words text file to convert to bpe
+        :param Path|str bpe_codes: bpe codes file
+        :param Path|str|None: bpe_vocab: if provided, then merge operations that produce OOV are reverted
+        """
+        self.text_file = text_file
+        self.bpe_codes = bpe_codes
+        self.bpe_vocab = bpe_vocab
+
+        self.subword_nmt_repo = CloneGitRepositoryJob(
```
curufinwe, line 115: I think it's better to have this as an argument to the job. It makes it a bit more complicated to use the job, but you gain more flexibility. Eventually there might be changes in the NMT repo that make you want to update to a different version.

bpe/apply.py
```diff
@@ -95,3 +95,45 @@ def run(self):
...
+            "--input",
+            tk.uncached_path(self.text_file),
+            "--codes",
+            tk.uncached_path(self.bpe_codes),
+            "--output",
+            self.out_bpe_text.get_path(),
+        ]
+        if self.bpe_vocab:
+            cmd += ["--vocabulary", tk.uncached_path(self.bpe_vocab)]
+        sp.run(cmd)
```
curufinwe, line 139: use `util.create_executable` and use `check=True`

bpe/apply.py
```diff
@@ -22,7 +22,7 @@ def __init__(self, bpe_code, lexicon_path, vocabulary_path=None):
 
         self.subword_nmt_repo = CloneGitRepositoryJob(
```
curufinwe, line 23: see comment below

bpe/train.py
```diff
@@ -29,7 +30,7 @@ def __init__(
 
         self.subword_nmt_repo = CloneGitRepositoryJob(
```
curufinwe, line 31: same here

bpe/train.py
```diff
@@ -79,52 +80,64 @@ def __init__(self, text_file, bpe_size, unk_label="UNK"):
...
         """
         self.text_file = text_file
         self.bpe_size = bpe_size
         self.unk_label = unk_label
 
-        self.bpe_codes = self.output_path("bpe.codes")
-        self.bpe_vocab = self.output_path("bpe.vocab")
-        self.vocab_size = self.output_var("vocab_size")
-
         self.subword_nmt_repo = CloneGitRepositoryJob(
```
curufinwe, line 91: again

bpe/train.py
```diff
@@ -79,52 +80,64 @@ def __init__(self, text_file, bpe_size, unk_label="UNK"):
...
+            "--txt",
+            tk.uncached_path(self.text_file),
+            "--bpe",
+            self.out_bpe_codes.get_path(),
+            "--unk",
+            self.unk_label,
+            "--out",
+            self.out_bpe_vocab.get_path(),
+        ]
+        sp.run(bpe_vocab_cmd)
```
curufinwe, line 136: see above + set `check=True`

curufinwe (changes requested):

bpe/apply.py
mmz33: We agreed that this won't change. That's why I also use a fixed commit hash. This anw might get deprecated later as we plan in the future to use `sentencepiece` software which can generates BPEs, Sentencepices, etc. This is more generic and already support by RETURNN.

JackTemaki: Still this should definitely be outside of the job

mmz33: hmm ok. this will make it a bit complicated but ok i can just pass the repo as an argument.

albertz: Shouldn't we handle this like we handle other tools or external dependencies like RETURNN, RASR? In the usual case, I would not want to explicitly pass this as an argument, but instead just install it somewhere and then Sisyphus would pick my default.

christophmluscher: Does a switch to `sentencepiece` mean you want to extend the job, deprecate subword_nmt or create new jobs?
I am not sure if we want to deprecate subword_nmt in a year or so, since then people will be relying on these jobs..

> Shouldn't we handle this like we handle other tools or external dependencies like RETURNN, RASR?

the default can be specified in `settings.py` but you can also pass a specific path to the job

albertz: > Does a switch to `sentencepiece` mean you want to extend the job, deprecate subword_nmt or create new jobs?

The job cannot really be extended/changed when `subword_nmt` is an argument, right?

SentencePiece also creates other types of files. So it probably doesn't make sense to use the same job.

> I am not sure if we want to deprecate subword_nmt in a year or so, since then people will be relying on these jobs..
> 
> > Shouldn't we handle this like we handle other tools or external dependencies like RETURNN, RASR?
>
> the default can be specified in `settings.py` but you can also pass a specific path to the job

You mean for Subword-NMT and SentencePiece?

mmz33: > Does a switch to sentencepiece mean you want to extend the job, deprecate subword_nmt or create new jobs?
I am not sure if we want to deprecate subword_nmt in a year or so, since then people will be relying on these jobs..
No this would be new job. this will not affect this job so the discussion is independent to this PR.

mmz33: ok so how should we proceed here? I can define the subword-nmt repo in `settings.py` and then use it in the job as `gs.SUBWORD_NMT` or so. but also i can keep the repo as parameter and set it to default to None. So if this was `None`, it will just use the one defined in `settings.py`

JackTemaki: Yes, I would do it exactly like the RETURNN jobs do it

bpe/train.py
```diff
@@ -74,57 +74,71 @@ def run(self):
 
 
 class ReturnnTrainBpeJob(Job):
-    def __init__(self, text_file, bpe_size, unk_label="UNK"):
+    def __init__(self, text_file, bpe_size, subword_nmt_repo=None, unk_label="UNK"):
         """
         Create Bpe codes and vocab files
         NOTE: This uses Albert's subword-nmt fork which is compatible to RETURNN BytePairEncoding class
```
curufinwe, line 80: concrete reference to git-repo would be more helpful here, now that the repo is not pulled automatically anymore.

curufinwe (changes requested):

curufinwe (approved):

PR #42, opened 2021-06-09
Added DynamicSplitSegmentFileJob;

erreurt:
Added DynamicSplitSegmentFileJob (see SplitSegmentFileJob) to avoid output paths conflicts with earlier settings. 
DynamicSplitSegmentFileJob receives a **Delayed** object for argument concurrent. 
Only an output directory called **out_split_dir** should be accessible after job creation.

### DIFF
--- a/corpus/segments.py
+++ b/corpus/segments.py
@@ -4,6 +4,7 @@ __all__ = [
     "SegmentCorpusByRegexJob",
     "ShuffleAndSplitSegmentsJob",
     "SplitSegmentFileJob",
+    "DynamicSplitSegmentFileJob",
     "SortSegmentsByLengthAndShuffleJob",
     "UpdateSegmentsWithSegmentMapJob",
 ]
@@ -258,6 +259,33 @@ class SplitSegmentFileJob(Job):
             start = end
             end += n // self.concurrent + (1 if i <= m else 0)
             with open(self.single_segments[i].get_path(), "wt") as f:
+                f.writelines(lines[start:end])
+
+
+class DynamicSplitSegmentFileJob(Job):
+
+    def __init__(self, segment_file, concurrent=1):
+        self.segment_file = segment_file
+        self.concurrent = concurrent
+        self.segment_dir = self.output_path("split", directory=True)
+
+    def tasks(self):
+        yield Task("run", resume="run", mini_task=True)
+
+    def run(self):
+        with open(tk.uncached_path(self.segment_file), "rt") as f:
+            lines = [l for l in f.readlines() if len(l.strip()) > 0]
+
+        nb_seg = len(lines)
+        self.concurrent = self.concurrent.get()
+        seg_per_split = nb_seg // self.concurrent
+        nb_rest_seg = nb_seg % self.concurrent
+        end = 0
+        for i in range(1, self.concurrent + 1):
+            start = end
+            fpath = '{}/segments.{}'.format(self.segment_dir, i)
+            end += seg_per_split + (1 if i <= nb_rest_seg else 0)
+            with open(fpath, "wt") as f:
                 f.writelines(lines[start:end])
 
 
--- a/corpus/segments.py
+++ b/corpus/segments.py
@@ -271,7 +271,7 @@ class DynamicSplitSegmentFileJob(Job):
 
     def __init__(self, segment_file, concurrent):
         """
-        :param tk.Union[tk.Path, str] segment_file: segment file
+        :param tk.Path|str segment_file: segment file
         :param tk.Delayed concurrent: number of splits
         """
         self.segment_file = segment_file
--- a/corpus/segments.py
+++ b/corpus/segments.py
@@ -265,8 +265,8 @@ class SplitSegmentFileJob(Job):
 class DynamicSplitSegmentFileJob(Job):
 
     """
-    This split the segments to concurrent many shares.
-    It is a variant to the existing SplitSegmentFileJob, which requires a delayed variable for argument concurrent.
+    Split the segments to concurrent many shares. It is a variant to the existing SplitSegmentFileJob.
+    This requires a tk.Delayed variable (instead of int) for the argument concurrent.
     """
 
     def __init__(self, segment_file, concurrent):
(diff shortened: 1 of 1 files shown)


corpus/segments.py
```diff
@@ -261,6 +262,33 @@ def run(self):
...
+
+    def tasks(self):
+        yield Task("run", resume="run", mini_task=True)
+
+    def run(self):
+        with open(tk.uncached_path(self.segment_file), "rt") as f:
+            lines = [l for l in f.readlines() if len(l.strip()) > 0]
+
+        nb_seg = len(lines)
+        self.concurrent = self.concurrent.get()
```
curufinwe, line 280: this will fail when concurrent keeps it's default value of 1, please do a check for Delayed here

corpus/segments.py
```diff
@@ -261,6 +262,33 @@ def run(self):
                 f.writelines(lines[start:end])
 
 
+class DynamicSplitSegmentFileJob(Job):
+
+    def __init__(self, segment_file, concurrent=1):
```
curufinwe, line 267: missing documentation, check for example the returnn jobs for correct format

corpus/segments.py
```diff
@@ -261,6 +262,33 @@ def run(self):
...
+        self.segment_file = segment_file
+        self.concurrent = concurrent
+        self.segment_dir = self.output_path("split", directory=True)
```
curufinwe, line 270: output members should have prefix out_ in the name

curufinwe (changes requested):
Please reformat code with black (see https://github.com/rwth-i6/i6_core/blob/main/CONTRIBUTUNG.md )

corpus/segments.py
JackTemaki: Also, if this job should only be used for a "Dynamic" concurrent value, you can remove the default "1" here and set the type to `Delayed` only.

corpus/segments.py
erreurt: I hope we can write those constrains/requirements (also that regarding documentation) as a guideline. This will save a lot of time in the future.

JackTemaki: We have this here: https://github.com/rwth-i6/i6_core/blob/main/CONTRIBUTUNG.md,
but of course this can be extended if something is unclear.

erreurt: sorry for missing the info you already mentioned above

corpus/segments.py
```diff
@@ -261,6 +262,43 @@ def run(self):
...
+
+    def tasks(self):
+        yield Task("run", resume="run", mini_task=True)
+
+    def run(self):
+        with open(tk.uncached_path(self.segment_file), "rt") as f:
+            lines = [l for l in f.readlines() if len(l.strip()) > 0]
+
+        nb_seg = len(lines)
+        if isinstance(self.concurrent, tk.Delayed):
```
mmz33, line 289: you don't need this now

mmz33 (changes requested):

corpus/segments.py
```diff
@@ -261,6 +262,43 @@ def run(self):
...
+        """
+        self.segment_file = segment_file
+        self.concurrent = concurrent
+        self.out_split_dir = self.output_path("split", directory=True)
+
+    def tasks(self):
+        yield Task("run", resume="run", mini_task=True)
+
+    def run(self):
+        with open(tk.uncached_path(self.segment_file), "rt") as f:
```
mmz33, line 285: use `util.uopen` instead of `open`. see other usages

mmz33 (changes requested):

corpus/segments.py
JackTemaki: Did we already put this in the guidelines file? If not I will add this

corpus/segments.py
```diff
@@ -261,6 +262,43 @@ def run(self):
...
+
+        nb_seg = len(lines)
+        if isinstance(self.concurrent, tk.Delayed):
+            self.concurrent = self.concurrent.get()
+        seg_per_split = nb_seg // self.concurrent
+        nb_rest_seg = nb_seg % self.concurrent
+        end = 0
+        for i in range(1, self.concurrent + 1):
+            start = end
+            fpath = '{}/segments.{}'.format(self.out_split_dir, i)
```
mmz33, line 296: `self.out_split_dir` is a `tk.Path` object. Does it work with string `format`?

mmz33 (changes requested):

corpus/segments.py
mmz33: yeah it works.

JackTemaki: This does work, but I am not sure if `__str__` returns the correct thing here (maybe this calls the caching...)

mmz33: ok then this need to be verified

corpus/segments.py
mmz33: ok thanks

erreurt: should I change correspondingly for all other Jobs in that file which has this issue?

JackTemaki: No this is not needed, I already planned to do it for all Jobs at once, it is just that new submissions should already follow this.

corpus/segments.py
```diff
@@ -261,6 +262,42 @@ def run(self):
...
+        """
+        self.segment_file = segment_file
+        self.concurrent = concurrent
+        self.out_split_dir = self.output_path("split", directory=True)
+
+    def tasks(self):
+        yield Task("run", resume="run", mini_task=True)
+
+    def run(self):
+        with uopen(tk.uncached_path(self.segment_file), "rt") as f:
```
curufinwe, line 285: uopen will call `tk.uncached_path`for you

curufinwe (changes requested):
Only one minor thing left.

corpus/segments.py
erreurt: true! will do

curufinwe (approved):

curufinwe (changes requested):
code formatting check fails

PR #43, opened 2021-06-11
Added utility function to get all speakers from a corpus and its subc…

mattiadg:
…orpora


mattiadg:
It is useful to improve the usability of the class without forcing the user to know all the implementation details

curufinwe (approved):

PR #45, opened 2021-06-11
refactor score_features.py

JackTemaki:
The doscstring ScoreFeaturesJob is still empty, because I am not 100% sure what this job actually can do, so here someone else should fill this in.

### DIFF
--- a/am/score_features.py
+++ b/am/score_features.py
@@ -12,7 +12,11 @@ Path = setup_path(__package__)
 
 
 class ScoreFeaturesJob(rasr.RasrCommand, Job):
-    """ """
+    """
+
+    This job uses a (multiple) RASR process(es) to forward all the data and calculates the average softmax output.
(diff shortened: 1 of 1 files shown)


michelwi:
This job uses a (multiple) RASR process(es) to forward all the data and calculates the average softmax output. This is to calculate the prior based on the "Povey" method.

I guess this job should be deprecated as it is simpler and faster to use returnn +GPU to calculate this prior. (If returnn still supports the task "compute_prior")

JackTemaki:
> This job uses a (multiple) RASR process(es) to forward all the data and calculates the average softmax output. This is to calculate the prior based on the "Povey" method.

Done

> I guess this job should be deprecated as it is simpler and faster to use returnn +GPU to calculate this prior. (If returnn still supports the task "compute_prior")

Do you have a job for this? I did not see one yet in the `i6_core`

curufinwe (approved):

PR #47, opened 2021-06-11
refactor cart jobs

JackTemaki:
Here it would be again nice if someone could add a descriptive docstring for the jobs, I am really not familiar with CART.

### DIFF
--- a/cart/estimate.py
+++ b/cart/estimate.py
@@ -5,13 +5,16 @@ import shutil
 
 from sisyphus import *
 
-Path = setup_path(__package__)
-
+from i6_core.cart.questions import BasicCartQuestions
 import i6_core.rasr as rasr
 import i6_core.util as util
 
+Path = setup_path(__package__)
+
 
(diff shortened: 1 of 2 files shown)


michelwi:
Nobody is an expert in CART any more, but let me try. Maybe we want to replace "triphone state" with "allophone" since this term is used more often in the recipes. I am open to discuss.

AccumulateCartStatisticsJob: 
Goes over all training data and for each triphone state accumulates the values and squared values of the given feature flow

EstimateCartJob:
This job estimates a phonetic decision tree. Given a set of accumulated (squared) feature values a single gaussian model is estimated per triphone state. Then iteratively states are merged according to the provided questions such that the log-likelihood of the resulting models is minimized. Finally states which have a low number of occurrences are merged into the closes cluster.

Source:
https://www-i6.informatik.rwth-aachen.de/publications/download/262/Beulen--1999.pdf

curufinwe (approved):

PR #48, opened 2021-06-14
Rename parameter names of corpus jobs

JackTemaki:
The jobs still needs general refactoring, this is mostly about streamlining the corpus job names.

### DIFF
--- a/corpus/convert.py
+++ b/corpus/convert.py
@@ -187,14 +187,14 @@ class CorpusToTextDictJob(Job):
     Extract the Text from a Bliss corpus to fit a "{key: text}" structure (e.g. for RETURNN)
     """
 
-    def __init__(self, bliss_corpus, segments=None, invert_match=False):
+    def __init__(self, bliss_corpus, segment_file=None, invert_match=False):
         """
         :param Path bliss_corpus: bliss corpus file
-        :param Path|None segments: a segment file as optional whitelist
+        :param Path|None segment_file: a segment file as optional whitelist
         :param bool invert_match: use segment file as blacklist (needs to contain full segment names then)
         """
         self.bliss_corpus = bliss_corpus
-        self.segments_file_path = segments
+        self.segment_file = segments
         self.invert_match = invert_match
 
         self.out_dictionary = self.output_path("text_dictionary.py")
@@ -209,8 +209,8 @@ class CorpusToTextDictJob(Job):
         dictionary = {}
 
         segments = None
-        if self.segments_file_path:
-            with uopen(self.segments_file_path) as f:
+        if self.segment_file:
+            with uopen(self.segment_file) as f:
                 segments = set(line.decode().strip() for line in f)
 
         for segment in c.segments():
--- a/corpus/convert.py
+++ b/corpus/convert.py
@@ -194,7 +194,7 @@ class CorpusToTextDictJob(Job):
         :param bool invert_match: use segment file as blacklist (needs to contain full segment names then)
         """
         self.bliss_corpus = bliss_corpus
-        self.segment_file = segments
+        self.segment_file = segment_file
         self.invert_match = invert_match
 
         self.out_dictionary = self.output_path("text_dictionary.py")
--- a/corpus/data_augmentation.py
+++ b/corpus/data_augmentation.py
@@ -22,16 +22,16 @@ class SelfNoiseCorpusJob(Job):
     WARNING: This Job uses /dev/shm for performance reasons, please be cautious
     """
 
-    def __init__(self, corpus_file, snr, corpus_name, n_noise_tracks=1, seed=0):
+    def __init__(self, bliss_corpus, snr, corpus_name, n_noise_tracks=1, seed=0):
         """
 
-        :param Path corpus_file: Bliss corpus with wav files
+        :param Path bliss_corpus: Bliss corpus with wav files
         :param float snr: signal to noise ratio in db, positive values only
         :param str corpus_name: name of the new corpus
         :param int n_noise_tracks: number of random (parallel) utterances to add
         :param int seed: seed for random utterance selection
         """
-        self.corpus_file = corpus_file
+        self.bliss_corpus = bliss_corpus
         self.snr = snr
         self.corpus_name = corpus_name
         self.n_noise_tracks = n_noise_tracks
@@ -59,7 +59,7 @@ class SelfNoiseCorpusJob(Job):
         nc = corpus.Corpus()
         segment_file_names = []
 
-        c.load(tk.uncached_path(self.corpus_file))
+        c.load(tk.uncached_path(self.bliss_corpus))
         nc.name = self.corpus_name
         nc.speakers = c.speakers
         nc.default_speaker = c.default_speaker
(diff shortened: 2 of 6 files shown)


corpus/data_augmentation.py
```diff
@@ -59,7 +59,7 @@ def run(self):
         nc = corpus.Corpus()
         segment_file_names = []
 
-        c.load(tk.uncached_path(self.corpus_file))
+        c.load(tk.uncached_path(self.bliss_corpus))
```
mmz33, line 62: use `.get_path()`

corpus/data_augmentation.py
```diff
@@ -204,7 +204,7 @@ def run(self):
...
```
mmz33, line 207: use `.get_path()`

corpus/transform.py
```diff
@@ -25,14 +25,14 @@
...
+    def __init__(self, bliss_corpus, ctm_path, remove_empty_segments=True):
         self.set_vis_name("Replace Transcription from CTM file")
 
-        self.corpus_path = corpus_path
+        self.corpus_path = bliss_corpus
         self.ctm_path = ctm_path
         self.remove_empty_segments = remove_empty_segments
 
-        gzip_output = tk.uncached_path(corpus_path).endswith(".gz")
+        gzip_output = tk.uncached_path(bliss_corpus).endswith(".gz")
```
mmz33, line 35: use `.get_path()`

mmz33 (changes requested):

curufinwe (approved):

curufinwe:
OK from my point of view, but @christophmluscher @Marvin84 @michelwi please also comment on this one.

mmz33 (approved):

corpus/filter.py
```diff
@@ -174,16 +174,18 @@ def run(self):
...
 class FilterCorpusBySegmentsJob(Job):
-    def __init__(self, corpus_file, segment_list, compressed=False, invert_match=False):
+    def __init__(
+        self, bliss_corpus, segment_file, compressed=False, invert_match=False
+    ):
         """
-        :param Path corpus_file:
-        :param Path segment_list:
+        :param Path bliss_corpus:
+        :param Path segment_file:
```
michelwi, line 182: This job was written with a list of segment files in mind and later extended to include the option of having a single segment file.

I agree that the "single file" case can be the expected default and we can name the parameter `segment_file`, but we should include the possibility to give a list of files in the doc string

corpus/filter.py
```diff
@@ -174,16 +174,18 @@ def run(self):
...
         :param bool compressed:
         :param bool invert_match:
         """
-        self.in_corpus = corpus_file
+        self.in_corpus = bliss_corpus
         self.segment_list = (
```
michelwi, line 187: `segment_list` implies to me a python-list of segment names. Maybe we should rename this to `self.segment_file_list` ?

michelwi (changes requested):
Found one problem in your changes, the other comments for your consideration.

Also one more comment about "bliss_*" naming scheme:
* in your changes it is only used in the init functions but not consistently throughout the Jobs
* having the tag at the fron makes it look like a qualifier for corpus (bliss_corpus vs. audio_corpus or text_corpus) rather than a file format (corpus_bilss_file vs. corpus_xml_file or corpus_text_file)
* it adds no additional information here, as we exclusively support "bliss" file format

so I would argue to consistently name everything to `corpus_file` and `lexicon_file` (or maybe `corpus_X_file` to specify file format requirements).

curufinwe:
Actually we might support other corpus/lexica formats via RETURNN, so the addition of the file format is a good idea IMO.

JackTemaki:
> Actually we might support other corpus/lexica formats via RETURNN, so the addition of the file format is a good idea IMO.

This is why I added `bliss_` in front of all `corpus` and `lexicon` parameters. When I asked on slack for naming suggestions no one but Chris answered, so I took this for "sufficiently discussed"

> in your changes it is only used in the init functions but not consistently throughout the Jobs

I already have a commit for that, I can add it here

corpus/convert.py
```diff
@@ -187,14 +187,14 @@ class CorpusToTextDictJob(Job):
...
+    def __init__(self, bliss_corpus, segment_file=None, invert_match=False):
         """
         :param Path bliss_corpus: bliss corpus file
-        :param Path|None segments: a segment file as optional whitelist
+        :param Path|None segment_file: a segment file as optional whitelist
         :param bool invert_match: use segment file as blacklist (needs to contain full segment names then)
         """
         self.bliss_corpus = bliss_corpus
-        self.segments_file_path = segments
+        self.segment_file = segments
```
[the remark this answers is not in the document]
JackTemaki: Hmm, PyCharm should have highlighted this...

corpus/filter.py
JackTemaki: I would go for `segment_files` then instead.

michelwi (approved):

corpus/filter.py
```diff
@@ -174,16 +174,18 @@ def run(self):
...
 class FilterCorpusBySegmentsJob(Job):
-    def __init__(self, corpus_file, segment_list, compressed=False, invert_match=False):
+    def __init__(
+        self, bliss_corpus, segment_files, compressed=False, invert_match=False
+    ):
         """
-        :param Path corpus_file:
-        :param Path segment_list:
+        :param Path bliss_corpus:
+        :param list[Path]|Path segment_files: a single segment file or a list of segment files
```
christophmluscher, line 182: In `CorpusToTextDictJob` class `segment_file` was used, which only accepted a single segment file. Do we want to change this behavior?

corpus/transform.py
```diff
@@ -139,7 +139,7 @@ def tasks(self):
 
     def run(self):
         c = corpus.Corpus()
-        c.load(str(self.corpus))
+        c.load(str(self.bliss_corpus))
```
christophmluscher, line 142: `str` should be replaced

corpus/transform.py
```diff
@@ -437,10 +437,6 @@ def run(self):
 class ShiftCorpusSegmentStartJob(Job):
     """
     Shifts the start time of a corpus to change the fft window offset
-
-    :param Path corpus_file: path to a bliss corpus file
-    :param str corpus_name: name of the new corpus
-    :param float shift: shift in seconds
     """
 
     def __init__(self, corpus_file, corpus_name, shift):
```
christophmluscher, line 442: Here refactoring `corpus_file` to `bliss_corpus` missing

corpus/transform.py
```diff
@@ -437,10 +437,6 @@ def run(self):
 class ShiftCorpusSegmentStartJob(Job):
     """
     Shifts the start time of a corpus to change the fft window offset
-
-    :param Path corpus_file: path to a bliss corpus file
-    :param str corpus_name: name of the new corpus
-    :param float shift: shift in seconds
```
christophmluscher, line 443: was this supposed to be removed or moved to `__init__`?

christophmluscher (changes requested):

corpus/transform.py
JackTemaki: Ah yes, I will fix this. But anyway I will have to do a second pass of refactoring where I fix all the other issues, I wanted to start with the parameter names first.

corpus/filter.py
JackTemaki: We could do this later, actually this is a good point. Maybe it should be singular here (as it takes both), so that if we enable "multiple" inputs at other locations, we do not have to change the name.

christophmluscher: I am fine with that. seems a bit easier to use singular and allow both. especially if jobs get extended from singular to multiple input

christophmluscher (approved):

Issue #55, opened 2021-06-16
Add output_mode parameter to ReturnnSearchJob as in ReturnnSearchFromFileJob


jvhoffbauer:
I think it would make sense to add an `output_mode` parameter to `ReturnnSearchJob`. `ReturnnSearchFromFileJob` has this parameter already and it specifies whether Returnn outputs the search result in `'txt'` or `'py'` mode. 

Currently `ReturnnSearchJob` has `'py'` as default, while Returnn uses `'txt'` as its default. This can lead to problems when migrating a pipeline. So I would also think about changing the default to `'txt'`

curufinwe:
First part sounds OK to me. Second part would break already existing sisyphus setups, so we'd have to do it this week or not at all.
@mmz33 @JackTemaki  thoughts?

JackTemaki:
The default was `py` because `ReturnnComputeWERJob` (direct WER), and `SearchWordsToCTMJob`  (for sclite WER) need the py format, so I do not think it is a good idea to have 'txt' as default, even if this is the case for RETURNN.

christophmluscher:
>  Second part would break already existing Sisyphus setups

I think both parts would break the hash or not? Or do we want to include `__sis_hash_exclude__` already?

@jvhoffbauer which pipelines are using `txt` format?

mmz33:
I would also agree with @JackTemaki . I am not sure what you mean here by migrating a pipeline. You would need to use the jobs in this repo which are consistent with `py` format.

JackTemaki:
Yes both parts would break the hash, but I think adding the option for the output format is fine (when done this week).

jvhoffbauer:
Ok. I will commit today.

jvhoffbauer:
> I would also agree with @JackTemaki . I am not sure what you mean here by migrating a pipeline. You would need to use the jobs in this repo which are consistent with py format.

I just wanted to point out, that the Returnn and the recipe default are different here and this might be unintuitive. But anyways, I also think the py output format should be the default as it is easier to parse and integrates with he rest of the i6_core recipes.

curufinwe:
> @JackTemaki I agree.
> 
> So adding the option: yes. Changing the default: no. @curufinwe yes?
> 
> @jvhoffbauer can you make the pull request today or tomorrow?

correct

christophmluscher:
> Yes both parts would break the hash

[...]

PR #56, opened 2021-06-16
Make output mode configurable in ReturnnSearchJob

jvhoffbauer:
As discussed in #55

### DIFF
--- a/returnn/search.py
+++ b/returnn/search.py
@@ -32,6 +32,7 @@ class ReturnnSearchJob(Job):
         model_checkpoint,
         returnn_config,
         *,
+        output_mode="py",
         log_verbosity=3,
         device="gpu",
         time_rqmt=4,
@@ -55,6 +56,9 @@ class ReturnnSearchJob(Job):
         self.search_data = search_data
 
         self.model_checkpoint = model_checkpoint
+
+        assert output_mode in ["py", "txt"]
+        self.output_mode = output_mode
 
         self.returnn_python_exe = (
             returnn_python_exe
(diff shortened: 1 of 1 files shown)


returnn/search.py
```diff
@@ -32,6 +32,7 @@ def __init__(
         model_checkpoint,
         returnn_config,
         *,
+        output_mode="py",
```
mmz33, line 35: this should be also added to `hash` function below.

mmz33 (changes requested):

returnn/search.py
christophmluscher: the `output_mode` gets added to the return_config param in hash. that should suffice, right!?

returnn/search.py
```diff
@@ -56,6 +57,9 @@ def __init__(
 
         self.model_checkpoint = model_checkpoint
 
+        assert output_mode in ["py", "txt"]
+        self.output_mode = output_mode
```
christophmluscher, line 61: maye we could pass this directly to the `create_returnn_config`!? so that if necessary the `create_returnn_config` could also easily be called from outside the job?

returnn/search.py
curufinwe: correct

returnn/search.py
curufinwe: Yes, this would be nice, but we can also do this in another PR, that one would not break anything.

curufinwe (approved):

returnn/search.py
mmz33: ah yes right.

dthulke: It gets added to self.returnn_config but this is not used in the hash (the original version from the kwargs is used there). So the output_mode currently does not affect the hash.

Created a PR to fix this: https://github.com/rwth-i6/i6_core/pull/62

PR #57, opened 2021-06-17
Make merging mixtures more disk-efficient

curufinwe:
Before we used to delete the temp mixtures after everything was merged. Now the strategy is to delete the mixtures as soon as they are merged (except for the original accumulated mixtures).

### DIFF
--- a/mm/mixtures.py
+++ b/mm/mixtures.py
@@ -107,6 +107,7 @@ class MergeMixturesJob(rasr.RasrCommand, Job):
                 if e in tmp_files:
                     logging.info("unlink %s" % e)
                     os.unlink(e)
+                    tmp_files.remove(e)
 
             return tmp_merge_file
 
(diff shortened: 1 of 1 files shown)


mm/mixtures.py
```diff
@@ -100,6 +103,11 @@ def merge_helper(elements):
                 "merge.log.%d" % merge_num, self.merge_log_file[merge_num].get_path()
             )
 
+            for e in elements:
+                if e in tmp_files:
+                    logging.info("unlink %s" % e)
+                    os.unlink(e)
+
```
michelwi, line 110: after unlinking the files should you remove them from `tmp_files`?

curufinwe: possible, but not necessary. tmp_files is just an allow-list for elements to delete.

michelwi (approved):

PR #59, opened 2021-06-18
fix dataset behavior in rasr training

JackTemaki:
This was accidentally removed when refactoring the `ReturnnTrainingJob`, but also the old previous behavior was not optimal.

The question that remains here is: should `partition_epochs` stay like this, or do we want to remove it, make it more explicit or something else?

### DIFF
--- a/returnn/rasr_training.py
+++ b/returnn/rasr_training.py
@@ -72,14 +72,16 @@ class ReturnnRasrTrainingJob(ReturnnTrainingJob):
         :param disregarded_classes:
         :param class_label_file:
         :param buffer_size:
-        :param partition_epochs:
+        :param dict[str, int] partition_epochs:
         :param extra_rasr_config:
         :param extra_rasr_post_config:
         :param additional_rasr_config_files:
         :param additional_rasr_post_config_files:
         :param use_python_control:
         """
-        datasets = self.create_dataset_config(train_crp, partition_epochs)
+        datasets = self.create_dataset_config(
+            train_crp, returnn_config, partition_epochs
+        )
         returnn_config.config["train"] = datasets["train"]
         returnn_config.config["dev"] = datasets["dev"]
         super().__init__(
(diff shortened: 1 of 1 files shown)


JackTemaki:
Also relevant for @mattiadg and @papar22

michelwi (approved):

michelwi:
> The question that remains here is: should `partition_epochs` stay like this, or do we want to remove it, make it more explicit or something else?

We could absorb it into the crp although it is indeed not a RASR parameter.. so, keep it like this?

returnn/rasr_training.py
```diff
@@ -232,9 +234,22 @@ def create_config(
         return config, post_config
 
     @classmethod
-    def create_dataset_config(cls, train_crp, partition_epochs):
+    def create_dataset_config(cls, train_crp, returnn_config, partition_epochs):
```
christophmluscher, line 237: do we have the situation, where `train_crp` and `dev_crp` rely on different RASR binaries? So no problem taking RASR for both train/dev from `train_crp` right?

returnn/rasr_training.py
```diff
@@ -232,9 +234,22 @@ def create_config(
...
+        """
+        :param rasr.CommonRasrParameters train_crp:
+        :param ReturnnConfig returnn_config:
+        :param dict[str, int] partition_epochs:
+        :return:
+        """
         datasets = {}
 
+        assert not (
+            "partition_epoch" in returnn_config.config["train"] and partition_epochs
```
mattiadg, line 247: What about making `partition_epochs` an empty dict by default and then check
```
assert not (
            "partition_epoch" in returnn_config.config["train"] and "train" in partition_epochs
)
```
and the same for dev?

I know, it would be strange to define partition_epoch partially in one variable and partially in another variable, but at least you remove the bad None default.

JackTemaki: Why is the `None` default worse than an empty dict?

mattiadg: Because every time we don't let something be None is a potential source of errors removed.

returnn/rasr_training.py
curufinwe: I cannot imagine a situation where you'd need to use different RASR binaries for train and dev.

returnn/rasr_training.py
curufinwe: It is not good style to have mutable objects as default parameters, since the object will be created once and reused across all calls. If for example the constructor were to change the dict it would also change for all future evaluations. This is why we usually use None as a default parameter.

curufinwe (changes requested):
Adjust doc-string and also give a description for it, i.e. that the keys of that dict are train and dev.

returnn/rasr_training.py
mattiadg: Ok sorry, I didn't know that default parameters are shared between function calls.

curufinwe (approved):

PR #60, opened 2021-06-21
add parameter blacklist to ReturnnTrainingJob

JackTemaki:
Setting these parameters directly in the 'returnn_config` leads to incorrect hashes and causes inconsistencies in the dumped config, so doing this should be prohibited.

### DIFF
--- a/returnn/training.py
+++ b/returnn/training.py
@@ -97,6 +97,7 @@ class ReturnnTrainingJob(Job):
         :param Path|str returnn_root: file path to the RETURNN repository root folder
         """
         assert isinstance(returnn_config, ReturnnConfig)
+        self.check_blacklisted_parameters(returnn_config)
         kwargs = locals()
         del kwargs["self"]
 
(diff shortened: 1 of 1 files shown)


papar22 (approved):

christophmluscher (approved):

michelwi (approved):

returnn/training.py
```diff
@@ -370,6 +371,28 @@ def create_returnn_config(
...
+        blacklisted_keys = [
+            "log_verbosity",
+            "device",
+            "num_epochs",
+            "save_interval",
+            "keep_epochs",
+        ]
+        for key in blacklisted_keys:
+            assert not (
+                key in returnn_config.config or key in returnn_config.post_config
```
curufinwe, line 390: please use config.get() for this instead of manually checking config and post_config

curufinwe (changes requested):

curufinwe (approved):

PR #61, opened 2021-06-21
Removes out_ prefix for bliss corpus segments in FilterCorpus jobs

dthulke:
Probably a search and replace error while refactoring the jobs


JackTemaki:
Ah yes sorry, sometimes there are side effects even when renaming via PyCharm refactoring instead of search and replace (I do not use that). I guess I did not pay attention there...

PR #62, opened 2021-06-22
Adds the output_mode to the hash of the ReturnnSearchJob

dthulke:
The output_mode in the ReturnnSearchJob currently does not affect the hash.

This changes the hash. Is this still okay or should I add a `__sis_hash_exclude__`?

### DIFF
--- a/returnn/search.py
+++ b/returnn/search.py
@@ -171,6 +171,7 @@ class ReturnnSearchJob(Job):
             "returnn_root": kwargs["returnn_root"],
             "model_checkpoint": kwargs["model_checkpoint"],
             "search_data": kwargs["search_data"],
+            "output_mode": kwargs["output_mode"],
         }
         return super().hash(d)
 


christophmluscher:
the output_mode is already in the hash. This is not needed.

dthulke:
It's not in the hash. In create_returnn_config the output_mode gets added to self.returnn_config but that's not used in the hash.
Also, I tested this and the hash does not change when changing the output_mode.

christophmluscher:
ah right. the hash uses the keyword arg. I thought it used the output from the `create_returnn_config`. But wouldn't it be better to use that in the hash?

mmz33:
> It's not in the hash. In create_returnn_config the output_mode gets added to self.returnn_config but that's not used in the hash.

but if it is added to the config then this should change the hash. Actually, this might be an issue:
```
        config.update(copy.deepcopy(original_config))  # update with the original config
```
So this can overwrite the output_mode. Do you have `search_output_file_format` already in your train config?

dthulke:
> Do you have search_output_file_format already in your config?

No, it's not in my config (but it might be useful to check for this / add a similar blacklist as in #60)

> but if it is added to the config then this should change the hash

It is added to the config created in `create_returnn_config` (which create a copy of the original config from the kwargs), but, as Chris wrote, the original unmodified config from the kwargs is used in the hash.

But, I agree with Chris that it may be nicer to use the output of `create_returnn_config` in the hash. Therefore we would need to make the method static (if we move the assignment of search_output_file out of the method this is no problem) and change this also in the ReturnnTrainingJob for consistency.

I can create a PR for this.

PR #63, opened 2021-06-23
Use create_returnn_config to get the hash in the ReturnnSearchJob and ReturnnTrainingJob

dthulke:
Alternative to #62 (see the discussion for context)

The output of create_returnn_config is now used to get the hash in the ReturnnSearchJob and ReturnnTrainingJob.
This ensures that all changes made to the config in this method change the hash.

**This will change all hashes of train and search jobs.**

### DIFF
--- a/returnn/rasr_training.py
+++ b/returnn/rasr_training.py
@@ -273,7 +273,11 @@ class ReturnnRasrTrainingJob(ReturnnTrainingJob):
         train_config, train_post_config = cls.create_config(**kwargs)
         kwargs["crp"] = dev_crp
         dev_config, dev_post_config = cls.create_config(**kwargs)
-        returnn_config = kwargs["returnn_config"]
+
+        datasets = cls.create_dataset_config(train_crp, kwargs["partition_epochs"])
+        kwargs["returnn_config"].config["train"] = datasets["train"]
+        kwargs["returnn_config"].config["dev"] = datasets["dev"]
+        returnn_config = ReturnnTrainingJob.create_returnn_config(**kwargs)
         d = {
             "train_config": train_config,
             "dev_config": dev_config,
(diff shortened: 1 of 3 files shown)


JackTemaki:
I think this might not be a bad idea, as then there can be no conflicts when trying to override. But we need to make sure that the `config` that is set internally is exactly as we want it to be, so we would need to discuss if `"target": "classes"` needs to be removed, and if 
```        
 "need_data": False,
 "search_do_eval": 0,
```
is still correct.

JackTemaki:
Also the question would be if still blacklisting is not the better option, because now you can still do user errors. So if you say in the Job parameter `output_mode = "txt"` and in your returnn config `"search_output_file_format": "py"` it will hash "py" correctly, but this still is inconsistent...

christophmluscher:
> But we need to make sure that the config that is set internally is exactly as we want it to be

I agree.

> because now you can still do user errors

IMHO the most important part is that the config actually used in training/search is used to compute the hash. So that change in search always results in a different hash. David showed this is not the case.

Additionally, I think it is good to enforce that the user gets what she expects. So blacklist and/or warning system would be useful in this case.

JackTemaki:
> Additionally, I think it is good to enforce that the user gets what she expects. So blacklist and/or warning system would be useful in this case.

Okay, so we do all, fix the hash here, add a config consistency check (I already started code for this) and add the parameter blacklist.

JackTemaki (changes requested):
Please also fix this for the hash function in ReturnnRasrTrainingJob

JackTemaki:
Also see #64

dthulke:
> Please also fix this for the hash function in ReturnnRasrTrainingJob

Done

curufinwe (approved):

PR #64, opened 2021-06-23
returnn config consistency check

JackTemaki:
Adds a check that will prohibit accidentally overwriting hashed parameters (in `config`) with unhashed parameters (in `post_config`)

as you did not appear in the reviewer list, also @dthulke and @mattiadg

### DIFF
--- a/returnn/config.py
+++ b/returnn/config.py
@@ -109,6 +109,7 @@ class ReturnnConfig:
             f.write(self.serialize())
 
     def serialize(self):
+        self.check_consistency()
         config = self.config
         config.update(self.post_config)
 
@@ -185,6 +186,15 @@ class ReturnnConfig:
         if inspect.isclass(code):
             return inspect.getsource(code)
         raise RuntimeError("Could not serialize %s" % code)
+
+    def check_consistency(self):
+        """
+        check that there is no config key overwritten by post_config
+        """
+        for key in self.config:
+            assert (
+                key not in self.post_config
+            ), "%s in post_config would overwrite existing entry in config"
 
     def _sis_hash(self):
         h = {
--- a/returnn/config.py
+++ b/returnn/config.py
@@ -192,9 +192,9 @@ class ReturnnConfig:
         check that there is no config key overwritten by post_config
         """
         for key in self.config:
-            assert (
-                key not in self.post_config
-            ), "%s in post_config would overwrite existing entry in config"
+            assert key not in self.post_config, (
+                "%s in post_config would overwrite existing entry in config" % key
+            )
 
     def _sis_hash(self):
         h = {
--- a/returnn/search.py
+++ b/returnn/search.py
@@ -160,6 +160,7 @@ class ReturnnSearchJob(Job):
         res = copy.deepcopy(returnn_config)
         res.config = config
         res.post_config = post_config
+        res.check_consistency()
 
         return res
 
--- a/returnn/training.py
+++ b/returnn/training.py
@@ -367,6 +367,7 @@ class ReturnnTrainingJob(Job):
 
         res.config = config
         res.post_config = post_config
+        res.check_consisteny()
 
         return res
 


returnn/config.py
```diff
@@ -186,6 +187,15 @@ def __parse_python(self, code, name=None):
...
         raise RuntimeError("Could not serialize %s" % code)
 
+    def check_consistency(self):
+        """
+        check that there is no config key overwritten by post_config
+        """
+        for key in self.config:
+            assert (
+                key not in self.post_config
+            ), "%s in post_config would overwrite existing entry in config"
```
mmz33, line 197: missing `% key`

mmz33 (changes requested):

mmz33 (approved):

returnn/search.py
```diff
@@ -160,6 +160,7 @@ def create_returnn_config(self, returnn_config, log_verbosity, device):
         res = copy.deepcopy(returnn_config)
         res.config = config
         res.post_config = post_config
+        res.check_consistency()
```
mattiadg, line 163: Wouldn't it be better to call `check_consistency()` only internally in ReturnnConfig when it's constructed and when config or post_config are modified?
It's not a client's responsibility to check that everything is correct

returnn/training.py
```diff
@@ -367,6 +367,7 @@ def create_returnn_config(
 
         res.config = config
         res.post_config = post_config
+        res.check_consisteny()
```
mattiadg, line 370: same as above

mattiadg (commented):
The idea to prevent bugs by explicit checking is good, but the client code shouldn't know about it, or the same errors are introduced by simply not calling the new function

JackTemaki:
> The idea to prevent bugs by explicit checking is good, but the client code shouldn't know about it, or the same errors are introduced by simply not calling the new function

it is also checked in the config itself, during "serialize". The difference is here, that it then will crash during a task  ( e.g. create_files) and not during the graph construction....

mattiadg:
> > The idea to prevent bugs by explicit checking is good, but the client code shouldn't know about it, or the same errors are introduced by simply not calling the new function
> 
> it is also checked in the config itself, during "serialize". The difference is here, that it then will crash during a task ( e.g. create_files) and not during the graph construction....

wouldn't it be more natural to check it when the object is constructed and when either `config` or `post_config` are modified?
of course  this would require some further modifications like making them properties or adding a new method to change them.
So, it requires to modify some code but it doesn't delegate responsibilities to the client code.

JackTemaki:
> > > The idea to prevent bugs by explicit checking is good, but the client code shouldn't know about it, or the same errors are introduced by simply not calling the new function
> > 
> > 
> > it is also checked in the config itself, during "serialize". The difference is here, that it then will crash during a task ( e.g. create_files) and not during the graph construction....
> 
> wouldn't it be more natural to check it when the object is constructed and when either `config` or `post_config` are modified?
> of course this would require some further modifications like making them properties or adding a new method to change them.
> So, it requires to modify some code but it doesn't delegate responsibilities to the cl

JackTemaki:
> > wouldn't it be more natural to check it when the object is constructed and when either `config` or `post_config` are modified?
> > of course this would require some further modifications like making them properties or adding a new method to change them.
> > So, it requires to modify some code but it doesn't delegate responsibilities to the client code.

(Sorry for the misclick)

but if you just "alter" the post_config, the class can not detect this easily, or am I missing something?

mattiadg:
> but if you just "alter" the post_config, the class can not detect this easily, or am I missing something?

That's why I say that it would require some additional changes, like making it a property.

curufinwe (commented):
I'm not sure if this is really a good idea. Sometimes you might have a legacy config and you want to update it in post_config without breaking the hash. I did this for RASR configs once or twice.

michelwi:
> I'm not sure if this is really a good idea. Sometimes you might have a legacy config and you want to update it in post_config without breaking the hash. I did this for RASR configs once or twice

I don't understand.. If a parameter changes the output then we would want to rerun the jobs and change the hashes. and if it doesn't, then no need to change it in already finished jobs.?

curufinwe:
Well, sometimes you might have parameters in there that determine the logging or the API changed and you don't want to rerun the previous jobs. It's a hacky way of preserving your jobs.

returnn/search.py
curufinwe: This is a bit too much, if we construct dozends of configs with hundreds of keys that will slow down config loading. So if it's done this would be a good place to do it. But I dont think it should be done at all.

christophmluscher (approved):
the questions raised about where to call `check_consistency` should be addressed before merging

JackTemaki:
> It's a hacky way of preserving your jobs.
@curufinwe what do you think about adding this as a parameter which is true by default but can be disabled? That way, if you are really sure about what you are doing, you can still override it.

michelwi:
> sometimes you might have parameters in there that determine the logging 

Then it should be in post_config anyway

> or the API changed 

In this case see below

> > It's a hacky way of preserving your jobs.
> 
> @curufinwe what do you think about adding this as a parameter which is true by default but can be disabled? That way, if you are really sure about what you are doing, you can still override it.

I would say we do not need a special parameter for that. If one is really sure about keeping the jobs upon a change of the hash, then the sisyphus migrate_graph feature should be used.

michelwi (approved):

curufinwe (changes requested):
Add also to SprintTraining

JackTemaki:
> Add also to SprintTraining

The check is already part of the SprintTraining via the super() call in the init.

curufinwe (approved):

PR #65, opened 2021-06-24
Set bellman-ford as the default best path algorithm

Marvin84:
Differently to Dijkstra, the Bellmann-Ford algorithm operates on graphs with negative edges. This is important for the lattice to ctm conversion job, when we have negative scores.

### DIFF
--- a/recognition/conversion.py
+++ b/recognition/conversion.py
@@ -19,7 +19,7 @@ class LatticeToCtmJob(rasr.RasrCommand, Job):
         parallelize=False,
         encoding="utf-8",
         fill_empty_segments=False,
-        best_path_algo=None,  # defaults to dijkstra, which does not work on graphs with negative edge-weights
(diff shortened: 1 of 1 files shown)


recognition/conversion.py
```diff
@@ -19,7 +19,7 @@ def __init__(
         parallelize=False,
         encoding="utf-8",
         fill_empty_segments=False,
-        best_path_algo=None,  # defaults to dijkstra, which does not work on graphs with negative edge-weights
+        best_path_algo="bellman-ford",  # default is dijkstra, which does not work on graphs with negative edge-weights
```
curufinwe, line 22: please remove the comment, it is not relevant anymore ("default" might confuse the reader as the default in this job is now bellman-ford)

curufinwe (changes requested):

michelwi (approved):

curufinwe (approved):

PR #68, opened 2021-06-25
Replaced rasr.PathWithPrefixFlowAttribute with the plain path

### LINKED ISSUE #67: EstimateLDAMatrixJob wrong usage of rasr.PathWithPrefixFlowAttribute?
This is the error that led me to find the problem in #66.
In https://github.com/rwth-i6/i6_core/blob/88a4f225a9e4609a76d5e33c247d5ced103d9079/lda/estimate.py#L341 `between_class_scatter_matrix_file` and `within_class_scatter_matrix_file` are assigned as 
```
 config.acoustic_model_trainer.lda_estimator.between_class_scatter_matrix_file = rasr.PathWithPrefixFlowAttribute(
            "xml:", between_class_scatter_matrix
        )
        config.acoustic_model_trainer.lda_estimator.within_class_scatter_matrix_file = (
            rasr.PathWithPrefixFlowAttribute("xml:", within_class_scatter_matrix)
        )
```
which are then printed in the config as python objects and not strings. Also, these files are not xml.

mattiadg:
fixes #67

### DIFF
--- a/lda/estimate.py
+++ b/lda/estimate.py
@@ -338,11 +338,11 @@ class EstimateLDAMatrixJob(rasr.RasrCommand, Job):
         )
         config.acoustic_model_trainer.lda_estimator.output_precision = 20
 
-        config.acoustic_model_trainer.lda_estimator.between_class_scatter_matrix_file = rasr.PathWithPrefixFlowAttribute(
-            "xml:", between_class_scatter_matrix
+        config.acoustic_model_trainer.lda_estimator.between_class_scatter_matrix_file = (
+            between_class_scatter_matrix
         )
         config.acoustic_model_trainer.lda_estimator.within_class_scatter_matrix_file = (
-            rasr.PathWithPrefixFlowAttribute("xml:", within_class_scatter_matrix)
+            within_class_scatter_matrix
         )
         config.acoustic_model_trainer.lda_estimator.projector_matrix_file = (
             "xml:lda.matrix"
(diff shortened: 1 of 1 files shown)


michelwi (changes requested):
I think you still need to specify the file format with the prefix `xml:`

curufinwe:
Not needed anymore due to #69

mattiadg:
Not really, the code as it is still print the object in the config instead of the path
```
[acoustic-model-trainer.lda-estimator]
between-class-scatter-matrix-file = <i6_core.rasr.flow.PathWithPrefixFlowAttribute object at 0x7efe568d63c8>
output-precision                  = 20
projector-matrix-file             = xml:lda.matrix
reduced-dimesion                  = 40
results.channel                   = output-channel
within-class-scatter-matrix-file  = <i6_core.rasr.flow.PathWithPrefixFlowAttribute object at 0x7efe568d6470>
```
At the very least, we should call `get`, which is not done now.
Also, are you sure that we should use a FlowAttribute in the config?

JackTemaki:
@curufinwe there are two cases for this scenario:
- the xml prefix is not needed, and we can add a "path" directly 
- the xml prefix is needed, then we can use the already built-in `DelayedAdd` or `DelayedFormat` of Sisyphus, instead of introducing a new class like `PathWithPrefixFlowAttribute`.

curufinwe:
I see. Then let's just use the DelayedFormat object and completely remove PathWithPrefixFlowAttribute

michelwi (approved):

curufinwe (approved):

PR #82, opened 2021-07-06
add DownloadJob

JackTemaki:
I added this job to download arbitrary files (e.g. the 4-gram.arpa.gz file)

### DIFF
--- a/tools/download.py
+++ b/tools/download.py
@@ -36,7 +36,7 @@ class DownloadJob(Job):
         yield Task("run", mini_task=True)
 
     def run(self):
-        sp.check_call(["wget", self.url])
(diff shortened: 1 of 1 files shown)


tools/download.py
```diff
@@ -0,0 +1,48 @@
...
+        self.out_file = self.output_path(self.target_filename)
+
+        if self.target_filename:
+            self.set_vis_name(target_filename)
+
+    def tasks(self):
+        yield Task("run", mini_task=True)
+
+    def run(self):
+        sp.check_call(["wget", self.url])
```
curufinwe, line 39: target_filename is not considered here

curufinwe (changes requested):

curufinwe (approved):

Issue #85, opened 2021-07-07
create_returnn_config can create invalid config


mattiadg:
In the referenced lines. if `config` has keys that are present in the default `post_config`, then the new `ReturnnConfig` won't pass the `check_consistency` assert.

I understand that maybe those keys are supposed to be in the post_config, but maybe it's better to have a way not to change the hash of existing jobs if the keys were in the config.

https://github.com/rwth-i6/i6_core/blob/9c5d66010b84dd18b3631e473c6bb2953841616f/returnn/training.py#L365-L367

michelwi:
maybe we could extend the line 
https://github.com/rwth-i6/i6_core/blob/9c5d66010b84dd18b3631e473c6bb2953841616f/returnn/config.py#L196
to

```
assert key not in self.post_config or self.post_config[key] == self.config[key], {
```

JackTemaki:
> maybe we could extend the line
> 
> https://github.com/rwth-i6/i6_core/blob/9c5d66010b84dd18b3631e473c6bb2953841616f/returnn/config.py#L196
> 
> to
> 
> ```
> assert key not in self.post_config or self.post_config[key] == self.config[key], {
> ```

Hmm... I am not really happy with this, as you then can get different (inconsistent) hashes for the same job, just because you repeated the same setting in `config`...

michelwi:
Well, the other solution is to say
"@mattiadg if your `config` entry is in the default `post_config`, then it will not affect the outcome of the job and should not have been in `config` in the first place. Would you please clean up your `config` and then use the `sisyphus` `migrate_graph` feature to fix the hashes."

JackTemaki:
> Well, the other solution is to say

To clarify how that would look like:

- use the "old/last working" version of `i6_core` without the consistency check
- open the Sisyphus console and use `tk.export_graph(<graph_filename>)` to export the graph to a file
- close and update i6_core to the newest version
- open Sisyphus console again
- use `tk.migrate_graph(<graph_filename>) to do a "dryrun" import of the previous jobs
- check if the jobs are found and mapped correctly
- run `tk.migrate_graph` with link mode or so (move or copy can be valid options as well...)

curufinwe:
@mattiadg is your issue resolved?

mattiadg:
Not really, @JackTemaki told me that I cannot do that process because I don't have a `main()` in my config.
In the end I started a new process anyway.

PR #86, opened 2021-07-07
Fix consistency of generated ReturnnConfig

mattiadg:
Attempt of solution to #85


JackTemaki:
This PR should not be needed, see discussion in #85

PR #88, opened 2021-07-07
Add ExtractVocabularyFromCorpusJob

curufinwe:


michelwi (approved):

corpus/stats.py
```diff
@@ -46,3 +47,31 @@ def run(self):
...
+
+    def run(self):
+        c = libcorpus.Corpus()
+        c.load(self.bliss_corpus.get_path())
+
+        words = set()
+        for s in c.segments():
+            words.update(set(s.orth.strip().split()))
+
+        with open(self.out_vocabulary.get_path(), "wt") as f:
```
JackTemaki, line 76: I thought we only wanted to use `uopen` everywhere...

JackTemaki (approved):

mattiadg (commented):
Don't we want a vocabulary with word frequency?

curufinwe:
This is just a utility job for me to know what words appear in a corpus, but we can also add frequencies here.

Issue #91, opened 2021-07-13
Cached lattices in OptimizeAMandLMScaleJob


vieting:
#80 removed the hard-coded reference to the cache manager from the `OptimizeAMandLMScaleJob`. Now, if the lattice path is a `Path` object with `cached=True`, we end up with a string as expected in `self.run_script`. However, the `subprocess.check_call` does not seem to handle it correctly, I assume because of the space in `cf /u/...`.

I end up with the error
>   Archive file "`cf" does not exist.

JackTemaki:
close with #92

Issue #38, opened 2021-05-31
Beginning of stable period


curufinwe:
We wanted to be finished with setup/hash breaking changes by end of May. That's today. Does anyone of you have more breaking changes that you'd like to integrate? If so, should we agree on a timetable until when these changes need to be done?

mmz33:
Yes I still need to fix the BPE train and apply jobs. We agreed to use a fixed commit hash for `CloneGitRepository`. I can do a PR for that soon.

JackTemaki:
Yes, also for me there are more jobs incoming. I would say (with looking at the upcoming holidays) we need at least 2 more weeks.

christophmluscher:
@JackTemaki new jobs or refactoring?
new jobs should not influence beginning of stable period in my opinion

JackTemaki:
There is something that needs to be fixed for the RETURNN hashes, so definitely critical.

christophmluscher:
the 2 weeks factors in ICASSP?

JackTemaki:
> the 2 weeks factors in ICASSP?

Ah right...

curufinwe:
We could also do this in two stages:
1. no breaking hashes but may need to refactor setups
2. setups must run without modification

christophmluscher:
I think the points raised here, have been addressed. Are they completed now?
Or are there still hash breaking changes missing?

mmz33:
From my side, I think all jobs needed to train/recognize attention models at least are merged except for the BPE jobs which are in PR currently.

curufinwe:
Actually I also want to make a few changes to the recognition job. Would June 18th as a final stop on hash breaking changes be OK? (1 week after ICASSP).

JackTemaki:
As @mattiadg  and @papar22 started using this actively and pointed out some bugs, I propose that we extend the duration by one more week. @curufinwe what do you think?

curufinwe:
OK

curufinwe:
The time has come and nobody asked for another extension. Thus we can no go to a more stable regime where changes must not break existing hashes. Thanks to everyone contributing to the refactoring effort!

PR #92, opened 2021-07-13
Fix for cached lattices in OptimizeAMandLMScaleJob

vieting:
As described in #91, there are currently problems if a cached lattice path is used in the `OptimizeAMandLMScaleJob`. This is a possible fix by writing the path directly into the config, where it should be handled correctly.

### DIFF
--- a/recognition/optimize_parameters.py
+++ b/recognition/optimize_parameters.py
@@ -64,7 +64,6 @@ class OptimizeAMandLMScaleJob(rasr.RasrCommand, Job):
         lm_scale = self.initial_lm_scale
 
         result_cache = collections.OrderedDict()
-        lattice_cache = self.lattice_cache
 
         def calc_wer(params):
             def clip_float(f):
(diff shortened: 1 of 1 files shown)


recognition/optimize_parameters.py
```diff
@@ -147,7 +145,7 @@ def cleanup_before_run(self, cmd, retry, *args):
         util.backup_if_exists("lm_and_state_tree.log")
 
     @classmethod
-    def create_config(cls, crp, extra_config, extra_post_config, **kwargs):
+    def create_config(cls, crp, extra_config, extra_post_config, lattice_cache, **kwargs):
```
curufinwe, line 148: order of parameters here should match order of parameters in __init__

curufinwe (changes requested):
please format code with black

vieting:
> please format code with black
Yes, sorry. Done.

curufinwe (approved):

vieting:
Btw, this is hash-breaking. So not sure how we handle this after the beginning of the stable period (#38).

PR #94, opened 2021-07-14
Add ReturnnComputePriorJob and ReturnnRasrComputePriorJob

michelwi:
Adds Jobs to compute the state prior of a hybrid model based in the underlying neural network.

### DIFF
--- a/returnn/extract_prior.py
+++ b/returnn/extract_prior.py
@@ -4,18 +4,17 @@ __all__ = [
     "ReturnnRasrComputePriorJob",
 ]
 
+import copy
 import math
 import os
 import subprocess as sp
 
-import copy
 import h5py
 import numpy as np
 from sisyphus import *
 
 import i6_core.rasr as rasr
 import i6_core.util as util
-
 from i6_core.returnn.config import ReturnnConfig
 from i6_core.returnn.rasr_training import ReturnnRasrTrainingJob
 from i6_core.returnn.training import ReturnnTrainingJob
--- a/returnn/extract_prior.py
+++ b/returnn/extract_prior.py
@@ -4,13 +4,13 @@ __all__ = [
     "ReturnnRasrComputePriorJob",
 ]
 
-import copy
-import h5py
 import math
-import numpy as np
 import os
 import subprocess as sp
 
+import copy
+import h5py
+import numpy as np
 from sisyphus import *
 
 import i6_core.rasr as rasr
@@ -120,13 +120,14 @@ class ReturnnComputePriorJob(Job):
             returnn_root if returnn_root is not None else gs.RETURNN_ROOT
         )
 
+        self.returnn_config = ReturnnComputePriorJob.create_returnn_config(**kwargs)
+
         self.out_returnn_config_file = self.output_path("returnn.config")
 
         self.out_prior_txt_file = self.output_path("prior.txt")
         self.out_prior_xml_file = self.output_path("prior.xml")
         self.out_prior_png_file = self.output_path("prior.png")
 
-        self.returnn_config = ReturnnComputePriorJob.create_returnn_config(**kwargs)
         self.returnn_config.post_config["output_file"] = self.out_prior_txt_file
 
         self.rqmt = {
@@ -145,12 +146,7 @@ class ReturnnComputePriorJob(Job):
         config = self.returnn_config
         config.write(self.out_returnn_config_file.get_path())
 
-        cmd = [
-            tk.uncached_path(self.returnn_python_exe),
-            os.path.join(tk.uncached_path(self.returnn_root), "rnn.py"),
-            self.out_returnn_config_file.get_path(),
-        ]
-
+        cmd = _get_run_cmd()
         util.create_executable("rnn.sh", cmd)
 
         # check here if model actually exists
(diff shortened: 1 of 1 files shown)


returnn/extract_prior.py
```diff
@@ -58,3 +71,333 @@ def run(self):
...
+        *,  # args below are keyword only
+        log_verbosity=3,
+        device="gpu",
+        time_rqmt=4,
+        mem_rqmt=4,
+        cpu_rqmt=1,
+        returnn_python_exe=None,
+        returnn_root=None,
+        # these are new parameters
+        num_classes=None,
```
JackTemaki, line 278: Are theses parameters all relevant for the PriorJob, or is it fine to leave them with their default values from the ReturnnRasrTrainingJob?

returnn/extract_prior.py
```diff
@@ -58,3 +71,333 @@ def run(self):
...
+    def hash(cls, kwargs):
+        d = {
+            "returnn_config": ReturnnComputePriorJob.create_returnn_config(**kwargs),
+            "returnn_python_exe": kwargs["returnn_python_exe"],
+            "returnn_root": kwargs["returnn_root"],
+        }
+        return super().hash(d)
+
+
+class ReturnnRasrComputePriorJob(ReturnnComputePriorJob, ReturnnRasrTrainingJob):
```
JackTemaki, line 259: A minimal docstring would be nice at least, something like "Prior job that uses a Rasr Dataset" or so....

returnn/extract_prior.py
```diff
@@ -58,3 +71,333 @@ def run(self):
...
+            returnn_root if returnn_root is not None else gs.RETURNN_ROOT
+        )
+
+        self.out_returnn_config_file = self.output_path("returnn.config")
+
+        self.out_prior_txt_file = self.output_path("prior.txt")
+        self.out_prior_xml_file = self.output_path("prior.xml")
+        self.out_prior_png_file = self.output_path("prior.png")
+
+        self.returnn_config = ReturnnComputePriorJob.create_returnn_config(**kwargs)
```
JackTemaki, line 129: This should be before defining the outputs

JackTemaki (changes requested):
I think it looks mostly good, I like that you re-used the existing classes.

returnn/extract_prior.py
michelwi: Some of these parameters we afaik never set (e.g. `disregarded_classes`, `buffer_size`), but they are settable in the train job and I'd much rather not have them inconsistent.

Others (I mainly think of `partition_epochs`) make sense to set explicitly different than in the corresponding train Job

returnn/extract_prior.py
```diff
@@ -58,3 +71,333 @@ def run(self):
...
+
+        config.update(copy.deepcopy(original_config))  # update with the original config
+
+        # override always
+        config["task"] = "compute_priors"
+
+        if prior_data is not None:
+            config["train"] = {
+                **original_config["train"].copy(),
+                **prior_data,
```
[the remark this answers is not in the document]
michelwi: here I was trying to be consistent with
https://github.com/rwth-i6/i6_core/blob/4ba665186b169733319177d08fac79c832e3caf2/returnn/search.py#L156-L162

But I see in favor of your argument, that the training set is usually set and for a different purpose than prior computation, whereas the `search_data` would always be intentionally set for search.

returnn/extract_prior.py
```diff
@@ -58,3 +71,333 @@ def run(self):
...
+        """
+        assert device in ["gpu", "cpu"]
+        original_config = returnn_config.config
+        assert "network" in original_config
+
+        config = {
+            "load": model_checkpoint.ckpt_path,
+        }
+
+        config.update(copy.deepcopy(original_config))  # update with the original config
```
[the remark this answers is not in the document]
michelwi: here I was being consistent with 
https://github.com/rwth-i6/i6_core/blob/4ba665186b169733319177d08fac79c832e3caf2/returnn/search.py#L143-L150

But we can do it the other way.

returnn/extract_prior.py
```diff
@@ -1,12 +1,25 @@
-__all__ = ["ExtractPriorFromHDF5Job"]
+__all__ = [
+    "ExtractPriorFromHDF5Job",
+    "ReturnnComputePriorJob",
+    "ReturnnRasrComputePriorJob",
+]
 
+import copy
+import h5py
```
curufinwe, line 8: h5py import should be together with other non-python-standard library imports. See https://github.com/rwth-i6/i6_core/blob/main/CONTRIBUTUNG.md (Import order). Also check other imports.

returnn/extract_prior.py
curufinwe: If you are looking for consistency with the train job my suggestion would be to add **kwargs to the constructor and pass everything in kwargs to the train job. This way you also get forward compatibility.

curufinwe (changes requested):

returnn/extract_prior.py
michelwi: from https://github.com/rwth-i6/i6_core/blob/cef8a57cf0983b810c85c8f5507cc05a34ba4d19/returnn/rasr_training.py#L32 I thought we would get problems with Sisyphus when using the **kwargs here. Or is it ok in this case?

curufinwe: Good question, I don't remember why passing **kwargs is/(was?) a problem at that point.

returnn/extract_prior.py
```diff
@@ -1,12 +1,25 @@
...
+    "ReturnnComputePriorJob",
+    "ReturnnRasrComputePriorJob",
+]
 
 import math
+import os
+import subprocess as sp
 
-import numpy as np
+import copy
```
curufinwe, line 11: copy is also part of std-lib

curufinwe (approved):

Issue #93, opened 2021-07-14
ReturnnTrainingJob hashes with function in python_prolog


vieting:
Assume I am running a `ReturnnTrainingJob` and the `ReturnnConfig` has a `python_prolog` which is a dict which contains a function, e.g. in my case the function `get_network()` which is defined in a helper file. This function is serialized to be written in the config file, however, not for creating the hash. So if I change something in the function, the hash remains unchanged while the behavior in RETURNN would obviously be very different.

From my perspective, it would be intuitive if the hash changes when changing the function. This would have the drawback that if you add a new argument to the function which does not change the behavior when the default value is used, still all hashes change. But for this, you might consider using `python_prolog_hash`.

What do you guys think?

michelwi:
But in https://github.com/rwth-i6/i6_core/blob/e0971f14f43a1d73ca0fa4965ff775fcd72cb16d/returnn/config.py#L95-L97 the `python_prolog_hash` is set to the `python_prolog` so it should be considered for hashing. 

Ah, but if you hand a function object instead of the actual source code, then maybe this object serializes (for the purpose of hashing) always the same even if you change the code.

So we should instead use 
```
        self.python_prolog_hash = (
            python_prolog_hash if python_prolog_hash is not None else self.__parse_python(python_prolog)
        )
```
which would keep the hash of all text-based `python_prolog`'s intact.

vieting:
Exactly, something like this would be my suggestion. However, I wasn't sure if the current behavior was intended. (And this change would break some hashes.)

michelwi:
I think this behavior was not intended. And it should imho be consistent for text-based and function-object-based prologs.

So I'd say go forward with a PR (and don't forget the epilog)

JackTemaki:
> I think this behavior was not intended

The problem is this behavior was indeed intended, and nobody complained about it. So the idea was to either change the function name (= new hash) or do it manually.

JackTemaki:
it is true that nobody changed the "status: unstable" yet (I guess I should do that), but we did not want to break the hashes anymore, which this change will definitely do.... I think for the Optimize Job this was an exception, because there was a crash in certain cases and it comes usually at the end of pipelines...

curufinwe:
It was not intended from my side. I assumed that sisyphus would handle hashing of functions correctly.

curufinwe:
[...]

PR #95, opened 2021-07-14
allow returnn configs with get_network function

vieting:
RETURNN allows using configs which define a function `get_network()` instead of the usual `network`. This should also be allowed in the training job.

### DIFF
--- a/returnn/training.py
+++ b/returnn/training.py
@@ -340,7 +340,10 @@ class ReturnnTrainingJob(Job):
         **kwargs,
     ):
         assert device in ["gpu", "cpu"]
-        assert "network" in returnn_config.config
+        assert "network" in returnn_config.config or "get_network" in [
+            getattr(obj, "__name__", None)
+            for obj in returnn_config.python_prolog.values()
+        ]
 
         res = copy.deepcopy(returnn_config)
 


michelwi (changes requested):
No, this should not be allowed, because it directly contradicts our conclusion in #93 that a function name should be unique to what the function does.

So either the function code is hashed or we cannot allow to define the network via a function object.

vieting:
Well, `network = get_network()` is allowed but leads to exactly the same problem.

vieting:
So if I use functions, I have to take care of the hash manually anyway.

michelwi:
There definitively should never be the case that two different trainings with different network structures have the same hash.

> I have to take care of the hash manually anyway.

this is not how it should be. Then we should instead add a check, that the function `get_network()`, if used, is added via a string and not as a function object, such that it is correctly hashed.

JackTemaki:
This is actually a good point that I did not thing of yet.

In the end we wanted to have a way where we define a "dynamic" (with respect to the epoch) network object in our Sisyphus setup, but then print each "fixed" network into the config file. But for this we actually need to use `get_network()`. I will think of a solution for this, as I will need it anyway for the new TTS construction that my Hiwi is currently implementing.

@vieting if you can wait a little bit, we can discuss this together, and then do a new PR about this.

vieting:
>There definitively should never be the case that two different trainings with different network structures have the same hash.

I agree, but this contradicts the outcome of the discussion in #93.

> @vieting if you can wait a little bit, we can discuss this together, and then do a new PR about this.

Sure, let's do this :)

michelwi:
> > There definitively should never be the case that two different trainings with different network structures have the same hash.

> I agree, but this contradicts the outcome of the discussion in #93.

Then we should reopen the discussion or find another solution.

returnn/training.py
```diff
@@ -340,7 +340,10 @@ def create_returnn_config(
         **kwargs,
     ):
         assert device in ["gpu", "cpu"]
-        assert "network" in returnn_config.config
+        assert "network" in returnn_config.config or "get_network" in [
+            getattr(obj, "__name__", None)
+            for obj in returnn_config.python_prolog.values()
```
curufinwe, line 345: also check in python_epilog

vieting:
I discussed this with @JackTemaki and he'll create a PR with a different solution instead.

PR #98, opened 2021-07-19
Add LJSpeech corpus

JackTemaki:
Add a download and a bliss creation job for the LJSpeech corpus.

### DIFF
--- a/datasets/ljspeech.py
+++ b/datasets/ljspeech.py
@@ -1,0 +1,96 @@
+__all__ = [
+    "DownloadLJSpeechCorpusJob",
+]
+
+import os
+import shutil
+import subprocess
+import wave
+
+from sisyphus import *
+
+from i6_core.lib import corpus
+from i6_core.util import uopen, check_file_checksum
+
+
+class DownloadLJSpeechCorpusJob(Job):
+    """
+    Downloads, checks and extracts the LJSpeech corpus.
+    """
+
+    def __init__(self):
+        self.out_ljspeech_folder = self.output_path("LJSpeech-1.1")
+        self.out_audio_folder = self.output_path("LJSpeech-1.1/wavs")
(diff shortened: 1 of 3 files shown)


util.py
```diff
@@ -252,3 +252,16 @@ def create_executable(filename, command):
         | stat.S_IXGRP
         | stat.S_IXOTH,
     )
+
+
+def check_file_checksum(filename, reference_checksum):
```
curufinwe, line 257: if sha256 is hardcoded in the function it should be added to the name (for better readability). i.e. `check_file_sha256sum`

curufinwe (changes requested):

datasets/ljspeech.py
```diff
@@ -0,0 +1,96 @@
+__all__ = [
+    "DownloadLJSpeechCorpusJob",
+]
```
Atticus1806, line 3: Does it make sense to just hase the Download Job here? Maybe also add the other job.

datasets/ljspeech.py
```diff
@@ -0,0 +1,94 @@
...
+        self.out_ljspeech_folder = self.output_path("LJSpeech-1.1")
+        self.out_audio_folder = self.output_path("LJSpeech-1.1/wavs")
+        self.out_metadata = self.output_path("LJSpeech-1.1/metadata.csv")
+
+    def tasks(self):
+        yield Task("run", mini_task=True)
+
+    def run(self):
+        subprocess.check_call(
+            ["wget", "https://data.keithito.com/data/speech/LJSpeech-1.1.tar.bz2"]
```
michelwi, line 29: Do we have the option to download to a node-local temp location instead of the work folder?

curufinwe: In the code right now not, but in principle I see no problem implementing it that way unless we want to be able to resume downloads and add the corresponding flags to wget.

PR #99, opened 2021-07-19
Add LexiconFromTextFileJob

JackTemaki:
This is my try to replicate the behavior of the "Pavel"-style Lexicon for LibriSpeech, which means kind of emulating the behavior of very ancient RASR tools that are no longer supported.

As the lexicon format is a quite critical decision for the future I added all of you as reviewers (I could not find Simon and Zijian here)

You can review the files under `/u/rossenbach/stuff/pavel_ls_lexicon_test`:
 - `librispeech-lexicon.txt` the input LibriSpeech lexcion as downloaded from the webpage
 - `orig.xml` the lexicon used until now
 - `test.xml` the lexicon produced by the new job
 
 The documentation for the Bliss-Lexicon can be found here:
 https://www-i6.informatik.rwth-aachen.de/rwth-asr/manual/index.php/Lexicon

### DIFF
--- a/lexicon/conversion.py
+++ b/lexicon/conversion.py
@@ -142,24 +142,24 @@ class LexiconUniqueOrthJob(Job):
 
 class LexiconFromTextFileJob(Job):
     """
-    Create a bliss lexicon from space or tab separated files.
+    Create a bliss lexicon from a regular text file, where each line contains:
+    < WORD > < PHONEME1 > < PHONEME2 > ...
+    separated by tabs or spaces.
+    The lemmata will be added in the order they appear in the text file,
+    the phonemes will be sorted alphabetically.
+    No special lemmata or phonemes are added.
 
     As the splitting is taken from RASR and not fully tested,
     it might not work in all cases so do not use this job
-    without checking the output manually on new lexicas.
+    without checking the output manually on new lexica.
     """
 
-    def __init__(self, text_file, add_unknown, add_noise, compressed=True):
+    def __init__(self, text_file, compressed=True):
         """
-
         :param Path text_file:
-        :param bool add_unknown: add [UNKNOWN] lemma
-        :param bool add_noise: add [NOISE] lemma
         :param compressed: save as .xml.gz
         """
         self.text_file = text_file
-        self.add_unknown = add_unknown
-        self.add_noise = add_noise
 
         self.out_bliss_lexicon = self.output_path(
             "lexicon.xml.gz" if compressed else "lexicon.xml"
@@ -170,47 +170,6 @@ class LexiconFromTextFileJob(Job):
 
     def run(self):
         lex = lexicon.Lexicon()
-
-        lex.add_lemma(
-            lexicon.Lemma(
-                orth=["[SILENCE]"],
-                phon=["[SILENCE]"],
-                synt=[""],
-                special="silence",
-                eval=[""],
-            )
-        )
-
-        lex.add_lemma(
-            lexicon.Lemma(
-                orth=["[SENTENCE-BEGIN]"], synt=[["<s>"]], special="sentence-begin"
-            )
-        )
-        lex.add_lemma(
-            lexicon.Lemma(
-                orth=["[SENTENCE-END]"], synt=[["</s>"]], special="sentence-end"
-            )
-        )
-
-        if self.add_unknown:
-            lex.add_lemma(
-                lexicon.Lemma(
-                    orth=["[UNKNOWN]"],
-                    phon=["[UNKNOWN]"],
-                    synt=[["<UNK>"]],
-                    special="unknown",
-                )
-            )
-            lex.add_phoneme("[UNKNOWN]", variation="none")
-        if self.add_noise:
-            lex.add_lemma(
-                lexicon.Lemma(
-                    orth=["[NOISE]"],
-                    phon=["[NOISE]"],
-                    synt=[["<NOISE>"]],
-                    special="unknown",
-                )
-            )
 
         phonemes = set()
         last_lemma = None
@@ -234,14 +193,6 @@ class LexiconFromTextFileJob(Job):
                 else:
                     lex.add_lemma(lemma)
                     last_lemma = lemma
-
-        if self.add_noise:
-            lex.add_phoneme("[NOISE]", variation="none")
-
-        lex.add_phoneme("[SILENCE]", variation="none")
-
-        if self.add_unknown:
-            lex.add_phoneme("[UNKNOWN]", variation="none")
 
         for phoneme in sorted(phonemes):
             lex.add_phoneme(phoneme)
--- a/lexicon/conversion.py
+++ b/lexicon/conversion.py
@@ -9,7 +9,7 @@ from sisyphus import *
 Path = setup_path(__package__)
 
 import i6_core.lib.lexicon as lexicon
-from i6_core.util import uopen
+from i6_core.util import uopen, write_xml
 
 
 class LexiconToWordListJob(Job):
@@ -185,8 +185,7 @@ class LexiconFromTextFileJob(Job):
                 ]
                 for phon_variant in phon_variants:
                     for phon in phon_variant:
-                        if phon not in phonemes:
-                            phonemes.add(phon)
+                        phonemes.add(phon)
                 phon = [" ".join(v) for v in phon_variants]
                 lemma = lexicon.Lemma(orth=[orth], phon=phon)
                 if last_lemma and lemma.orth[0] == last_lemma.orth[0]:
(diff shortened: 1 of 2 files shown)


lexicon/conversion.py
```diff
@@ -140,6 +140,118 @@ def run(self):
             lexicon_file.write(ET.tostring(root, "unicode"))
 
 
+class LexiconFromTextFileJob(Job):
+    """
+    Create a bliss lexicon from space or tab separated files.
```
michelwi, line 145: Is this maybe the same as https://github.com/rwth-i6/i6_core/blob/cef8a57cf0983b810c85c8f5507cc05a34ba4d19/g2p/convert.py#L44

See one example g2p Lexicon here:
/work/asr4/michel/setups-data/librispeech/g2p/convert/BlissLexiconToG2PLexiconJob.AFMZQsJUiOIA/output/g2p.lexicon

JackTemaki: Ah no, it is definitely different!

lexicon/conversion.py
```diff
@@ -140,6 +140,118 @@ def run(self):
...
+
+    def tasks(self):
+        yield Task("run", mini_task=True)
+
+    def run(self):
+        lex = lexicon.Lexicon()
+
+        lex.add_lemma(
+            lexicon.Lemma(
+                orth=["[SILENCE]"],
```
michelwi, line 176: special lemma "silence" should have an empty orth tag. Otherwise no silence will be aligned between words if not explicitly transcribed.
```suggestion
                orth=["[SILENCE]", ""],
```

michelwi (changes requested):
Please add the empty orth for silence.

Another thing we might discuss:
If this job is meant to be general, should we then maybe make the silence/noise phoneme and lemmata more configurable? E.g. in the Quaero lexicon we have silence=sil and unknown=mul, noise=[Gfil,Gsta,Gint,Gspe] ; SWB also contains laughter, vocalized-noise

lexicon/conversion.py
vieting, line 145: Maybe describe the input format a little bit more, something like:
Each line contains
< WORD > < PHONEME1 > < PHONEME2 > ...
separated by tabs or spaces

lexicon/conversion.py
```diff
@@ -140,6 +140,118 @@ def run(self):
             lexicon_file.write(ET.tostring(root, "unicode"))
 
 
+class LexiconFromTextFileJob(Job):
+    """
+    Create a bliss lexicon from space or tab separated files.
+
+    As the splitting is taken from RASR and not fully tested,
+    it might not work in all cases so do not use this job
+    without checking the output manually on new lexicas.
```
vieting, line 149: lexica/lexicons

lexicon/conversion.py
vieting: Also the code below assumes that the entries are sorted alphabetically.

lexicon/conversion.py
```diff
@@ -140,6 +140,118 @@ def run(self):
...
+    """
+
+    def __init__(self, text_file, add_unknown, add_noise, compressed=True):
+        """
+
+        :param Path text_file:
+        :param bool add_unknown: add [UNKNOWN] lemma
+        :param bool add_noise: add [NOISE] lemma
```
vieting, line 157: There could also be other special lemmas, e.g., [LAUGHTER] or [VOCALIZEDNOISE]. We could have a list as input instead where you can specify which special lemmas to use.

lexicon/conversion.py
vieting: Not sure if my original comment is still visible for you, so I copy it again here:

>Maybe describe the input format a little bit more, something like:
Each line contains
< WORD > < PHONEME1 > < PHONEME2 > ...
separated by tabs or spaces

JackTemaki:
After your feedback I came to the conclusion that it might be better to split the job into two different ones, one that does the "conversion", and one that adds static phonemes/lemmata. I will update the code here, and then add a second PR.

lexicon/conversion.py
```diff
@@ -140,6 +140,69 @@ def run(self):
             lexicon_file.write(ET.tostring(root, "unicode"))
 
 
+class LexiconFromTextFileJob(Job):
+    """
+    Create a bliss lexicon from a regular text file, where each line contains:
+    < WORD > < PHONEME1 > < PHONEME2 > ...
```
vieting, line 146: We could also remove the spaces inside the brackets, I just had to insert them because otherwise nothing would show up in the comment here.

lexicon/conversion.py
```diff
@@ -140,6 +140,69 @@ def run(self):
...
+    separated by tabs or spaces.
+    The lemmata will be added in the order they appear in the text file,
```
vieting, line 148: Again: The code below assumes that the lines of the file are sorted alphabetically, right? (Because of the comparison with `last_lemma`.) If the file is not sorted alphabetically, there might be separate entries for the same lemma with different pronunciations which might or might not cause problems.

lexicon/conversion.py
```diff
@@ -140,6 +140,69 @@ def run(self):
...
+                # splitting is taken from RASR
+                # src/Tools/Bliss/blissLexiconLib.py#L185
+                s = line.split(None, 1)
+                orth = s[0].split("\\", 1)[0]
+                phon_variants = [
+                    tuple(p.split()) for p in s[1].split("\\") if p.strip()
+                ]
+                for phon_variant in phon_variants:
+                    for phon in phon_variant:
+                        if phon not in phonemes:
```
michelwi, line 187: This check is not strictly necessary as `set.add` already takes care of not adding duplicates.

lexicon/conversion.py
```diff
@@ -140,6 +140,69 @@ def run(self):
...
+                else:
+                    lex.add_lemma(lemma)
+                    last_lemma = lemma
+
+        for phoneme in sorted(phonemes):
+            lex.add_phoneme(phoneme)
+
+        with uopen(self.out_bliss_lexicon.get_path(), "wb") as lexicon_file:
+            s = ET.tostring(lex.to_xml(), "unicode")
+            pretty_s = minidom.parseString(s).toprettyxml(indent=" ", encoding="utf-8")
```
vieting, line 202: We have a function in `util` which you might consider for writing the xml file.

https://github.com/rwth-i6/i6_core/blob/main/util.py#L204

lexicon/conversion.py
```diff
@@ -140,6 +140,69 @@ def run(self):
...
+                if last_lemma and lemma.orth[0] == last_lemma.orth[0]:
+                    last_lemma.phon += phon
+                else:
+                    lex.add_lemma(lemma)
+                    last_lemma = lemma
+
+        for phoneme in sorted(phonemes):
+            lex.add_phoneme(phoneme)
+
+        with uopen(self.out_bliss_lexicon.get_path(), "wb") as lexicon_file:
```
michelwi, line 200: Maybe this whole logic should be moved to a `Lexicon.dump` function.

vieting (approved):

michelwi (approved):

lexicon/conversion.py
```diff
@@ -140,6 +140,66 @@ def run(self):
...
+            for line in f:
+                # splitting is taken from RASR
+                # src/Tools/Bliss/blissLexiconLib.py#L185
+                s = line.split(None, 1)
+                orth = s[0].split("\\", 1)[0]
+                phon_variants = [
+                    tuple(p.split()) for p in s[1].split("\\") if p.strip()
+                ]
+                for phon_variant in phon_variants:
+                    for phon in phon_variant:
```
curufinwe, line 187: Better: `phonemes.update(phon_variant)`

lexicon/conversion.py
```diff
@@ -140,6 +140,66 @@ def run(self):
...
+    <WORD> <PHONEME1> <PHONEME2> ...
+    separated by tabs or spaces.
+    The lemmata will be added in the order they appear in the text file,
+    the phonemes will be sorted alphabetically.
+    Phoneme variants of the same word need to appear next to each other.
+    No special lemmata or phonemes are added.
+
+    As the splitting is taken from RASR and not fully tested,
+    it might not work in all cases so do not use this job
+    without checking the output manually on new lexica.
```
curufinwe, line 155: Please also add a warning that the generated lexicon is missing special lemmas and thus not directly usable in RASR.

curufinwe (changes requested):

lexicon/conversion.py
```diff
@@ -140,6 +140,66 @@ def run(self):
...
+        with uopen(self.text_file.get_path()) as f:
+            for line in f:
+                # splitting is taken from RASR
+                # src/Tools/Bliss/blissLexiconLib.py#L185
+                s = line.split(None, 1)
+                orth = s[0].split("\\", 1)[0]
+                phon_variants = [
+                    tuple(p.split()) for p in s[1].split("\\") if p.strip()
+                ]
+                phonemes.update(phon_variants)
```
curufinwe, line 188: what I meant was:
```
for phon_variant in phon_variants:
    phonemes.update(phon_variant)
```

curufinwe (changes requested):

lexicon/conversion.py
JackTemaki: Oh s... seems I did not pay attention here!

PR #101, opened 2021-07-22
extra data augmentation methods

Paubaquero:
two additional data augmentation methods using the audiomentations library: impulse responses and noise mixing


corpus/data_augmentation_audiomentations.py
```diff
@@ -0,0 +1,127 @@
...
+        self.add_impulse_response = am.AddImpulseResponse(ir_path=ir_path, p=p)
+
+    def tasks(self):
+        yield Task("run", rqmt=self.rqmt)
+
+    def run(self):
+        self.id = os.path.basename(self.job_id())
+        if not os.path.isdir(f"/dev/shm/{self.id}"):
+            os.mkdir(f"/dev/shm/{self.id}")
+        c = corpus.Corpus()
```
michelwi, line 35: maybe use more expressive variable names e.g. `old_corpus` and `new_corpus`

corpus/data_augmentation_audiomentations.py
```diff
@@ -0,0 +1,127 @@
...
+        nc = corpus.Corpus()
+        segment_file_names = []
+
+        c.load(tk.uncached_path(self.corpus_file))
+        nc.name = self.new_corpus_name
+        nc.speakers = c.speakers
+        nc.default_speaker = c.default_speaker
+        nc.speaker_name = c.speaker_name
+
+        for r in c.recordings:
```
michelwi, line 45: use the function `all_recordings()` to also capture recordings of subcorpora. cf. https://github.com/rwth-i6/i6_core/blob/a62a3175a1c1a3dae49c5dbf686dd0466a86dedf/lib/corpus.py#L173-L176
```suggestion
        for r in c.all_recordings():
```

corpus/data_augmentation_audiomentations.py
```diff
@@ -0,0 +1,127 @@
+__all__ = ["AddImpulseResponseJob", "MixNoiseJob"]
+
+import os
+import audiomentations as am
```
michelwi, line 4: maybe rename `am` to a different alias to avoid confusion with the `i6_core.am` module.

corpus/data_augmentation_audiomentations.py
```diff
@@ -0,0 +1,127 @@
...
+            nc.add_recording(pr)
+            for s in pr.segments:
+                segment_file_names.append(nc.name + '/' + pr.name + '/' + s.name + '\n')
+
+        nc.dump(self.out_corpus.get_path())
+
+        with open(tk.uncached_path(self.out_segment_file), 'w') as segments_outfile:
+            segments_outfile.writelines(segment_file_names)
+
+        shutil.rmtree(f"/dev/shm/{self.id}")
```
michelwi, line 67: How was this folder used?

corpus/data_augmentation_audiomentations.py
```diff
@@ -0,0 +1,127 @@
...
+        self.id = os.path.basename(self.job_id())
+        if not os.path.isdir(f"/dev/shm/{self.id}"):
+            os.mkdir(f"/dev/shm/{self.id}")
+        c = corpus.Corpus()
+        nc = corpus.Corpus()
+        segment_file_names = []
+
+        c.load(tk.uncached_path(self.corpus_file))
+        nc.name = self.new_corpus_name
+        nc.speakers = c.speakers
```
michelwi, line 41:
```suggestion
        nc.speakers = list(c.all_speakers())
```
maybe we want to use `all_speakers()` to capture speakers defined in subcorpora?

corpus/data_augmentation_audiomentations.py
```diff
@@ -0,0 +1,127 @@
...
+
+        nc.dump(self.out_corpus.get_path())
+
+        with open(tk.uncached_path(self.out_segment_file), 'w') as segments_outfile:
+            segments_outfile.writelines(segment_file_names)
+
+        shutil.rmtree(f"/dev/shm/{self.id}")
+
+
+class MixNoiseJob(Job):
```
michelwi, line 70: please provide a short 1-3 sentence docstring what this job does

(Edit: basically all comments on the previous job also apply here)

corpus/data_augmentation_audiomentations.py
```diff
@@ -0,0 +1,127 @@
...
+        self.impulse_responses_dir = ir_path 
+        self.p_impulse_response = p
+
+        self.out_audio_folder = self.output_path("out_audio/", directory=True)
+        self.out_corpus = self.output_path("ir_convoluted.corpus.xml.gz")
+        self.out_segment_file = self.output_path("ir_convoluted.segments")
+
+        self.rqmt = {"time": 12, "cpu": 2}
+
+        self.add_impulse_response = am.AddImpulseResponse(ir_path=ir_path, p=p)
```
michelwi, line 26: Do we want this here as a member of `AddImpulseResponseJob` or should it be local to the `run` method?

If yes, then we don't need to store `ir_path` and `p`.

corpus/data_augmentation_audiomentations.py
```diff
@@ -0,0 +1,127 @@
...
+        self.corpus_file = corpus_file
+        self.new_corpus_name = new_corpus_name
+        self.impulse_responses_dir = ir_path 
+        self.p_impulse_response = p
+
+        self.out_audio_folder = self.output_path("out_audio/", directory=True)
+        self.out_corpus = self.output_path("ir_convoluted.corpus.xml.gz")
+        self.out_segment_file = self.output_path("ir_convoluted.segments")
+
+        self.rqmt = {"time": 12, "cpu": 2}
```
michelwi, line 24: time requirement should be configurable or at least dependent on corpus length.

corpus/__init__.py
```diff
@@ -1,6 +1,7 @@
 from .convert import *
 from .costa import *
 from .data_augmentation import *
+from .data_augmentation_audiomentations import *
```
michelwi, line 4: would it be sensible to include this in `corpus/data_augmentation` instead of creating a new module?

corpus/data_augmentation_audiomentations.py
```diff
@@ -0,0 +1,127 @@
...
+class AddImpulseResponseJob(Job):
+    def __init__(self, corpus_file, new_corpus_name, ir_path, p):
+        self.corpus_file = corpus_file
+        self.new_corpus_name = new_corpus_name
+        self.impulse_responses_dir = ir_path 
+        self.p_impulse_response = p
+
+        self.out_audio_folder = self.output_path("out_audio/", directory=True)
+        self.out_corpus = self.output_path("ir_convoluted.corpus.xml.gz")
+        self.out_segment_file = self.output_path("ir_convoluted.segments")
```
michelwi, line 22: This segment file does not provide any additional information. In the spirit of "one job does exactly one thing" I would vote to remove it here and instead obtain it from a `SegmentCorpusJob`.

michelwi (changes requested):
Hi, thanks for your contribution!

I do request changes on the handling of subcorpora in your jobs. see inline comments

There are many other comments for your consideration / our discussion.

michelwi:
Also please take a look at https://github.com/rwth-i6/i6_core/blob/main/CONTRIBUTUNG.md about black formatting of code and the order of imports (standard library, external libraries (incl. sisyphus), recipes + alphabetical) in your document.
Module `logging` does not seem to be used.

corpus/__init__.py
Paubaquero: I created a new module because this functionality depends on a different external Python library (audiomentations), so that it doesn't break the code of the old module for people who don't have this library installed.

JackTemaki: I would actually to prefer to have it in `data_augmentation`, you can use try/except statements to check for missing imports and just give a warning. You can check at the top of adaptation/ivector.py for an example. But this is not a necessity, just my personal opinion.

corpus/data_augmentation_audiomentations.py
```diff
@@ -0,0 +1,162 @@
...
+        self.corpus_file = corpus_file
+        self.new_corpus_name = new_corpus_name
+        self.impulse_responses_dir = ir_path
+
+        self.out_audio_folder = self.output_path("out_audio/", directory=True)
+        self.out_corpus = self.output_path("ir_convoluted.corpus.xml.gz")
+
+        self.rqmt = {"time": time_rqmt, "cpu": 2}
+
+        self.add_impulse_response = audiomentations.AddImpulseResponse(
```
JackTemaki, line 39: The init function should not perform any tasks, because this will be called each time with the manager. So instead also save `ir_path` and `augment_prob` and call this later in the run.

I know the documentation on Sisyphus jobs does not cover this kind of things yet, I will try to improve this as soon as possible.

corpus/data_augmentation_audiomentations.py
```diff
@@ -0,0 +1,162 @@
...
+
+    def tasks(self):
+        yield Task("run", rqmt=self.rqmt)
+
+    def run(self):
+        job_id = os.path.basename(self.job_id())
+        orig_corpus = corpus.Corpus()
+        new_corpus = corpus.Corpus()
+
+        orig_corpus.load(tk.uncached_path(self.corpus_file))
```
JackTemaki, line 51: `self.corpus_file.get_path()`, we want to get rid of string casts, see https://github.com/rwth-i6/i6_core/issues/32

corpus/data_augmentation_audiomentations.py
```diff
@@ -0,0 +1,162 @@
...
+            perturbed_audio_name = "perturbed_" + r.audio.split("/")[-1]
+            (
+                samples,
+                sample_rate,
+            ) = audiomentations.core.audio_loading_utils.load_wav_file(
+                r.audio, sample_rate=None
+            )
+            augmented_samples = self.add_impulse_response(samples, sample_rate)
+            wavfile.write(
+                str(self.out_audio_folder) + "/" + perturbed_audio_name,
```
JackTemaki, line 67: `self.out_audio_folder.get_path()`, see comment above

corpus/data_augmentation_audiomentations.py
```diff
@@ -0,0 +1,162 @@
...
+                augmented_samples,
+            )
+
+            pr = corpus.Recording()
+            pr.name = r.name
+            pr.segments = r.segments
+            pr.speaker_name = r.speaker_name
+            pr.speakers = r.speakers
+            pr.default_speaker = r.default_speaker
+            pr.audio = str(self.out_audio_folder) + "/" + perturbed_audio_name
```
JackTemaki, line 78: `self.out_audio_folder.get_path()`, see comment above

corpus/data_augmentation_audiomentations.py
```diff
@@ -0,0 +1,162 @@
...
+        self.noise_files_dir = noise_files_dir
+        self.new_corpus_name = new_corpus_name
+
+        self.out_audio_folder = self.output_path("out_audio/", directory=True)
+        self.out_corpus = self.output_path("noise_mixed.corpus.xml.gz")
+        self.out_segment_file = self.output_path("noise_mixed.segments")
+
+        self.rqmt = {"time": time_rqmt, "cpu": 2}
+
+        self.mix_noise = audiomentations.AddBackgroundNoise(
```
JackTemaki, line 120: As mentioned above, please move this into a Task

corpus/data_augmentation_audiomentations.py
```diff
@@ -0,0 +1,162 @@
...
+
+    def tasks(self):
+        yield Task("run", rqmt=self.rqmt)
+
+    def run(self):
+        self.id = os.path.basename(self.job_id())
+        c = corpus.Corpus()
+        nc = corpus.Corpus()
+
+        c.load(tk.uncached_path(self.corpus_file))
```
JackTemaki, line 132: `.get_path()`, see above

corpus/data_augmentation_audiomentations.py
```diff
@@ -0,0 +1,162 @@
...
+            perturbed_audio_name = "perturbed_" + r.audio.split("/")[-1]
+            (
+                samples,
+                sample_rate,
+            ) = audiomentations.core.audio_loading_utils.load_wav_file(
+                r.audio, sample_rate=None
+            )
+            augmented_samples = self.mix_noise(samples, sample_rate)
+            wavfile.write(
+                str(self.out_audio_folder) + "/" + perturbed_audio_name,
```
JackTemaki, line 148: see above

corpus/data_augmentation_audiomentations.py
```diff
@@ -0,0 +1,162 @@
...
+                augmented_samples,
+            )
+
+            pr = corpus.Recording()
+            pr.name = r.name
+            pr.segments = r.segments
+            pr.speaker_name = r.speaker_name
+            pr.speakers = r.speakers
+            pr.default_speaker = r.default_speaker
+            pr.audio = str(self.out_audio_folder) + "/" + perturbed_audio_name
```
JackTemaki, line 159: see above

corpus/data_augmentation_audiomentations.py
```diff
@@ -0,0 +1,162 @@
...
+        self, corpus_file, new_corpus_name, ir_path, augment_prob, time_rqmt=12
+    ):
+        """
+
+        :param Path corpus_file: Bliss corpus with wav files
+        :param str new_corpus_name: name of the new corpus
+        :param ir_path: path to the directory containing impulse responses
+        :param augment_prob: probability of each audio to be augmented by impulse response convolution
+        :param time_rqmt: time requirement for the Sisypus job
+        """
```
JackTemaki, line 29: Some types are still missing

corpus/data_augmentation_audiomentations.py
```diff
@@ -0,0 +1,162 @@
...
+
+        :param Path corpus_file: Bliss corpus with wav files
+        :param str new_corpus_name: name of the new corpus
+        :param noise_files_dir: path to the directory containing noise files
+        :param min_snr: minimum signal-to-noise ratio to be used
+        :param max_snr: maximum signal-to-noise ratio to be used
+        :param augment_prob: probability of each audio to be augmented by noise mixing
+        :param time_rqmt: time requirement for the Sisypus job
+        """
```
JackTemaki, line 109: some types are still missing

JackTemaki (changes requested):

corpus/data_augmentation_audiomentations.py
```diff
@@ -0,0 +1,162 @@
...
+        c = corpus.Corpus()
+        nc = corpus.Corpus()
+
+        c.load(tk.uncached_path(self.corpus_file))
+        nc.name = self.new_corpus_name
+        nc.speakers = c.speakers
+        nc.default_speaker = c.default_speaker
+        nc.speaker_name = c.speaker_name
+
+        for r in c.recordings:
```
michelwi, line 138: To also include recordings from sub corpora
```suggestion
        for r in c.all_recordings():
```

corpus/data_augmentation_audiomentations.py
```diff
@@ -0,0 +1,162 @@
...
+        yield Task("run", rqmt=self.rqmt)
+
+    def run(self):
+        self.id = os.path.basename(self.job_id())
+        c = corpus.Corpus()
+        nc = corpus.Corpus()
+
+        c.load(tk.uncached_path(self.corpus_file))
+        nc.name = self.new_corpus_name
+        nc.speakers = c.speakers
```
michelwi, line 134: To also include speakers from sub corpora
```suggestion
        nc.speakers = list(c.all_speakers())
```

corpus/__init__.py
michelwi: Now after merging the files, this line should be removed.

corpus/data_augmentation.py
```diff
@@ -236,3 +243,167 @@ def run(self):
...
+        self.noise_files_dir = noise_files_dir
+        self.new_corpus_name = new_corpus_name
+
+        self.out_audio_folder = self.output_path("out_audio/", directory=True)
+        self.out_corpus = self.output_path("noise_mixed.corpus.xml.gz")
+        self.out_segment_file = self.output_path("noise_mixed.segments")
+
+        self.min_snr = min_snr
+        self.max_snr = max_snr
+        self.augment_prob = augment_prob
```
michelwi, line 363: Sorting of the `__init__`: first set the inputs, then the outputs.

```suggestion
        self.corpus_file = corpus_file
        self.new_corpus_name = new_corpus_name
        self.noise_files_dir = noise_files_dir
        self.min_snr = min_snr
        self.max_snr = max_snr
        self.augment_prob = augment_prob

        self.out_audio_folder = self.output_path("out_audio/", directory=True)
        self.out_corpus = self.output_path("noise_mixed.corpus.xml.gz")
        self.out_segment_file = self.output_path("noise_mixed.segments")
```

corpus/data_augmentation.py
```diff
@@ -236,3 +243,167 @@ def run(self):
...
+        """
+        self.corpus_file = corpus_file
+        self.new_corpus_name = new_corpus_name
+        self.impulse_responses_dir = ir_path
+
+        self.out_audio_folder = self.output_path("out_audio/", directory=True)
+        self.out_corpus = self.output_path("ir_convoluted.corpus.xml.gz")
+
+        self.ir_path = ir_path
+        self.augment_prob = augment_prob
```
michelwi, line 276: Sorting of the `__init__`: first set the inputs, then the outputs.
`ir_path` was set twice. please remove either `self.ir_path` or `self.impulse_responses_dir` and refactor below.

```suggestion
        self.corpus_file = corpus_file
        self.new_corpus_name = new_corpus_name
        self.impulse_responses_dir = ir_path
        self.augment_prob = augment_prob

        self.out_audio_folder = self.output_path("out_audio/", directory=True)
        self.out_corpus = self.output_path("ir_convoluted.corpus.xml.gz")
```

corpus/data_augmentation.py
```diff
@@ -236,3 +243,167 @@ def run(self):
...
+    """
+
+    def __init__(
+        self, corpus_file, new_corpus_name, ir_path, augment_prob, time_rqmt=12
+    ):
+        """
+
+        :param Path corpus_file: Bliss corpus with wav files
+        :param str new_corpus_name: name of the new corpus
+        :param str ir_path: path to the directory containing impulse responses
```
JackTemaki, line 264: A question here: it seems the ir_path will contain multiple IRs, so will only one be picked randomly or will all be applied?

corpus/data_augmentation.py
```diff
@@ -236,3 +243,167 @@ def run(self):
...
+        :param float augment_prob: probability of each audio to be augmented by impulse response convolution
```
JackTemaki, line 265: Is it possible to make this randomness deterministic? (e.g. provide a seed to the job). Because otherwise it should be clearly marked that this job is not deterministic.

corpus/data_augmentation.py
```diff
@@ -236,3 +243,167 @@ def run(self):
...
+        :param int time_rqmt: time requirement for the Sisypus job
+        """
+        self.corpus_file = corpus_file
+        self.new_corpus_name = new_corpus_name
+        self.impulse_responses_dir = ir_path
+
+        self.ir_path = ir_path
+        self.augment_prob = augment_prob
+
+        self.rqmt = {"time": time_rqmt, "cpu": 2}
```
JackTemaki, line 275: If you have a reasonable estimation for the memory consumption, you could also add it.

corpus/data_augmentation.py
```diff
@@ -236,3 +243,167 @@ def run(self):
...
+        new_corpus = corpus.Corpus()
+
+        orig_corpus.load(tk.get_path(self.corpus_file))
+        new_corpus.name = self.new_corpus_name
+        new_corpus.speakers = list(orig_corpus.all_speakers())
+        new_corpus.default_speaker = orig_corpus.default_speaker
+        new_corpus.speaker_name = orig_corpus.speaker_name
+
+        for r in orig_corpus.all_recordings():
+            perturbed_audio_name = "perturbed_" + r.audio.split("/")[-1]
```
JackTemaki, line 299: The `perturbed_` prefix comes from the speed perturbation job I guess... I would refactor this, I can do this later for all jobs at once.

corpus/data_augmentation.py
```diff
@@ -236,3 +243,167 @@ def run(self):
...
+        yield Task("run", rqmt=self.rqmt)
+
+    def run(self):
+        add_impulse_response = audiomentations.AddImpulseResponse(
+            ir_path=self.ir_path, p=self.augment_prob
+        )
+
+        job_id = os.path.basename(self.job_id())
+        orig_corpus = corpus.Corpus()
+        new_corpus = corpus.Corpus()
```
JackTemaki, line 290: I will test if it is not better to modify the corpus in-place instead of creating a new one, that way also sub-corpus structures would be preserved (which would currently break).

JackTemaki (changes requested):
Those are more questions than a "change request". Do not forget to klick on "re-review" (next to the reviewers names) if you want people to look at a PR again!

corpus/data_augmentation.py
michelwi: >   sub-corpus structures [...] would currently break

They would be flattened but otherwise the functionality would stay the same.
But I agree nothing speaks against preserving the original structure.

JackTemaki: But if they are flattened the "full" segment names change, and then this could cause issues.

corpus/data_augmentation.py
```diff
@@ -236,3 +243,167 @@ def run(self):
...
+        :param str new_corpus_name: name of the new corpus
+        :param str ir_path: path to the directory containing impulse responses
+        :param float augment_prob: probability of each audio to be augmented by impulse response convolution
+        :param int time_rqmt: time requirement for the Sisypus job
+        """
+        self.corpus_file = corpus_file
+        self.new_corpus_name = new_corpus_name
+        self.impulse_responses_dir = ir_path
+
+        self.ir_path = ir_path
```
michelwi, line 272: These lines are still redundant and `self.impulse_responses_dir` does not seem to be used.
```suggestion
        self.ir_path = ir_path
```

corpus/data_augmentation.py
michelwi: Ah, you are right, I agree.

michelwi (changes requested):

corpus/data_augmentation.py
Paubaquero: One will be picked randomly for each audio.

corpus/data_augmentation.py
Paubaquero: Yes, both the randomness to choose an impulse response and the randomness to whether anything is done to a given audio come from the Python interpreter via the 'random' library. Setting the random.seed will make it deterministic. I'm not sure if I should put it in the job description string documentation, or add a new parameter to the call.

corpus/data_augmentation.py
```diff
@@ -236,3 +243,165 @@ def run(self):
...
+            pr.speaker_name = r.speaker_name
+            pr.speakers = r.speakers
+            pr.default_speaker = r.default_speaker
+            pr.audio = self.out_audio_folder.get_path() + "/" + perturbed_audio_name
+            new_corpus.add_recording(pr)
+
+        new_corpus.dump(self.out_corpus.get_path())
+
+
+class MixNoiseJob(Job):
```
curufinwe, line 323: This class is basically a copy of the previous one, except for the different audiomentations object used. It would be cleaner to refactor these classes into one abstract base-class which does the transformations and two child child classes which create the audiomentations transformations object.

curufinwe (changes requested):

corpus/data_augmentation.py
```diff
@@ -245,82 +252,116 @@ def run(self):
...
-    ):
+    def __init__(self, corpus_file, new_corpus_name):
         """
 
         :param Path corpus_file: Bliss corpus with wav files
         :param str new_corpus_name: name of the new corpus
-        :param str ir_path: path to the directory containing impulse responses
-        :param float augment_prob: probability of each audio to be augmented by impulse response convolution
-        :param int time_rqmt: time requirement for the Sisypus job
+        :param str out_corpus_path: name of the new corpus file
```
michelwi, line 265: `out_corpus_path` is not a parameter of the `__init__` ?

corpus/data_augmentation.py
```diff
@@ -245,82 +252,116 @@ def run(self):
...
 
-    def run(self):
-        add_impulse_response = audiomentations.AddImpulseResponse(
-            ir_path=self.ir_path, p=self.augment_prob
-        )
-
-        job_id = os.path.basename(self.job_id())
-        orig_corpus = corpus.Corpus()
-        new_corpus = corpus.Corpus()
+    def _run(self, audiomentations_transformation, perturbed_string):
```
michelwi, line 277: maybe rename `perturbed_string` to to a more descriptive name e.g.  `perturbed_audio_name_prefix`

michelwi (commented):
I am wondering about using this job in practice. You would provide the original corpus and get a perturbed corpus back. Now in training you would like to use (maybe several versions of) the perturbed corpus plus the original, so you would use a MergeCorpusJob. 
But since we here only change the top-level corpus name and keep the recording and segment names as is, only the merge strategy SUBCORPUS can be used, right? And this would invalidate all existing alignments of the original corpus since the segment fullname changed.

@Paubaquero did you already try this job? Is this a problem in practice?
@JackTemaki  how do you handle it for the output of the ChangeCorpusSpeedJob?

curufinwe (approved):

curufinwe:
It has not been used for training yet. To make it work with training you'd have to also add a job that renames the files in the alignment caches.

JackTemaki:
@michelwi @curufinwe can I close this PR? I assume we do not need this anymore.

JackTemaki:
Closed due to inactivity

PR #104, opened 2021-07-27
Add WriteLexiconJob MergeLexiconJob

JackTemaki:
The `StaticLexiconJob` can be used to create a lexicon containing specific entries, such as silence or unknown. Those can then be individually specified in the sisyphus setup config for certain corpora. 

The `MergeLexiconJob` can then do the merging of e.g. a static lexicon for special cases and a lexicon containing the actual words.

### DIFF
--- a/bpe/apply.py
+++ b/bpe/apply.py
@@ -45,7 +45,7 @@ class ApplyBPEModelToLexiconJob(Job):
         lexicon.load(lexicon_path)
 
         lm_tokens = set()
-        for l in lexicon.lemmas:
+        for l in lexicon.lemmata:
             for orth in l.orth:
                 lm_tokens.add(orth)
             for synt in l.synt:
@@ -83,7 +83,7 @@ class ApplyBPEModelToLexiconJob(Job):
 
         w2b = {w: b for w, b in zip(lm_tokens, bpe_tokens)}
 
-        for l in lexicon.lemmas:
+        for l in lexicon.lemmata:
             if l.special is None and len(l.orth) > 0:
                 if len(l.synt) == 0 and len(l.eval) == 0:
                     o = l.orth[0]
--- a/bpe/apply.py
+++ b/bpe/apply.py
@@ -45,7 +45,7 @@ class ApplyBPEModelToLexiconJob(Job):
         lexicon.load(lexicon_path)
 
         lm_tokens = set()
-        for l in lexicon.lemma:
+        for l in lexicon.lemmas:
             for orth in l.orth:
                 lm_tokens.add(orth)
             for synt in l.synt:
@@ -83,7 +83,7 @@ class ApplyBPEModelToLexiconJob(Job):
 
         w2b = {w: b for w, b in zip(lm_tokens, bpe_tokens)}
 
-        for l in lexicon.lemma:
+        for l in lexicon.lemmas:
             if l.special is None and len(l.orth) > 0:
                 if len(l.synt) == 0 and len(l.eval) == 0:
                     o = l.orth[0]
--- a/lexicon/modification.py
+++ b/lexicon/modification.py
@@ -42,7 +42,7 @@ class WriteLexiconJob(Job):
     """
 
     def __init__(
-        self, static_lexicon, sort_phonemes=False, sort_lemmata=False, compressed=False
+        self, static_lexicon, sort_phonemes=False, sort_lemmata=False, compressed=True
     ):
         """
         :param lexicon.Lexicon static_lexicon: A Lexicon object
--- a/lexicon/modification.py
+++ b/lexicon/modification.py
@@ -1,4 +1,4 @@
-__all__ = ["StaticLexiconJob", "MergeLexiconJob"]
+__all__ = ["WriteLexiconJob", "MergeLexiconJob"]
 
 from collections import OrderedDict
 
@@ -8,7 +8,7 @@ from i6_core.lib import lexicon
 from i6_core.util import write_xml
 
 
-class StaticLexiconJob(Job):
+class WriteLexiconJob(Job):
     """
     Create a bliss lexicon file from a static Lexicon.
 
@@ -40,7 +40,9 @@ class StaticLexiconJob(Job):
         static_lexicon.add_phoneme("[UNKNOWN]", variation="none")
     """
 
-    def __init__(self, static_lexicon, sort_phonemes, sort_lemmata, compressed=False):
+    def __init__(
+        self, static_lexicon, sort_phonemes=False, sort_lemmata=False, compressed=False
+    ):
         """
         :param lexicon.Lexicon static_lexicon: A Lexicon object
         :param bool sort_phonemes: sort phoneme inventory alphabetically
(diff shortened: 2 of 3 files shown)


bpe/apply.py
```diff
@@ -45,7 +45,7 @@ def run(self):
         lexicon.load(lexicon_path)
 
         lm_tokens = set()
-        for l in lexicon.lemma:
+        for l in lexicon.lemmas:
```
michelwi, line 48: can we please call it by the Greek plural `lemmata`

lexicon/modification.py
```diff
@@ -0,0 +1,156 @@
...
+
+from collections import OrderedDict
+
+from sisyphus import Job, Task
+
+from i6_core.lib import lexicon
+from i6_core.util import write_xml
+
+
+class StaticLexiconJob(Job):
```
michelwi, line 11: I think a better name would be `WriteLexiconJob` because it effectively only writes out a given lexicon.

lexicon/modification.py
```diff
@@ -0,0 +1,156 @@
...
+    """
+    Merge multiple bliss lexica into a single bliss lexicon.
+
+    Phonemes and lemmas can be individually sorted alphabetically or kept as is.
+
+    When merging a lexicon with a static lexicon, putting the static lexicon first
+    and only sorting the phonemes will result in the "typical" lexicon structure.
+    """
+
+    def __init__(self, bliss_lexica, sort_phonemes, sort_lemmas, compressed=False):
```
michelwi, line 95: the ` sort_phonemes` and  `sort_lemmata` params can have a default value and it should imho be `False`

Maybe we want to add a warning that if sorting of lemmata is enabled, any alignment based on the old lexicon might be invalidated

lexicon/modification.py
```diff
@@ -0,0 +1,156 @@
...
+        else:
+            merged_lex.phonemes = merged_phonemes
+
+        # combine the lemmas
+        if self.sort_lemmas:
+            lemma_dict = {}
+            for lex in lexica:
+                for lemma in lex.lemmas:
+                    # sort by first orth entry
+                    lemma_dict[lemma.orth[0]] = lemma
```
michelwi, line 150: If one first orth exists in two lexica the old lemma is overwritten. Would we want/expect that?

lexicon/modification.py
```diff
@@ -0,0 +1,156 @@
...
+            merged_lex.lemmas = [lemma_dict[key] for key in sorted(lemma_dict.keys())]
+        else:
+            for lex in lexica:
+                merged_lex.lemmas.extend(lex.lemmas)
```
michelwi, line 154: If a lemma existed in more than one lexicon, then it would now appear twice in the merged lexicon. This is different from the sorted case.

michelwi (changes requested):
I strongly suggest the plural of lemma to be lemmata.

Also I think the inconsistency in case a lemma appears in more than one lexicon should be addressed.

The other comments are only suggestions.

bpe/apply.py
curufinwe: Using the English plural is also accepted by the Oxford dictionary.

lexicon/modification.py
curufinwe: No, this is inconsistent with the behavior of not sorting lemmas. There no merging takes place. I.e. here we should make lemma_dict a `str -> list[Lemma]` dict.

lexicon/modification.py
curufinwe: only sorting of phonemes might be affected. sorting lemmas will not break alignments.

lexicon/modification.py
curufinwe: agree

lexicon/modification.py
```diff
@@ -0,0 +1,156 @@
...
+                phon=["[UNKNOWN]"],
+                synt=[["<UNK>"]],
+                special="unknown",
+            )
+        )
+        static_lexicon.add_phoneme("[SILENCE]", variation="none")
+        static_lexicon.add_phoneme("[UNKNOWN]", variation="none")
+    """
+
+    def __init__(self, static_lexicon, sort_phonemes, sort_lemmas, compressed=False):
```
curufinwe, line 43: here sorting of phonemes and lemmas definitely should have default values (False) and I'd set compressed's default to True.

curufinwe (changes requested):

lexicon/modification.py
michelwi: no, in fact the other way round. changing the order of the lemmata might in some cases change the order of the allophone table. (e.g. when no complete allophone file is used)

lexicon/modification.py
JackTemaki: Yes, this is inconsistent, but the question is how would we want that? I would just add an assert for now that crashes the job if there is any overlap.

curufinwe: I'd just keep the two lemmas as is and include them both. Just like in the non-sorted case.

JackTemaki: But can this ever be the desired result? So how will RASR treat two lemmata with the same orth?

curufinwe: It will emit a warning saying that it will use the first lemma when mapping orth -> lemma. But in decoding this is not a problem.

curufinwe: Also what do you do when the orth is empty? E.g. for silence? Sentence-begin/Sentence-end might also have empty orth. This needs to be considered here (i.e. you cannot access lemma.orth[0] without checking for the length of orth).

JackTemaki: An empty orth is a zero length string, so the list will still contain one entry

JackTemaki: > It will emit a warning saying that it will use the first lemma when mapping orth -> lemma. But in decoding this is not a problem.

But if it uses the first, we would not want to have a lemma in there that is never used, or not? I would rather go for the version that it crashes when there is a duplicate, and we later add an option that allows merging if the actually get the case.

curufinwe: But you already now might have multiple lemmas with empty orth (i.e. silence, sentence-end, sentence-begin)

curufinwe: Or you might have lemmas without any orth.

michelwi (approved):
You may still consider changing the default for compressed to True.

JackTemaki:
> You may still consider changing the default for compressed to True.

I only changed it for the merge job, but yes I can do it for the write job as well

lexicon/modification.py
```diff
@@ -0,0 +1,168 @@
...
+            for phoneme_tuple in sorted_phoneme_list:
+                merged_lex.add_phoneme(
+                    symbol=phoneme_tuple[0], variation=phoneme_tuple[1]
+                )
+        else:
+            merged_lex.phonemes = merged_phonemes
+
+        # combine the lemmata
+        if self.sort_lemmata:
+            lemma_dict = defaultdict(lambda: [])
```
curufinwe, line 154: better:
`defaultdict(list)`

curufinwe (changes requested):
Only one minor thing. Commit msg should be something like: "Add WriteLexiconJob MergeLexiconJob #104", i.e. fix code style and merge yourself with this commit msg.

JackTemaki:
> Commit msg should be something like: "Add WriteLexiconJob MergeLexiconJob #104"

Yes, the PR could also have been this.

Ok, will do that

PR #105, opened 2021-07-28
prohibit caching twice in pipeline

JackTemaki:
When a variable already had caching enabled, it would apply the caching twice when accessing it via the `__str__` representation.

I would like keep the (historical) caching default for the Pipeline though, so I replaced `str()` by `.get_path()`

### DIFF
--- a/text/processing.py
+++ b/text/processing.py
@@ -75,9 +75,9 @@ class PipelineJob(Job):
             pipeline.append("gzip")
         pipe = " | ".join([str(i) for i in pipeline])
         if isinstance(self.text_files, (list, tuple)):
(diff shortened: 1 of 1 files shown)


michelwi (approved):

curufinwe:
Why is the job calling gs.file_caching anyway? This should be a property of the Path. No?

JackTemaki:
> Why is the job calling gs.file_caching anyway? This should be a property of the Path. No?

Hmm... caching is in general handled quite inconsistent so far. But yes, maybe it would be better for the future to use `get_cached_path()` here.

curufinwe (approved):

PR #106, opened 2021-07-28
utf-8 fix

christophmluscher:
changes the parsing to the xml file to handle utf-8

I had problems with special utf-8 symbols not being handled correctly and the job crashing.

### DIFF
--- a/corpus/stats.py
+++ b/corpus/stats.py
@@ -31,13 +31,15 @@ class ExtractOovWordsFromCorpusJob(Job):
 
     def run(self):
         with uopen(self.bliss_lexicon, "rt", encoding="utf-8") as f:
-            tree = ET.parse(f)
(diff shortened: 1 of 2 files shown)


curufinwe:
Looking at the documentation [1] of ET parser it says:
If encoding 1 is given, the value overrides the encoding specified in the XML file.
I'd prefer not to override it by default. Maybe you just need to have the correct header in your xml file?

[1] https://docs.python.org/3/library/xml.etree.elementtree.html#xml.etree.ElementTree.XMLParser

christophmluscher:
Yes, you are right. I had "utf8" instead of "utf-8". Because the corpus dump function has "utf8" see new commit

curufinwe:
Oh, OK. I thought that utf8 and utf-8 are synonyms, but it looks like I was wrong.

PR #109, opened 2021-08-03
Switchboard 300h Training Data Preparation Jobs

mmz33:
This PR contains jobs to create switchboard 300h dataset training data corpus from scratch. Since the audio files are not public, this need to be given. I used `/u/corpora/speech/switchboard-1/audio`. 

For preprocessing, I followed Kaldi's recipe. Also this is consistent to what we had before. I compared to Zoltan corpus and it should be exactly the same. 

Concerning speakers mapping, there are some statistics downloaded which contains information about some recordings. If a recording has no speaker info, it is assumed that it has a unique speaker and a unique id is used. btw, i found a bug in the old scripts:
```
                3326B   1.7150          263.6050        0.23    4.36    4.98    ws97_train
                3346B   3.6350          287.9650        0.56    4.73    5.00    WS97_TRAIN
                3413B   11.0150         270.9250        3.12    4.33    4.99     ws97_train
 ```
 this is taken from the statistics file. A `grep` command was applied and the row with `WS97_TRAIN` was skipped because of case sensitivity. The speakers mapping should be consistent always as I sort the recordings when adding the unique ids in case the speaker is unknown.

### DIFF
--- a/datasets/switchboard.py
+++ b/datasets/switchboard.py
@@ -35,7 +35,7 @@ class DownloadSwitchboardTranscriptionAndDictJob(Job):
     def run(self):
         zipped_dirname = "switchboard_word_alignments.tar.gz"
         subprocess.check_call(
-            ["wget", os.path.join("http://www.openslr.org/resources/5", zipped_dirname)]
+            ["wget", "http://www.openslr.org/resources/5/" + zipped_dirname]
         )
         subprocess.check_call(
             [
--- a/datasets/switchboard.py
+++ b/datasets/switchboard.py
@@ -58,10 +58,15 @@ class DownloadSwitchboardSpeakersStatsJob(DownloadJob):
     recording has a unique speaker and a unique id is used for those recordings with unknown speakers info
     """
 
-    def __init__(self, checksum=None):
+    def __init__(self):
         super(DownloadSwitchboardSpeakersStatsJob, self).__init__(
-            url="http://www.isip.piconepress.com/projects/switchboard/doc/statistics/ws97_speaker_stats.text"
+            url="http://www.isip.piconepress.com/projects/switchboard/doc/statistics/ws97_speaker_stats.text",
+            checksum="64f538839073dbbdf46027fff40cec57a11c5de1eed4e8b22b50ed86038d9e90",
         )
+
+    @classmethod
+    def hash(cls, parsed_args):
+        return Job.hash(parsed_args)
 
 
 class CreateSwitchboardSpeakersListJob(Job):
--- a/datasets/switchboard.py
+++ b/datasets/switchboard.py
@@ -1,6 +1,7 @@
 """
 Switchboard is conversational telephony speech with 8 Khz audio files. The training data consists of
 300h hours.
+Reference: https://catalog.ldc.upenn.edu/LDC97S62
 
 number of recordings: 4876
 number of segments: 249624
--- a/datasets/switchboard.py
+++ b/datasets/switchboard.py
@@ -20,6 +20,7 @@ import re
 
 from i6_core.lib import corpus
 from i6_core.util import uopen
+from i6_core.tools.download import DownloadJob
 
 
 class DownloadSwitchboardTranscriptionAndDictJob(Job):
@@ -50,26 +51,16 @@ class DownloadSwitchboardTranscriptionAndDictJob(Job):
         shutil.move("swb_ms98_transcriptions", self.out_trans_and_dict_dir)
 
 
-class DownloadSwitchboardSpeakersStatsJob(Job):
+class DownloadSwitchboardSpeakersStatsJob(DownloadJob):
     """
     Note that this does not contain the speaker info for all recordings. We assume later that each
     recording has a unique speaker and a unique id is used for those recordings with unknown speakers info
     """
 
-    def __init__(self):
-        self.out_speakers_stats_file = self.output_path("speakers_stats.txt")
-
-    def tasks(self):
-        yield Task("run", mini_task=True)
-
-    def run(self):
-        subprocess.check_call(
-            [
-                "wget",
-                "http://www.isip.piconepress.com/projects/switchboard/doc/statistics/ws97_speaker_stats.text",
-            ]
+    def __init__(self, checksum=None):
+        super(DownloadSwitchboardSpeakersStatsJob, self).__init__(
+            url="http://www.isip.piconepress.com/projects/switchboard/doc/statistics/ws97_speaker_stats.text"
         )
-        shutil.move("ws97_speaker_stats.text", self.out_speakers_stats_file)
 
 
 class CreateSwitchboardSpeakersListJob(Job):
--- a/datasets/switchboard.py
+++ b/datasets/switchboard.py
@@ -1,0 +1,293 @@
+"""
+Switchboard is conversational telephony speech with 8 Khz audio files. The training data consists of
+300h hours.
+
+number of recordings: 4876
+number of segments: 249624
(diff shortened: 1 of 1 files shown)

### CODE REVIEW

File: datasets/switchboard.py [hunk 1]
```diff
@@ -0,0 +1,293 @@
...
+
+        rec_to_speaker = {}
+        with uopen(self.speakers_list_file) as f:
+            for line in f:
+                l = line.strip().split()
+                assert len(l) == 3
+                assert (
+                    l[2] not in rec_to_speaker
+                ), "duplicate recording name: {}?".format(l[2])
+                assert l[1] in ["F", "M"], "Gender info must be either M or F"
```
michelwi, line 150: Lets hope no one from the LBGTQ+ community reads our code xP

datasets/switchboard.py
```diff
@@ -0,0 +1,293 @@
+"""
+Switchboard is conversational telephony speech with 8 Khz audio files. The training data consists of
```
michelwi, line 2: maybe give the link
https://catalog.ldc.upenn.edu/LDC97S62
in the description to make sure it is clear which version of SWB is used. Or at least give the LDC catalogue number.

datasets/switchboard.py
```diff
@@ -0,0 +1,293 @@
...
+    def __init__(self):
+        self.out_trans_and_dict_dir = self.output_path("swb_trans_and_dict")
+
+    def tasks(self):
+        yield Task("run", mini_task=True)
+
+    def run(self):
+        zipped_dirname = "switchboard_word_alignments.tar.gz"
+        subprocess.check_call(
+            ["wget", os.path.join("http://www.openslr.org/resources/5", zipped_dirname)]
```
curufinwe, line 37: this is not a filesystem path, don't use `os.path.join` here

datasets/switchboard.py
```diff
@@ -0,0 +1,293 @@
...
+                "-xf",
+                zipped_dirname,
+                "-C",
+                ".",
+            ]
+        )
+        shutil.move("swb_ms98_transcriptions", self.out_trans_and_dict_dir)
+
+
+class DownloadSwitchboardSpeakersStatsJob(Job):
```
curufinwe, line 51: This job is superflous, you can use https://github.com/rwth-i6/i6_core/blob/cef8a57cf0983b810c85c8f5507cc05a34ba4d19/tools/download.py#L12 for this. But if you want you can subclass that job and hardcode the url.

michelwi (approved):
Will you later also add jobs for hub5'00 hub5'01 etc in this file?
What about Fisher train data?

datasets/switchboard.py
mmz33: > Will you later also add jobs for hub5'00 hub5'01 etc in this file?
> What about Fisher train data?

yes Hub5'00 and Hub5'01 will be added later. I am not sure if will add Fisher.

datasets/switchboard.py
```diff
@@ -0,0 +1,293 @@
...
+        rec_to_segs = {}  # typing: dict[str, List[str]]
+        for trans_file in glob.glob(
+            os.path.join(self.trans_dir.get_path(), "*/*/*-trans.text")
+        ):
+            with uopen(trans_file, "rt") as f:
+                for line in f:
+                    seg_info = line.strip().split(" ", 3)  # name start end orth
+                    assert len(seg_info) == 4
+                    rec_name = seg_info[0].split("-")[0]
+                    if rec_name not in rec_to_segs:
```
curufinwe, line 290: Here it's usually nicer to use a defaultdict instead of manually checking for the existence of some key, See https://docs.python.org/3/library/collections.html#collections.defaultdict

mmz33:
> Will you later also add jobs for hub5'00 hub5'01 etc in this file?
> What about Fisher train data?

@michelwi  yes Hub5'00 and Hub5'01 will be added later. I am not sure if will add Fisher.

curufinwe (changes requested):

datasets/switchboard.py
mmz33: this won't work since i need to override the hash function because it checks for checksum. so i will get keyerror. To fix this, i need to override the hash but then i can't call the hash function for the `Job` class right?

mmz33: this works if i pass `checksum=None` to the constructor. i can do this if this is fine.

mmz33: done

datasets/switchboard.py
```diff
@@ -0,0 +1,284 @@
...
+        subprocess.check_call(
+            [
+                "tar",
+                "-xf",
+                zipped_dirname,
+                "-C",
+                ".",
+            ]
+        )
+        shutil.move("swb_ms98_transcriptions", self.out_trans_and_dict_dir)
```
curufinwe, line 51: To save a bit of disk space it would be good to delete the downloaded tar file.

datasets/switchboard.py
```diff
@@ -0,0 +1,284 @@
...
+    """
+
+    def __init__(self):
+        self.out_trans_and_dict_dir = self.output_path("swb_trans_and_dict")
+
+    def tasks(self):
+        yield Task("run", mini_task=True)
+
+    def run(self):
+        zipped_dirname = "switchboard_word_alignments.tar.gz"
```
curufinwe, line 38: dirname = directory name, but this refers to a file, so better `filename`

curufinwe (changes requested):

JackTemaki (changes requested):

datasets/switchboard.py
```diff
@@ -0,0 +1,284 @@
...
+        shutil.move("swb_ms98_transcriptions", self.out_trans_and_dict_dir)
+
+
+class DownloadSwitchboardSpeakersStatsJob(DownloadJob):
+    """
+    Note that this does not contain the speaker info for all recordings. We assume later that each
+    recording has a unique speaker and a unique id is used for those recordings with unknown speakers info
+    """
+
+    def __init__(self, checksum=None):
```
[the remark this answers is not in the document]
michelwi: should the checksum not rather be given to the `super().__init__` call?
```
    def __init__(self, checksum=None):
            super(DownloadSwitchboardSpeakersStatsJob, self).__init__(
            url="http://www.isip.piconepress.com/projects/switchboard/doc/statistics/ws97_speaker_stats.text",
            checksum=checksum,
        )
```

datasets/switchboard.py
```diff
@@ -0,0 +1,284 @@
...
+            if token.lower() not in special_tokens:
+                all_special = False
+                break
+        if all_special:
+            return ""
+
+        out = " ".join(filtered_orth)
+
+        # replace &
+        out = out.replace("AT&T's", "AT and T")
```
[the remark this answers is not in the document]
michelwi: If it is ok to keep the `'s` then the next line would also cover the AT&T's case:
`"AT&T's" -> "AT and T's"`

datasets/switchboard.py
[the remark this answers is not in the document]
mmz33: we should not use `checksum` at all since it will be never used. I have it just as workaround.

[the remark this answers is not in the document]
mmz33: but i can generate the checksum since the file is fixed and use this

datasets/switchboard.py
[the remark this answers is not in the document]
mmz33: No the `'s` need to be removed to be consistent with the current corpus (what we have from Zoltan).

curufinwe (approved):

PR #112, opened 2021-08-05
Fix bug in LexiconWriteFromFile job to allow multiple pronunciation variants

mmz33:
This PR fixes the bug in writing lexicon from file by allowing multiple pronunciation variants.

### DIFF
--- a/lexicon/conversion.py
+++ b/lexicon/conversion.py
@@ -188,12 +188,12 @@ class LexiconFromTextFileJob(Job):
                 for phon_variant in phon_variants:
                     phonemes.update(phon_variant)
                 phon = [" ".join(v) for v in phon_variants]
-                lemma = lexicon.Lemma(orth=[orth], phon=phon)
-                if last_lemma and lemma.orth[0] == last_lemma.orth[0]:
+                if last_lemma and orth == last_lemma.orth[0]:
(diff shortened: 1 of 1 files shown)


mmz33:
btw, i was thinking also if we should use a dict to check for already seen lemma instead of `last_lemma`. Because this requires that the orth need to be sorted in the file or at least this need to be done.

curufinwe (approved):
I think the only bug here is in `if p not in lemma.phon:` where we should look at `p not in last_lemma.phon`. Everything else could remain the same. But this code also works. So OK from my side.

curufinwe:
@JackTemaki  what do you think? (IIRC you wrote the original version)

mmz33:
> I think the only bug here is in `if p not in lemma.phon:` where we should look at `p not in last_lemma.phon`. Everything else could remain the same. But this code also works. So OK from my side.

yes but also no need to create the `lemma` object outside the else branch

JackTemaki (approved):
Yeah, this was a really bad mistake from my side, the new version is definitely correct, I just tested it again. It seems I did not check properly after fixing the repeated entry issue...

mmz33:
@curufinwe can this be merged?

JackTemaki (approved):

PR #113, opened 2021-08-05
Extended BlissLexiconToG2PLexiconJob 

mattiadg:
Extended BlissLexiconToG2PLexiconJob to output more phonemes for the same word

### DIFF
--- a/g2p/convert.py
+++ b/g2p/convert.py
@@ -16,17 +16,15 @@ class BlissLexiconToG2PLexiconJob(Job):
     Convert a bliss lexicon into a g2p compatible lexicon for training
     """
 
-    __sis_hash_exclude = {"write_first_phoneme_only": True}
+    __sis_hash_exclude = {"include_pronunciation_variants": False}
 
-    def __init__(self, bliss_lexicon, write_first_phoneme_only=True):
+    def __init__(self, bliss_lexicon, include_pronunciation_variants=False):
         """
         :param Path bliss_lexicon:
-        :param bool write_first_phoneme_only: In case of multiple phoneme representations for one lemma, when this is true it outputs only the first phoneme
+        :param bool include_pronunciation_variants: In case of multiple phoneme representations for one lemma, when this is false it outputs only the first phoneme
         """
         self.bliss_lexicon = bliss_lexicon
-
-        self.write_first_phoneme_only = write_first_phoneme_only
-
+        self.include_pronunciation_variants = include_pronunciation_variants
         self.out_g2p_lexicon = self.output_path("g2p.lexicon")
 
     def tasks(self):
@@ -47,11 +45,7 @@ class BlissLexiconToG2PLexiconJob(Job):
 
                 orth = lemma.find("orth").text.strip()
 
-                if self.write_first_phoneme_only:
-                    phon = lemma.find("phon").text.strip()
-                    out.write("%s %s\n" % (orth, phon))
-
-                else:
+                if self.include_pronunciation_variants:
                     phons = lemma.findall("phon")
                     phon_single = []
                     for phon in phons:
@@ -59,6 +53,9 @@ class BlissLexiconToG2PLexiconJob(Job):
                         if p not in phon_single:
                             phon_single.append(p)
                             out.write("%s %s\n" % (orth, p))
+                else:
+                    phon = lemma.find("phon").text.strip()
+                    out.write("%s %s\n" % (orth, phon))
 
 
 class G2POutputToBlissLexiconJob(Job):
--- a/g2p/convert.py
+++ b/g2p/convert.py
@@ -16,13 +16,15 @@ class BlissLexiconToG2PLexiconJob(Job):
     Convert a bliss lexicon into a g2p compatible lexicon for training
     """
 
-    def __init__(self, bliss_lexicon):
+    def __init__(self, bliss_lexicon, all_phons=False):
         """
         :param Path bliss_lexicon:
         """
         self.bliss_lexicon = bliss_lexicon
 
         self.out_g2p_lexicon = self.output_path("g2p.lexicon")
+
+        self.all_phons = all_phons
 
     def tasks(self):
         yield Task("run", mini_task=True)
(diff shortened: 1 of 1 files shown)


g2p/convert.py
```diff
@@ -16,14 +16,16 @@ class BlissLexiconToG2PLexiconJob(Job):
     Convert a bliss lexicon into a g2p compatible lexicon for training
     """
 
-    def __init__(self, bliss_lexicon):
+    def __init__(self, bliss_lexicon, all_phons=False):
```
JackTemaki, line 19: This will break existing hashes. What you can do is add:
```python
__sis_hash_exclude = {'all_phons': False}
```
above the init.

But I think the parameter name is not really nice. I would do it the other way round, and call it `write_first_phoneme_only` or so with "legacy" default True.

mmz33, line 19: missing docs for `all_phons`

g2p/convert.py
```diff
@@ -16,14 +16,16 @@ class BlissLexiconToG2PLexiconJob(Job):
...
         """
         :param Path bliss_lexicon:
         """
         self.bliss_lexicon = bliss_lexicon
 
         self.out_g2p_lexicon = self.output_path("g2p.lexicon")
 
+        self.all_phons = all_phons
```
JackTemaki, line 27: this needs to be put above the output definition, see https://github.com/rwth-i6/i6_core/blob/main/CONTRIBUTUNG.md

mmz33, line 27: output should be defined at the end. see job constructor order here: https://github.com/rwth-i6/i6_core/blob/main/CONTRIBUTUNG.md

mmz33 (changes requested):

JackTemaki (changes requested):

g2p/convert.py
```diff
@@ -16,15 +16,18 @@ class BlissLexiconToG2PLexiconJob(Job):
     Convert a bliss lexicon into a g2p compatible lexicon for training
     """
 
-    def __init__(self, bliss_lexicon, all_phons=False):
+    __sis_hash_exclude = {"write_first_phoneme_only": True}
+
+    def __init__(self, bliss_lexicon, write_first_phoneme_only=True):
```
Marvin84, line 21: Here we are considering the different pronunciations and this terminology (phoneme, writing the first phoneme) is not clear. I would call this boolean `include_pronunciation_variants`

g2p/convert.py
```diff
@@ -39,15 +42,15 @@ def run(self):
```
Marvin84, line 42: If one passes a file that does not adhere to the desired xml structure, `tree.findall(".//lemma")` would be an empty array and the resulting output file is an empty file. I would add an assertion on the length before starting this loop.

Marvin84 (changes requested):

g2p/convert.py
mattiadg: Can we all agree on this before I change again @mmz33 @JackTemaki?

JackTemaki: @curufinwe ? I would agree to Tinas suggestion.

g2p/convert.py
mattiadg: Added  this check in the last commit

g2p/convert.py
curufinwe: I also agree with Tina's suggestion.

g2p/convert.py
```diff
@@ -16,12 +16,17 @@ class BlissLexiconToG2PLexiconJob(Job):
...
 
-    def __init__(self, bliss_lexicon):
+    __sis_hash_exclude = {"write_first_phoneme_only": True}
+
+    def __init__(self, bliss_lexicon, write_first_phoneme_only=True):
         """
         :param Path bliss_lexicon:
+        :param bool write_first_phoneme_only: In case of multiple phoneme representations for one lemma, when this is true it outputs only the first phoneme
         """
         self.bliss_lexicon = bliss_lexicon
```
curufinwe, line 27: no newline between setting members from parameters (this should be one logical block)

g2p/convert.py
```diff
@@ -31,14 +36,29 @@ def run(self):
...
                     continue
 
                 orth = lemma.find("orth").text.strip()
-                phon = lemma.find("phon").text.strip()
 
-                out.write("%s %s\n" % (orth, phon))
+                if self.write_first_phoneme_only:
+                    phon = lemma.find("phon").text.strip()
+                    out.write("%s %s\n" % (orth, phon))
+
```
curufinwe, line 53: no new line before else (see: https://www.python.org/dev/peps/pep-0008/#blank-lines)

g2p/convert.py
```diff
@@ -31,14 +36,29 @@ def run(self):
...
+                else:
+                    phons = lemma.findall("phon")
+                    phon_single = []
```
curufinwe, line 56: phon_single is used as a lookup datastructure here. In that case it's better to make it a set instead of a list as list has lookup time of O(n) vs. set's (amortized) O(1). In this case we do not expect many clashes or even entries it does not matter much, but I want to point out the general principle here.

curufinwe (changes requested):

g2p/convert.py
```diff
@@ -16,12 +16,15 @@ class BlissLexiconToG2PLexiconJob(Job):
...
+    __sis_hash_exclude = {"include_pronunciation_variants": False}
+
+    def __init__(self, bliss_lexicon, include_pronunciation_variants=False):
         """
         :param Path bliss_lexicon:
+        :param bool include_pronunciation_variants: In case of multiple phoneme representations for one lemma, when this is false it outputs only the first phoneme
         """
         self.bliss_lexicon = bliss_lexicon
-
+        self.include_pronunciation_variants = include_pronunciation_variants
```
curufinwe, line 27: Actually the new line between the input parameters and output variables was OK as these are two different logical blocks. I.e. before you have a newline between `self.bliss_lexicon = ...` and `self.out_g2p_lexicon = ...` because these are two parts of the constructor. As `self.include_pronunciation_variants` is an input member it should be in the first block (without separation of the members within the block), but the second block is still separated by a newline from the first one.

curufinwe (approved):

PR #114, opened 2021-08-05
Add BuildGlobalCacheJob

curufinwe:
First step in refactoring recognition pipeline dependencies. Two things that I want to change:
1. recognizer parameters are currently very restricted in the AdvancedTreeSearchJob. This is fixed by providing the recognizer_config via crp.
2. AdvancedTreeSearchLmImageAndGlobalCacheJob does too many things (building global cache + building lm image). Ideally these are two separate steps, s.t. changes to the search space do not trigger a rebuild of the LM image.

This pull request addresses 1. by adding a new member to crp and 2. by adding a separate job for building the global cache.
Next step will be a CreateLmImageJob and after that a AdvancedTreeSearchV2Job.

### DIFF
--- a/recognition/advanced_tree_search.py
+++ b/recognition/advanced_tree_search.py
@@ -3,6 +3,7 @@ __all__ = [
     "AdvancedTreeSearchJob",
     "AdvancedTreeSearchWithRescoringJob",
     "BidirectionalAdvancedTreeSearchJob",
+    "BuildGlobalCacheJob",
 ]
 
 from sisyphus import *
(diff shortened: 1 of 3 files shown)


recognition/advanced_tree_search.py
```diff
@@ -854,3 +855,75 @@ def hash(cls, kwargs):
...
+        config._update(extra_config)
+        post_config._update(extra_post_config)
+
+        return config, post_config
+
+    @classmethod
+    def hash(cls, kwargs):
+        config, post_config = cls.create_config(**kwargs)
+        return super().hash(
+            {"config": config, "exe": kwargs["crp"].speech_recognizer_exe}
```
mmz33, line 928: should `exe` be consistent with `self.exe` as in the constructor? checking the code this will use the default provided in case `kwargs["crp"].speech_recognizer_exe` was `None` right?

mmz33: ah no nevermind i overlooked into this

mmz33 (approved):

michelwi (approved):

PR #115, opened 2021-08-05
Allow lexicon file with unsorted words in `LexiconFromTextFileJob`

mmz33:
Currently, `LexiconFromTextFileJob` job requires that the orth or words in the file need to be sorted. This is because it uses the variable `last_lemma` to check if the current lemma is seen before. So this requires sorted order of the words. But this does not need to be the case from a user perspective. The user can provide a file with any order of words.


mmz33:
Note that I also had this case when building the switchboard lexicon. Now it is constructed correctly.

curufinwe (approved):

PR #116, opened 2021-08-05
Add job to create SWB lexicon text file

mmz33:
This PR adds a job to create preprocess SWB lexicon text-based format. This will be used later with `LexiconWriteFromFile` job to create the xml lexicon.

### DIFF
--- a/datasets/switchboard.py
+++ b/datasets/switchboard.py
@@ -71,7 +71,7 @@ class DownloadSwitchboardTranscriptionAndDictJob(Job):
     """
 
     def __init__(self):
-        self.out_raw_dict = self.output_path("raw_dict.txt")
+        self.out_raw_dict = self.output_path("swb_trans/sw-ms98-dict.text")
         self.out_trans_dir = self.output_path("swb_trans")
 
     def tasks(self):
@@ -92,10 +92,6 @@ class DownloadSwitchboardTranscriptionAndDictJob(Job):
             ]
         )
         shutil.move("swb_ms98_transcriptions", self.out_trans_dir.get_path())
-        shutil.copy(
-            os.path.join(self.out_trans_dir.get_path(), "sw-ms98-dict.text"),
-            self.out_raw_dict.get_path(),
-        )
         os.remove(zipped_filename)
 
 
--- a/datasets/switchboard.py
+++ b/datasets/switchboard.py
@@ -329,4 +329,4 @@ class CreateSwitchboardLexiconTextFileJob(Job):
                     continue
                 token = parts[0].replace("&amp;", "&")  # e.g A&amp;E -> A&E
                 mapped_token = _map_token(token)  # preprocessing as corpus
-                out_f.write(mapped_token + " " + " ".join(parts[1:]) + "\n")
+                out_f.write(mapped_token + " " + parts[1] + "\n")
--- a/datasets/switchboard.py
+++ b/datasets/switchboard.py
@@ -23,6 +23,48 @@ from i6_core.util import uopen
 from i6_core.tools.download import DownloadJob
 
 
+def _map_token(token):
+    """
+    This function applies some mapping rules for Switchboard transcription words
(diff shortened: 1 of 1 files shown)


datasets/switchboard.py
```diff
@@ -288,3 +298,35 @@ def _get_rec_to_segs_map(self):
...
+        with uopen(raw_dict_file) as read_f, uopen(self.out_dict, "w") as out_f:
+            for line in read_f.readlines()[1:]:
+                if line.startswith("#"):  # skip comment
+                    continue
+                parts = line.strip().split(" ", 1)
+                if len(parts) < 2:
+                    continue
+                token = parts[0].replace("&amp;", "&")  # e.g A&amp;E -> A&E
+                mapped_token = _map_token(token)  # preprocessing as corpus
+                out_f.write(mapped_token + " " + " ".join(parts[1:]) + "\n")
```
curufinwe, line 332: why join here? you will have exactly 2 parts here

datasets/switchboard.py
```diff
@@ -288,3 +298,35 @@ def _get_rec_to_segs_map(self):
...
+        :param tk.Path dict_dir: path containing the raw dictionary text file
+        """
+        self.dict_dir = dict_dir
+        self.out_dict = self.output_path("dict.txt")
+
+    def tasks(self):
+        yield Task("run", mini_task=True)
+
+    def run(self):
+        raw_dict_file = os.path.join(self.dict_dir.get_path(), "sw-ms98-dict.text")
```
curufinwe, line 322: Passing a dict and expecting a file name sw-ms98-dict.text to be present there creates a very tight coupling between DownloadSwitchboardTranscriptionAndDictJob and CreateSwitchboardLexiconTextFileJob. I think it would be better to have an additional output for the lexicon in DownloadSwitchboardTranscriptionAndDictJob and then use that path here.

curufinwe (changes requested):

datasets/switchboard.py
mmz33: ah yeah right

datasets/switchboard.py
```diff
@@ -90,7 +91,11 @@ def run(self):
                 ".",
             ]
         )
-        shutil.move("swb_ms98_transcriptions", self.out_trans_and_dict_dir)
+        shutil.move("swb_ms98_transcriptions", self.out_trans_dir.get_path())
+        shutil.copy(
```
curufinwe, line 95: No need to copy the data, your second output path can also be within the swb_trans folder. E.g.:
```
self.out_raw_dict = self.output_path("swb_trans/sw-ms98-dict.text")
```

mmz33: but then what would be the default value for `self.out_raw_dict` in the constructor? `None` is ok?

mmz33: The path need to be defined beforehand right? so it should be some `self.output_path` already in the constructor.

michelwi: the suggestion is that after the move
```
shutil.move("swb_ms98_transcriptions", self.out_trans_dir.get_path())
```

there exists the file `swb_trans/sw-ms98-dict.text`
which you would copy to a new name. No need to do this if you define this name directly above because the file will then be used directly. (cf. code suggestion in other comment)

datasets/switchboard.py
```diff
@@ -23,13 +23,56 @@
...
+
+
 class DownloadSwitchboardTranscriptionAndDictJob(Job):
     """
     Downloads switchboard training transcriptions and dictionary (or lexicon)
     """
 
     def __init__(self):
-        self.out_trans_and_dict_dir = self.output_path("swb_trans_and_dict")
+        self.out_raw_dict = self.output_path("raw_dict.txt")
```
michelwi, line 74:
```suggestion
        self.out_raw_dict = self.output_path("swb_trans/sw-ms98-dict.text")
```

datasets/switchboard.py
mmz33: ah i see

curufinwe (approved):

PR #119, opened 2021-08-06
allow keeping dict order in returnn config

vieting:
When large networks are printed into the config file, I'd prefer not to sort the layers alphabetically but to keep the order. This makes it easier to see what the network does since the order from top to bottom could correspond to the logical order of the layers in the model.

Therefore, I added an option which allows not sorting dicts when the config is written. This only works for python >=3.8, see https://docs.python.org/3.8/library/pprint.html. We might add a fallback to sorted dicts for <3.8 if you think this is necessary.

### DIFF
--- a/returnn/config.py
+++ b/returnn/config.py
@@ -78,6 +78,7 @@ class ReturnnConfig:
         python_epilog="",
         python_epilog_hash=None,
         hash_full_python_code=False,
+        pprint_kwargs=None,
     ):
         """
 
@@ -91,6 +92,8 @@ class ReturnnConfig:
         :param str|None python_epilog_hash: sets a specific hash for the python_epilog
         :param bool hash_full_python_code: By default, function bodies are not hashed. If set to True, the full content
             of python pro-/epilog is parsed and hashed.
+        :param dict|None pprint_kwargs: kwargs for pprint, e.g. {"sort_dicts": False} to print dicts in given order for
+            python >= 3.8
         """
         self.config = config
         self.post_config = post_config if post_config is not None else {}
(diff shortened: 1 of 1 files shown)


returnn/config.py
```diff
@@ -78,6 +78,7 @@ def __init__(
         python_epilog="",
         python_epilog_hash=None,
         hash_full_python_code=False,
+        sort_dicts=False,
```
michelwi, line 81: Default value should be old behaviour i.e. sorting enabled?
```suggestion
        sort_dicts=True,
```

returnn/config.py
```diff
@@ -91,6 +92,7 @@ def __init__(
         :param str|None python_epilog_hash: sets a specific hash for the python_epilog
         :param bool hash_full_python_code: By default, function bodies are not hashed. If set to True, the full content
             of python pro-/epilog is parsed and hashed.
+        :param bool sort_dicts: if True, the dicts in the written config file are sorted alphabetically by keys
```
michelwi, line 95: Maybe its more interesting to document what happens if it is set to False (i.e. printed in order of insertion)

returnn/config.py
```diff
@@ -128,7 +131,7 @@ def serialize(self):
         config_lines = []
         unreadable_data = {}
 
-        pp = pprint.PrettyPrinter(indent=2, width=150)
+        pp = pprint.PrettyPrinter(indent=2, width=150, sort_dicts=self.sort_dicts)
```
curufinwe, line 134: This breaks for pythons < 3.8. how about passing arbitrary arguments to PrettyPrinter, i.e. change `sort_dicts` to `pprint_kwargs`. Then you can set it in your setup, but it will still work with older pythons for setups that don't use this option.

curufinwe (changes requested):

returnn/config.py
vieting: Yes, I think that's a good idea. I'll update the PR accordingly.

curufinwe (approved):

michelwi (approved):

JackTemaki (approved):

PR #123, opened 2021-08-09
Refactor segment jobs

JackTemaki:
It seems a complete file was missed during refactoring, and I think we should fix this now (or never).

I think it would be best that all people that already build setups with this confirm this PR, so that I can see that everyone is aware of the API (not hash) changes.

### DIFF
--- a/adaptation/linear_adaptation_layer.py
+++ b/adaptation/linear_adaptation_layer.py
@@ -181,8 +181,8 @@ class MetaLinearAdaptationJob(Job):
         for key, seg in self.single_segments.items():
             print("{}: Adding training {}".format(self.name, key))
             new_segments = corpus_recipes.ShuffleAndSplitSegmentsJob(seg)
-            seg_train = new_segments.new_segments["train"]
-            seg_dev = new_segments.new_segments["dev"]
+            seg_train = new_segments.out_segments["train"]
+            seg_dev = new_segments.out_segments["dev"]
 
             train_corpus = copy.deepcopy(self.corpus)
             train_corpus.segment_path = seg_train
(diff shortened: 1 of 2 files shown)


michelwi (approved):
I wasn't using it yet and would be fine with the change

mmz33 (approved):

vieting (changes requested):
There is a usage of the `ShuffleAndSplitSegmentsJob` in `linear_adaptation_layer.py` which should be updated. Also, I guess we need an associate PR for i6_experiments?

curufinwe (dismissed):

mattiadg:
> It seems a complete file was missed during refactoring, and I think we should fix this now (or never).
> 
> I think it would be best that all people that already build setups with this confirm this PR, so that I can see that everyone is aware of the API (not hash) changes.

by refactoring here, you mean updating to the new conventions?

JackTemaki:
> > It seems a complete file was missed during refactoring, and I think we should fix this now (or never).
> > I think it would be best that all people that already build setups with this confirm this PR, so that I can see that everyone is aware of the API (not hash) changes.
> 
> by refactoring here, you mean updating to the new conventions?

Yes exactly

JackTemaki:
> There is a usage of the `ShuffleAndSplitSegmentsJob` in `linear_adaptation_layer.py` which should be updated. Also, I guess we need an associate PR for i6_experiments?

Done, and yes I will try to do a PR that also fixes this for `i6_experiments/common`

curufinwe (approved):

mattiadg (approved):

vieting (approved):

Marvin84 (approved):

michelwi (approved):

christophmluscher (approved):

papar22 (approved):

mmz33 (approved):

PR #126, opened 2021-08-11
allow empty pipe in PipelineJob

JackTemaki:
As the docstring also says now: "[pipe] can be empty to use the job for concatenation or gzip compression only"

### DIFF
--- a/text/processing.py
+++ b/text/processing.py
@@ -2,6 +2,7 @@ __all__ = ["PipelineJob", "ConcatenateJob", "HeadJob", "TailJob"]
 
 import os
 from sisyphus import Job, Task, Path, global_settings as gs
+from sisyphus.delayed_ops import DelayedBase
 
 
 class PipelineJob(Job):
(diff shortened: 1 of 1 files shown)


christophmluscher (approved):

text/processing.py
```diff
@@ -171,15 +176,15 @@ def tasks(self):
 
     def run(self):
         if self.ratio:
-            assert not self.lines
+            assert not self.num_lines
```
mmz33, line 179: `self.num_lines` is not defined?

text/processing.py
```diff
@@ -171,15 +176,15 @@ def tasks(self):
 
     def run(self):
         if self.ratio:
```
mmz33, line 178: `self.ratio` is not defined?

mmz33 (changes requested):

text/processing.py
JackTemaki: This is because of `self.set_attrs(locals())`, so it is defined but PyCharm does not see that. I can change this, this is still legacy code.

mmz33 (approved):

curufinwe (approved):

PR #127, opened 2021-08-13
update Checkpoint class and minor changes

JackTemaki:
There is no need to give the Checkpoint class directly both inputs. I now refactored it in a way that there is no `.get_path()` call in the init function (which is not nice), and the path is created only when the object is accessed (which makes `Checkpoint` kind of a "Delayed" object). 

~~This is a draft, because it is not yet fully tested.~~ Tested now!

### DIFF
--- a/returnn/training.py
+++ b/returnn/training.py
@@ -37,20 +37,22 @@ class ReturnnModel:
 
 class Checkpoint:
     """
-    Checkpoint object which holds the checkpoint path prefix as string,
-    and the index file path as tk.Path
+    Checkpoint object which holds the (Tensorflow) index file path as tk.Path,
+    and will return the checkpoint path as common prefix of the .index/.meta/.data[...]
     """
 
-    def __init__(self, ckpt_path, index_path):
+    def __init__(self, index_path):
         """
-        :param str ckpt_path:
         :param Path index_path:
         """
-        self.ckpt_path = ckpt_path
         self.index_path = index_path
 
     def _sis_hash(self):
         return self.index_path._sis_hash()
+
+    @property
+    def ckpt_path(self):
+        return self.index_path.get_path()[: -len(".index")]
 
     def __str__(self):
         return self.ckpt_path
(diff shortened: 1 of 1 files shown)


returnn/training.py
```diff
@@ -136,21 +151,14 @@ def __init__(
...
                 for k in stored_epochs
                 if k in self.keep_epochs
                 for index_path in [self.output_path("models/epoch.%.3d.index" % k)]
             }
         self.out_plot_se = self.output_path("score_and_error.png")
         self.out_plot_lr = self.output_path("learning_rate.png")
 
-        self.returnn_config.post_config["model"] = os.path.join(
-            self.out_model_dir.get_path(), "epoch"
-        )
```
michelwi, line 149: I am a bit hesitant with this change because if you move this part to the `create_files` method, the config of the Job object at creation will be different from the config at runtime.

If other Jobs depended on this returnn config they would have to delay config handling until the job has actually started.

michelwi (commented):
I like and support your changes to the `Checkpoint` class.

Some comments on the minor changes.

returnn/training.py
JackTemaki: I really hope no one is accessing the actual return config object from the job again.... This is actually something I thought of, that this kind of members should be strictly private. Because the Job is construction this only for itself, it should not be used outside.The model path is added only for this job, if other jobs want to use this they can use the same config that was used as input, and access the models via the job outputs. They should definitely not access this locally created config.

curufinwe: But why move it from the constructor to the create_files method in the first place? This looks unrelated to the Checkpoint changes to me.

JackTemaki: It is unrelated, I just did it because I think that any code that does need to be in the `__init__` should not be there, as it slows down the graph construction. In this specific case I also do not like that it would mean calling `.get_path()` during the graph construction, which is okay here but should be discouraged in general.

curufinwe: I don't see the problem here. The setting of that property is very fast (it's not like we are looping over a large data-structure here). Accessing the path that the job created itself is also not an anti-pattern in my mind.
BUT: usually the job's do construct all of the config / flow networks / etc. in the constructor and not afterwards. I see it a as an anti-pattern to modify the config after the ctor if it's not absolutely necessary (because the value is not available before / depends on concrete inputs) as you'd have to look at the generated files and could not look at the python object. In this specific case it's not too tragic, but I'd still prefer the old way.

returnn/training.py
```diff
@@ -222,6 +234,7 @@ def tasks(self):
         yield Task("plot", resume="plot", mini_task=True)
 
     def create_files(self):
+
```
michelwi, line 237: this newline is probably not intended

```suggestion
```

michelwi (approved):

curufinwe (approved):

PR #132, opened 2021-08-23
add ExtractDatasetStatisticsJob

JackTemaki:
This job is used to compute the global mean and variance, both scalar as `tk.Variable` or per feature as file).

### DIFF
--- a/returnn/dataset.py
+++ b/returnn/dataset.py
@@ -75,17 +75,18 @@ class ExtractDatasetMeanStddevJob(Job):
         total_mean = 0
         total_var = 0
 
-        mean_file = open("stats.mean.txt")
-        std_dev_file = open("stats.std_dev.txt")
+        with open("stats.mean.txt") as mean_file, open(
+            "stats.std_dev.txt"
+        ) as std_dev_file:
 
-        # compute the total mean and std-dev in an iterative way
-        for i, (mean, std_dev) in enumerate(zip(mean_file, std_dev_file)):
-            mean = float(mean)
-            var = float(std_dev.strip()) ** 2
-            mean_variance = (total_mean - mean) ** 2
-            adjusted_mean_variance = mean_variance * i / (i + 1)
-            total_var = (total_var * i + var + adjusted_mean_variance) / (i + 1)
-            total_mean = (total_mean * i + mean) / (i + 1)
+            # compute the total mean and std-dev in an iterative way
+            for i, (mean, std_dev) in enumerate(zip(mean_file, std_dev_file)):
+                mean = float(mean)
+                var = float(std_dev.strip()) ** 2
+                mean_variance = (total_mean - mean) ** 2
+                adjusted_mean_variance = mean_variance * i / (i + 1)
+                total_var = (total_var * i + var + adjusted_mean_variance) / (i + 1)
+                total_mean = (total_mean * i + mean) / (i + 1)
 
-        self.out_mean.set(total_mean)
-        self.out_std_dev.set(numpy.sqrt(total_var))
+            self.out_mean.set(total_mean)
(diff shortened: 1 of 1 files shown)


mmz33 (approved):

returnn/dataset.py
```diff
@@ -0,0 +1,90 @@
...
+import numpy
+import os
+import shutil
+import subprocess
+
+from i6_core.returnn.config import ReturnnConfig
+from i6_core.util import create_executable
+
+
+class ExtractDatasetStatisticsJob(Job):
```
curufinwe, line 14: If job only computes mean + std-dev it should be called like that. Current name is a bit confusing as I would expect Dataset statistics to be things like #number of sequences, average length, etc.

returnn/dataset.py
```diff
@@ -0,0 +1,90 @@
...
+            returnn_python_exe
+            if returnn_python_exe is not None
+            else gs.RETURNN_PYTHON_EXE
+        )
+        self.returnn_root = (
+            returnn_root if returnn_root is not None else gs.RETURNN_ROOT
+        )
+        self.out_mean = self.output_var("mean_var")
+        self.out_std_dev = self.output_var("std_dev_var")
+
```
curufinwe, line 47: no newline here

returnn/dataset.py
```diff
@@ -0,0 +1,90 @@
...
+        self.returnn_config = returnn_config
+        self.returnn_python_exe = (
+            returnn_python_exe
+            if returnn_python_exe is not None
+            else gs.RETURNN_PYTHON_EXE
+        )
+        self.returnn_root = (
+            returnn_root if returnn_root is not None else gs.RETURNN_ROOT
+        )
```
curufinwe, line 44: add newline

returnn/dataset.py
```diff
@@ -0,0 +1,90 @@
+__all__ = ["ExtractDatasetStatisticsJob"]
+
+from sisyphus import *
+
+import numpy
```
curufinwe, line 5: numpy is not an python buildin lib and thus should be in it's own block (after the python libs)

curufinwe (changes requested):

returnn/dataset.py
```diff
@@ -0,0 +1,90 @@
...
+            "returnn.config",
+            "--endseq -1",
+            "--stats",
+            "--dump_stats stats",
+        ]
+
+        create_executable("rnn.sh", command)
+        subprocess.check_call(["./rnn.sh"])
+
+        shutil.copy("stats.mean.txt", self.out_mean_file.get_path())
```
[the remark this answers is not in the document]
JackTemaki: The file is very small, still move is better than copy

returnn/dataset.py
```diff
@@ -0,0 +1,90 @@
...
+        shutil.copy("stats.std_dev.txt", self.out_std_dev_file.get_path())
+
+        total_mean = 0
+        total_var = 0
+
+        mean_file = open("stats.mean.txt")
+        std_dev_file = open("stats.std_dev.txt")
+
+        # compute the total mean and std-dev in an iterative way
+        for i, (mean, std_dev) in enumerate(zip(mean_file, std_dev_file)):
```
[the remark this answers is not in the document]
JackTemaki: Well... this part gave me a lot of headaches in the past. The problem is the RETURNN tool only computes mean and variance per feature axis, but for TTS I needed it globally. So I did some research how I can transform multiple means and variances into a single mean and variance. I think I should add a test case for this to confirm it is correct. I tested it locally, but this might be better be somewhere explicit.

returnn/dataset.py
JackTemaki: Ok

curufinwe (approved):

PR #137, opened 2021-09-06
Fix FilterLexiconByWordListJob for gzip

### LINKED ISSUE #136: FilterLexiconByWordListJob does not work with gzip
https://github.com/rwth-i6/i6_core/blob/bbe3adfa821ccb02258794cc7401d9e08873a53e/lexicon/conversion.py#L78-L80

This write to a file doesn't work when the target file is a `.gz`.

mattiadg:
Fix #136

### DIFF
--- a/lexicon/conversion.py
+++ b/lexicon/conversion.py
@@ -75,7 +75,7 @@ class FilterLexiconByWordListJob(Job):
             ):
                 root.append(lemma)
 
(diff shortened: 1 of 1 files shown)


michelwi (approved):
`gzip.open` with mode "w" defaults to binary mode while `open` defaults to text mode.
Since text mode is required it is save always to specify "wt", even for `open`.

curufinwe (approved):

PR #144, opened 2021-09-16
Mfcc add output port

### LINKED ISSUE #143: mfcc_flow not creating output port "features" if normalize is True
https://github.com/rwth-i6/i6_core/blob/d27320747b0d7ebe40b8fec6a96cdacb7a961aa5/features/mfcc.py#L104
When normalize is `True`, the output of this network is connected to `network:features` but it is never defined.

mattiadg:
Fix #143

### DIFF
--- a/corpus/filter.py
+++ b/corpus/filter.py
@@ -37,12 +37,12 @@ class FilterSegmentsByListJob(Job):
         self.invert_match = invert_match
 
         num_segment_lists = len(self.segment_files)
-        self.single_segment_files = dict(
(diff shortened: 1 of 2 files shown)


michelwi (approved):
In the else part the output is created by the interconnect_output function. Here, I guess we have to do it manually.

curufinwe:
See comments in the issue. This needs to be solved by a flag as the proposed solution will break hashes.

curufinwe (approved):

PR #148, opened 2021-09-23
Add concatenate strategy merge corpora job

### LINKED ISSUE #146: MergeCorporaJob with MergeStrategy.FLAT produces invalid corpora if subcorpora are present
When using MergeStrategy.FLAT with corpus files structured as : 
corpus 
  subcorpus
    speaker-description*
    recording*

the result eliminates the `subcorpus` tags, doesn't copy the `speaker-description`s and loses the attribute name of subcorpus.
This results in a corpus file that is not compatible with the source files.

After discussing with @curufinwe I'm going to add a new strategy MergeStrategy.CONCATENATE that concatenates the subcorpus tags so that everything stays consistent with the source.

mattiadg:
Fix #146

### DIFF
--- a/corpus/transform.py
+++ b/corpus/transform.py
@@ -351,6 +351,8 @@ class MergeCorporaJob(Job):
                     merged_corpus.add_subcorpus(subcorpus)
                 for rec in c.top_level_recordings():
                     merged_corpus.add_recording(rec)
+                for speaker in c.top_level_speakers():
+                    merged_corpus.add_speaker(speaker)
             else:
                 assert False, "invalid merge strategy"
 
--- a/corpus/transform.py
+++ b/corpus/transform.py
@@ -311,6 +311,7 @@ class CompressCorpusJob(Job):
 class MergeStrategy(enum.Enum):
     SUBCORPORA = 0
     FLAT = 1
+    CONCATENATE = 2
 
 
 class MergeCorporaJob(Job):
@@ -345,6 +346,11 @@ class MergeCorporaJob(Job):
                 for rec in c.all_recordings():
                     merged_corpus.add_recording(rec)
                 merged_corpus.speakers.update(c.speakers)
+            elif self.merge_strategy == MergeStrategy.CONCATENATE:
+                for subcorpus in c.all_subcorpora():
+                    merged_corpus.add_subcorpus(subcorpus)
+                for rec in c.top_level_recordings():
+                    merged_corpus.add_recording(rec)
             else:
                 assert False, "invalid merge strategy"
 
--- a/corpus/transform.py
+++ b/corpus/transform.py
@@ -347,7 +347,7 @@ class MergeCorporaJob(Job):
                     merged_corpus.add_recording(rec)
                 merged_corpus.speakers.update(c.speakers)
             elif self.merge_strategy == MergeStrategy.CONCATENATE:
-                for subcorpus in c.all_subcorpora():
+                for subcorpus in c.top_level_subcorpora():
                     merged_corpus.add_subcorpus(subcorpus)
                 for rec in c.top_level_recordings():
                     merged_corpus.add_recording(rec)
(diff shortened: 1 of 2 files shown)


lib/corpus.py
```diff
@@ -170,6 +170,10 @@ def segments(self):
         for sc in self.subcorpora:
             yield from sc.segments()
 
+    def all_subcorpora(self):
```
curufinwe, line 173: this would be `top_level_subcorpora`, not `subcorpora`(as subcorpora can be nested).

curufinwe (changes requested):

lib/corpus.py
```diff
@@ -180,6 +180,12 @@ def all_speakers(self):
         for sc in self.subcorpora:
             yield from sc.all_speakers()
 
+    def top_level_recordings(self):
+        yield from self.recordings
+
+    def top_level_subcorpora(self):
+        yield from self.subcorpora
+
```
michelwi, line 188: Suggested implementation for `top_level_speakers()` (see prev comment)
```suggestion

    def top_level_speakers(self):
        yield from self.speakers.values()
```

Only semi related to this PR, but if `self.speakers` is a dict, `yield from self.speakers` would give the keys (speaker name) and not the values (speaker object), as would be necessary, right?
This should be fixed in `all_speakers` as well.

corpus/transform.py
```diff
@@ -345,6 +346,11 @@ def run(self):
                 for rec in c.all_recordings():
                     merged_corpus.add_recording(rec)
                 merged_corpus.speakers.update(c.speakers)
+            elif self.merge_strategy == MergeStrategy.CONCATENATE:
+                for subcorpus in c.top_level_subcorpora():
+                    merged_corpus.add_subcorpus(subcorpus)
+                for rec in c.top_level_recordings():
+                    merged_corpus.add_recording(rec)
```
michelwi, line 353: Do we also have to add the top level speakers?
```suggestion
                    merged_corpus.add_recording(rec)
                for speaker in c.top_level_speakers():
                    merged_corpus.add_speaker(speaker)
```

(with a suitable implementation of `Corpus.top_level_speakers()`)

lib/corpus.py
mattiadg: Related to this question, why is `self.speakers` a dictionary, if the value also contain the key (`speaker.name`)?

mattiadg: Maybe this should be discussed in a new issue and not here

mattiadg:
@curufinwe do you see more problems here?

lib/corpus.py
curufinwe: `top_level_speakers` should behave the same as `all_speakers`. It's very counter-intuitive that one returns values and the other keys. And as `all_speakers` already exists the one that needs to be changed is `top_level_speakers`.

mattiadg: Are you sure that `all_speakers` is not buggy as it is now?

michelwi: I agree that `top_level_speakers` should behave the same as `all_speakers`. However, I think the current way how `all_speakers` works is wrong:

`all_speakers` returns the keys of `self.speakers` which are of type `str`. 
But `Corpus.add_speaker` expects a `Speaker` object (in fact there is an assertion), so the code
```
for speaker in c.all_speakers():
     merged_corpus.add_speaker(speaker)
```
will not work.

I therefore propose to check if `all_speakers` is already used (and if returning the key is expected) and if not change both functions to return the values instead.

curufinwe: Looks like the `all_speakers` code was added here: https://github.com/rwth-i6/i6_core/pull/43 And I did not find a use in i6_core or the AppTek repo for it, so it's probably safe to change it.

mattiadg:
Shall I also do it in this PR?

Il ven 24 set 2021, 18:36 Eugen Beck ***@***.***> ha scritto:

lib/corpus.py
mattiadg: ok, so I probably didn't realize that `self.speakers` is a dict, thus the implementation

curufinwe (approved):

PR #151, opened 2021-09-24
Add search_numpy_blas and blas_lib parameters to CompileNativeOpJob

curufinwe:
tk.uncached_path used for blas_lib because the location in my setup is in a singularity image and thus the external sisyphus manager cannot check for it.

### DIFF
--- a/returnn/compile.py
+++ b/returnn/compile.py
@@ -116,16 +116,22 @@ class CompileNativeOpJob(Job):
     Compile a RETURNN native op into a shared object file.
     """
 
+    __sis_hash_exclude__ = {"search_numpy_blas": True, "blas_lib": None}
+
     def __init__(
         self,
         native_op,
         returnn_python_exe=None,
         returnn_root=None,
+        search_numpy_blas=True,
+        blas_lib=None,
     ):
         """
         :param str native_op: Name of the native op to compile (e.g. NativeLstm2)
         :param Path|str returnn_python_exe: file path to the executable for running returnn (python binary or .sh)
         :param Path|str returnn_root: file path to the RETURNN repository root folder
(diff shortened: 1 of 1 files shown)


returnn/compile.py
```diff
@@ -116,16 +116,22 @@ class CompileNativeOpJob(Job):
...
         returnn_root=None,
+        search_numpy_blas=True,
+        blas_lib=None,
     ):
         """
         :param str native_op: Name of the native op to compile (e.g. NativeLstm2)
         :param Path|str returnn_python_exe: file path to the executable for running returnn (python binary or .sh)
         :param Path|str returnn_root: file path to the RETURNN repository root folder
+        :param bool search_numpy_blas: search for blas lib in numpy's .libs folder
+        :param Path|str blas_lib: explicit path to the blas library to use
```
JackTemaki, line 134: Can we not establish the convention that paths always have to be `tk.Path`? Then you also do not need the `tk.uncached_path()` below, and for constant stuff you can just wrap it with `Path(...)` in the setup...

JackTemaki (changes requested):

returnn/compile.py
curufinwe: In this case I cannot pass a path from the outside as the blas_lib is only inside a singularity container and not visible from the manager.

curufinwe: i.e. the manager will complain about missing input, but the Path is there during the runtime of the job.

JackTemaki: Then it is okay for this case, but maybe we should add a "no check" flag to a Path object then...

JackTemaki (approved):

returnn/compile.py
curufinwe: Yes, that would be a more principled thing.

PR #152, opened 2021-09-24
Do not try to create ARPA images that already exist in AdvancedTreeSearchLmImageAndGlobalCacheJob

curufinwe:


recognition/advanced_tree_search.py
```diff
@@ -77,12 +77,24 @@ def cleanup_before_run(self, cmd, retry, *args):
...
         elif lm_config.type == "combine":
             for i in range(1, lm_config.num_lms + 1):
                 sub_lm_config = lm_config["lm-%d" % i]
-                if sub_lm_config.type == "ARPA":
+                sub_lm_config = (
+                    lm_post_config["lm-%d" % i] if lm_post_config is not None else None
+                )
+                if sub_lm_config.type == "ARPA" and not has_image(
+                    sub_lm_config, sub_lm_post_config
+                ):
```
michelwi, line 97: Maybe more elegant and also catering for nested "combine" LMs
```suggestion
                result.append(find_arpa_lms(sub_lm_config,sub_lm_post_config))
```
(plus delete the following `result.append` part.)

michelwi (changes requested):
typo config->post_config

plus one suggestion for consideration.

recognition/advanced_tree_search.py
```diff
@@ -77,20 +77,22 @@ def cleanup_before_run(self, cmd, retry, *args):
...
-                            sub_lm_config,
-                            lm_post_config["lm-%d" % i]
-                            if lm_post_config is not None
-                            else None,
-                        )
-                    )
+                sub_lm_post_config = (
+                    lm_post_config["lm-%d" % i] if lm_post_config is not None else None
+                )
+                result.append(find_arpa_lms(sub_lm_config, sub_lm_post_config))
```
michelwi, line 95: Did you test this?^^ 
I think it should be 
```suggestion
                result.append(cls.find_arpa_lms(sub_lm_config, sub_lm_post_config))
```

No wait. `find_arpa_lms` returns already a list (empty if no LMs are found), so we should not append but concatenate
```suggestion
                result += cls.find_arpa_lms(sub_lm_config, sub_lm_post_config)
```

(sorry, my error previously.)

michelwi (changes requested):

recognition/advanced_tree_search.py
curufinwe: I have not. Don't have a setup that uses the combine-lm case^^

michelwi (approved):
If you are confident about my changes, then I approve.

PR #154, opened 2021-09-29
fix for ExtractDatasetMeanStddevJob

JackTemaki:
I am not sure why it was wrong here, so this is now the fix...


returnn/oggzip.py
```diff
@@ -78,20 +81,26 @@ def run(self):
...
 
         sp.check_call(args)
         relink("out.ogg.zip", self.out_ogg_zip.get_path())
 
     @classmethod
     def hash(cls, parsed_args):
         del parsed_args["returnn_python_exe"]
+        # hash compatibility
+        if parsed_args["no_audio"] == False:
+            del parsed_args["no_audio"]
```
michelwi, line 105: Would you not rather do this via a `__sis_hash_exclude__` instead of coding in the hash function such that it can be found easier for future hash correction dates and/or is more obvious why the `no_audio == False` case is treated specially.

JackTemaki: It seems I combined the wrong commits into the branch :D This does not belong here...

michelwi (approved):

curufinwe (approved):

PR #155, opened 2021-10-01
Mark out file

mattiadg:
This job wasn't included in the renaming

### DIFF
--- a/corpus/costa.py
+++ b/corpus/costa.py
@@ -23,7 +23,7 @@ class CostaJob(rasr.RasrCommand, Job):
         )
         self.audio_flow = raw_audio_flow(crp.audio_format) if eval_recordings else None
 
-        self.log_file = self.output_path(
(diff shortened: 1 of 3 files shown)


michelwi (changes requested):
Please also change the name where it is used. I found these positions:
https://github.com/rwth-i6/i6_core/blob/0ca5462aa8ff7650390674a6c44a977128cedf85/meta/system.py#L425
https://github.com/rwth-i6/i6_core/blob/34f82dd09f932da9d6c28d9781d7f20d4df888af/corpus/costa.py#L51

mattiadg:
I didn't find other usages but I've found another class with exactly the same problem

michelwi (approved):
There are some more classes with this problem (cf. https://github.com/rwth-i6/i6_core/search?q=log_file), but these two seem to be corrected consistently.

curufinwe (approved):

PR #159, opened 2021-10-11
extend docstrings of lexicon lib

michelwi:
Adds some real documentation. Also two arguments (synt, eval) were wrong.

### DIFF
--- a/lib/lexicon.py
+++ b/lib/lexicon.py
@@ -21,10 +21,10 @@ class Lemma:
         :param Optional[list[str]] orth: list of spellings used in the training data
         :param Optional[list[str]] phon: list of pronunciation variants. Each str should
             contain a space separated string of phonemes from the phoneme-inventory.
-        :param Optional[list[list[str]]] synt: list of language model representations.
+        :param Optional[list[list[str]]] synt: list of token sequences, for which each token sequence is used as as possible language model representation.
             Each sublist should contain words from the language model vocabulary.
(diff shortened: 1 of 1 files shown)


lib/lexicon.py
```diff
@@ -18,11 +18,17 @@ class Lemma:
...
         """
-        :param Optional[list[str]] orth:
-        :param Optional[list[str]] phon:
-        :param Optional[list[str]] synt:
-        :param Optional[list[str]] eval:
-        :param Optional[str] special:
+        :param Optional[list[str]] orth: list of spellings used in the training data
+        :param Optional[list[str]] phon: list of pronunciation variants. Each str should
+            contain a space separated string of phonemes from the phoneme-inventory.
+        :param Optional[list[list[str]]] synt: list of language model representations.
```
JackTemaki, line 24: I still have no idea why this needs to be a list of lists....

michelwi: alright, feel free to "insert a suggestion" for a docstring that you find more informative for the user.

lib/lexicon.py
```diff
@@ -18,11 +18,17 @@ class Lemma:
...
+            Each sublist should contain words from the language model vocabulary.
+        :param Optional[list[list[str]]] eval: list of output representations. Each
+            sublist should contain one possible transcription of this lemma
```
JackTemaki, line 27:
```suggestion
            sublist should contain one possible transcription (token sequence) of this lemma
```

curufinwe (approved):

PR #160, opened 2021-10-12
System: add set_hub5_scorer

vieting:
The `Hub5ScoreJob` is already in the recipes, but `System.set_hub5_scorer` is missing.

### DIFF
--- a/meta/system.py
+++ b/meta/system.py
@@ -46,6 +46,7 @@ class System:
         self.cart_questions = {}
         self.normalization_matrices = {}
         self.stm_files = {}
+        self.glm_files = {}
 
         self.scorers = {}
         self.scorer_args = {}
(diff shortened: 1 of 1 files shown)


michelwi (approved):

meta/system.py
```diff
@@ -46,6 +46,7 @@ def __init__(self):
         self.cart_questions = {}
         self.normalization_matrices = {}
         self.stm_files = {}
+        self.glm_files = {}
```
JackTemaki, line 49: There is a `create_stm_from_corpus`, is there no need for something similar for the glm?

michelwi: > Because now you would need to set this by explicitely accesing the variable...

we could create a generic setter function for the glm files if you prefere it.

JackTemaki: > we could create a generic setter function for the glm files if you prefere it.

I would suggest this or have a verbose error message if the glm file is not set yet

JackTemaki: But still we can merge this now

vieting: > I would suggest this or have a verbose error message if the glm file is not set yet

I added that. If you have no objections to the message, we can merge the PR.

Issue #166, opened 2021-10-20
Set operations on word lists?


mattiadg:
I have two files, each containing a list of words (two vocabularies). I want to perform set difference between their words, is there any Job in here that allows me to do it?

michelwi:
It would probably live in https://github.com/rwth-i6/i6_core/blob/main/text/processing.py and there is only
https://github.com/rwth-i6/i6_core/blob/8e2dcbc9310ee823fec5b75bdad745a718c9ff1c/text/processing.py#L1


You could misuse a pipeline job via
```
cat file1 | comm -23 - file2 > output.txt
```
but then `file2` wouldn't be an input but part of the pipeline, which might put the absolute path in the hash if you are not careful.

JackTemaki:
Just as a note, you can add inputs to every step of the Pipeline!

```python
    tokenizer_command = DelayedFormat("{} -a -l en -no-escape", tokenizer_tts)
    [...]
    tokenized_text = PipelineJob(
        bliss_text,
        [whitelist_pipe,
         tokenizer_command,
    [...]
```
`DelayedFormat` does the job! Here it is a binary that is added dynamically, but it can also be other files.

PR #168, opened 2021-10-21
Vocab difference

mattiadg:
Computes the set difference between two files. I didn't subclass PipelineJob (as discussed in #166) because it always starts with `zcat input` and I didn't work to do workarounds for such a simple job.

### DIFF
--- a/text/processing.py
+++ b/text/processing.py
@@ -211,9 +211,16 @@ class SetDifferenceJob(Job):
     Return the set difference of two text files, where one line is one element.
     """
 
-    def __init__(self, file1, file2, gzipped=False):
-        self.file1 = file1
-        self.file2 = file2
+    def __init__(self, minuend, subtrahend, gzipped=False):
+        """
+        This job performs the set difference minuend - subtrahend. Unlike the bash utility comm, the two files
+        do not need to be sorted.
+        :param Path minuend: left-hand side of the set subtraction
+        :param Path subtrahend: right-hand side of the set subtraction
+        :param bool gzipped: whether the output should be compressed in gzip format
+        """
+        self.minuend = minuend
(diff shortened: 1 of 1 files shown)


text/processing.py
```diff
@@ -202,3 +204,28 @@ def run(self):
             self.lines = int(length * self.ratio)
 
         self.sh("zcat -f {text_file} | tail -n {num_lines} | gzip > {out}")
+
+
+class DiffJob(Job):
```
michelwi, line 209: I think we better name it differently (e.g. `SetDifferenceJob`) because after `HeadJob` and `TailJob` on might assume that `DiffJob` would be a wrapper around the GNU diff utillity.

text/processing.py
```diff
@@ -202,3 +204,28 @@ def run(self):
...
+    """
+    Return the set difference of two text files, where one line is one element.
+    """
+
+    def __init__(self, file1, file2):
```
michelwi, line 214: pls. add a doc string for file1 and file2
e.g. so that the user knows the diff is file1 - file2 and not the other way round.

Maybe there is better names to make sure which file is which (e.g. minuend and subtrahend - ok, this is set difference not subtraction so not entirely correct nomenclature)... But we can also leave it like this.

text/processing.py
```diff
@@ -202,3 +204,28 @@ def run(self):
...
+        self.file1 = file1
+        self.file2 = file2
+
+        self.out_file = self.output_path("diff.txt")
+
+        self.rqmt = {"cpu": 1, "time": 1, "mem": 1}
```
michelwi, line 220: If we want to use this job on larger files, then maybe we can think about allowing gzipped outputs and custom rqmt.

michelwi (approved):

PR #170, opened 2021-11-04
extension of concat job

papar22:
Here is an extension to "ConcatenateJob: to support:

1. gzip or raw output
2. user-specific name to be comparable with RASR

### DIFF
--- a/text/processing.py
+++ b/text/processing.py
@@ -134,7 +134,7 @@ class ConcatenateJob(Job):
             if zip_out:
                 self.out = self.output_path(out_name + ".gz")
             else:
-                self.out = self.output_path("out_name")
+                self.out = self.output_path(out_name)
 
         for input in text_files:
             assert isinstance(input, Path) or isinstance(
--- a/text/processing.py
+++ b/text/processing.py
@@ -112,9 +112,11 @@ class ConcatenateJob(Job):
     Concatenate all given input files (gz or raw)
     """
 
-    def __init__(self, text_files):
+    def __init__(self, text_files, zip_out=True, out_name="out"):
         """
         :param list[Path] text_files: input text files
+        :param bool zip_out: apply gzip to the output
+        :param str out_name: user specific name
         """
         assert text_files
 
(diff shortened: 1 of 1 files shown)


text/processing.py
```diff
@@ -129,21 +131,28 @@ def __init__(self, text_files):
         if len(text_files) == 1:
             self.out = text_files.pop()
         else:
-            self.out = self.output_path("out.gz")
+            if zip_out:
+                self.out = self.output_path(out_name + ".gz")
+            else:
+                self.out = self.output_path("out_name")
```
JackTemaki, line 137:
```suggestion
                self.out = self.output_path(out_name)
```

JackTemaki (changes requested):

JackTemaki:
>     2\. user-specific name to be comparable with RASR

Just out of curiosity, why is this needed?

papar22:
> > ```
> > 2\. user-specific name to be comparable with RASR
> > ```
> 
> Just out of curiosity, why is this needed?

to create "bundle" files which is a hard-coded name in RASR as far as I know.

JackTemaki (approved):

text/processing.py
```diff
@@ -112,9 +112,11 @@ class ConcatenateJob(Job):
     Concatenate all given input files (gz or raw)
     """
 
-    def __init__(self, text_files):
+    def __init__(self, text_files, zip_out=True, out_name="out"):
```
michelwi, line 115: Do we need an hash exclude for the new parameters or is this job sufficiently unused?

JackTemaki: Oh missed this, yes for sure this is required!

JackTemaki (changes requested):
See @michelwi comment

text/processing.py
papar22: Why do we have no has exclude for the "out_zip" in PipelineJob `https://github.com/rwth-i6/i6_core/blob/main/text/processing.py#L103` ? The output format can change the hash?

JackTemaki: By hash exclude we do not mean to exclude `out_zip` from the hash computation, but to exclude the old default. Example:
```python3
j = SomeJob(old_feature=xyz)
```
And now you have
```python3
j = SomeJob(old_feature=xyz, new_feature=False)
```
You want the same hash for both.

This is why you need to include:
```python3
__sis_hash_exclude = {'new_feature': False}
```

And yes, the output format should definitely change the hash, because you are getting a different file.

papar22: got it! thanks

JackTemaki (approved):

michelwi (approved):

curufinwe (approved):

PR #173, opened 2021-11-05
Add doc

christophmluscher:
add doc string for functions

### DIFF
--- a/meta/system.py
+++ b/meta/system.py
@@ -57,9 +57,10 @@ class System:
     def set_corpus(self, name, corpus, concurrent, segment_path=None):
         """
         :param str name:
-        :param tk.Object corpus:
+        :param CorpusObject corpus:
         :param int concurrent:
         :param util.MultiOutputPath segment_path:
+        :return:
         """
         self.crp[name] = rasr.CommonRasrParameters(base=self.crp["base"])
         rasr.crp_set_corpus(self.crp[name], corpus)
@@ -83,6 +84,11 @@ class System:
         self.jobs[name] = {}
 
     def add_overlay(self, origin, name):
+        """
+        :param str origin: name of the original common rasr parameters with associated alignments, features, etc
+        :param str name: name of the new overlay over the original
+        :return:
+        """
         self.crp[name] = rasr.CommonRasrParameters(base=self.crp[origin])
 
         self.alignments[name] = {}
@@ -107,6 +113,7 @@ class System:
         :param System origin_system:
         :param str origin_name: Name of the dataset in another system
         :param str target_name: Name of the dataset in current system (optional) (default=origin_name)
+        :return:
         """
         if not target_name:
             target_name = origin_name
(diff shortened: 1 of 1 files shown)


meta/system.py
```diff
@@ -487,6 +603,16 @@ def align(self, name, corpus, flow, feature_scorer, **kwargs):
     def estimate_mixtures(
         self, name, corpus, flow, old_mixtures=None, alignment=None, prefix="", **kwargs
     ):
+        """
+        :param str name:
+        :param str corpus:
+        :param str flow:
+        :param str|list[str]|tuple[str]|rasr.FlagDependentFlowAttribute old_mixtures:
```
michelwi, line 610: Sure that `rasr.FlagDependentFlowAttribute` describes a mixture file?

meta/system.py
```diff
@@ -647,6 +807,19 @@ def recog_and_optimize(
         prefix="",
         **kwargs,
     ):
+        """
+        :param str name:
+        :param str corpus:
+        :param str|list[str]|tuple[str]|rasr.FlagDependentFlowAttribute flow:
+        :param str|list[str]|tuple[str]|rasr.FeatureScorer feature_scorer:
+        :param float pronunciation_scale:
+        :param  floatlm_scale:
```
michelwi, line 816: typo
```suggestion
        :param float lm_scale:
```

meta/system.py
```diff
@@ -727,6 +912,18 @@ def create_nn_feature_scorer(
         prior_file=None,
         **kwargs,
     ):
+        """
+        :param str name:
+        :param str corpus:
+        :param int|types.FunctionType feature_dimension:
+        :param int|types.FunctionType output_dimension:
+        :param str|list[str]|tuple[str]|rasr.FlagDependentFlowAttribute prior_mixtures:
```
michelwi, line 920: is `FlagDependentFlowAttribute` correct for mixture file?

michelwi (changes requested):

meta/system.py
```diff
@@ -647,6 +807,19 @@ def recog_and_optimize(
...
     ):
+        """
+        :param str name:
+        :param str corpus:
+        :param str|list[str]|tuple[str]|rasr.FlagDependentFlowAttribute flow:
+        :param str|list[str]|tuple[str]|rasr.FeatureScorer feature_scorer:
+        :param float pronunciation_scale:
+        :param  floatlm_scale:
+        :param bool parallelize_conversion:
+        :param dictlattice_to_ctm_kwargs:
```
michelwi, line 818: typo
```suggestion
        :param dict lattice_to_ctm_kwargs:
```

meta/system.py
christophmluscher: You are right. It is not. It should be Sisyphus Path object

michelwi (approved):

JackTemaki (approved):
I am not sure I would find errors even if there are, so this is just an approve from me...

PR #174, opened 2021-11-10
fix speaker gender tag

JackTemaki:
There was an error in the speaker tags that nobody noticed, which causes warnings when running RASR. I will fix this manually in all jobs that are already stored openly for everyone.

### DIFF
--- a/datasets/librispeech.py
+++ b/datasets/librispeech.py
@@ -169,7 +169,7 @@ class LibriSpeechCreateBlissCorpusJob(Job):
                 continue
             speaker = corpus.Speaker()
             speaker.name = speaker_id
-            speaker.attribs["gender"] = speaker_info[0].lower()
+            speaker.attribs["gender"] = "male" if speaker_info[0] == "M" else "female"
             c.add_speaker(speaker)
 
         c.dump(self.out_corpus.get_path())


datasets/librispeech.py
```diff
@@ -169,7 +169,7 @@ def run(self):
                 continue
             speaker = corpus.Speaker()
             speaker.name = speaker_id
-            speaker.attribs["gender"] = speaker_info[0].lower()
+            speaker.attribs["gender"] = ["male" if speaker_info[0] == "M" else "female"]
```
christophmluscher, line 172: I think this should be a `str` not a `list[str]`, right?

christophmluscher (changes requested):

datasets/librispeech.py
JackTemaki: Oh... yes sure

vieting (commented):
I never used the gender attribute, if you need to have `"male"` and `"female"` then this makes sense of course. Just one question: Is it guaranteed that `speaker_info[0]` will always be `"M"` or `"F"`? You might consider an assertion here.

JackTemaki:
> Is it guaranteed
Yes, because this is taken from the LibriSpeech metadata file, this is a very fixed case.

vieting (approved):

christophmluscher (approved):

PR #175, opened 2021-11-12
Separate load lm

Nintendo-Wei:
Pre-compute LM image without generating global cache

### DIFF
--- a/lm/lm_image.py
+++ b/lm/lm_image.py
@@ -20,26 +20,18 @@ class LmImageJob(rasr.RasrCommand, Job):
         extra_config=None,
         extra_post_config=None,
         encoding="utf-8",
-        renormalize=False,
+        mem=2,
     ):
         kwargs = locals()
         del kwargs["self"]
 
-        self.text_file = text_file
-        self.renormalize = renormalize
-
-        self.config, self.post_config, self.num_images = LmImageJob.create_config(
-            **kwargs
-        )
+        self.config, self.post_config = LmImageJob.create_config(**kwargs)
         self.exe = self.select_exe(crp.lm_util_exe, "lm-util")
 
         self.log_file = self.log_file_output_path("lm_image", crp, False)
-        self.lm_images = {
-            i: self.output_path(f"lm-{i}.image", cached=True)
-            for i in range(1, self.num_images + 1)
-        }
+        self.out = self.output_path("lm.image", cached=True)
 
-        self.rqmt = {"time": 1, "cpu": 1, "mem": 2}
+        self.rqmt = {"time": 1, "cpu": 1, "mem": mem}
 
     def tasks(self):
         yield Task("create_files", resume="create_files", mini_task=True)
@@ -59,32 +51,9 @@ class LmImageJob(rasr.RasrCommand, Job):
 
     def run(self):
         self.run_script(1, self.log_file)
-        for i in range(1, self.num_images + 1):
-            shutil.move(f"lm-{i}.image", self.lm_images[i].get_path())
 
     def cleanup_before_run(self, cmd, retry, *args):
         util.backup_if_exists("lm_image.log")
-
-    @classmethod
-    def find_arpa_lms(cls, lm_config, lm_post_config=None):
-        result = []
-
-        def has_image(c, pc):
-            res = c._get("image") is not None
-            res = res or (pc is not None and pc._get("image") is not None)
-            return res
-
-        if lm_config.type == "ARPA":
-            if not has_image(lm_config, lm_post_config):
-                result.append((lm_config, lm_post_config))
-        elif lm_config.type == "combine":
-            for i in range(1, lm_config.num_lms + 1):
-                sub_lm_config = lm_config["lm-%d" % i]
-                sub_lm_post_config = (
-                    lm_post_config["lm-%d" % i] if lm_post_config is not None else None
-                )
-                result += cls.find_arpa_lms(sub_lm_config, sub_lm_post_config)
-        return result
 
     @classmethod
     def create_config(
@@ -93,7 +62,6 @@ class LmImageJob(rasr.RasrCommand, Job):
         extra_config,
         extra_post_config,
         encoding,
-        renormalize,
         **kwargs,
     ):
         config, post_config = rasr.build_config_from_mapping(
@@ -104,17 +72,15 @@ class LmImageJob(rasr.RasrCommand, Job):
         )  # scale not considered here, delete to remove ambiguity
 
         config.lm_util.action = "load-lm"
-        config.lm_util.file = crp.language_model_config.file
         config.lm_util.encoding = encoding
         config.lm_util.batch_size = 100
-        config.lm_util.renormalize = renormalize
 
         config._update(extra_config)
         post_config._update(extra_post_config)
 
-        arpa_lms = cls.find_arpa_lms(
-            config,
-            post_config if post_config else None,
-        )
+        return config, post_config
 
-        return config, post_config, len(arpa_lms)
+    @classmethod
+    def hash(cls, kwargs):
+        config, post_config = cls.create_config(**kwargs)
+        return super().hash({"config": config, "exe": kwargs["crp"].lm_util_exe})
--- a/lm/lm_image.py
+++ b/lm/lm_image.py
@@ -10,10 +10,13 @@ Path = setup_path(__package__)
 
 
 class LmImageJob(rasr.RasrCommand, Job):
+    """
+    pre-compute LM image without generating global cache
+    """
+
     def __init__(
         self,
         crp,
-        text_file,
         extra_config=None,
         extra_post_config=None,
         encoding="utf-8",
@@ -44,10 +47,13 @@ class LmImageJob(rasr.RasrCommand, Job):
 
     def create_files(self):
         self.write_config(self.config, self.post_config, "lm_image.config")
-        device = "gpu" if self.rqmt.get("gpu", 0) > 0 else "cpu"
         extra_code = (
-            f":${{THEANO_FLAGS:="
-            '}\nexport THEANO_FLAGS="$THEANO_FLAGS,device={device},force_device=True"\nexport TF_DEVICE="{device}"'
+            ":${{THEANO_FLAGS:="
+            "}}\n"
+            'export THEANO_FLAGS="$THEANO_FLAGS,device={0},force_device=True"\n'
+            'export TF_DEVICE="{0}"'.format(
+                "gpu" if self.rqmt.get("gpu", 0) > 0 else "cpu"
+            )
         )
         self.write_run_script(self.exe, "lm_image.config", extra_code=extra_code)
 
@@ -84,7 +90,6 @@ class LmImageJob(rasr.RasrCommand, Job):
     def create_config(
         cls,
         crp,
-        text_file,
         extra_config,
         extra_post_config,
         encoding,
@@ -99,7 +104,7 @@ class LmImageJob(rasr.RasrCommand, Job):
         )  # scale not considered here, delete to remove ambiguity
 
         config.lm_util.action = "load-lm"
-        config.lm_util.file = text_file
+        config.lm_util.file = crp.language_model_config.file
         config.lm_util.encoding = encoding
         config.lm_util.batch_size = 100
         config.lm_util.renormalize = renormalize
@@ -107,6 +112,9 @@ class LmImageJob(rasr.RasrCommand, Job):
         config._update(extra_config)
         post_config._update(extra_post_config)
 
-        arpa_lms = cls.find_arpa_lms(config, post_config if post_config else None,)
+        arpa_lms = cls.find_arpa_lms(
+            config,
+            post_config if post_config else None,
+        )
 
         return config, post_config, len(arpa_lms)
--- a/lm/lm_image.py
+++ b/lm/lm_image.py
@@ -1,6 +1,7 @@
 __all__ = ["LmImageJob"]
 
 import shutil
+import logging
 
 from sisyphus import *
 from i6_core import rasr
@@ -73,6 +74,11 @@ class LmImageJob(rasr.RasrCommand, Job):
         )  # scale not considered here, delete to remove ambiguity
 
         assert config.lm_util.lm.type == "ARPA"
+
+        if post_config.lm_util.lm.image:
+            logging.warning(
+                "The LM image already exists, but a new one will be recreated."
+            )
 
         config.lm_util.action = "load-lm"
         config.lm_util.lm.image = "lm.image"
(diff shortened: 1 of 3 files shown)


lm/lm_image.py
```diff
@@ -0,0 +1,120 @@
...
+        crp,
+        extra_config=None,
+        extra_post_config=None,
+        encoding="utf-8",
+        renormalize=False,
+    ):
+        kwargs = locals()
+        del kwargs["self"]
+
+        self.text_file = text_file
```
michelwi, line 28: `text_file` is not defined? Is this a missing argument for `__init__`?

(edit:) Does not seem to be used. Can this line be deleted?

lm/lm_image.py
```diff
@@ -0,0 +1,120 @@
...
+        self.renormalize = renormalize
+
+        self.config, self.post_config, self.num_images = LmImageJob.create_config(
+            **kwargs
+        )
+        self.exe = self.select_exe(crp.lm_util_exe, "lm-util")
+
+        self.log_file = self.log_file_output_path("lm_image", crp, False)
+        self.lm_images = {
```
michelwi, line 37: if this is an output of the Job, maybe we want to prefix the member with "out_"?
```suggestion
        self.out_lm_images = {
```

lm/lm_image.py
```diff
@@ -0,0 +1,120 @@
...
+    pre-compute LM image without generating global cache
+    """
+
+    def __init__(
+        self,
+        crp,
+        extra_config=None,
+        extra_post_config=None,
+        encoding="utf-8",
+        renormalize=False,
```
michelwi, line 23: What does `renormalize` do here? I mean it will somehow normalize the LM, but why do we have it as parameter in this Job, whos job it is to compute the image to an existing LM?

Will the LM file be modified? Will only the image be normalized? Is the image now inconsistent to the text file?

In the favor of "one Job should do one thing" I would argue that we can have a NormalizeLmJob if need be, and that the LmImageJob only does write out the image.

lm/lm_image.py
```diff
@@ -0,0 +1,120 @@
...
+        )
+        self.exe = self.select_exe(crp.lm_util_exe, "lm-util")
+
+        self.log_file = self.log_file_output_path("lm_image", crp, False)
+        self.lm_images = {
+            i: self.output_path(f"lm-{i}.image", cached=True)
+            for i in range(1, self.num_images + 1)
+        }
+
+        self.rqmt = {"time": 1, "cpu": 1, "mem": 2}
```
michelwi, line 42: Maybe mem rqmt should be exposed to the user because I guess it can be different depending on the size of the LM to be processed.

lm/lm_image.py
```diff
@@ -0,0 +1,120 @@
...
+        config.lm_util.batch_size = 100
+        config.lm_util.renormalize = renormalize
+
+        config._update(extra_config)
+        post_config._update(extra_post_config)
+
+        arpa_lms = cls.find_arpa_lms(
+            config,
+            post_config if post_config else None,
+        )
```
michelwi, line 118: This reads as if this job takes a LM-config and prepares images for all the LMs. 
Would it not be smarter/more efficient/more modular if this job only ever did one image at a time:

E.g. if you have a 3 LMs and want to do all possible combinations (1, 2, 3, 1+2, 1+3, 2+3, 1+2+3) , then there would be 7 different `LmImageJob`s (because the config differs) and each individual image would be written 3 times.

So better to have the Recognition Job do the logic of determining the number and identity of the LMs and then create n separate single `LmImageJob`s

lm/lm_image.py
```diff
@@ -0,0 +1,120 @@
...
+    ):
+        config, post_config = rasr.build_config_from_mapping(
+            crp, {"lexicon": "lm-util.lexicon", "language_model": "lm-util.lm"}
+        )
+        del (
+            config.lm_util.lm.scale
+        )  # scale not considered here, delete to remove ambiguity
+
+        config.lm_util.action = "load-lm"
+        config.lm_util.file = crp.language_model_config.file
```
michelwi, line 107: This is probably due to the RASR implementation of lm-util, but why the explicit `config.lm_util.file` and the implicit (via `rasr.build_config_from_mapping` and `crp.language_model_config`) `config.lm_util.lm.file`? What happens if there are multiple LMs defined (e.g. combined LM with sub-LMs) what if the LM is not in ARPA format and does not have a `.file`?

lm/lm_image.py
```diff
@@ -0,0 +1,120 @@
...
+import shutil
+
+from sisyphus import *
+from i6_core import rasr
+from i6_core import util
+
+Path = setup_path(__package__)
+
+
+class LmImageJob(rasr.RasrCommand, Job):
```
michelwi, line 12: This Job does not have an explicit hash function, therefore all parameters of the `__init__` function will be included in the hash, including `extra_post_config`.

But also you would not want to hash all entries of the crp but only those relevant to the LM image.

michelwi (changes requested):
I think there are some design problems with this job we should discuss or I misunderstand your intentions.

See inline comments for further discussion points.

lm/lm_image.py
Nintendo-Wei: Yes, this should be deleted, sorry

lm/lm_image.py
Nintendo-Wei: Sounds fair, I will remove this parameter.

lm/lm_image.py
Nintendo-Wei: Will add it!

lm/lm_image.py
Nintendo-Wei: Thanks for the advise! I will change the logic here

lm/lm_image.py
Nintendo-Wei: I was not sure about this part, I thought `file` was a necessary parameter for the `lm_util` job, so I added it ... It seemed weird to me too...

Nintendo-Wei: @michelwi Is this `file` parameter not a necessary one for `lm_util` in `RASR`?

lm/lm_image.py
```diff
@@ -0,0 +1,86 @@
...
+        mem=2,
+    ):
+        kwargs = locals()
+        del kwargs["self"]
+
+        self.config, self.post_config = LmImageJob.create_config(**kwargs)
+        self.exe = self.select_exe(crp.lm_util_exe, "lm-util")
+
+        self.log_file = self.log_file_output_path("lm_image", crp, False)
+        self.out = self.output_path("lm.image", cached=True)
```
michelwi, line 32: In my personal opinion, the output member should have a more descriptive name than simply `out`; e.g. `out_image` or `out_lm_image`

lm/lm_image.py
```diff
@@ -0,0 +1,86 @@
...
+            ":${{THEANO_FLAGS:="
+            "}}\n"
+            'export THEANO_FLAGS="$THEANO_FLAGS,device={0},force_device=True"\n'
+            'export TF_DEVICE="{0}"'.format(
+                "gpu" if self.rqmt.get("gpu", 0) > 0 else "cpu"
+            )
+        )
+        self.write_run_script(self.exe, "lm_image.config", extra_code=extra_code)
+
+    def run(self):
```
michelwi, line 52: I guess you still need one move statement for the one LM image you are creating.

michelwi (changes requested):

lm/lm_image.py
michelwi: I do not know as I never used it. From https://github.com/rwth-i6/rasr/blob/843d1eb5f2137b57dcc6cf4bc9dce44e85d66f27/src/Tools/Lm/LmUtilityTool.cc#L93 as well as Lines 132 and following, I gather that the `config.lm_util.file` is a plain text file, that is used to compute the perplexities of the given LM on.

Also from https://github.com/rwth-i6/rasr/blob/843d1eb5f2137b57dcc6cf4bc9dce44e85d66f27/src/Tools/Lm/LmUtilityTool.cc#L119-L123 I don't see the action `"load-lm"` defined. Did/Will you/someone implement this functionality?

michelwi: maybe @curufinwe who wrote the lm-util tool can help clarify.

Nintendo-Wei: I am using RASR in `/home/ebeck/dev/full_rasr/rasr/` and there we can find a `load-lm` at line 91 in `/home/ebeck/dev/full_rasr/rasr/src/Tools/Lm/LmUtilityTool.cc`. Maybe the changes haven't been pushed yet?

michelwi: no, the option does exist in the closed source repo of AppTek. If you want to use it here, we should ask permission to port it to the public repo. @curufinwe

lm/lm_image.py
```diff
@@ -0,0 +1,87 @@
...
+        self.run_script(1, self.log_file)
+        shutil.move("lm.image.1", self.out_image.get_path())
```
michelwi, line 54: please verify that this is (always) the correct name according to the config

michelwi (commented):
I think this PR is good now. But since the functionality does not exist in the official RASR repo I will withhold acceptance until it is backported.

lm/lm_image.py
```diff
@@ -0,0 +1,87 @@
...
+        mem=2,
+    ):
+        kwargs = locals()
+        del kwargs["self"]
+
+        self.config, self.post_config = LmImageJob.create_config(**kwargs)
+        self.exe = self.select_exe(crp.lm_util_exe, "lm-util")
+
+        self.log_file = self.log_file_output_path("lm_image", crp, False)
+        self.out_image = self.output_path("lm.image", cached=True)
```
curufinwe, line 32: The number of images and their location can vary. I'd use the code from recognition/advanced_tree_search.py here to detect the images correctly and create all of them. In addition to the image paths themselves it would be good to have an updated config here as output (i.e. the original lm config + a post config with the images set correctly).

curufinwe (changes requested):

lm/lm_image.py
michelwi: There used to be logic to do just that and then I argued against it. Here is the original argument from the other thread:

>This reads as if this job takes a LM-config and prepares images for all the LMs.
Would it not be smarter/more efficient/more modular if this job only ever did one image at a time:
>
>E.g. if you have a 3 LMs and want to do all possible combinations (1, 2, 3, 1+2, 1+3, 2+3, 1+2+3) , then there would be 7 different LmImageJobs (because the config differs) and each individual image would be written 3 times.
>
>So better to have the Recognition Job do the logic of determining the number and identity of the LMs and then create n separate single LmImageJobs

Feel free to reopen the discussion.

curufinwe: OK, you have a point there. But then we should add an assert that `lm_config.type == "arpa"`. And fix the name below.

lm/lm_image.py
michelwi: then I suggest to name this one 
```suggestion
        shutil.move("lm.image", self.out_image.get_path())
```

and below in the `create_config` method, after asserting there is only exactly one arpa lm, we explicitly overwrite `lm-util.lm.image = "lm.image` to make sure the name is correct.

michelwi (commented):
I approve of the changes you did.
But since the functionality does not exist in the official RASR repo I will withhold acceptance until it is backported.

curufinwe:
Done: https://github.com/rwth-i6/rasr/commit/473d202c4bd9040593d7dc861317b731ec9fb724

michelwi (approved):

lm/lm_image.py
```diff
@@ -0,0 +1,90 @@
...
+            crp, {"lexicon": "lm-util.lexicon", "language_model": "lm-util.lm"}
+        )
+        del (
+            config.lm_util.lm.scale
+        )  # scale not considered here, delete to remove ambiguity
+
+        assert config.lm_util.lm.type == "ARPA"
+
+        config.lm_util.action = "load-lm"
+        config.lm_util.lm.image = "lm.image"
```
curufinwe, line 78: You also need to check if image is set in the lm_post_config. I'm a bit undecided what to do in that case. Either keep the old image (and create a link to it as output of this job) or to recreate the image. @michelwi @Nintendo-Wei  what do you think?

curufinwe (changes requested):

lm/lm_image.py
michelwi: how would the image from post config be created in the first place? If I understand correctly, once this job is being used, it is the only place where images are created within the recipes. 

So if all works correctly if a second `LmImageJob` is called for the same LM, then it should be the same instance as the one that created the image and the config will be unmodified / no second image created.
In case the existing image is an external file, I would recreate it within the graph.
If the image was created by running RASR and accidentally creating it, then it will be in the work folder of the other Job and not accessible / possibly cleaned away later.
Now the only other possibility would be the image comes from the now existing `LmImageAndGlobalCacheJob` then not sure what to do.. depends on your/our plans how to handle the transition phase.

Independent of this, yes, we should check for `post_config.lm_util.lm.image` and delete it. Or maybe write lm.image in the post config in the first place. mh.. its anyway constant so doesn't matter.

Nintendo-Wei: According to line 416 in `i6_core/recognition/advanced_tree_search.py`: `lm_config[1].image = ...` the image is actually saved to the post_config right? According to the definition of `find_arpa_lms`, `lm_config[0]` is config and `lm_config[1]` is post_config.
So I did a similar thing for the new PR in `apptek_asr`: https://bitbucket.org/omnifluent/apptek_asr/pull-requests/93 line 162-164.
Is that what we wanted? Please correct me if I was wrong, thanks!

michelwi: in the `find_arpa_lms` function, anyway only LMs without existing image are returned. so for the use case in `apptek_asr` it is save.

Maybe we add a warning if image already exists (this should only happen in case of user error?) and then create a new one anyway?

Nintendo-Wei: Would something like 
```
if post_config.lm_util.lm.image:
    logging.warning("The LM image already exists, but a new one will be recreated.")
```
be enough?
Or we would like to delete the image file with `os.remove()`?

curufinwe: It's not about how this Job would be used inside a recognition job. But about the interface of the job on it's own. A user is allowed to put the image parameter in post_config and then this job is broken. I thought about it more and I'd say the most common use case is to specifically create the LM-image. But we should then in the recognition job only call it for LM's that don't have an image yet. The job should thus be renamed `CreateLmImageJob` to better communicate this purpose.

curufinwe: But most importantly: DO NOT call os.remove on the old image.

michelwi: >  I thought about it more and I'd say the most common use case is to specifically create the LM-image. 

I agree.

> But we should then in the recognition job only call it for LM's that don't have an image yet. 

yes, it is done like this in the associated PR to the `apptek_asr` recipes

> The job should thus be renamed CreateLmImageJob to better communicate this purpose.

ok, we can do it

> DO NOT call os.remove on the old image.

yes!

curufinwe (approved):

PR #176, opened 2021-11-15
Log name fixes in advanced_tree_search.py

Nintendo-Wei:
Change the log file names so that they are consistent in `log_file_output_path` and `cleanup_before_run`


michelwi (changes requested):
First please check the branch you use for doing this PR as there are numerous commits of the other PR in it as well.

Secondly, instead if fixing the logfile name, we should change the `util.backup_if_exists("logfile_name")` call instead to keep the naming of log files consistent.

michelwi (approved):

curufinwe (approved):

PR #179, opened 2021-11-17
Added spelling conversion Job for Lexicon;

erreurt:


lexicon/conversion.py
```diff
@@ -280,3 +279,109 @@ def run(self):
...
+                    # remove source orth from source lemma
+                    if source in source_lemma.orth:
+                        source_lemma.orth.remove(source)
+
+                    # remove source lemma from lexicon if no orth remaining
+                    if not source_lemma.orth:
+                        lex.remove_lemma(source_lemma)
+
+                # add source orth as synt (for LM query) to final lemma
+                final_lemma.synt.append([source])
```
michelwi, line 372: maybe check if source already had a synt set and then add this instead the orth.

similarly, if target had no synt defined, then setting it here will overwrite the default behavior of (if empty then use orth) i.e. the target lemma will now always use the source orth for LM lookup and never the original target orth.

With the current level of indentation the source orth will be added to synt of target even if source does not exist. Is this intended? I think it should not be.

lexicon/conversion.py
```diff
@@ -280,3 +279,109 @@ def run(self):
...
+                        continue
+                    source, target = line.split(self.mapping_delimiter)
+                    if self.reverse_mapping:
+                        source, target = target, source
+                    mapping[source] = target
+
+        # load input lexicon and build orthography to lemma dict
+        lex = lexicon.Lexicon()
+        lex.load(self.bliss_lexicon.get())
+        orth2lemma = {o: lemma for lemma in lex.lemmata for o in lemma.orth}
```
michelwi, line 346: if two lemmata have the same orth, second one will overwrite the first and first one will not be mapped. Not sure if this occurs in practice

lexicon/conversion.py
```diff
@@ -280,3 +279,109 @@ def run(self):
...
+                    # remove source lemma from lexicon if no orth remaining
+                    if not source_lemma.orth:
+                        lex.remove_lemma(source_lemma)
+
+                # add source orth as synt (for LM query) to final lemma
+                final_lemma.synt.append([source])
+                continue
+
+            # elif source lemma is existing
+            if source in orth2lemma:
```
michelwi, line 376: your comment says `elif` but the code says `if`. so if target exists then still also the source lemma is modified again. Is this intended? I think it should not be.

lexicon/conversion.py
```diff
@@ -280,3 +279,109 @@ def run(self):
...
+    """Convert lexicon to a new one with other regional spellings
+    e.g. US -> GB spelling
+    """
+
+    def __init__(
+        self,
+        bliss_lexicon,
+        orth_mapping_file,
+        reverse_mapping=False,
+        mapping_delimiter=" ",
```
michelwi, line 294: maybe better default would be "any whitespace character" to also incorporate tabs and the like. you may decide.

michelwi (changes requested):
Some comments about RASR quirks of the lexicon. Please check if the behavior you implemented is the desired behavior

lexicon/conversion.py
erreurt: there is a continue in last if block, so this is equivalent to elif

lexicon/conversion.py
erreurt: the idea of conversion job is: we have US-English biased system, especially the lexicon and language model are so, we want to only simply convert the lexicon such that the ASR system will output UK-spelling. 

what done here is basically, make US-spelling as UK-spelling's lemma's synt, such that decoder will search US-spelling in the LM, but output UK-spelling.

lexicon/conversion.py
erreurt: the key of the mapping is orth token, as long as orths in the original lemma are all unique, the mapping is there.

erreurt: hmm, you are right! in what case will this happen? if this is affecting the idea of converting spelling, then I will better make the value a list

lexicon/conversion.py
erreurt: if a lemma already has some synt, what will happen during decoding when we append US-spelling to the synt list?

michelwi: > we want to only simply convert the lexicon such that the ASR system will output UK-spelling.

Then you should not do any of this, but instead add the UK-spelling as `lemma.eval` token. This way internally the US spelling is used consistently for pronunciation, AM, and LM and only the final output is modified to UK spelling.

lexicon/conversion.py
michelwi: Sorry, you are right, I didn't spot this.

then I weakly suggest to change "continue; if" to "elif" for better readability and/or future extension of if cases.

erreurt:
Will add a new PR.

PR #181, opened 2021-11-18
removes tk.uncached from hash call

christophmluscher:
OptimizeAMandLMScaleJob has a `tk.uncached_path` call in the hash creation, making the hash dependent on an absolute path. call to `tk.uncached_path` removed.

### DIFF
--- a/recognition/optimize_parameters.py
+++ b/recognition/optimize_parameters.py
@@ -216,7 +216,7 @@ class OptimizeAMandLMScaleJob(rasr.RasrCommand, Job):
             "exe": kwargs["crp"].flf_tool_exe,
             "initial_am_scale": kwargs["initial_am_scale"],
             "initial_lm_scale": kwargs["initial_lm_scale"],
-            "lattice_cache": tk.uncached_path(kwargs["lattice_cache"]),
+            "lattice_cache": kwargs["lattice_cache"],
             "scorer_cls": kwargs["scorer_cls"],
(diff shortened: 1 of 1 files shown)


JackTemaki (approved):

michelwi:
The new one is the correct behavior, but this changes hashes of running setups. So we should make it optional.

JackTemaki:
> The new one is the correct behavior, but this changes hashes of running setups. So we should make it optional.

Yes, it changes the hashes of running setups. But the hashes are incorrect, and changing when moving a setup. Also importing is impossible for that reason. And as the optimizing job is very late in pipelines, Chris and I came to the conclusion this should be fixed in place.

christophmluscher:
> The new one is the correct behavior, but this changes hashes of running setups. So we should make it optional.

I agree with Nick on this. But if necessary I would add an option as an input to the job. But I would like to avoid that :D

michelwi:
I personally do not mind to change the hashes, as I am not affected.

>so what todo?

in the past we had this case in the `EstimateLDAMatrixJob` and introduced a global option that could be triggered by the user.
(cf. https://p.i6.informatik.rwth-aachen.de/diffusion/ASR/browse/master/lda/estimate.py$164-166)

mattiadg:
I also wouldn't add more complexity to prevent changing hashes that are going to change anyway

christophmluscher:
OK. I won't add the global option. If necessary people can add that option themselves.

PR #183, opened 2021-11-19
pretty print cart question file

christophmluscher:
dumps the cart question in a readable format

### DIFF
--- a/cart/questions.py
+++ b/cart/questions.py
@@ -99,7 +99,10 @@ class BasicCartQuestions:
 
     def write_to_file(self, file):
         from xml.dom.minidom import parseString
-        xmlstr = parseString(ET.tostring(self.get_questions())).toprettyxml(indent="   ")
+
+        xmlstr = parseString(ET.tostring(self.get_questions())).toprettyxml(
+            indent="   "
+        )
         with open(file, "wt", encoding="utf-8") as f:
             f.write(xmlstr)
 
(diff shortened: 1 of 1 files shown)


cart/questions.py
```diff
@@ -287,5 +292,10 @@ def process_questions(root, questions):
         return root
 
     def write_to_file(self, file):
```
curufinwe, line 294: This override is not necessary.

cart/questions.py
```diff
@@ -98,8 +98,13 @@ def get_questions(self):
...
-        tree = ET.ElementTree(self.get_questions())
-        tree.write(file, encoding="unicode")
+        from xml.dom.minidom import parseString
```
curufinwe, line 101: There is already a write_xml function with a prettify option in util.py. Please use that.

curufinwe (changes requested):

cart/questions.py
christophmluscher: do you mean to completely remove `write_to_file`? I am not sure where that is used. Probably only CartEstimateJob.

michelwi: As `CMUCartQuestions` inherits from `BasicCartQuestions` and the same implementation is already present in `BasicCartQuestions`, this function `write_to_file` will override the exact same function `write_to_file` from `BasicCartQuestions` with exactly the same implementation.

Therefore, deleting the function here will lead to the exact same behavior.

christophmluscher: But the function `write_to_xml` belongs to `PythonCartQuestions` and not to `CMUCartQuestions`. And `PythonCartQuestions` does not inherit from `BasicCartQuestions`

curufinwe (approved):

Issue #177, opened 2021-11-16
upper in ExtractOovWordsFromCorpusJob


christophmluscher:
In `ExtractOovWordsFromCorpusJob` we upper case for every and all lemmas. This causes problems for german because python treats "ß" like this: `"ß".upper() -> SS".
I think this should be this should be configurable via job parameter. Add param to sis hash ignore? has this Job been used?
@curufinwe @JackTemaki @michelwi

michelwi:
> I think this should be this should be configurable via job parameter. 

I think default should be to not change the casing and have a `TextToUpperJob` or `PipelineJob(["sed 's/[:lower:]/[:upper:]/g')` if need be. Now that it already is that way, adding a configurable parameter is the goto solution.

> has this Job been used?

Don't know, not by me. 

> Add param to sis hash ignore? 

If in doubt yes.

JackTemaki:
@christophmluscher Is this resolved?

christophmluscher:
fixed #184

PR #184, opened 2021-11-19
add choice: upper, lower, None

christophmluscher:
Addressing: #177

### DIFF
--- a/corpus/stats.py
+++ b/corpus/stats.py
@@ -1,5 +1,6 @@
 __all__ = ["ExtractOovWordsFromCorpusJob", "CountCorpusWordFrequenciesJob"]
 
+import logging
 from collections import Counter
 import xml.etree.cElementTree as ET
 
@@ -20,7 +21,7 @@ class ExtractOovWordsFromCorpusJob(Job):
         "casing": "upper",
     }
 
-    def __init__(self, bliss_corpus, bliss_lexicon, casing="none"):
+    def __init__(self, bliss_corpus, bliss_lexicon, casing="upper"):
         """
         :param Union[Path, str] bliss_corpus: path to corpus file
         :param Union[Path, str] bliss_lexicon: path to lexicon
@@ -31,6 +32,11 @@ class ExtractOovWordsFromCorpusJob(Job):
         self.bliss_corpus = bliss_corpus
         self.bliss_lexicon = bliss_lexicon
         self.casing = casing
+
+        if self.casing != "none":
+            logging.warning(
+                "The orthography/lemma casing is changed. Is this what you want? Normally you should set this to 'none'. For legacy reasons this is set to 'upper'."
+            )
 
         self.out_oov_words = self.output_path("oov_words")
 
--- a/corpus/stats.py
+++ b/corpus/stats.py
@@ -20,11 +20,11 @@ class ExtractOovWordsFromCorpusJob(Job):
         "casing": "upper",
     }
 
-    def __init__(self, bliss_corpus, bliss_lexicon, casing="upper"):
+    def __init__(self, bliss_corpus, bliss_lexicon, casing="none"):
         """
         :param Union[Path, str] bliss_corpus: path to corpus file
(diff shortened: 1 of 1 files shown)


corpus/stats.py
```diff
@@ -32,18 +40,46 @@ def tasks(self):
...
+                    for orth in tree.findall(".//lemma/orth")
+                    if orth.text
+                }
+            elif self.casing == "lower":
+                iv_words = {
+                    orth.text.lower()
+                    for orth in tree.findall(".//lemma/orth")
+                    if orth.text
+                }
+            else:
```
michelwi, line 55: maybe check for `None` (or do you mean `"none"`?) explicitly and add an assertion that the user does not request an unknown casing mode.

michelwi (commented):
thats a lot of code duplication.. maybe we can define a function `some_name(w)` that returns `w.upper()`, `w.lower()` or `w` depending on mode and then use this instead of the case distinctions.

curufinwe (changes requested):
Better do something like:
```
modifier = lambda x: x
if casing == "upper":
  modifier = lambda x: x.upper()
elif casing == "lower":
  modifier = lambda x: x.lower()
```
And then use modifier instead of branching during creation of iv_words and oov_words

christophmluscher:
> Better do something like:
> 
> ```
> modifier = lambda x: x
> if casing == "upper":
>   modifier = lambda x: x.upper()
> elif casing == "lower":
>   modifier = lambda x: x.lower()
> ```
> 
> And then use modifier instead of branching during creation of iv_words and oov_words

Yes. I agree.
One question/modification. I thought that assigning a lambda statement is discouraged!? I checked PyCharm also complains about this. So I would use a `def` instead of `lambda`.

curufinwe:
Sure, no problem.

michelwi (approved):

corpus/stats.py
```diff
@@ -16,24 +16,44 @@ class ExtractOovWordsFromCorpusJob(Job):
     Extracts the out of vocabulary words based on a given corpus and lexicon
     """
 
-    def __init__(self, bliss_corpus, bliss_lexicon):
+    __sis_hash_exclude__ = {
+        "casing": "upper",
+    }
+
+    def __init__(self, bliss_corpus, bliss_lexicon, casing="none"):
```
curufinwe, line 23: Why is default "none" but sis_hash_exclude has "upper"? This will break old setups.

christophmluscher: yes. I missed that.

When I think about this.. the old behavior was upper. so the default value and sis_hash_exclude should be "upper" correct?
@JackTemaki you agree?

christophmluscher: but we set this in the Librispeech pipeline to "upper".. and that for that case the hash should not change. but for every new setup we want the "none" behavior.. not sure how to handle this...

JackTemaki: But would it not be correct to have "upper" as default, and exclude this from the hash? Then it does not change anything anywhere.

JackTemaki: >  but for every new setup we want the "none" behavior
New behavior newer works without breaking old stuff, this is why RETURNN got this `behavior_version` parameter...

JackTemaki: What if we add a "second" version of the job with changed defaults...? I have no better idea....

JackTemaki: It would inherit completely from the first one except for that single change

christophmluscher: maybe have "upper" as default but add a warning for command line or log file!?

christophmluscher: "upper" is now default and included in sis_hash_exclude to not break legacy settings. but I added a warning

christophmluscher: maybe move the warning to the run function??

curufinwe (changes requested):
Move warning into run function.

curufinwe (approved):

curufinwe (approved):

PR #185, opened 2021-11-19
Changed synt's type to List[str], only allowing 1 sequence of synt tokens

erreurt:
Fix synt's type.
Replaced file opening with uopen in Lexicon.load

### DIFF
--- a/lib/lexicon.py
+++ b/lib/lexicon.py
@@ -6,9 +6,10 @@ For format details visit: `https://www-i6.informatik.rwth-aachen.de/rwth-asr/man
 __all__ = ["Lemma", "Lexicon"]
 
 from collections import OrderedDict
-import gzip
 from typing import Optional, List
 import xml.etree.ElementTree as ET
+
+from i6_core.util import uopen
 
 
 class Lemma:
@@ -21,8 +22,8 @@ class Lemma:
         :param Optional[list[str]] orth: list of spellings used in the training data
         :param Optional[list[str]] phon: list of pronunciation variants. Each str should
             contain a space separated string of phonemes from the phoneme-inventory.
-        :param Optional[list[list[str]]] synt: list of token sequences, for which each token sequence is used as as possible language model representation.
-            Each sublist should contain words from the language model vocabulary.
+        :param Optional[list[str]] synt: list of tokens, this token sequence
+            is used as as possible language model representation.
         :param Optional[list[list[str]]] eval: list of output representations. Each
             sublist should contain one possible transcription (token sequence) of this lemma
             that is scored against the reference transcription.
@@ -49,11 +50,13 @@ class Lemma:
         for p in self.phon:
             el = ET.SubElement(res, "phon")
             el.text = p
-        for s in self.synt:
+        if self.synt and any(token.strip() for token in self.synt):
             el = ET.SubElement(res, "synt")
-            for t in s:
+            for token in self.synt:
+                if not token.strip():
+                    continue
                 el2 = ET.SubElement(el, "tok")
-                el2.text = t
+                el2.text = token
         for e in self.eval:
             el = ET.SubElement(res, "eval")
             for t in e:
@@ -136,9 +139,7 @@ class Lexicon:
         """
         :param str path: bliss lexicon .xml or .xml.gz file
         """
-        open_fun = gzip.open if path.endswith(".gz") else open
-
-        with open_fun(path, "rt") as f:
+        with uopen(path, "rt") as f:
             root = ET.parse(f)
 
         for phoneme in root.findall(".//phoneme-inventory/phoneme"):
--- a/lib/lexicon.py
+++ b/lib/lexicon.py
@@ -22,8 +22,8 @@ class Lemma:
         :param Optional[list[str]] orth: list of spellings used in the training data
         :param Optional[list[str]] phon: list of pronunciation variants. Each str should
             contain a space separated string of phonemes from the phoneme-inventory.
-        :param Optional[list[str]] synt: list of tokens, this token sequence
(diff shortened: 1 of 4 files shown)


lib/lexicon.py
```diff
@@ -21,8 +22,8 @@ def __init__(self, orth=None, phon=None, synt=None, eval=None, special=None):
         :param Optional[list[str]] orth: list of spellings used in the training data
         :param Optional[list[str]] phon: list of pronunciation variants. Each str should
             contain a space separated string of phonemes from the phoneme-inventory.
-        :param Optional[list[list[str]]] synt: list of token sequences, for which each token sequence is used as as possible language model representation.
-            Each sublist should contain words from the language model vocabulary.
+        :param Optional[list[str]] synt: list of tokens, this token sequence
+            is used as as possible language model representation.
```
michelwi, line 26: Typo: "as as" -> "as a"

But since I'm on it may I suggest slight rewording to make clear that only one single token sequence is accepted.

```suggestion
        :param Optional[list[str]] synt: list of LM tokens that form a single token sequence.
            This sequence is used as the language model representation.
```

lib/lexicon.py
```diff
@@ -49,11 +50,13 @@ def to_xml(self):
...
-        for s in self.synt:
+        if self.synt and any(token.strip() for token in self.synt):
             el = ET.SubElement(res, "synt")
-            for t in s:
+            for token in self.synt:
+                if not token.strip():
+                    continue
                 el2 = ET.SubElement(el, "tok")
-                el2.text = t
+                el2.text = token
```
michelwi, line 59: not sure, but would you also use `token.strip()` here as you test for this beforehand?

michelwi (commented):
According to the implementation in RASR only a single syntactic token sequence is supported. This PR changes the type of `synt` accordingly

lib/lexicon.py
```diff
@@ -49,11 +50,13 @@ def to_xml(self):
         for p in self.phon:
             el = ET.SubElement(res, "phon")
             el.text = p
-        for s in self.synt:
+        if self.synt and any(token.strip() for token in self.synt):
```
curufinwe, line 53: I would not modify the token sequence here. The lexicon is a data-structure that is written to disk. Having these sort of modifications done here makes it harder to reason about what's going on.

curufinwe (changes requested):
Please also check for usages of the Lexicon in the recipes and adjust to new API. Or provide some backward-compatible mode that does not break existing usage.

lib/lexicon.py
erreurt: Im not sure whether I understand the question. Now synt should be a single sequence of tokens, thus a list of strings. Here what added is only a check to avoid empty token written to final lexicon file. Is empty token in synt fine with RASR?

lib/lexicon.py
erreurt: Yes makes sense! Will add that

lib/lexicon.py
erreurt: 👌 will do

lib/lexicon.py
michelwi: What you say is true. But I think curufinwes argument is, that the check for empty tokens should not happen here in this method, but at the point where the `Lemma` object is created.
Here we would assume that `self.synt`, if given, is correctly formed.

curufinwe: Yes, @michelwi is right. The main point is that now for a given lexicon `lex` we do not have `lex == load(store(lex))` as the act of storing changes the data-structure. This is not good behavior in general as you should be able to assume that storing and loading a datastructure should give you the same object.

erreurt: @michelwi @curufinwe Got it, thanks!

michelwi (approved):

michelwi:
> Please also check for usages of the Lexicon in the recipes and adjust to new API. Or provide some backward-compatible mode that does not break existing usage.

I found some usages in i6_core, but also check https://github.com/rwth-i6/i6_experiments as some corpus specific lexica are created here and the typing has to be updated.

Notifying @JackTemaki about this.

curufinwe (changes requested):
One more example that needs to be changed:
https://github.com/rwth-i6/i6_core/blob/main/lexicon/modification.py#L36

Also this made me notice that we need to distinguish between no synt token given and an emtpy synt token sequence (e.g. for silence). We need to change the default for synt to None (that will represent not given) whereas empty list represents an empty synt token sequence.

Also these two need modification:
https://github.com/rwth-i6/i6_core/blob/main/lexicon/conversion.py#L248
https://github.com/rwth-i6/i6_core/blob/main/lexicon/conversion.py#L270

curufinwe (approved):

JackTemaki:
It seems you did not really pay attention here... It also my fault for not having a closer look, but of course this change here broke my complete setup. You did not only change the outside type, but the internal representation as well, which means all hashes changed. I will directly revert this and then we have a closer look.

JackTemaki:
@erreurt as we had to revert this commit, can you open a new PR where you do similar changes, but where the internal representation does not change. Otherwise the hashes will change.

I am not 100% sure about what is needed, but I guess it is already sufficient if you do something like:
`self.synt = [synt]` , so that the constructor type is correct, but the internal representation stays exactly the same as before

curufinwe:
@JackTemaki in which job is the internal representation of the lexicon used? Also we'd ideally only change the hash computation for the Job and not the internal representation.

JackTemaki:
The `WriteLexiconJob` to add static Lemmas. Changing the hash calculation of this job only is a good idea. I can take care of that.

PR #186, opened 2021-11-24
Revert "Changed synt's type to List[str], only allowing 1 sequence of synt tokens"

JackTemaki:
Reverts rwth-i6/i6_core#185


curufinwe:
Why?

JackTemaki:
Because the merge changed the hashes of exactly ALL new i6 setups...

curufinwe (approved):

PR #188, opened 2021-11-26
Change Lexicon to only allow a single sequence of synt tokens - Updated

JackTemaki:
I now decided to still "break" hashes in a sense that without adapting the lexicon it will have different hashes, but it gives an assertion warning so people can adjust to the correct type, and then the hashes are the same as before.


JackTemaki:
@erreurt

curufinwe (approved):

curufinwe:
Thank you @JackTemaki !

erreurt:
Nice! Then I can continue with the LexiconSpellingConversionJob. Thank you both! @JackTemaki @curufinwe

PR #189, opened 2021-11-26
fixed alignment type mismatch

JackTemaki:
I am not sure if the type was defined wrong or the assignment was wrong. Maybe someone can clarify what is correct, and then I can adapt the PR.


curufinwe:
Both is allowed and select_element https://github.com/rwth-i6/i6_core/blob/main/meta/system.py#L963 will do the right thing

JackTemaki:
Ok that works, then I will add this to the documentation!

PR #191, opened 2021-11-29
changed LexiconUniqueOrthJob

erreurt:
After discussion with Willy, we decided to extend this job to support merging-lemmata-with-multi-orth mode; 
Also the new implementation is updated to use the newest Lexicon and Lemma IO interface;

### DIFF
--- a/lexicon/conversion.py
+++ b/lexicon/conversion.py
@@ -156,9 +156,10 @@ class LexiconUniqueOrthJob(Job):
                 # CAUTION: synt token sequence is only copied when
                 # 1) the final lemma does not already have one
                 # 2) and the current to-be-merged-lemma has one
+                # ** note an empty sequence (i.e. []) counts to "having a synt"
                 # this could lead to information loss if all the
                 # to-be-merged-lemmata have different synt token sequences
-                if not final_lemma.synt and lemma.synt:
+                if final_lemma.synt is None and lemma.synt is not None:
                     final_lemma.synt = lemma.synt
 
                 for eval in lemma.eval:
--- a/lexicon/conversion.py
+++ b/lexicon/conversion.py
@@ -98,8 +98,8 @@ class LexiconUniqueOrthJob(Job):
             orth/phon/eval
                 all orth/phon/eval elements are merged together
             synt
-                synt element is only copied to final lemma when
-                    1) the final lemma does not already have one
+                synt element is only copied to target lemma when
+                    1) the target lemma does not already have one
                     2) and the rest to-be-merged-lemmata have any synt element
                     ** having a synt <=> synt is not None
                 this could lead to INFORMATION LOSS if there are several
@@ -109,6 +109,12 @@ class LexiconUniqueOrthJob(Job):
     __sis_hash_exclude__ = {"merge_multi_orths_lemmata": False}
 
     def __init__(self, bliss_lexicon, merge_multi_orths_lemmata=False):
+        """
+        :param tk.Path bliss_lexicon: lexicon file to be handeled
+        :param bool merge_multi_orths_lemmata: if True, also lemmata containing
+            multiple orths are merged based on their primary orth. Otherwise
+            they are ignored.
+        """
         self.set_vis_name("Make Lexicon Orths Unique")
 
         self.bliss_lexicon = bliss_lexicon
--- a/lexicon/conversion.py
+++ b/lexicon/conversion.py
@@ -81,30 +81,7 @@ class FilterLexiconByWordListJob(Job):
 
 
 class LexiconUniqueOrthJob(Job):
-    """Merge lemmata with the same orthography.
-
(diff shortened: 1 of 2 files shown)


erreurt:
@curufinwe @michelwi

erreurt:
@michelwi I have adjusted the comments etc, feel free to help rephrase the docstring!

lexicon/conversion.py
```diff
@@ -81,10 +81,38 @@ def run(self):
...
+                    1) the final lemma does not already have one
+                    2) and the rest to-be-merged-lemmata have any synt element
+                    ** having a synt <=> synt is not None
+                this could lead to INFORMATION LOSS if there are several
+                different synt token sequences in the to-be-merged lemmata
+    """
+
+    __sis_hash_exclude__ = {"merge_multi_orths_lemmata": False}
+
+    def __init__(self, bliss_lexicon, merge_multi_orths_lemmata=False):
```
michelwi, line 111:
```suggestion
    def __init__(self, bliss_lexicon, merge_multi_orths_lemmata=False):
            """
        :param tk.Path bliss_lexicon: lexicon file to be handeled
        :param bool merge_multi_orths_lemmata: if True, also lemmata containing
            multiple orths are merged based on their primary orth. Otherwise they
            are ignored.
        """
```

michelwi (commented):
thanks @erreurt

erreurt:
> if there are multiple lemmata which have a synth, which is selected?

@michelwi 
This is tricky, and there is probably no better way than simply taking the first one that is available (as the order how lexicon was loaded)

lexicon/conversion.py
```diff
@@ -81,10 +81,44 @@ def run(self):
 
 
 class LexiconUniqueOrthJob(Job):
-    def __init__(self, bliss_lexicon):
+    """Merge lemmata with the same orthography.
+
+    Merging strategy (only natural proper word lemmata are affected)
```
curufinwe, line 86: This description should probably go into the constructor, where `merge_multi_orths_lemmata` is defined. Having it here is kind of confusing.

lexicon/conversion.py
```diff
@@ -94,50 +128,53 @@ def tasks(self):
...
 
-        lemmas = collections.OrderedDict()
-        for lemma in old_lexicon.findall("lemma"):
-            if "special" in lemma.attrib:
-                root.append(lemma)
-            elif len(lemma.findall("orth")) != 1:
-                root.append(lemma)
-            else:
-                orth = lemma.find("orth").text
+        for lemma in lex.lemmata:
```
curufinwe, line 137: Please not newline after for loop line

lexicon/conversion.py
```diff
@@ -94,50 +128,53 @@ def tasks(self):
...
+                continue
 
-            o = ET.SubElement(el, "orth")
-            o.text = orth
+            orth2lemmata[lemma.orth[0]].append(lemma)
 
-            for phon in lemma["phon"]:
-                p = ET.SubElement(el, "phon")
-                p.text = phon
+        for orth, lemmata in orth2lemmata.items():
```
curufinwe, line 151: no newline

lexicon/conversion.py
```diff
@@ -94,50 +128,53 @@ def tasks(self):
...
-            for eval in lemma["eval"]:
-                ev = ET.SubElement(el, "eval")
-                ev.text = eval
+            final_lemma = lemmata[0]
 
-        with uopen(self.out_bliss_lexicon.get_path(), "w") as lexicon_file:
-            lexicon_file.write('<?xml version="1.0" encoding="utf-8"?>\n')
-            lexicon_file.write(ET.tostring(root, "unicode"))
+            for lemma in lemmata[1:]:
+
```
curufinwe, line 158: no newline

lexicon/conversion.py
```diff
@@ -94,50 +128,53 @@ def tasks(self):
...
+                        final_lemma.phon.append(phon)
+
+                if final_lemma.synt is None and lemma.synt is not None:
+                    final_lemma.synt = lemma.synt
+
+                for eval in lemma.eval:
+                    if eval not in final_lemma.eval:
+                        final_lemma.eval.append(eval)
+
+                if lemma in lex.lemmata:
```
curufinwe, line 174: why is this check necessary? Wouldn't the lemma always be in lex.lemmata? You just picked it out from there.

curufinwe (changes requested):

erreurt:
> why is this check necessary? Wouldn't the lemma always be in lex.lemmata? You just picked it out from there.
this check was there when we formerly NOT ONLY consider primary lemmata (a single lemma could appear in several times in orth2lemmata.values())
but now this check is not necessary

curufinwe (approved):

michelwi (approved):

PR #192, opened 2021-11-30
Add pro and epilog to rasr config

Nintendo-Wei:


rasr/config.py
```diff
@@ -203,13 +219,17 @@ def __repr__(self):
             max_len = max(map(lambda t: len(t[1]), g))
             for t in g:
                 buf.append("%-*s = %s" % (max_len, t[1], t[2]))
-
+        buf.append(self._epilog)
```
curufinwe, line 222: same as with prolog

rasr/config.py
```diff
@@ -203,13 +219,17 @@ def __repr__(self):
...
         return "\n".join(buf)
 
     def __sis_state__(self):
         result = {"tree": self.__dict, "value": self._value}
         return result
 
+    def _sis_hash(self):
```
curufinwe, line 229: This is probably not right. Now the config values themselves are not hashed. Also it is essential that old hashes do not change after the addition of prolog/epilog.

curufinwe (changes requested):

rasr/config.py
michelwi, line 229: Will this not ignore the regular contents of the config? Or is it complementary to the `__sis_state__` ?

Also, to preserve the hashes of older configs, if the value is `""` it should be excluded from the hash.

michelwi (changes requested):

michelwi (approved):

curufinwe (approved):

PR #194, opened 2021-12-02
Fix hash in StringWrapper

JackTemaki:
This is a hash-breaking PR, but in the way the StringWrapper is currently used you can have absolute paths hashed which makes setups non-consistent. Maybe it is fine to just keep it as it is, but then everybody should change to `DelayedFormat` or so for modifying paths that are added to RASR configs.

### DIFF
--- a/rasr/config.py
+++ b/rasr/config.py
@@ -307,9 +307,22 @@ class ConfigBuilder:
 
 
 class StringWrapper:
+    """
+    Allows to define custom strings for RASR configs with correct hashing,
+    this is somewhat redundant with DelayedFormat from Sisyphus
+    """
+
     def __init__(self, string, hidden=None):
+        """
+
+        :param str string: some string based on the hashing object
+        :param Any hidden: hashing object
+        """
         self.string = string
         self.hidden = hidden
 
     def __str__(self):
         return self.string
+
+    def __sis_state__(self):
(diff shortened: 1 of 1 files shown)


michelwi:
Can you give examples where an absolute path is used in the string?
I think in the most prominent cases this is replaced by the new returnn.Model and returnn.Checkpoint classes.

In any case, if a user resolves an absolute path in the manager to put into the string, then they are doing a user error. If some Jobs use this erroneously, then let's fix them.

Completely removing the string from the hash computation could lead to a situation where a result-altering change results in the same hash.

Stupid example:
```
random_object = Path("/var/tmp")
config1.lm.scale = StringWrapper("10.0", random_object)
config2.lm.scale = StringWrapper("1.0", random_object)
```

JackTemaki:
> Stupid example:

Good point, then we should replace each occurrence with the correct `DelayedFormat` which does correct hashing of both the format string and the passed path. `StringWrapper` should not be used at all.

curufinwe (changes requested):
OK, then let's deprecate StringWrapper in favor of DelayedFormat, but keep it's current behavior. The docstrings that you added are still usefull. So just remove the __sis_state__ function and we can merge them.

curufinwe (approved):

PR #196, opened 2021-12-08
make ApplyG2pJob not a mini_task, since it can be quite long

mattiadg:
This job can take also hours to finish, according to the vocabulary size. 
I think it's better not to have it as a mini_task

### DIFF
--- a/g2p/apply.py
+++ b/g2p/apply.py
@@ -57,7 +57,7 @@ class ApplyG2PModelJob(Job):
         self.rqmt = {"cpu": 1, "mem": 1, "time": 2}
 
     def tasks(self):
-        yield Task("run")
+        yield Task("run", rqmt=self.rqmt)
 
     def run(self):
         with uopen(self.out_g2p_lexicon, "wt") as out:
--- a/g2p/apply.py
+++ b/g2p/apply.py
@@ -54,6 +54,8 @@ class ApplyG2PModelJob(Job):
 
         self.out_g2p_lexicon = self.output_path("g2p.lexicon")
 
+        self.rqmt = {"cpu": 1, "mem": 1, "time": 2}
+
     def tasks(self):
         yield Task("run")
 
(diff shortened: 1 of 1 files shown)


g2p/apply.py
```diff
@@ -55,7 +55,7 @@ def __init__(
         self.out_g2p_lexicon = self.output_path("g2p.lexicon")
 
     def tasks(self):
-        yield Task("run", mini_task=True)
+        yield Task("run")
```
JackTemaki, line 58: If you have a regular job (no mini_task), you need to define a `self.rqmt` with some reasonable defaults. 
See https://github.com/rwth-i6/i6_core/blob/main/CONTRIBUTING.md (we should extend maybe the details in that document)

JackTemaki (changes requested):

g2p/apply.py
michelwi: reasonable defaults can/should be task dependent (e.g. on the size of word list) or if that is not always realizable we can expose the rqmt to the user so he can in-/decrease them as needed.
In this case also make sure it does not enter the hash.

mattiadg: Can the rqmt enter the hash?
And isn't the rqmt always exposed since it is a public attribute of Job?

JackTemaki: In the usual way that there is only `self.rqmt = { some defaults }` if will not enter the hash and will be exposed for outside adjustment. I think Willi referred to the case that some jobs that often change requirements have them as actual constructor parameter, and then you need to make sure it is not part of the hash.

curufinwe (changes requested):
Agree with @JackTemaki and @michelwi

mattiadg:
Is it ok like this?

curufinwe:
You still have to pass the requirements to the run Task.

g2p/apply.py
michelwi: To use the rqmt for the job:
```suggestion
        yield Task("run", rqmt=self.rqmt)
```

michelwi (changes requested):
If you think 2h are enough for all/most cases, we can keep the defined rqmt's like this.

curufinwe (approved):

michelwi (approved):

JackTemaki (approved):

PR #199, opened 2021-12-10
Add parallel and indirect options to FeatureExtractionJob

curufinwe:
The idea behind this PR is to have less write load on the fileservers while reading the feature caches and also have a mechanism to limit total load.

This PR might break some usage of the FeatureExtractionJob as the order of arguments has changed, but usage via meta.System is unaffected as additional parameters are passed as keyword arguments.

### DIFF
--- a/features/energy.py
+++ b/features/energy.py
@@ -19,11 +19,6 @@ def EnergyJob(crp, energy_options=None, **kwargs):
 
     port_name_mapping = {"energy": "energy"}
 
-    if "rtf" not in kwargs:
-        kwargs["rtf"] = 0.1
-    if "mem" not in kwargs:
-        kwargs["mem"] = 2
-
     return FeatureExtractionJob(
         crp,
         feature_flow,
--- a/features/energy.py
+++ b/features/energy.py
@@ -7,7 +7,7 @@ from .extraction import *
 import i6_core.rasr as rasr
 
 
-def EnergyJob(crp, energy_options=None, extra_config=None, extra_post_config=None):
+def EnergyJob(crp, energy_options=None, **kwargs):
     if energy_options is None:
         energy_options = {}
     else:
@@ -19,15 +19,17 @@ def EnergyJob(crp, energy_options=None, extra_config=None, extra_post_config=Non
 
     port_name_mapping = {"energy": "energy"}
 
+    if "rtf" not in kwargs:
+        kwargs["rtf"] = 0.1
+    if "mem" not in kwargs:
+        kwargs["mem"] = 2
+
     return FeatureExtractionJob(
         crp,
         feature_flow,
         port_name_mapping,
         job_name="Energy",
-        rtf=0.1,
-        mem=2,
-        extra_config=extra_config,
-        extra_post_config=extra_post_config,
+        **kwargs,
     )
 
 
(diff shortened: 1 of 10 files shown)


features/energy.py
```diff
@@ -19,15 +19,17 @@ def EnergyJob(crp, energy_options=None, extra_config=None, extra_post_config=Non
 
     port_name_mapping = {"energy": "energy"}
 
+    if "rtf" not in kwargs:
+        kwargs["rtf"] = 0.1
+    if "mem" not in kwargs:
+        kwargs["mem"] = 2
+
```
michelwi, line 26: no need to explicitly set them here (and in the other classes) as these values will already be the default in `FeatureExtractionJob`

JackTemaki (approved):
I did not see any change of any parameter order except for the `FeatureExtractionJob` it self, and I guess there people should have used named arguments anyway. So for me this is fine.

Maybe a side question: Was `parallel` always an allowed option for Tasks? I have never seen it.

michelwi (approved):
I don't like it. It will probably work as it is intended, but it violates the spirit of sisyphus and the ASR recipes in general.

e.g. writing and accessing RASR configs with star selector, writing a custom solution for a particular Job.

I would say the preferred solution would be to have an inversed_caching method in Sisyphus for Job outputs, that performs the local write caching transparently on the fly (just like `cf -d FILENAME` at the chair). Then we could use this also for recognition lattices, sequence training lattices, big LM images.. etc.

This having said, as I see the current FS issues as an urgent problem I will accept the merge.

features/extraction.py
```diff
@@ -131,6 +158,11 @@ def create_config(
         config, post_config = rasr.build_config_from_mapping(
             crp, {"corpus": "extraction.corpus"}, parallelize=True
         )
+        post_config[
+            "*"
+        ].OUTPUT_DIR = (
+            "./"  # make path very unspecific to allow easy override via command-line
```
michelwi, line 164:
```suggestion
            "."  # make path very unspecific to allow easy override via command-line
```
as the `/` will be added by `out_path = os.path.join(out_dir, out_path)`

curufinwe:
> Maybe a side question: Was `parallel` always an allowed option for Tasks? I have never seen it.
Yes, every task supports this.

curufinwe:
> I don't like it. It will probably work as it is intended, but it violates the spirit of sisyphus and the ASR recipes in general.
> 
> e.g. writing and accessing RASR configs with star selector, writing a custom solution for a particular Job.

This OUTPUT-PATH variable can be seen as an argument for the config (like LOGFILE and TASK). The only difference here is that we have a default value for it in the config and not in the run script. I guess I could also move it to the run-script if you like. I agree that having extra arguments is strongly discouraged as it violates readability, but in this case it's unavoidable as we will only know the path when running the job on the specific machine.

> I would say the preferred solution would be to have an inversed_caching method in Sisyphus for Job outputs, that performs the local write caching transparently on the fly (just like `cf -d FILENAME` at the chair). Then we could use this also for recognition lattices, sequence training lattices, big LM images.. etc.

Yes, that would also be my preferred solution, but it would take more time to fix.

> This having said, as I see the current FS issues as an urgent problem I will accept the merge.

michelwi:
> This OUTPUT-PATH variable [...] I guess I could also move it to the run-script if you like. 

Would it be possible to have the usual path in `feature.flow`, no changes in the config, but add a `--*.feature-extraction.cache.file=/var/tmp.xyz..` to the RASR call at runtime? This way the configs/flows would remain consistent and "nice" and the indirect write would be more transparent for the user.

curufinwe:
Yes, that should be possible with a customized run script.

michelwi (approved):

PR #200, opened 2021-12-13
Orthography spelling conversion

erreurt:
1. Issue 1: for us2gb, I've got a list of 1706 mapping pairs, but still I have noticed that this is far from complete. There are rules like "zation" --> "sation" to be applied, should I add this mechanism to this job?
2. Issue 2: should I remove the unused historical imports in this PR or someone else can do it in another minor PR?

### DIFF
--- a/lexicon/conversion.py
+++ b/lexicon/conversion.py
@@ -1,9 +1,10 @@
 import collections
-import gzip
-import os.path
 import json
+import logging
+import os.path
+import re
+import xml.dom.minidom
 import xml.etree.ElementTree as ET
-import xml.dom.minidom as minidom
 
 from sisyphus import *
 
@@ -306,8 +307,9 @@ class SpellingConversionJob(Job):
         self,
         bliss_lexicon,
         orth_mapping_file,
-        flip_mapping=False,
         mapping_delimiter=" ",
+        mapping_rules=None,
+        invert_mapping=False,
     ):
         """
         :param Path bliss_lexicon:
@@ -318,23 +320,47 @@ class SpellingConversionJob(Job):
             in case of plain text file
                 one can adjust mapping_delimiter
                 a line starting with "#" is a comment line
-        :param bool flip_mapping:
-            flip the orth mapping
         :param str mapping_delimiter:
             delimiter of source and target orths
             relevant only if mapping is provided with a plain text file
+        :param Optional[List[Tuple[str, str, str]]] mapping_rules
+            a list of mapping rules, each rule is represented by 3 strings
+                (source orth-substring, target orth-substring, pos)
+                where
+                pos should be one of ["leading", "trailing", "any"]
+            e.g. the rule ("zation", "sation", "trailing") will convert orth
+            ending with -zation to orth ending with -sation
+            set this ONLY when it's clearly defined rules which can not
+            generate any kind of ambiguities
+        :param bool invert_mapping:
+            invert the input orth mapping
+            NOTE: this also affect the pairs which are inferred from mapping_rules
         """
         self.set_vis_name("Convert Between Regional Orth Spellings")
 
         self.bliss_lexicon = bliss_lexicon
         self.orth_mapping_file = orth_mapping_file
-        self.flip_mapping = flip_mapping
+        self.invert_mapping = invert_mapping
         self.mapping_delimiter = mapping_delimiter
+        self.mapping_rules = mapping_rules
 
         self.out_bliss_lexicon = self.output_path("lexicon.xml.gz")
 
     def tasks(self):
         yield Task("run", mini_task=True)
+
+    @staticmethod
+    def _lemma_to_str(lemma, description: str):
+        """
+        :param Lemma lemma:
+        :return:
+        """
+        xml_string = xml.dom.minidom.parseString(
+            ET.tostring(lemma.to_xml())
+        ).toprettyxml(indent=" " * 2)
+        lemma_str = "\n".join(xml_string.split("\n")[1:])
+        lemma_str = description + "\n" + lemma_str
+        return lemma_str
 
     def run(self):
         # load mapping from json or plain text file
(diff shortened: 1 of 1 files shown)


erreurt:
@michelwi @curufinwe @mattiadg

lexicon/conversion.py
```diff
@@ -302,3 +303,109 @@ def run(self):
...
+                        if phon not in target_lemma.phon:
+                            target_lemma.phon.append(phon)
+                    for eval in source_lemma.eval:
+                        if eval not in target_lemma.eval:
+                            target_lemma.eval.append(eval)
+                    if not target_lemma.synt:
+                        if source_lemma.synt:
+                            target_lemma.synt = source_lemma.synt
+                        else:
+                            target_lemma.synt = source_orth.split()
```
michelwi, line 400: I am wondering if we should overwrite the `target_lemma.synt` even if it is set..?

michelwi:
to Issue 1:
sure, you could add a param `maping_rules: List[str]` and for each word in vocab and element in mapping_rules a new mapping pair is created when `re.sub(mapping, old_orth) != old_orth` or smth like that. Not sure how save it is, maybe log all created mappings so we can check if anything gets wrong..

to issue 2: sure why not.. you can do it.

lexicon/conversion.py
```diff
@@ -302,3 +303,109 @@ def run(self):
...
+            if self.flip_mapping:
+                mapping = {v: k for k, v in mapping.items()}
+        else:
+            mapping = dict()
+            with uopen(self.orth_mapping_file, "rt") as f:
+                for line in f:
+                    line = line.strip()
+                    if not line or line.startswith("#"):
+                        continue
+                    source_orth, target_orth = line.split(self.mapping_delimiter)
```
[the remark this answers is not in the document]
erreurt: Good point!

erreurt:
@michelwi 
The main functionality body becomes large

lexicon/conversion.py
erreurt: Makes sense for me to overwrite it anyway.
Words that have synt are mostly those which do not look like natural words, e.g. 'coz etc. 
Setting the synt anyway to American spelling cannot harm that much as we know the American spelling is definitely a valid synt we desire for this Job's original purpose. What do you think? @michelwi

michelwi: I think it makes sense to overwrite it. I don't think it will happen in practice.

michelwi:
> The main functionality body becomes large

You think? I would say the length is adequate. You could make it shorter at the expense of readability.. but no, this is fine to me.

lexicon/conversion.py
erreurt: 👍

michelwi (approved):

erreurt:
@michelwi 
Sorry, please check again. I noticed that the typing and docstring of _lemma_to_str was not consistent with the rest of the file.

lexicon/conversion.py
curufinwe: One question here: If I understand correctly you want to keep the original orthography for the language model (as presumably the LM is not updated when switching the spelling). Why would you override an already existing synt token with the original orthography? If the lemma already has a synt token isn't that what is expected to be in the LM and changing it to the split of the orthography is a bad idea?

curufinwe (commented):
Code looks OK, but see comment in existing conversation above.

lexicon/conversion.py
erreurt: > Words that have synt are mostly those which do not look like natural words, e.g. 'coz etc.

In this Job we are dealing with natural words, so we think overwriting does no harm. (We need more evidence/observation though)

If @curufinwe @michelwi you both think we should revert to earlier decision, I can revert them to what it was like.

michelwi: I don't think this case will happen in practice and I don't have a strong opinion on which synt to keep.
We could also assert that only one synt is present and then the user has to decide on a per case basis.

curufinwe: OK, then let's keep the original synt if it was defined already. Thanks!

erreurt: @curufinwe Done.

michelwi:
> Code looks OK, but see comment in existing conversation above.

I think this comment has been addressed, therefore this is a gentle reminder to @curufinwe to review again.

JackTemaki:
I have some questions regarding what the job does. I see there are `orth_mapping_file` and `mapping_rules`. Is there are reason why one is file based and the other one is given as direct paramters? Also it would be good if also for the mapping file it is written which exact format both .txt or .json files should have in order to work correctly.

michelwi:
> I see there are orth_mapping_file and mapping_rules. Is there are reason why one is file based and the other one is given as direct paramters? 

This job should map the `<orth>` of a lexicon to a specific regional spelling variant. This allows to reuse the existing ASR system (AM, LM). One example would be American vs British English. 
I assume the reason for the mapping being a file rather than a python list is that the number of entries can be quite large (1k+).

I guess the mapping rules could also be given as a file, but are usually fewer in nature.

Now that I think of it, we can maybe unify both params if we allow and parse wildcarts in the mapping file (e.g. `*zation -> *sation`) or make the orth_mapping_file a list of regex (e.g. `'s/^fries$/chips/'` would be valid single word mapping and `'s/(.+)zation$/\1sation/'` would transform all words ending in "zation" to "sation")
Thus we would only need one parameter

But then the repeated application of many regexes might slow down the code.

JackTemaki (approved):
I would say we just merge this, this seems to be a valid use case, documentation is sufficient and I have no general objections.

michelwi:
> I would say we just merge this, this seems to be a valid use case, documentation is sufficient and I have no general objections.

agreed.

As this is rather old, can we rebase onto current head and rerun the pipelines before merging?

christophmluscher:
rebased on main

michelwi:
> AppTek test fails

ah yes, this is because this pr comes from a fork and not a branch. I think we can proceed anyway as this change does not change existing jobs.

christophmluscher (approved):
see @michelwi comment

PR #204, opened 2022-01-05
first draft of job tests

JackTemaki:
As this was discussed already for a longer time, I now wanted to present a first draft on how to actually introduce some kind of very simple tests that work for not so complex jobs.

Feel free to add suggestions, changes and comments here.

The rough idea is that for each job we can have some example input and (desired) output files that the job is run on, verifying that the result is still the same. The demo files I added might already be way too large (I just copied them from existing jobs), but this was just for the demo purpose here.


JackTemaki:
I would recommend that every ASR Sisyphus user actually has a look on this, so that the test cases are simple enough that everybody would be fine with contributing to it when adding/changing jobs.

tests/job_tests/returnn/test_search.py
```diff
@@ -0,0 +1,37 @@
...
+def test_search_bpe_to_words_single():
+    search_out = Path("files/search_out_single")
+    reference_word_search_results = Path("files/word_search_results_single.py")
+    bpe_to_words_job = SearchBPEtoWordsJob(search_out)
+    bpe_to_words_job.out_word_search_results = Path(tempfile.mktemp(".py"))
+    bpe_to_words_job.run()
+
+    reference_checksum = compute_file_sha256_checksum(
+        reference_word_search_results.get_path()
+    )
```
michelwi, line 19: Do we actually need the reference file if we ever only compare against the checksum?

Probably better yes, so we can debug easier if the test fails.

tests/job_tests/returnn/test_search.py
```diff
@@ -0,0 +1,37 @@
...
+from i6_core.util import check_file_sha256_checksum, compute_file_sha256_checksum
+
+Path = setup_path(__package__)
+
+
+def test_search_bpe_to_words_single():
+    search_out = Path("files/search_out_single")
+    reference_word_search_results = Path("files/word_search_results_single.py")
+    bpe_to_words_job = SearchBPEtoWordsJob(search_out)
+    bpe_to_words_job.out_word_search_results = Path(tempfile.mktemp(".py"))
```
michelwi, line 14: Why is this necessary?

util.py
```diff
@@ -260,14 +260,22 @@ def create_executable(filename, command):
...
+    :return: checksum
+    :rtype:str
     """
     checksum_command_output = sp.check_output(["sha256sum", filename])
-    file_checksum = checksum_command_output.decode().strip().split(" ")[0]
-    assert file_checksum == reference_checksum
+    return checksum_command_output.decode().strip().split(" ")[0]
+
+
+def check_file_sha256_checksum(filename, reference_checksum):
```
vieting, line 275: For the use cases you have above, you could also add a small function which directly takes two files as input and computes and compares their checksums.

tests/job_tests/returnn/test_search.py
Atticus1806: Follow up: Do we always need to calculate the checksum each time we test? Since I assume this file does not change wouldn't it be easier to calculate the checksum once and then only compare to that? 
Or is this too prone to errors?

tests/job_tests/returnn/test_search.py
```diff
@@ -0,0 +1,37 @@
...
+
+def test_search_bpe_to_words_nbest():
+    search_out = Path("files/search_out_nbest")
+    reference_word_search_results = Path("files/word_search_results_nbest.py")
+    bpe_to_words_job = SearchBPEtoWordsJob(search_out)
+    bpe_to_words_job.out_word_search_results = Path(tempfile.mktemp(".py"))
+    bpe_to_words_job.run()
+
+    reference_checksum = compute_file_sha256_checksum(
+        reference_word_search_results.get_path()
```
Atticus1806, line 33: Same as above, does this need to be calculated each time?

util.py
```diff
@@ -260,14 +260,22 @@ def create_executable(filename, command):
...
+    """
+    Validates the sha256sum for a file against the target checksum
+
+    :param str filename: a single file to be checked
+    """
+    assert compute_file_sha256_checksum(filename) == reference_checksum
```
Atticus1806, line 281: Since this is only one assert, maybe it would be easier to have this in the tests themselves? Because I imagine right now if this assert fails the error will be given in the util file and then you have to search the trace for the call. This might make debugging / finding out what failed more tedious. 
If this logic is extended / larger then I agree with sourcing it out.

Atticus1806 (commented):
Looks good, would be fine like this, see my comments for possible changes, but nothing crucial.

tests/job_tests/returnn/test_search.py
JackTemaki: The files should be ideally quite small (I will also reduce the file size used here for the test), so calculating the checksum is not expensive. I think it is safer like this, because if you only have the checksum and not the file you can easily make errors. Of course we could include the file in the git and still use a fixed checksum, but I think it is not worth saving that little time on the checks.

util.py
JackTemaki: As the testing is run with `pytest`, it will directly show the name of the tests that are failing. But yes, the assert could also be part of the test, maybe this is better style.

tests/job_tests/returnn/test_search.py
vieting: I agree. Also, if you have the file in the git but still use a fixed checksum, you would need to manually change it if the file ever changes (should usually not happen but who knows...), which can easily be overlooked.

Atticus1806: If the computation is fast I dont see a reason not to do it, agreed.  :+1: 

We can still keep this in mind though if there are any tests with larger files, but thats for the future.

util.py
Atticus1806: Ahh okay. If the unit shows which test fails its not that big of a deal, then its just personal preference/style. Then we should not over think this. Just decide on something, both are fine I guess.

mattiadg: pytest also tells you the line of the failing assert

.github/workflows/job_tests.yml
```diff
@@ -0,0 +1,35 @@
...
+        with:
+          repository: "rwth-i6/i6_core"
+          path: "i6_core"
+      - uses: actions/checkout@v2
+        with:
+          repository: "rwth-i6/sisyphus"
+          path: "sisyphus"
+      - uses: actions/setup-python@v2
+        with:
+          python-version: 3.7
```
mattiadg, line 23: Shouldn't we test also with python 3.6.9?

tests/job_tests/returnn/test_search.py
mattiadg: En passant, `tempfile.mktemp` is deprecated https://docs.python.org/3/library/tempfile.html#deprecated-functions-and-variables

util.py
```diff
@@ -260,14 +260,22 @@ def create_executable(filename, command):
     )
 
 
-def check_file_sha256_checksum(filename, reference_checksum):
+def compute_file_sha256_checksum(filename):
```
mattiadg, line 263: Many libraries, such as numpy, tensorflow and many others, have a "tests" package to allow their users to test their code using the library itself. Maybe doing something like that can be more explicit than putting this code into a "util.py" file

util.py
mattiadg: Are you sure that testing the checksum is the best way to perform the test? This way, it also takes into account the output formatting and not only the content of the data structures

.github/workflows/job_tests.yml
JackTemaki: As far as I remember the recipe code we are using includes some 3.7 features, so at some point 3.6.9 might not work.

pyproject.toml
```diff
@@ -1,3 +1,4 @@
 [tool.black]
 line-length = 88
 target-version = ["py36"]
```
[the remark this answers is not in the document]
JackTemaki: Hmm this should be changed then.

tests/job_tests/returnn/test_search.py
JackTemaki: Good to know, I thought there would be warning in PyCharm if you use deprecated methods.

JackTemaki: > Can it be possible to add some settings in sisyphus to change the target directory to a temporary one for test purpose?

I am not sure how simple this is, you can have a look at it if you want.

util.py
JackTemaki: This function is not only used in testing, but also at other jobs (e.g. the "Download" job). So this is why it was in util.py already.

util.py
JackTemaki: I think this depends on the kind of job. If you are checking the data structures the tests will be more complicated, and then people will not add tests. This is why I thought a checksum is the easiest solution for a start.

JackTemaki:
I switched to content based tests not, but I keep the changes for the checksum computation for further tests.

michelwi:
In the n-best case, are you currently comparing list[tuple] with a simple == operation? Does this work and not compare by reference?

Maybe there exists or we should write a deep_compare function to check if all keys and values are exactly the same for any (nested) dict/list/tuple.

JackTemaki:
> Does this work

Yes, this does work, it seems all primitive types have a well defined `__eq__()`.

michelwi:
> Yes, this does work, it seems all primitive types have a well defined `__eq__()`.

Would then not also 
```
assert job_dict == reference_dict
```
work and would actually be better, since in your implementation 
```
for key, string in reference_dict.items():
    assert job_dict[key] == string
```
additional keys in job_dict that are not in reference dict, are not testet?

JackTemaki:
> > Yes, this does work, it seems all primitive types have a well defined `__eq__()`.
> 
> Would then not also
> 
> ```
> assert job_dict == reference_dict
> ```
> 
> work and would actually be better, since in your implementation

Surely yes, I changed it!

michelwi (approved):

JackTemaki:
@curufinwe @mattiadg can you verify that this is okay for a start?

christophmluscher (approved):

mattiadg:
Sorry, I completely missed the updates here. I think the test code looks good, I only have some minor remarks:

1. it's better to have the expected result inside the test, and not read it from an external file, in particular with `eval`. This would also encompass having shorter references.
2. the test code should be responsible for deleting the temporary directory and all the files in it
3. the process of replacing the job outputs with temporary directories can be automated so that it becomes basically impossible to do it wrong

JackTemaki:
> Sorry, I completely missed the updates here. I think the test code looks good, I only have some minor remarks:
> 
>     1. it's better to have the expected result inside the test, and not read it from an external file, in particular with `eval`. This would also encompass having shorter references.
> 
>     2. the test code should be responsible for deleting the temporary directory and all the files in it
> 
>     3. the process of replacing the job outputs with temporary directories can be automated so that it becomes basically impossible to do it wrong

I fixed number 2, and would leave 1 and 3 for future work. As there is no further feedback, I will merge this now.

PR #205, opened 2022-01-06
fix keep_epochs behavior in ReturnnTrainingJob

JackTemaki:
I am not sure why this issue did not come up earlier, but the `keep_epochs` parameter of the job conflicted with any `cleanup_old_models` setting in the Returnn config.

With the new code, Returnn is responsible for cleaning, so that the fixed epochs can be combined with keeping the last and best (according to dev score) epochs. This allows e.g. to do recognition on fixed checkpoints but also automatically on the best checkpoints (as soon as all pipelines support this) without wasting a lot of file space.

One missing part is that currently the `self.out_checkpoints` are limited to the checkpoints defined in the `keep_epochs` list. Maybe we just drop this condition, and have "empty" outputs in case the model gets deleted. As long as the following jobs have an `update` check this is fine. 

This is the reason this is still a draft (and I am waiting for my tests to finish)

### DIFF
--- a/returnn/training.py
+++ b/returnn/training.py
@@ -348,6 +348,7 @@ class ReturnnTrainingJob(Job):
         device,
         num_epochs,
         save_interval,
+        keep_epochs,
         horovod_num_processes,
         **kwargs,
     ):
@@ -376,6 +377,16 @@ class ReturnnTrainingJob(Job):
         if returnn_config.post_config is not None:
             post_config.update(copy.deepcopy(returnn_config.post_config))
 
+        if "cleanup_old_models" in post_config and keep_epochs:
+            if isinstance(post_config["cleanup_old_models"], bool):
+                post_config["cleanup_old_models"] = {"keep": keep_epochs}
+            elif isinstance(post_config["cleanup_old_models"], dict):
+                post_config["cleanup_old_models"]["keep"] = keep_epochs
+            else:
+                assert False, "invalid type of cleanup_old_models: %s" % type(
+                    post_config["cleanup_old_models"]
+                )
+
         res.config = config
         res.post_config = post_config
         res.check_consistency()
@@ -400,6 +411,12 @@ class ReturnnTrainingJob(Job):
             assert returnn_config.get(key) is None, (
                 "please define %s only as parameter to ReturnnTrainingJob directly"
                 % key
+            )
+
+        config_blacklisted_keys = ["cleanup_old_models"]
+        for key in config_blacklisted_keys:
+            assert returnn_config.config.get(key) is None, (
+                "please define %s only as parameter in the post_config" % key
             )
 
     @classmethod
--- a/returnn/training.py
+++ b/returnn/training.py
@@ -369,7 +369,7 @@ class ReturnnTrainingJob(Job):
         if returnn_config.post_config is not None:
             post_config.update(copy.deepcopy(returnn_config.post_config))
 
-        if keep_epochs:
+        if keep_epochs is not None:
             if not "cleanup_old_models" in post_config or isinstance(
                 post_config["cleanup_old_models"], bool
             ):
--- a/returnn/training.py
+++ b/returnn/training.py
@@ -368,9 +368,17 @@ class ReturnnTrainingJob(Job):
             post_config.update(copy.deepcopy(returnn_config.post_config))
 
         if keep_epochs:
-            if isinstance(post_config["cleanup_old_models"], bool):
+            if not "cleanup_old_models" in post_config or isinstance(
+                post_config["cleanup_old_models"], bool
+            ):
+                assert (
+                    post_config.get("cleanup_old_models", True) == True
+                ), "'cleanup_old_models' can not be False if 'keep_epochs' is specified"
                 post_config["cleanup_old_models"] = {"keep": keep_epochs}
             elif isinstance(post_config["cleanup_old_models"], dict):
+                assert (
+                    "keep" not in post_config["cleanup_old_models"]
+                ), "you can only provide either 'keep_epochs' or 'cleanup_old_models/keep', but not both"
                 post_config["cleanup_old_models"]["keep"] = keep_epochs
             else:
                 assert False, "invalid type of cleanup_old_models: %s" % type(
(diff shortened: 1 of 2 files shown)


returnn/training.py
```diff
@@ -402,6 +403,12 @@ def check_blacklisted_parameters(self, returnn_config):
                 % key
             )
 
+        config_blacklisted_keys = ["cleanup_old_models"]
+        for key in config_blacklisted_keys:
+            assert returnn_config.config.get(key) is None, (
+                "please define %s only as parameter in the post_config" % key
+            )
+
```
michelwi, line 411: Would this not be caught by `Config.check_consistency` or, if not, would we rather have this check added to the `check_consistency` method?

returnn/training.py
```diff
@@ -376,6 +367,16 @@ def create_returnn_config(
         if returnn_config.post_config is not None:
             post_config.update(copy.deepcopy(returnn_config.post_config))
 
+        if keep_epochs:
+            if isinstance(post_config["cleanup_old_models"], bool):
+                post_config["cleanup_old_models"] = {"keep": keep_epochs}
+            elif isinstance(post_config["cleanup_old_models"], dict):
+                post_config["cleanup_old_models"]["keep"] = keep_epochs
```
michelwi, line 374: Would we maybe check if already some keep_epochs are defined here and if yes (use union? give warning? use defined value? overwrite anyway?)

returnn/training.py
JackTemaki: Yes this would be caught, I just wanted to add it here to remind a user to not put the cleanup parameter in the hashable part.

returnn/training.py
JackTemaki: Ah right, this is missing. I would say an Exception is the best solution...

JackTemaki:
I just figured that this change is quite critical, so the reviews should be done very carefully. In my first version I missed one path where the code would not work if "keep_epochs" is set but "cleanup_old_models" is not.

returnn/training.py
```diff
@@ -376,6 +367,24 @@ def create_returnn_config(
         if returnn_config.post_config is not None:
             post_config.update(copy.deepcopy(returnn_config.post_config))
 
+        if keep_epochs:
+            if not "cleanup_old_models" in post_config or isinstance(
+                post_config["cleanup_old_models"], bool
+            ):
+                assert (
+                    post_config.get("cleanup_old_models", True) == True
+                ), "'cleanup_old_models' can not be False if 'keep_epochs' is specified"
```
michelwi, line 376: Mh.. If `cleanup_old_models=False` then all epochs are kept, especially also the ones defined in `keep_epochs`. Maybe in this case we could ignore `keep_epochs` and leave it to `False`?

Just opening the discussion, I don have a strong opinion here.

michelwi:
If you look at lines https://github.com/rwth-i6/i6_core/blob/86ce2273be52a73fd324ec69affeb012745f7e9f/returnn/training.py#L128-L136 would it worry us, that `self.keep_epochs` might be different from the argument `keep_epochs` that is given to `create_returnn_config`?

mattiadg:
> I just figured that this change is quite critical, so the reviews should be done very carefully. In my first version I missed one path where the code would not work if "keep_epochs" is set but "cleanup_old_models" is not.

why don't you add tests for all the conditions you thought of, and the ones that will come during the discussion? Nobody can be as precise as the python interpreter

JackTemaki:
> > I just figured that this change is quite critical, so the reviews should be done very carefully. In my first version I missed one path where the code would not work if "keep_epochs" is set but "cleanup_old_models" is not.
> 
> why don't you add tests for all the conditions you thought of, and the ones that will come during the discussion? Nobody can be as precise as the python interpreter

We do not have the test environment yet, and I am not sure how long this will take. But ideally yes.

JackTemaki:
> If you look at lines
> 
> https://github.com/rwth-i6/i6_core/blob/86ce2273be52a73fd324ec69affeb012745f7e9f/returnn/training.py#L128-L136
> would it worry us, that `self.keep_epochs` might be different from the argument `keep_epochs` that is given to `create_returnn_config`?

Not at all, because if `keep_epochs` is `None`, `self.keep_epochs` will be exactly the interval steps. And those are passed to RETURNN, so this should be absolutely fine, there is no need to alter the `cleanup_old_models` entry then.

returnn/training.py
JackTemaki: But wouldn't this mean that there are some settings that the user probably did not intend to have? But yes, ignoring would be an option...

returnn/training.py
michelwi: I am all in to have a set of config options defined that should never be hashed and have an explicit check for that.
I just still think as this check is solely about the state of the config, it should live in the config object itself (e.g. the `check_consistency` method, but can be a different place) and can be used by all returnn_config processing jobs.

This is not so important for cleanup, but e.g. `log_verbosity` should not be hashed for training, decoding, prior estimation. So it would be nice to have the check at a central place.

returnn/training.py
michelwi: Yes, this would probably indicate a configuration error and we should alert the user with a failing assertion.

michelwi (approved):

Atticus1806 (approved):

returnn/training.py
```diff
@@ -376,6 +367,24 @@ def create_returnn_config(
         if returnn_config.post_config is not None:
             post_config.update(copy.deepcopy(returnn_config.post_config))
 
+        if keep_epochs:
```
mattiadg, line 370: I always prefer not to abuse Python power and keep things more readable, so here I would write explicitly
`if keep_epochs is not None`

mattiadg:
The solution is coherent with the design adopted in other similar problems.
Maybe I don't see something but it looks like `self.out_checkpoints` cannot access to the checkpoints if they are defined in the post-config and not as a job argument. Is this intended?

JackTemaki:
> The solution is coherent with the design adopted in other similar problems. Maybe I don't see something but it looks like `self.out_checkpoints` cannot access to the checkpoints if they are defined in the post-config and not as a job argument. Is this intended?

Well, yes this was intended by me. But we can disallow this if you think this is more reasonable.

mattiadg:
> Well, yes this was intended by me. But we can disallow this if you think this is more reasonable.
I don't know, do you think it is more surprising for the user if he specifies keep_epochs in the returnn config and then the job has them as output or not?

PR #208, opened 2022-01-11
BlissToOggZipJob: allow concurrenct jobs

vieting:
For larger corpora, running the `BlissToOggZipJob` can take quite long. I modified the job such that it runs in parallel if the provided `segments` are a `MultiOutputPath` with multiple segment files. The resulting splits are merged at the end in this case.

### DIFF
--- a/returnn/oggzip.py
+++ b/returnn/oggzip.py
@@ -36,8 +36,15 @@ class BlissToOggZipJob(Job):
         """
         use RETURNN to dump data into an ogg zip file
 
+        The output zip archive can be created in a single run or in a parallel way using a MultiOutputPath for segments,
+        e.g.
+        > segments = SegmentCorpusJob(corpus, concurrent).out_segment_path
+        > ogg_zip_job = BlissToOggZipJob(corpus, segments=segments)
+
         :param str|Path bliss_corpus: bliss corpus file
         :param str|Path|MultiOutputPath segments: RASR segment file
+            If a MultiOutputPath object is given, a single zip archive is created for each segment list split and all
+            archives are merged at the end.
         :param str|Path rasr_cache: feature rasr cache
         :param int raw_sample_rate: raw audio sampling rate
         :param int feat_sample_rate: feature sampling rate
--- a/returnn/oggzip.py
+++ b/returnn/oggzip.py
@@ -1,9 +1,12 @@
 __all__ = ["BlissToOggZipJob"]
 
 import os
+import shutil
 import subprocess as sp
+import tempfile
+import zipfile
 
-from i6_core.util import relink
+from i6_core.util import MultiOutputPath, relink
 
 from sisyphus import *
 
@@ -34,7 +37,7 @@ class BlissToOggZipJob(Job):
         use RETURNN to dump data into an ogg zip file
 
         :param str|Path bliss_corpus: bliss corpus file
-        :param str|Path segments: RASR segment file
+        :param str|Path|MultiOutputPath segments: RASR segment file
         :param str|Path rasr_cache: feature rasr cache
         :param int raw_sample_rate: raw audio sampling rate
         :param int feat_sample_rate: feature sampling rate
@@ -50,6 +53,9 @@ class BlissToOggZipJob(Job):
         self.feat_sample_rate = feat_sample_rate
         self.no_conversion = no_conversion
         self.no_audio = no_audio
+        self.concurrent = (
+            len(segments.hidden_paths) if isinstance(segments, MultiOutputPath) else 1
+        )
 
         self.returnn_python_exe = (
             returnn_python_exe
@@ -60,17 +66,39 @@ class BlissToOggZipJob(Job):
             returnn_root if returnn_root is not None else gs.RETURNN_ROOT
         )
 
+        self.zip_subarchives = (
+            MultiOutputPath(
+                self,
+                "ogg.$(TASK).zip",
+                {i: "ogg.{}.zip".format(i) for i in range(1, self.concurrent + 1)},
+                cached=True,
+            )
+            if self.concurrent > 1
+            else None
+        )
+
         self.out_ogg_zip = self.output_path("out.ogg.zip")
 
         self.rqmt = None
+        self.merge_rqmt = {"cpu": 1, "mem": 1, "time": 1}
 
     def tasks(self):
         if self.rqmt:
-            yield Task("run", rqmt=self.rqmt)
+            yield Task("run", rqmt=self.rqmt, args=range(1, self.concurrent + 1))
         else:
-            yield Task("run", mini_task=True)
+            yield Task("run", mini_task=True, args=range(1, self.concurrent + 1))
+        if self.concurrent > 1:
+            if self.merge_rqmt:
+                yield Task("merge", rqmt=self.merge_rqmt)
+            else:
+                yield Task("merge", mini_task=True)
 
-    def run(self):
+    def run(self, task_id):
+        output = (
+            self.zip_subarchives.hidden_paths[task_id]
+            if self.concurrent > 1
+            else "out.ogg.zip"
+        )
         args = [
             tk.uncached_path(self.returnn_python_exe),
             os.path.join(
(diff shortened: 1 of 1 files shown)


michelwi:
Task "Merge" is no mini_task, is this intended? It does also not have any rqmt set, maybe you want to define some (even if constant) `merge_rqmt`.

The whole unzipping and rezipping now happens on the file server right? We might want to do these operations locally on the node to reduce load on the file server and (if these are many small files) speedup the process.

christophmluscher:
> We might want to do these operations locally on the node to reduce load on the file server and (if these are many small files) speedup the process.
Yes, I would prefer if the merge is not on the fileserver :)

JackTemaki:
Just for my understanding, you need this because you are also converting wav to .ogg files in the job, right? Because usually it is better to do the conversion beforehand, because then you do not need to re-do all the conversion just because you are changing something in the text/orth. (well, this is a rather rare case for ASR, but very common for TTS).

vieting:
>Just for my understanding, you need this because you are also converting wav to .ogg files in the job, right?
Yes, I'm usually converting wav to ogg, but I'm also using the synchronization with a RASR feature cache.

vieting:
>We might want to do these operations locally on the node to reduce load on the file server and (if these are many small files) speedup the process.

That's a good idea.

vieting:
>Task "Merge" is no mini_task, is this intended?

I made it a mini_task now.

christophmluscher:
> >Task "Merge" is no mini_task, is this intended?

> I made it a mini_task now.

How large are disk space requirements during merge operation?? If this is computed on the local desktop machine, can this cause problems if the temp directory does not have enough space?

vieting:
>How large are disk space requirements during merge operation?? If this is computed on the local desktop machine, can this cause problems if the temp directory does not have enough space?

For the librispeech training set, the zip file has 15G and the unzipped folder has 51G. I guess that should usually be fine, what do you think?

christophmluscher:
I would submit this to the cluster.

/var/tmp at the chair is normally around 30GB.

returnn/oggzip.py
```diff
@@ -1,104 +1,200 @@
...
+                args.extend(["--raw_sample_rate", str(self.raw_sample_rate)])
+            if self.feat_sample_rate is not None:
+                args.extend(["--feat_sample_rate", str(self.feat_sample_rate)])
+
+        sp.check_call(args)
+        if self.concurrent == 1:
+            relink("out.ogg.zip", self.out_ogg_zip.get_path())
+
+    def merge(self):
+        with tempfile.TemporaryDirectory() as tmp_dir:
```
michelwi, line 129:
```suggestion
        with tempfile.TemporaryDirectory(prefix=gs.TMP_PREFIX) as tmp_dir:
```


see also https://github.com/rwth-i6/i6_core/pull/203
The location where the temporary dictionary is created depends on a couple of factors an is mostly controlled by environment variables. But those are read before Sisyphus cleans up or sets the env variables. With this change the folder will be created by a sisyphus controlled location.

michelwi: > if you use /local you should have more space available, which is 180GB, so you could use that..

With this change, you can specify in your settings e.g.
```
TMP_PREFIX = "/local/sis_"
```
and the local partition will be used.

christophmluscher: and the default env variable TMPDIR can remain the same (for example for the cluster)

at the chair you have to clean `/local` manually. for single-user desktop this might not be a problem for multi-user this could be an issue..

normally compute or io heavy jobs should be submitted to the cluster then you also don't need another variable in the `settings.py` which might cause problems if somebody uses the Job without the global settings variable

christophmluscher: lange rede kurzer sinn: easiest and safest would probably be just to submit to the cluster xD

michelwi: >you have to clean `/local` manually.

I understand this happens automatically as soon as the the `with ...` block ends.

> you also don't need another variable in the settings.py

This variable already exists. cf. https://github.com/rwth-i6/sisyphus/blob/fe8ad9fb112672546ec2fc4f180df7610c88b372/sisyphus/global_settings.py#L272
We would only have to overwrite it, if we don't want to use the default location.

christophmluscher: > I understand this happens automatically as soon as the the with ... block ends.

What if the job crashes??

ah I see. so setting `TMP_PREFIX` in `settings.py` would influence all other Jobs as well, or not? But we don't want different behaviour for cluster and desktop machines.

cluster: `/var/tmp` lots of space
desktop: `/local` "lots" of space

so we want to keep `/var/tmp` as temp location but only want to change it for one single job.. that seems like a lot of unnecessary effort and potential for errors compared to just submitting to cluster :)

michelwi: > safest would probably be just to submit to the cluster

I don't disagree. But I think still the suggestion should be added to add the configurability and to match the sisyohus mktemp mechanism (https://github.com/rwth-i6/sisyphus/blob/fe8ad9fb112672546ec2fc4f180df7610c88b372/sisyphus/toolkit.py#L82-L105)

christophmluscher: OK. I agree :)

submit to cluster and add the sisyphus mktemp mechanism

returnn/oggzip.py
```diff
@@ -1,104 +1,200 @@
...
+
+        self.out_ogg_zip = self.output_path("out.ogg.zip")
+
+        self.rqmt = None
+
+    def tasks(self):
+        if self.rqmt:
+            yield Task("run", rqmt=self.rqmt, args=range(1, self.concurrent + 1))
+        else:
+            yield Task("run", mini_task=True, args=range(1, self.concurrent + 1))
```
christophmluscher, line 88: these tasks could be heavy on I/O and disk space, even temporary. I would prefer these to be: `mini_task=False`

returnn/oggzip.py
```diff
@@ -1,104 +1,200 @@
...
+                    with open(sub_meta_file, "r") as smf:
+                        for line in smf.readlines():
+                            if not (line.startswith("[") or line.startswith("]")):
+                                mf.write(line)
+                    os.remove(sub_meta_file)
+                mf.write("]\n")
+
+            # move all folders to output folder
+            cmd = [
+                "rsync",
```
christophmluscher, line 162: I am wondering. `rsync` is a default ubuntu package, if I remember correctly. Is there any concern about other operating systems?

christophmluscher: Do we care?

returnn/oggzip.py
christophmluscher: ah sry only for the merge. for run there is already the choise

returnn/oggzip.py
christophmluscher: good point.. I ll check that

vieting: I added @michelwi's suggestion, also it's submitted to the cluster by default now. You could configure it to be a mini_task, but only from outside the job after initializing it (same as for the existing `run` task).

returnn/oggzip.py
vieting: `merge` is now not a mini task anymore by default.

returnn/oggzip.py
vieting: I would argue we don't as the job will use other packages (`ffmpeg` in the called RETURNN script, possibly more) as well which might not be supported on other systems either.

christophmluscher (approved):

michelwi (approved):
For the future, it would be nice if you could du multiple commits and push then to the branch instead of (I assume) amend the single commit and then force push. Then the reviewer can more quickly check only the new changes.

JackTemaki (changes requested):
Somehow the diff is broken. @vieting did you by any chance change the file format or the type of line breaks?

vieting:
>Somehow the diff is broken. @vieting did you by any chance change the file format or the type of line breaks?

It seems so, not sure why exactly but I fixed it now.

JackTemaki (changes requested):
Sorry for bothering again, but can you add some description to the docstring to what is actually happening when you provide a MultiOutputPath, and update the general docstring a little bit regarding this new feature?

JackTemaki (approved):

PR #212, opened 2022-01-16
added os.chmod to add reading permissions for all users

timur-sch:
Added a `os.chmod` call after `shutil.move` to add reading permissions for all users. Previously it sometimes occurred that users did not have any permission on the am.mix file created by the MexgeMixturesJob.

### DIFF
--- a/mm/mixtures.py
+++ b/mm/mixtures.py
@@ -10,7 +10,7 @@ import os
 import shutil
 import struct
 import tempfile
-
+import stat
 from sisyphus import *
 
 Path = setup_path(__package__)
@@ -117,6 +117,7 @@ class MergeMixturesJob(rasr.RasrCommand, Job):
             util.partition_into_tree(self.mixtures_to_combine, self.combine_per_step),
         )
         shutil.move(mixtures, self.out_mixtures.get_path())
+        os.chmod(self.out_mixtures.get_path, stat.S_IROTH)
 
     def cleanup_before_run(self, cmd, retry, *args):
         log = args[2][12:]
--- a/mm/mixtures.py
+++ b/mm/mixtures.py
@@ -8,9 +8,10 @@ __all__ = [
 import logging
 import os
 import shutil
+import stat
 import struct
 import tempfile
-import stat
+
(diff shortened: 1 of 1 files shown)


mm/mixtures.py
```diff
@@ -117,6 +117,7 @@ def merge_helper(elements):
             util.partition_into_tree(self.mixtures_to_combine, self.combine_per_step),
         )
         shutil.move(mixtures, self.out_mixtures.get_path())
+        os.chmod(self.out_mixtures.get_path, stat.S_IROTH)
```
michelwi, line 120:
```suggestion
        os.chmod(self.out_mixtures.get_path(), stat.S_IROTH)
```

Call the function.

michelwi (changes requested):
Check inline comment.

I guess your goal is to add read permission for everybody. 
A) you should add also read permission for group
B) I think the function os.chmod _sets_ the permissions, not _adds_ them (please verify). So actually you would want to give the mode as bitwise or of the appropriate stat things for owner rw, group r, other r.

mm/mixtures.py
timur-sch: Thanks for the reply. This should be done now with the new commit. You were right with the setting of the mode, so I did it bitwise instead as suggested.

mm/mixtures.py
```diff
@@ -10,7 +10,7 @@
 import shutil
 import struct
 import tempfile
-
+import stat
```
michelwi, line 13: Check the contributing guidelines: there should be three blocks of imports
1. From other recipes
2. From standard libraries
3. From other libraries (including Sisyphus)
each block sorted alphabetically. 

I think stat belongs to the standard libraries and should therefore be inserted between shutil and struct. And we keep the empty line separating sisyphus.

michelwi (approved):

mm/mixtures.py
timur-sch: Ah my bad, will read the guidelines more carefully next time.

JackTemaki (approved):

Issue #213, opened 2022-01-17
Filter out empty g2p translations


mattiadg:
I found out that our g2p tool can produce empty pronunciations. When the g2p output is transformed into a lexicon, the corresponding entry does not have a `<phon>` tag and this causes a failure of RASR when the lemma is encountered and should be transformed into its pronunciation.

What do you think it should be the best way to create a clean lexicon? Introduce a new job that filters out empty-pronunciation lemmas, or maybe add this option in other jobs like `G2POutputToBlissLexiconJob`?

michelwi:
quick n dirty: run the output of g2p apply job through a pipeline job that filters out lines with just one word (e.g. with awk)
more clean an nice: add an option `filter_empty_outputs` to the apply g2p job.

PR #214, opened 2022-01-17
Add option to filter out words that have been translated to an empty string

mattiadg:
Possible fix to #213

### DIFF
--- a/g2p/apply.py
+++ b/g2p/apply.py
@@ -18,14 +18,14 @@ class ApplyG2PModelJob(Job):
     __sis_hash_exclude__ = {"filter_empty_words": False}
 
     def __init__(
-            self,
-            g2p_model,
-            word_list_file,
-            variants_mass=1.0,
-            variants_number=1,
-            g2p_path=None,
-            g2p_python=None,
-            filter_empty_words=False,
+        self,
+        g2p_model,
+        word_list_file,
+        variants_mass=1.0,
+        variants_number=1,
+        g2p_path=None,
+        g2p_python=None,
+        filter_empty_words=False,
     ):
         """
         :param Path g2p_model:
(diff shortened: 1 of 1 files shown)


JackTemaki:
If this filtered output is a separate file, couldn't we just introduce the new output file without any parameter? @curufinwe @michelwi do we have conventions for this?

michelwi:
Do we really need the filtered and non-filtered lexicon at the same time? I'd say if filtering is enabled, the single output of this job should be the filtered lexicon.

This way all consecutive jobs can just access `ApplyG2PModelJob.out_g2p_lexicon` without having to know which output is present or desired.
Otherwise the information "job should filter the lexicon" has also be presented to other jobs/recipes in the vicinity of the original job to select the correct output.

mattiadg:
> Do we really need the filtered and non-filtered lexicon at the same time? I'd say if filtering is enabled, the single output of this job should be the filtered lexicon.
> 
> This way all consecutive jobs can just access `ApplyG2PModelJob.out_g2p_lexicon` without having to know which output is present or desired. Otherwise the information "job should filter the lexicon" has also be presented to other jobs/recipes in the vicinity of the original job to select the correct output.

I think you're right on this, I am going to fix it accordingly

michelwi (approved):

g2p/apply.py
```diff
@@ -81,3 +88,14 @@ def run(self):
                     stdout=out,
                     stderr=err,
                 )
+
+    def filter(self):
+        fd_out, tmp_path = mkstemp(dir=".", text=True)
+        with open(self.out_g2p_lexicon, "r") as lex:
```
michelwi, line 94:
```suggestion
        with uopen(self.out_g2p_lexicon, "rt") as lex:
```
to be consistent with the writing call above.?

JackTemaki (approved):

PR #215, opened 2022-01-17
Add an option for energy normalization

papar22:
Adding an option to enable energy normalization or not in the energy flow!

### DIFF
--- a/features/energy.py
+++ b/features/energy.py
@@ -30,9 +30,9 @@ def EnergyJob(crp, energy_options=None, **kwargs):
 
 def energy_flow(
     without_samples=False,
-    normalization_type="divide-by-mean",
     samples_options={},
     fft_options={},
+    normalization_type="divide-by-mean",
 ):
     net = rasr.FlowNetwork()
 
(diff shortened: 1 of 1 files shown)


JackTemaki (approved):

features/energy.py
```diff
@@ -28,7 +28,12 @@ def EnergyJob(crp, energy_options=None, **kwargs):
     )
 
 
-def energy_flow(without_samples=False, samples_options={}, fft_options={}):
+def energy_flow(
+    without_samples=False,
+    normalization_type="divide-by-mean",
```
JackTemaki, line 33: It might be better to add new parameters last, in case people did not call this function with named arguments, but for me this is ok.

features/energy.py
```diff
@@ -28,7 +28,12 @@ def EnergyJob(crp, energy_options=None, **kwargs):
...
+    samples_options={},
+    fft_options={},
```
michelwi, line 35:
```suggestion
    samples_options={},
    fft_options={},
    normalization_type="divide-by-mean",
```
this comment for your convenience.

michelwi (changes requested):
I would request the change that @JackTemaki brought up:
Move the new argument to the end.

michelwi (approved):

PR #218, opened 2022-02-02
New Job to replace orth from other corpus

Atticus1806:
Moved (and refactored) job from an old setup that takes two corpora and replaces the orth tags for sequences from one with orth tags from the other one, matched on the sequence key in contrast to ordering.

### DIFF
--- a/corpus/convert.py
+++ b/corpus/convert.py
@@ -1,8 +1,9 @@
 __all__ = [
-    "CorpusToStmJob",
-    "CorpusToTxtJob",
-    "CorpusReplaceOrthFromTxtJob",
     "CorpusReplaceOrthFromReferenceCorpus",
+    "CorpusReplaceOrthFromTxtJob",
+    "CorpusToStmJob",
+    "CorpusToTextDictJob",
+    "CorpusToTxtJob",
 ]
 
 import gzip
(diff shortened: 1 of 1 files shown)


JackTemaki:
`Only works for single segment recordings so far.` There is no need to have this limitation here, I think this is just there because I was lazy.

JackTemaki (changes requested):
Sorry I forgot to put this as change request. So please try to remove the limitation.

michelwi (commented):
What is the point of this Job?
why would two corpora have the same segments but different transcriptions and what would be the difference to then just using the second (reference) corpus?

I'm sure you have a valid reason in your setup. The motivation behind my question is rather: "This is a very specific Job with a specific use case, can we find a useful generalization of this Job?"

Atticus1806:
>What is the point of this Job?

In my case I have synthesized speech (from a TTS experiment) and want to add the original transcriptions back to it. The current corpus object does not have any orth tags. In an old version of this job the `bliss_corpus` corpus was named `empty_bliss_corpus` so maybe this naming is a bit more obvious? 

I tried doing this with `CorpusReplaceOrthFromTxtJob` but there ordering in the files matters which is not guaranteed in this case.

corpus/convert.py
```diff
@@ -1,4 +1,9 @@
-__all__ = ["CorpusToStmJob", "CorpusToTxtJob", "CorpusReplaceOrthFromTxtJob"]
+__all__ = [
+    "CorpusToStmJob",
+    "CorpusToTxtJob",
+    "CorpusReplaceOrthFromTxtJob",
+    "CorpusReplaceOrthFromReferenceCorpus",
```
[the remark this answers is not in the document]
Atticus1806: I can sure do that. For `__all__` I feel this is a good idea, not sure if ordering within the file is "overkill". I guess if we do it once for each file, enforcing it in later PRs is no problem if remembered.

Atticus1806:
>So can you try to extend this in a way that it works universally?

Added support for it. Not sure if this solves the problem correctly though, this assumes that each segment has a unique tag. I dont have a way to test this, so maybe someone who knows how a corpus with multiple segments will look like can confirm this is correct?

JackTemaki:
> Not sure if this solves the problem correctly though, this assumes that each segment has a unique tag

segment.name is not unique, segment.fullname() is, so you need to change that!

Atticus1806:
changed to fullname() and reordered. Additionally I found that the `CorpusToTextDictJob` was missing from `__all__` aswell.

@michelwi any further thoughts on the specificity of the job?

michelwi (approved):
> any further thoughts on the specificity of the job?

Nah.. It's fine.

Atticus1806:
@JackTemaki can you have a look again and approve / merge if you are happy or comment on changes, I think everything you stated is implemented now.

corpus/convert.py
```diff
@@ -12,6 +18,94 @@
...
+        self.reference_bliss_corpus = reference_bliss_corpus
+
+        self.out_corpus = self.output_path("corpus.xml.gz")
+
+    def tasks(self):
+        yield Task("run", mini_task=True)
+
+    def run(self):
+        orth_c = corpus.Corpus()
+        orth_c.load(tk.uncached_path(self.reference_bliss_corpus))
```
JackTemaki, line 41: I just noticed you are using `tk.uncached_path` here, please change that to `.get_path()` here and for other occurances.

corpus/convert.py
```diff
@@ -12,6 +18,94 @@
...
+                tag = r.segments[i].fullname()
+                orths[tag] = orth
+
+        c = corpus.Corpus()
+        c.load(tk.uncached_path(self.bliss_corpus))
+
+        for r in c.all_recordings():
+            for i in range(len(r.segments)):
+                tag = r.segments[i].fullname()
+                orth = orths[tag]
```
JackTemaki, line 56: For the `orths[tag]` you should give a more verbose error message in case the key does not exist.

JackTemaki (changes requested):

corpus/convert.py
Atticus1806: I noticed in the `CorpusReplaceOrthFromTxtJob` it was also using `uncached_path` so I changed it there aswell in a different commit (to allow rollback). If its fine changed there aswell this should be resolved then.

corpus/convert.py
Atticus1806: Okay like this or is assert overkill / do you want to just catch the Exception and change the error message

corpus/convert.py
```diff
@@ -12,6 +18,95 @@
...
+                tag = r.segments[i].fullname()
+                orths[tag] = orth
+
+        c = corpus.Corpus()
+        c.load(self.bliss_corpus.get_path())
+
+        for r in c.all_recordings():
+            for i in range(len(r.segments)):
+                tag = r.segments[i].fullname()
+                assert tag in orths.keys(), "Segment not found in Reference corpus"
```
JackTemaki, line 56: `assert tag in orths.keys(), "Segment %s not found in reference corpus" % s`

corpus/convert.py
```diff
@@ -12,6 +18,95 @@
...
+                tag = r.segments[i].fullname()
+                orths[tag] = orth
+
+        c = corpus.Corpus()
+        c.load(self.bliss_corpus.get_path())
+
+        for r in c.all_recordings():
+            for i in range(len(r.segments)):
+                tag = r.segments[i].fullname()
+               assert tag in orths.keys(), "Segment %s not found in reference corpus" % s
```
JackTemaki, line 56:
```suggestion
                assert tag in orths.keys(), "Segment %s not found in reference corpus" % s
```

JackTemaki (approved):

PR #220, opened 2022-02-21
support dicts in ReturnnDumpHDFJob

michelwi:
the docstring suggested that data should only be of type `str` while an inline comment suggested `dict|Path|str`.
the old implementation definitively worked for strings - although 
```
if isinstance(data, str):
    data = str(data)
```
seemed kinda weird to me - and otherwise assumed Path - again, I am unsure how a path defines a dataset. 

Anyway, now support for dicts is added while - I think - the handling of str and Path is unchanged.

Kudos to Jan Engler for the `resolve_sis_paths_in_dict` function. I think it might be useful in many places and therefore added it under util.

### DIFF
--- a/returnn/hdf.py
+++ b/returnn/hdf.py
@@ -33,7 +33,7 @@ class ReturnnDumpHDFJob(Job):
     ):
         """
 
-        :param dict|Path|str data: a string defining a RETURNN dataset
+        :param dict|Path|str data: a dict, path, or string defining a RETURNN dataset
         :param start_seq: first sequence to dump in the dataset
         :param end_seq: last sequence to dump in the dataset
         :param int epoch: epoch to dump
--- a/returnn/hdf.py
+++ b/returnn/hdf.py
@@ -5,7 +5,7 @@ import subprocess as sp
 
 from .rasr_training import ReturnnRasrTrainingJob
 import i6_core.rasr as rasr
-from i6_core.util import relink
+from i6_core.util import relink, instanciate_delayed
 
 from sisyphus import *
 
@@ -72,7 +72,7 @@ class ReturnnDumpHDFJob(Job):
     def run(self):
         data = self.data
         if isinstance(data, dict):
-            resolve_sis_paths_in_dict(data)
+            instanciate_delayed(data)
             data = str(data)
         elif isinstance(data, tk.Path):
             data = data.get_path()
--- a/returnn/hdf.py
+++ b/returnn/hdf.py
@@ -33,7 +33,7 @@ class ReturnnDumpHDFJob(Job):
     ):
         """
 
-        :param str data: a string defining a RETURNN dataset
+        :param dict|Path|str data: a string defining a RETURNN dataset
         :param start_seq: first sequence to dump in the dataset
         :param end_seq: last sequence to dump in the dataset
         :param int epoch: epoch to dump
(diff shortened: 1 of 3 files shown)


util.py
```diff
@@ -279,3 +279,10 @@ def check_file_sha256_checksum(filename, reference_checksum):
     :param str filename: a single file to be checked
     """
     assert compute_file_sha256_checksum(filename) == reference_checksum
+
```
michelwi, line 282:
```suggestion


```
for Black

returnn/hdf.py
```diff
@@ -33,7 +33,7 @@ def __init__(
     ):
         """
 
-        :param str data: a string defining a RETURNN dataset
+        :param dict|Path|str data: a string defining a RETURNN dataset
```
michelwi, line 36:
```suggestion
        :param dict|Path|str data: a dict, path, or string defining a RETURNN dataset
```
update docstring

michelwi (commented):
some inline changes

util.py
```diff
@@ -279,3 +279,11 @@ def check_file_sha256_checksum(filename, reference_checksum):
...
+
+def resolve_sis_paths_in_dict(dictionary):
```
michelwi, line 284:
```suggestion
def resolve_sis_paths_in_dict(dictionary):
    """
    Iterates through a (nested) dict and replaces all Path instances 
    with the respective file paths. Only to be used in the Worker.
    
    :param dict dictionary: a dict containing Path objects to be replaced
    """
```
@JackTemaki you like the docstring?

JackTemaki: > Only to be used in the Worker

Perfect!

util.py
```diff
@@ -279,3 +279,17 @@ def check_file_sha256_checksum(filename, reference_checksum):
...
```
vieting, line 284: sisyphus has a function `extract_paths` which does a very similar thing ([here](https://github.com/rwth-i6/sisyphus/blob/master/sisyphus/tools.py#L74)). Maybe that could be used here as well? On the other hand, this here is probably easier to read.

michelwi: I know about this function. The problem is, that it flattens the data structure in the process (I.e. Returned is a flat list of all path-strings).
I didn't find an elegant solution to write it back in the dict structure.

vieting: Alright, then let's keep it as it is.

vieting (approved):

JackTemaki:
Looks good, but if we are at it, shouldn't the function also cover lists and tuples as well? And maybe also `Variables`?

dthulke:
Doesn't `instanciate_delayed` do all of this already?
What about moving this function to util?

https://github.com/rwth-i6/i6_core/blob/970cd004a83d7220dd6aa288a72bb93a3a7a6373/returnn/config.py#L21-L28

michelwi:
> Doesn't `instanciate_delayed` do all of this already? What about moving this function to util?

Yes, thanks. Exactly what is needed. I am in favor of moving it to util.py as tis might be used by more than just the returnn config.

JackTemaki (approved):

michelwi:
`instanciate_delayed` is used in i6_experiments by @JackTemaki and @mmz33 here:
https://github.com/rwth-i6/i6_experiments/search?q=instanciate_delayed

do we have to change the import or will it still work as it is imported in `i6_core.returnn.config`?

JackTemaki:
> `instanciate_delayed` is used in i6_experiments by @JackTemaki and @mmz33 here: https://github.com/rwth-i6/i6_experiments/search?q=instanciate_delayed
> 
> do we have to change the import or will it still work as it is imported in `i6_core.returnn.config`?

As long as it is still imported further imports will work. @vieting can you review again so that we can merge?

PR #221, opened 2022-03-04
add WriteRasrConfigJob

JackTemaki:
There are some use cases where you just want to dump a config (e.g. to link them in a RETURNN training job with some custom rasr loss without ugly local job hacks)

### DIFF
--- a/rasr/config.py
+++ b/rasr/config.py
@@ -344,7 +344,7 @@ class StringWrapper:
         return self.string
 
 
-class WriteRasrConfigJob(Job, RasrCommand):
+class WriteRasrConfigJob(RasrCommand, Job):
     """
     Write a RasrConfig object into a .config file
     """
--- a/rasr/config.py
+++ b/rasr/config.py
@@ -11,6 +11,7 @@ import itertools as it
 from sisyphus import Job, Task
 
 from i6_core import util
+from i6_core.rasr.command import RasrCommand
 
 
 class RasrConfig:
@@ -343,7 +344,7 @@ class StringWrapper:
         return self.string
 
 
-class WriteRasrConfigJob(Job):
+class WriteRasrConfigJob(Job, RasrCommand):
     """
     Write a RasrConfig object into a .config file
     """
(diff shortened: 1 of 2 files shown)


rasr/config.py
```diff
@@ -338,3 +341,32 @@ def __init__(self, string, hidden=None):
...
+        self.config = config
+        self.post_config = post_config
+
+        self.out_config = self.output_path("rasr.config")
+
+    def tasks(self):
+        yield Task("run", mini_task=True)
+
+    def run(self):
+        self.config._update(self.post_config)
```
vieting, line 365: Should we make this method public?

rasr/config.py
```diff
@@ -338,3 +341,32 @@ def __init__(self, string, hidden=None):
...
+        with util.uopen(self.out_config, "wt") as f:
+            f.write(repr(self.config))
```
michelwi, line 367: Would it be possible/useful/elegant to inherit from `i6_core.rasr.RasrCommand` and use `self.write_config` instead?

At least to make config writing consistent to all other jobs in case the `RasrCommand.write_config` method ever changes.

vieting, line 367: We could add `RasrConfig.write(filename)` similar to `ReturnnConfig` and then use this at other occasions (e.g. in `AdvancedTreeSearchJob.write_config()` as well.

vieting (commented):
Looks good, just two general comments related two `RasrConfig`.

rasr/config.py
JackTemaki: It should be public, but that changes a lot of files...

rasr/config.py
JackTemaki: Good hint, will change

rasr/config.py
JackTemaki: I will do a new PR with this, if needed.

vieting: I think this is fine, but feel free to do that if you like of course.

JackTemaki:
I also added a test case, just because this should now be a standard for new jobs :) With this this we can potentially test some RasrConfig serialization issues if needed.

vieting (approved):

rasr/config.py
```diff
@@ -338,3 +342,34 @@ def __init__(self, string, hidden=None):
 
     def __str__(self):
         return self.string
+
+
+class WriteRasrConfigJob(Job, RasrCommand):
```
michelwi, line 347:
```suggestion
class WriteRasrConfigJob(RasrCommand, Job):
```
Its the other way round in AdvancedTreesearchJob. Does this make a difference?

JackTemaki: No it does not, but we can still do it for consistency.

PR #222, opened 2022-03-04
G2POutputToBlissLexiconJob use correct libs

JackTemaki:
The old code had the problem that all the added lemmas would be written in to one single line.

@christophmluscher pointed out that this is not nice.

Also it makes sense to use the "official" lexicon helper we have.

### DIFF
--- a/g2p/convert.py
+++ b/g2p/convert.py
@@ -6,7 +6,8 @@ import xml.etree.ElementTree as ET
 
 from sisyphus import *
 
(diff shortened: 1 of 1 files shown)


michelwi (approved):
> Also it makes sense to use the "official" lexicon helper we have.

I agree.

Also, I think the order of lemmata written to file will stay the same with this change.

christophmluscher (approved):

PR #223, opened 2022-03-10
fix ReturnnRasrDumpHDFJob

JackTemaki:
At least for me the old version did not work at all because of some missing parameters and incorrect dataset.

### DIFF
--- a/returnn/hdf.py
+++ b/returnn/hdf.py
@@ -116,7 +116,12 @@ class ReturnnRasrDumpHDFJob(ReturnnDumpHDFJob):
         alignment,
         num_classes,
         buffer_size=200 * 1024,
-        **kwargs,
+        cpu=2,
+        mem=8,
+        file_size=100,
+        time=4,
+        returnn_python_exe=None,
+        returnn_root=None,
     ):
         """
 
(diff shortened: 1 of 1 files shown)


returnn/hdf.py
```diff
@@ -125,11 +130,32 @@ def __init__(
         :param Path alignment:
         :param int num_classes:
         :param int buffer_size:
-        :param kwargs: parameters for ReturnnDumpHDFJob
+        :param int cpu: number of CPU cores required
+        :param int mem: RAM required
+        :param int file_size: request file space on compute node
+        :param int time: compute time
```
Atticus1806, line 136: Maybe it would make sense to add units here for these parameters in the docs.

Atticus1806 (commented):
Cant really comment on the RASR part looks fine in general, but maybe you could add documentation for `feature_flow` etc. aswell if youre already working on the job.

michelwi (approved):

christophmluscher (commented):
code looks good. I have one point where I am concerned and that is that the created hdf file is directly dumped to a file server...

michelwi:
> code looks good. I have one point where I am concerned and that is that the created hdf file is directly dumped to a file server...

true, but this should be part of a different PR for the underlying DumpHDFJob.

christophmluscher:
OK. that is fine for me :)

christophmluscher (approved):

PR #226, opened 2022-03-11
add ApplyLexiconToTranscriptionsJob

JackTemaki:
In some cases, when you are not using a bliss corpus with RASR but with RETURNN, it might be needed to convert the words into phonemes not on-the-fly but offline. Especially for TTS this is needed, but maybe there are other use cases as well.

### DIFF
--- a/corpus/transform.py
+++ b/corpus/transform.py
@@ -17,7 +17,7 @@ import os
 import wave
 import xml.etree.cElementTree as ET
 
-from i6_core.lib import corpus
+from i6_core.lib import corpus, lexicon
 from i6_core.util import uopen
 
 from sisyphus import *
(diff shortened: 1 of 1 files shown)


corpus/transform.py
```diff
@@ -491,3 +491,73 @@ def run(self):
...
+    def __init__(
+        self,
+        bliss_corpus,
+        bliss_lexicon,
+        word_separation_orth,
+        strategy=LexiconStrategy.PICK_FIRST,
+    ):
+        """
+        :param bliss_corpus: path to a bliss corpus xml
+        :param bliss_lexicon: path to a bliss lexicon file
```
Atticus1806, line 516: Maybe add typing here?

corpus/transform.py
```diff
@@ -491,3 +491,73 @@ def run(self):
...
+        lookup_dict = {}
+        for lemma in lex.lemmata:
+            for orth in lemma.orth:
+                if orth and self.strategy == LexiconStrategy.PICK_FIRST:
+                    if len(lemma.phon) > 0:
+                        lookup_dict[orth] = lemma.phon[0]
+
+        if self.word_separation_orth is not None:
+            word_separation_phon = lookup_dict[self.word_separation_orth]
+            print("using word separation symbold: %s" % word_separation_phon)
```
Atticus1806, line 549: typo

corpus/transform.py
```diff
@@ -491,3 +491,73 @@ def run(self):
...
+            separator = " %s " % word_separation_phon
```
Atticus1806, line 550: just to be sure: is the space behind the `word_separation_phon` intended?

Atticus1806 (commented):
Looks good, only small comments which should be easily fixable.

corpus/transform.py
```diff
@@ -491,3 +491,73 @@ def run(self):
 
         with open(str(self.out_segments), "w") as segments_outfile:
             segments_outfile.writelines(segment_file_names)
+
+
+class LexiconStrategy(enum.Enum):
+    PICK_FIRST = 0
+
+
+class ApplyLexiconToTranscriptionsJob(Job):
```
michelwi, line 500:
```suggestion
class ApplyLexiconToCorpusJob(Job):
```
I suggest a renaming to make more clear that this is applied to a bliss corpus (as opposed to plain text files)

corpus/transform.py
```diff
@@ -491,3 +491,73 @@ def run(self):
...
+    Use a bliss lexicon to convert all words in a bliss corpus into their phoneme representation.
+
+    Currently only supports picking the first phoneme.
+    """
+
+    def __init__(
+        self,
+        bliss_corpus,
+        bliss_lexicon,
+        word_separation_orth,
```
michelwi, line 511:
```suggestion
        word_separation_orth=None,
```
maybe make None the default.

corpus/transform.py
```diff
@@ -491,3 +491,73 @@ def run(self):
...
+        self.strategy = strategy
+
+        self.out_corpus = self.output_path("corpus.xml.gz")
+
+    def tasks(self):
+        yield Task("run", mini_task=True)
+
+    def run(self):
+        c = corpus.Corpus()
+        c.load(self.bliss_corpus.get_path())
```
michelwi, line 534: just for pure code coherence I would move these lines below to where the corpus is processed.

corpus/transform.py
JackTemaki: Yes, this is intended for `separator.join(words)`.

michelwi (commented):
some comments for your consideration

corpus/transform.py
JackTemaki: But "None" would not give valid results... one could use `" "` but I did not want to add that so that people have to think about what they need.

JackTemaki: Ok, will add `None`

Atticus1806 (approved):

PR #229, opened 2022-03-21
add arg for compile_tf_graph

christophmluscher:
add the option to set the device for tf graph compilation job

the script `compile_tf_graph.py` from `returnn` used to take the device from the given config. now it is an argument for the script and is by default set to "cpu". I think most people did not change the value in the returnn config. So the previous default would be "gpu".

### DIFF
--- a/returnn/compile.py
+++ b/returnn/compile.py
@@ -21,7 +21,7 @@ class CompileTFGraphJob(Job):
 
     """
 
-    __sis_hash_exclude__ = {"device": "gpu"}
+    __sis_hash_exclude__ = {"device": "cpu"}
 
     def __init__(
         self,
@@ -30,7 +30,7 @@ class CompileTFGraphJob(Job):
         eval=0,
         search=0,
         verbosity=4,
-        device="gpu",
+        device="cpu",
         summaries_tensor_name=None,
         output_format="meta",
         returnn_python_exe=None,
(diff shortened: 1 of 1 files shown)


christophmluscher:
1. I would argue for including it in the hash.
- For example comparing decoding methods on CPU and GPU becomes harder since you cannot distinguish for which architecture the graph was compiled.
- The result of the graph should not change but I thought the graph itself changes!? And then the output of the job changes.
2. I was thinking how not to change the current hash with the current setup. But I would be fine with Willis suggestion if we do not break hashes

JackTemaki:
> So the previous default would be "gpu".

But didn't most people run this job on CPU nodes and did not specify anything in the config? At least for me this was not set, because this would be set within jobs.

> or do you use this also for Weis LabelSyncSearch on GPU?
That is a clear yes from my side.

JackTemaki:
It should definitely be part of the hash, for the default I am really not sure if it should be "None" or "CPU", but I think CPU is good.

christophmluscher:
> But didn't most people run this job on CPU nodes and did not specify anything in the config? At least for me this was not set, because this would be set within jobs.

I think you are right. I set it manually in some configs but then I also ran decoding on GPU. 

I am fine with "cpu" as default

michelwi:
1. Ok, we include it in the hash.
2.  I am fine with both "cpu" and None.

JackTemaki (approved):

PR #231, opened 2022-03-24
Move `bpe` to `label` and add sentencepiece training

JackTemaki:
I thought it is a good idea to have all "label" related jobs in a single sub-package. Of course the old `bpe` location is kept, containing "linked" jobs so that any setup relying on this keeps working.

If there are any comments on the names of the jobs, now would be the time to change them.

I also added a job for Sentence-Piece training.


label/bpe/train.py
```diff
@@ -0,0 +1,160 @@
...
+
+import i6_core.util as util
+
+Path = setup_path(__package__)
+
+
+class TrainBPEModelJob(Job):
+    """
+    Create a bpe codes file using the official subword-nmt repo, either installed from pip
+    or https://github.com/rsennrich/subword-nmt
```
albertz, line 18: Btw, you can do the same with the sentencepiece software. I'm not sure if there are any differences. If there are, I expect the sentencepiece software to work better. At least, it also should be faster.

label/bpe/apply.py
```diff
@@ -0,0 +1,145 @@
...
+
+        elem = lexicon.to_xml()
+        tree = ET.ElementTree(elem)
+        with util.uopen(self.out_converted_lexicon.get_path(), "wb") as f:
+            tree.write(f, encoding="utf-8")
+
+
+class ApplyBPEToTextJob(Job):
+    """
+    Apply BPE codes on a text file
```
albertz, line 107: Btw, you can do the same with the sentencepiece software. I'm not sure if there are any differences. If there are, I expect the sentencepiece software to work better. At least, it also should be faster.

label/bpe/train.py
JackTemaki: Yes sure, but I would not trust that this results in the exact same BPE codes that are compatible with Kazukis LM models.

Of course we could add a comment that you should use the BPE mode of the Sentencepiece software for new setups.

albertz: Two ideas what we could (maybe should) do:

- At least mention this in the docstring here.
- Maybe rename the job to reflect that this is the job specifically for SubwordNMT.

The same applies for most other jobs here and in apply.py. I'm not repeating that now for all cases.

JackTemaki: Okay updated the docstrings and renamed the whole package to "subword_nmt" so that it is clear it is not about bpe but about this implementation in specific.

albertz (approved):

bpe/apply.py
```diff
@@ -1,145 +1,27 @@
...
-import subprocess as sp
-import os
-import sys
-import xml.etree.ElementTree as ET
 
-from sisyphus import *
+from i6_core.label.subword_nmt.apply import (
+    ApplyBPEModelToLexiconJob as _ApplyBPEModelToLexiconJob,
+)
+from i6_core.label.subword_nmt.apply import ApplyBPEToTextJob as _ApplyBPEToTextJob
```
michelwi, line 11: do we really need this
```
from i6_core.label.subword_nmt.apply import ApplyBPEToTextJob as _ApplyBPEToTextJob
class ApplyBPEToTextJob(_ApplyBPEToTextJob):
    pass
````
or would
```
from i6_core.label.subword_nmt.apply import ApplyBPEToTextJob
```
also be sufficient?

would there be implications for hashing (probably no as you tested this), comparison/isinstance test (probably no because of inheritance)?

File: label/sentencepiece/train.py [hunk 4]
```diff
@@ -0,0 +1,74 @@
...
+        self.model_type = model_type
+        self.character_coverage = character_coverage
+        self.additional_options = additional_options or {}
+
+        self.out_model = self.output_path("spm_out.model")
+        # there is no need to make a tk.Variable for this, as the vocab_size is exact
+        # just pass through
+        self.out_vocab_size = vocab_size
+
+        self.rqmt = {"cpu": 1, "mem": 2, "time": 4}
```
michelwi, line 50: will we need to update this at some point in time or is this universally enough.
maybe the `__init__` should then take some kwargs to modify this.

label/sentencepiece/train.py
```diff
@@ -0,0 +1,74 @@
...
+        sentencepiece.SentencePieceTrainer.Train(
+            input=training_text_path,
+            model_prefix="spm_out",
+            model_type=self.model_type.value,
+            vocab_size=self.vocab_size,
+            character_coverage=self.character_coverage,
+            **self.additional_options,
+        )
+
+        shutil.move("spm_out.model", self.out_model.get_path())
```
michelwi, line 74: would we maybe want some local caching for intermediate outputs or is this only a short computation?

label/subword_nmt/apply.py
```diff
@@ -0,0 +1,144 @@
+__all__ = ["ApplyBPEModelToLexiconJob", "ApplyBPEToTextJob"]
```
michelwi, line 1: these jobs probably don't work for sentencepiece models, right?

Will there be apply Jobs for the new module as well?

michelwi (commented):
I am a bit hesitant about the naming of this package: "label" seems to be a rather broad term. 
E.g. also cart classes from RASR are labels, speaker labels, language ID labels, etc.

what you currently have here are more like "subword units" (ok, maybe sentencepice can put together and then split multiple words).
Would e.g. Weis ADSM also go in this package? what other "labels" would you ideally like to include here?

label/sentencepiece/train.py
JackTemaki: I would only do it for jobs where you really want to do this for every instance, here it is usually sufficient, so changing it via `job_instance.rqmt = ....` should be fine.

bpe/apply.py
JackTemaki: Just importing should be fine... I just did this out of habit when moving/renaming jobs. I can test this.

label/sentencepiece/train.py
```diff
@@ -0,0 +1,74 @@
...
+
+        self.out_model = self.output_path("spm_out.model")
+        # there is no need to make a tk.Variable for this, as the vocab_size is exact
+        # just pass through
+        self.out_vocab_size = vocab_size
+
+        self.rqmt = {"cpu": 1, "mem": 2, "time": 4}
+
+    def tasks(self):
+        yield Task("run", rqmt=self.rqmt)
```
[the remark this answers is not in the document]
JackTemaki: Making it resumable would only make sense if it runs out of time or memory... hmm I guess that is okay,

label/subword_nmt/apply.py
JackTemaki: Currently they are not needed, but we can add them once they are. I can add a "TODO" issue for that if you want.

requirements.txt
```diff
@@ -3,4 +3,5 @@ h5py
 matplotlib
 numpy
 soundfile
+sentencepiece
```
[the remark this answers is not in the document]
JackTemaki: > I am a bit hesitant about the naming of this package: "label" seems to be a rather broad term.

I agree that label is too broad. What would you propose? 

@Atticus1806 @albertz @mmz33 what is your opinion on the naming?

JackTemaki:
> I am a bit hesitant about the naming of this package: "label" seems to be a rather broad term.

I agree that label is too broad. What would you propose? 

@Atticus1806 @albertz @mmz33 what is your opinion on the naming?

mmz33:
We could have the directory `label` under the `text` directory since all jobs create labels from the text so it is related.

In addition, ESPNet do it like this like here: https://github.com/espnet/espnet/tree/master/espnet2/text

JackTemaki:
@michelwi @albertz any objection to Mohammads suggestion? Otherwise I would fix the remaining issues and go with `text/label(s)`

albertz:
> @michelwi @albertz any objection to Mohammads suggestion? Otherwise I would fix the remaining issues and go with `text/label(s)`

Sounds good.

label/sentencepiece/train.py
```diff
@@ -0,0 +1,74 @@
...
+        self.training_text = training_text
+        self.vocab_size = vocab_size
+        self.model_type = model_type
+        self.character_coverage = character_coverage
+        self.additional_options = additional_options or {}
+
+        self.out_model = self.output_path("spm_out.model")
+        # there is no need to make a tk.Variable for this, as the vocab_size is exact
+        # just pass through
+        self.out_vocab_size = vocab_size
```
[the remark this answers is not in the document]
JackTemaki: We do not want Variables here, because otherwise we are forced to use async Sisyphus workflows when working with `returnn_common`... I will remove it then.

label/sentencepiece/train.py
JackTemaki: < 5 minutes and < 1 mb in writing, so we can safely ignore this.

michelwi (approved):
> @michelwi @albertz any objection to Mohammads suggestion? Otherwise I would fix the remaining issues and go with `text/label(s)`

yes we can do it this way. This would also separate text-labels from frame-labels or other labels

bpe/apply.py
albertz: Via: https://github.com/rwth-i6/i6_core/issues/255, https://github.com/rwth-i6/i6_core/pull/256

To add here: Just an import (alias) is not enough. It must be an own instance. It does not have the same hash because the hash depends on the full qual name of the class (which means including the package and module of the class where it is defined).

PR #232, opened 2022-03-24
set None as default for device arg of CompileTFGraphJob

michelwi:
to maintain compatibility to older returnn versions we need the option to not provide the device flag.


michelwi:
Closed in favor of #233

PR #233, opened 2022-03-24
fix to support backward compatibility for compile tf job

christophmluscher:
@michelwi does this fix the issue?

### DIFF
--- a/returnn/compile.py
+++ b/returnn/compile.py
@@ -21,7 +21,7 @@ class CompileTFGraphJob(Job):
 
     """
 
-    __sis_hash_exclude__ = {"device": "cpu"}
+    __sis_hash_exclude__ = {"device": None}
 
     def __init__(
         self,
@@ -30,7 +30,7 @@ class CompileTFGraphJob(Job):
         eval=0,
         search=0,
         verbosity=4,
-        device="cpu",
+        device=None,
         summaries_tensor_name=None,
         output_format="meta",
         returnn_python_exe=None,
@@ -43,7 +43,7 @@ class CompileTFGraphJob(Job):
         :param int eval:
         :param int search:
         :param int log_verbosity: RETURNN log verbosity from 1 (least verbose) to 5 (most verbose)
-        :param str device: optimize graph for cpu or gpu
+        :param str|None device: optimize graph for cpu or gpu
         :param summaries_tensor_name:
         :param str output_format: graph output format, one of ["pb", "pbtxt", "meta", "metatxt"]
         :param Path|str returnn_python_exe: file path to the executable for running returnn (python binary or .sh)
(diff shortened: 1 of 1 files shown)


christophmluscher:
@michelwi sry did not see yours

christophmluscher:
> which is actually not the case if you start the graph compilation while on a machine with active GPU, or am I wrong there?

I think the default in the old setup would be set in the `returnn.config`??? for example `device ="gpu"`

I am not sure if the availability of a GPU further influences this

michelwi:
This line 
https://github.com/rwth-i6/returnn/blob/2ec00d42baf5acdf0a6f4b341fd1e6c050945008/tools/compile_tf_graph.py#L1480
Indicates the (new) default if no flag is set is "cpu", which would proliferate to this job.

christophmluscher:
ah I thought we were talking about the old default. extend doc with `from commit HASH_date forward default is: CPU`??

JackTemaki:
> 

I never set this in the returnn config... so well anything can happen. Willi is right, for the "new" RETURNN it defaults to cpu, and for old Returnn it either takes the device specified in the config OR it depends on the availability on the GPU if nothing is specified.

michelwi:
Ok, I will abandon the other pr and then we can put as docstring whatever we see fit.

returnn/compile.py
```diff
@@ -43,7 +43,7 @@ def __init__(
         :param int eval:
         :param int search:
         :param int log_verbosity: RETURNN log verbosity from 1 (least verbose) to 5 (most verbose)
-        :param str device: optimize graph for cpu or gpu
+        :param str|None device: optimize graph for cpu or gpu
```
JackTemaki, line 46:
```suggestion
        :param str|None device: optimize graph for cpu or gpu. If `None`, defaults to cpu for current RETURNN. 
            For any RETURNN version before `cd4bc382`, the behavior will depend on the `device` entry in the
            `returnn_conig`, which will depend on the availability of the GPU if not defined at all. 
```

JackTemaki: maybe like this?

michelwi: > the `device` entry in the
            `returnn_conig`, which will depend on the availability of the GPU if not defined at all.

Well not the device entry depends on the availability. Therefore

```
            For any RETURNN version before `cd4bc382`, the behavior will depend on the `device` entry in the
            `returnn_conig`, or on the availability of a GPU on the execution host if not defined at all.
```

JackTemaki:
Oh damn, I did not see that black failed? Was the test that much delayed?

returnn/compile.py
```diff
@@ -92,12 +94,13 @@ def run(self):
...
             "--eval=%d" % self.eval,
             "--search=%d" % self.search,
-            "--device=%s" % self.device,
             "--verbosity=%d" % self.verbosity,
             "--output_file=%s" % self.out_graph.get_path(),
             "--output_file_model_params_list=model_params",
             "--output_file_state_vars_list=state_vars",
         ]
+        if self.device is not None:
+          args.extend(["--device=%s" % self.device])
```
michelwi, line 103: Wrong amount of whitespace. 
Also why extend with a single item list when there is append directly below?

PR #234, opened 2022-03-25
fix local file caching for ReturnnDumpHDFJob

michelwi:
I'm sorry I only now tested the changes of #225. Turned out:
* fd is a file handle, not a file object
* relink does not work across filesystem boundaries (obviously duh)


michelwi:
ok, black fails because of #233 but why is flask now missing for the job tests?

christophmluscher (approved):

PR #236, opened 2022-03-25
addresses issue #235

christophmluscher:
addresses issue #235 

fix white space and cleaner code

PR #237, opened 2022-03-28
install sisyphus via pip

JackTemaki:
To solve the test issues in #236  and #234


michelwi (approved):

michelwi (changes requested):
I think you should remove the changes of #236 in `returnn/compile.py` from this PR as they solve a separate issue.

michelwi (approved):

christophmluscher (approved):

PR #238, opened 2022-03-28
add cpu and lmgc_mem to AdvancedTreeSearch

JackTemaki:
As the default value was often wrong, I wanted to adjust both the cpu count as well as the memory for the caching directly as parameter.

### DIFF
--- a/recognition/advanced_tree_search.py
+++ b/recognition/advanced_tree_search.py
@@ -176,6 +176,8 @@ class AdvancedTreeSearchJob(rasr.RasrCommand, Job):
         use_gpu=False,
         rtf=10,
         mem=4,
+        cpu=1,
+        lmgc_mem=12,
         model_combination_config=None,
         model_combination_post_config=None,
         extra_config=None,
@@ -210,7 +212,7 @@ class AdvancedTreeSearchJob(rasr.RasrCommand, Job):
 
         self.rqmt = {
             "time": max(crp.corpus_duration * rtf / crp.concurrent, 0.5),
-            "cpu": 1,
+            "cpu": cpu,
             "gpu": 1 if self.use_gpu else 0,
             "mem": mem,
         }
@@ -274,6 +276,7 @@ class AdvancedTreeSearchJob(rasr.RasrCommand, Job):
         eval_single_best,
         eval_best_in_lattice,
         mem,
+        lmgc_mem,
         model_combination_config,
         model_combination_post_config,
         extra_config,
@@ -283,7 +286,7 @@ class AdvancedTreeSearchJob(rasr.RasrCommand, Job):
         lm_gc = AdvancedTreeSearchLmImageAndGlobalCacheJob(
             crp, feature_scorer, extra_config, extra_post_config
         )
-        lm_gc.rqmt["mem"] = mem
+        lm_gc.rqmt["mem"] = lmgc_mem
 
         search_parameters = cls.update_search_parameters(search_parameters)
 


recognition/advanced_tree_search.py
```diff
@@ -176,6 +176,8 @@ def __init__(
         use_gpu=False,
         rtf=10,
         mem=4,
+        cpu=1,
```
michelwi, line 179: independent of keeping the old default.. shouldn't it be at least 2 cpus used (at least when using TF models) because RASR will anyway use more than one cpu?

Marvin84: I agree with @michelwi

recognition/advanced_tree_search.py
```diff
@@ -176,6 +176,8 @@ def __init__(
...
+        lmgc_mem=12,
```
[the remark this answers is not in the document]
Marvin84: I think 8GB would be a safer default without asking too much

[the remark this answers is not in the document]
JackTemaki:
```
+------- EPILOGUE SCRIPT -----------------------------------------------
|
| Job ID ..............: 5313274.1 (of 1 tasks)
| Stopped at ..........: Mon Mar 21 16:43:40 CET 2022
| Resources requested .: scratch_free=5G,num_proc=1,s_core=0,h_vmem=512G,h_rt=3600,h_rss=16G,gpu=0,hostname=!*cluster-cn-21*,qname=!*GPU*,h_fsize=50G,pxe=ubuntu_16.04
| Resources used ......: cpu=00:16:31, mem=4276.22323 GB s, io=8.29875 GB, vmem=1.795G, maxvmem=8.759G, last_file_cache=3.690G, last_rss=3M, max-cache=6.948G
| Memory used .........: 10.638G / 16.000G (66.5%)
| Total time used .....: 0:18:26
|
+------- EPILOGUE SCRIPT -----------------------------------------------
```
This is from a regular librispeech recognition job with 4-gram LM. I would not consider Librispeech a "special case".

recognition/advanced_tree_search.py
JackTemaki: Most of the legacy RASR setups can actually just make use of a single CPU. Only with the correct Tensorflow settings and MKL support it can actually use more. I will update the cpu count once I add those correct flags to the job, but I would not do it before.

michelwi: I remember that the AdvancedTreeSearchJobs regularly ignored the `MKL_NUM_THREADS` and `OMP_NUM_THREADS` directives and still used 1.5 cores worth of computation causing overload on the nodes. maybe @christophmluscher remembers and can comment.

JackTemaki: I think what you mean is that Tensorflow spawned two threads independend of those settings. And you are right about this, tensorflow can (and will) do that in some scenarios (Tensorflow and MKL threads are something different). But afaik this can not "escape" the CPU assignment limitation.

recognition/advanced_tree_search.py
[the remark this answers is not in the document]
michelwi: Ah sorry, I was confused.. 4TB is definitively not the ram usage 😅 

Ok, then I am convinced with these defaults.

recognition/advanced_tree_search.py
christophmluscher: I disagree. TF can escape the SGE limit. We see this regularly for AdvancedTreeSearchJobs with Neural AM aka TF. This is not much but we get high load warnings if this is not set correctly.

christophmluscher: A SGE Job with `cpu=1` uses then for example `cpu=1.1`

JackTemaki: Wait, I remember I wanted to change this at some point but Eugen stopped this...

JackTemaki: Also @christophmluscher the jobs from your setup were queued with cpu=1 as well... so why is this wrong now?

christophmluscher: because for a GMM-HMM decoding the default is `1` and this is fine. only for decoding with neural AM this is a problem. So only with RASR+TF this occurs. so when we see this issue occur we ask the users to increase manually. instead of just increasing for everybody.

I would leave at `1`

michelwi: ok, If you think so, then I will concur

michelwi (approved):

christophmluscher (commented):
as an idea, we could check if the flow file contains a tf node and then set the `cpu>=2`?

JackTemaki:
> as an idea, we could check if the flow file contains a tf node and then set the `cpu>=2`?

I opened an issue for that.

Marvin84 (approved):

Marvin84:
> > as an idea, we could check if the flow file contains a tf node and then set the `cpu>=2`?
> 
> I opened an issue for that.

That would be great.

christophmluscher:
a heuristic for cpu default: #244

christophmluscher (approved):

PR #239, opened 2022-03-29
pin click dependency

JackTemaki:


requirements.txt
```diff
@@ -1,4 +1,5 @@
 black==21.4b2
+click==8.0.4
```
michelwi, line 2:
```suggestion
click==8.0.4
black==21.4b2
```
should the order not be reversed s.t. the correct version of click is installed before black?

JackTemaki: Well for a requirements.txt I am not sure if this affects the install order. I still think it will only be installed once, as pip resolves everything and then the installation is arbitrary. Can still change if you want though

michelwi (approved):

mmz33 (approved):

PR #240, opened 2022-03-29
AlignSplitAccumulateSequence: Alias and Score Reports

JackTemaki:
I now often came across the problem that I did not know at which stage my current GMM training is, or which steps were already removed when running the cleaner. With this PR, a monophone Alignment job could get the name e.g.:
`alias/experiments/librispeech/librispeech_100_gmm/gmm_tina/train/train-clean-100_mono_action_sequence/action_28_align`

I also added a function which computes a report file from which you can directly read the convergence rate of the GMM training using the `AMScoresFromAlignmentLogJob` (to be honest I never knew it exists). Then even if all job get cleaned (which one should definitely do) you still can read the final alignment scores.

The rest is some added documentation (which is definitely still incomplete).

### DIFF
--- a/meta/mm_sequence.py
+++ b/meta/mm_sequence.py
@@ -6,13 +6,19 @@ __all__ = [
     "split_and_accumulate_sequence",
     "first_split_acc_then_align_split_acc",
 ]
+import os
 
-from i6_core.mm.alignment import AlignmentJob
+from i6_core.mm.alignment import AlignmentJob, AMScoresFromAlignmentLogJob
+from i6_core.mm.flow import FlowNetwork
 from i6_core.mm.mixtures import EstimateMixturesJob
 from i6_core.rasr import FlagDependentFlowAttribute, DiagonalMaximumScorer
 
 
 class AlignSplitAccumulateSequence:
+    """
+    Creates a sequence of AlignmentJobs and EstimateMixturesJobs to do HMM-GMM training.
+    """
+
     def __init__(
         self,
         crp,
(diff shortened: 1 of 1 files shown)


meta/mm_sequence.py
```diff
@@ -175,6 +201,17 @@ def update_rqmt(rqmt, extra):
...
+    def get_alignment_score_report(self):
+        """
+        :return: report .txt file path containing the alignment scores in order of the job sequence
+        """
+
+        logs = []
+        for job, log in zip(self.all_jobs, self.all_logs):
+            if isinstance(job, AlignmentJob):
+                logs.append(log)
+        return AMScoresFromAlignmentLogJob(logs).out_report
```
michelwi, line 213: Maybe not a good idea to return the `out_report` as then the job instance is not accessible any more. 
(ok, maybe via creator of Path object)

Ideas:
* return the Job and then the caller can get the report out of it.
* add it as a member of the `AlignSplitAccumulateSequence` class
* you can even add it in the constructor of the class and as long as the report is not registered as output the Job will not be created.

michelwi (commented):
looks good overall

meta/mm_sequence.py
JackTemaki: > add it as a member of the AlignSplitAccumulateSequence class

I will go for this one

meta/mm_sequence.py
```diff
@@ -175,6 +203,20 @@ def update_rqmt(rqmt, extra):
...
+        :return: report .txt file path containing the alignment scores in order of the job sequence
+        """
+        if self.report_job is not None:
+            return self.report_job.out_report
+
+        logs = []
+        for job, log in zip(self.all_jobs, self.all_logs):
+            if isinstance(job, AlignmentJob):
+                logs.append(log)
+        self.report_job = AMScoresFromAlignmentLogJob(logs)
```
michelwi, line 217:
```suggestion
        if self.report_job is None:
            logs = []
            for job, log in zip(self.all_jobs, self.all_logs):
                if isinstance(job, AlignmentJob):
                    logs.append(log)
            self.report_job = AMScoresFromAlignmentLogJob(logs)
```

just a suggestion that I think is simpler but I am fine either way.

michelwi (approved):

JackTemaki:
@christophmluscher @Marvin84 what are your opinions on this?

Marvin84:
> @christophmluscher @Marvin84 what are your opinions on this?

I find it useful, and therefore, approve.

Marvin84 (approved):

christophmluscher (approved):

PR #243, opened 2022-03-30
check for empty prons in G2POutputToBlissLexiconJob

michelwi:
We already check if one specific line does not contain a pronunciation. 

But if an orthography never has a finite pronunciation, then still a lemma is added

```
<lemma>
    <orth> some_example </orth>
</lemma>
```

I propose to only add lemmata with at least one pronunciation.

Question: should this always happen or should it be a controllable parameter?

### DIFF
--- a/g2p/convert.py
+++ b/g2p/convert.py
@@ -114,7 +114,8 @@ class G2POutputToBlissLexiconJob(Job):
             g2p_lexicon.lemmata = iv_lexicon.lemmata
 
         for orth, prons in oov_words.items():
-            lemma = lexicon.Lemma(orth=[orth], phon=prons)
-            g2p_lexicon.add_lemma(lemma)
+            if len(prons) > 0:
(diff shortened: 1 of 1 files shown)


JackTemaki:
> I propose to only add lemmata with at least one pronunciation.
> 
> Question: should this always happen or should it be a controllable parameter?

I think for the g2p job it makes sense to only add lemmata with a pronunciation. A parameter that we could introduce is that we crash the job if not all words get a valid pronunciation... but this is maybe future work.

JackTemaki (approved):

michelwi:
> > I propose to only add lemmata with at least one pronunciation.
> > Question: should this always happen or should it be a controllable parameter?
> 
> I think for the g2p job it makes sense to only add lemmata with a pronunciation. A parameter that we could introduce is that we crash the job if not all words get a valid pronunciation... but this is maybe future work.

This check I would rather add to the apply job than this.

christophmluscher (approved):

PR #245, opened 2022-04-07
ReturnnConfig: parse delayed ops in prolog/epilog

vieting:
In one of my `ReturnConfig`'s prolog, I have a `DelayedFormat` object. Currently, this cannot be parsed, therefore I'd like to add it.

### DIFF
--- a/returnn/config.py
+++ b/returnn/config.py
@@ -229,7 +229,7 @@ class ReturnnConfig:
         if isinstance(code, str):
             return code
         if isinstance(code, DelayedBase):
-            return code.get()
+            return self.__parse_python(code.get())
         if isinstance(code, (tuple, list)):
             return "\n".join(self.__parse_python(c) for c in code)
         if isinstance(code, dict):


returnn/config.py
```diff
@@ -227,6 +227,8 @@ def __parse_python(self, code, name=None):
             return ""
         if isinstance(code, str):
             return code
+        if isinstance(code, DelayedBase):
+            return code.get()
```
michelwi, line 231: I am not sure that `DelayedBase.get()` always returned a string. 

should we assert that here?
would we like to handle other types? Maybe a delayed function should be turned into a function and then handled as source code?

vieting:
The import was lost when rebasing... I added it again.

returnn/config.py
vieting: That't true. We could do
```suggestion
        if isinstance(code, DelayedBase):
            return self.__parse_python(code.get())
```

JackTemaki:
The ReturnnConfig should already support this via `config = instanciate_delayed(config)` in the `_serialize` function. @vieting can you say for which case this does not work?

returnn/config.py
JackTemaki: This suggestion seems like the most consistent way.

michelwi (approved):

vieting:
@JackTemaki what do you think?

JackTemaki (approved):

JackTemaki:
I just noticed that this functionality is actually not hash-safe...., because the parse op is also called during the hash computation itself. So the hash will be different depending on if the variable exists or not.

@michelwi  shame on us I guess for not noticing this...

PR #248, opened 2022-04-14
ExtractDatasetMeanStddevJob, use uncached_path for returnn

albertz:


michelwi:
calling `get()` will call `get_path()` which I think already returns the path without caching even if you activated caching for the path object.  So this change shouldn't be necessary.
(cf. https://github.com/rwth-i6/sisyphus/blob/5f1373a0da21c8189ac316a60dea83a926bd2584/sisyphus/job_path.py#L174-L193)

In fact according to #32 we want to get rid of `tk.uncached_path` altogether.

albertz:
A `str` object does not have a `get` function.

albertz:
Also, shouldn't it be consistent to basically every other job?

michelwi (approved):
> A `str` object does not have a `get` function.

according to the doc string, only Path objects or None is allowed

> Also, shouldn't it be consistent to basically every other job? E.g. see ReturnnTrainingJob.

Yes, I can accept this argument. Although the solution should be to disallow `str` everywhere else. 
But for now, in the name of consistency I'll allow it.

albertz:
> according to the doc string, only Path objects or None is allowed

Yes, and I'm passing `None`, so it uses `gs.RETURNN_PYTHON_EXE`, which cannot be a `tk.Path` but must be a `str`.

But maybe we also should fix the docstring to make it consistent to all other jobs.

JackTemaki:
As a comment here, the non-support of `str` was on purpose, and `tk.uncached_path` should NEVER be put into new code.

albertz:
> As a comment here, the non-support of `str` was on purpose,

This PR here was about fixing the (default) case `returnn_python_exe=None` and `returnn_root=None`, not about adding `str` support.

> and `tk.uncached_path` should NEVER be put into new code.

I just did it for consistency to all the other jobs.

How would you suggest to do it otherwise?

But also, if we choose to do it otherwise, maybe we first should change it for the other jobs then?

PR #250, opened 2022-04-14
ReturnnConfig serialize, don't sort config when sort_dicts=False

albertz:


Atticus1806 (approved):

albertz:
> > I was actually even considering to just remove the sorted here. It should not be sorted.
> 
> Why is that? The order should not matter.

It matters a lot when you look and inspect the config file.

> And then we can have it sorted so that the order is consistent.

It is also consistent when it is not sorted. Insertion order makes much more sense than ordering by lexical name.

michelwi:
With the same argument I prefer it to be sorted.
People might insert things in an inconsistent order or add options later which would then appear at the bottom of the config and are easily overlooked when inspecting a config visually

christophmluscher:
If I remember correctly this was introduced before python dict kept the insert order. Which means that it was either random or sorted. With python 3.8 (?), dicts started to store insertion order, the prettyprint arg sort_dicts was introduced. For the network dict.
At the moment it seems to me that the default behaviour is sorted `returnn.config` with sorted or insert-order dict inside `returnn.config`. I think a lot of people prefer this (see Willi), so we should keep this option available.
If we want to change this, I think we should follow Willis suggestion:

> Mh.. Actually the pprint_kwargs are explicitly only for the PrettyPrinter class. I think if you want to have the config not sorted globally, then we should introduce an additional argument to the class.

albertz:
> With python 3.8 (?)

Python 3.7.

> I think a lot of people prefer this (see Willi)

I would expect most people would prefer insertion order.

returnn/config.py
```diff
@@ -79,6 +79,7 @@ def __init__(
         python_epilog="",
         python_epilog_hash=None,
         hash_full_python_code=False,
+        sort_dicts=True,
```
michelwi, line 82:
```suggestion
        sort_config=True,
```

returnn/config.py
```diff
@@ -96,6 +97,7 @@ def __init__(
         :param str|None python_epilog_hash: sets a specific hash for the python_epilog
         :param bool hash_full_python_code: By default, function bodies are not hashed. If set to True, the full content
             of python pro-/epilog is parsed and hashed.
+        :param bool sort_dicts: If set to True, dictionaries are sorted by key
```
michelwi, line 100:
```suggestion
        :param bool sort_config: If set to True, the dictionary part of the config is sorted by key
```

returnn/config.py
```diff
@@ -117,7 +119,9 @@ def __init__(
                 self.python_epilog_hash = self.__parse_python(python_epilog)
             else:
                 self.python_epilog_hash = python_epilog
+        self.sort_dicts = sort_dicts
```
michelwi, line 122:
```suggestion
        self.sort_config = sort_config
```

returnn/config.py
```diff
@@ -117,7 +119,9 @@ def __init__(
...
         self.pprint_kwargs = pprint_kwargs or {}
+        self.pprint_kwargs.setdefault("sort_dicts", sort_dicts)
```
michelwi, line 124:
```suggestion
        self.pprint_kwargs.setdefault("sort_dicts", sort_config)
```

returnn/config.py
```diff
@@ -185,7 +189,7 @@ def _serialize(self):
         unreadable_data = {}
 
         pp = pprint.PrettyPrinter(indent=2, width=150, **self.pprint_kwargs)
-        for k, v in sorted(config.items()):
+        for k, v in sorted(config.items()) if self.sort_dicts else config.items():
```
michelwi, line 192:
```suggestion
        for k, v in sorted(config.items()) if self.sort_config else config.items():
```

michelwi (commented):
Thanks, I consider my requested change to be fulfilled.

But one more thing: I think the name `sort_dicts` is not well named as it refers only to sorting the top level config items which are (in the file) rather `parameter = value` terms instead of dict entries.

I therefore propose the name `sort_config`.
What do the others think?

michelwi (approved):

PR #251, opened 2022-04-19
remove setdefault sort_dicts for PrettyPrinter kwargs

michelwi:
In python < 3.8 the `pprint.PrettyPrinter` class does not accept the keyword arg `sort_dicts`.
(c.f. https://docs.python.org/3.7/library/pprint.html)

If needed, this option can still be set via the `pprint_kwargs` manually.

### DIFF
--- a/returnn/config.py
+++ b/returnn/config.py
@@ -8,6 +8,7 @@ import os
 import pickle
 import pprint
 import string
+import sys
 import textwrap
 
 from sisyphus import *
@@ -121,6 +122,8 @@ class ReturnnConfig:
                 self.python_epilog_hash = python_epilog
         self.sort_config = sort_config
         self.pprint_kwargs = pprint_kwargs or {}
+        if sys.version_info[:2] >= (3, 8):
+            self.pprint_kwargs.setdefault("sort_dicts", sort_config)
         self.black_formatting = black_formatting
 
     def get(self, key, default=None):
--- a/returnn/config.py
+++ b/returnn/config.py
@@ -121,7 +121,6 @@ class ReturnnConfig:
                 self.python_epilog_hash = python_epilog
         self.sort_config = sort_config
         self.pprint_kwargs = pprint_kwargs or {}
-        self.pprint_kwargs.setdefault("sort_dicts", sort_config)
         self.black_formatting = black_formatting
 
     def get(self, key, default=None):


returnn/config.py
```diff
@@ -121,7 +121,6 @@ def __init__(
                 self.python_epilog_hash = python_epilog
         self.sort_config = sort_config
         self.pprint_kwargs = pprint_kwargs or {}
-        self.pprint_kwargs.setdefault("sort_dicts", sort_config)
```
albertz, line 124: I would still set it for Python >=3.8.

albertz (changes requested):
I would still set it for Python >=3.8.

returnn/config.py
albertz: Or, we could also maybe be more generic:
```python
if sys.version_info[:2] <= (3, 8):
    self.pprint_kwargs.pop("sort_dicts", None)  # only supported since Python >=3.8
```
So then Sis configs works for different Python versions.

michelwi: I would rather not introduce special behavior based on the python version used to run the recipe so that all configs and parameters are the same for all users.

Also, in my opinion, sorting the config and sorting dicts within the config are two reasonably different things so that having two separate, independent config parameters makes sense.

albertz: > Also, in my opinion, sorting the config and sorting dicts within the config are two reasonably different things so that having two separate, independent config parameters makes sense.

I disagree. I don't see any reason that you would want to sort only one of them. As you are not using this option anyway, I think we should ask the opinion of people who will use this option.

It should not be so complicated. It's not nice to have to set two options for sth which should even be the default (in my opinion).

Also, we should really not care too much about older Python versions. This is just a workaround to still support it. We could also maybe just drop Python <=3.7 support and then clean this up.

curufinwe: Python 3.6.9 is the default python that comes with ubuntu 18.04 which is still what we use at AppTek (cluster + test pipeline), which is currently broken due to that change.

albertz: So why not just use my proposed change? It should anyway not have an impact on the result. And if you anyway don't use it, it will not change anything for you.

Atticus1806:
>which is currently broken due to that change

It breaks via #250 and is fixed by removing? Or it breaks through this PR?

>Python 3.6.9 is the default python that comes with ubuntu 18.04

In general I think the question of which python to support is quite central. For `returnn_common` I think this is `3.7` , but `3.6.9` also makes sense if AppTek still uses it. 

I find it reasonable to assume that, if the sorting is not manually set, it mirrors the behaviour of the whole config. If it should not, setting it in `pprint_kwargs` is always an option. 
What I am unsure about is the version check. I see that its not an optimal solution, but I think the feature might be worth the "price", so as long as this version check is not breaking anything it would be fine with it.

albertz:
> > which is currently broken due to that change
> 
> It breaks via #250 and is fixed by removing? Or it breaks through this PR?

It breaks via #250 for Python <=3.7. This PR is supposed to fix it.

> > Python 3.6.9 is the default python that comes with ubuntu 18.04
> 
> In general I think the question of which python to support is quite central. For `returnn_common` I think this is `3.7` , but `3.6.9` also makes sense if AppTek still uses it.

What Python version to support in general is maybe orthogonal here or can be delayed, and we should have a separate discussion on this later maybe. This single small detail here should not be a reason to break Python <=3.7 support now, so such fix should be done in any case.

> I find it reasonable to assume that, if the sorting is not manually set, it mirrors the behaviour of the whole config.

What do you mean by manually set? And what option? You specifically refer to the `sort_config` option? It is either set to True (default because it was the old behavior) or False.

> If it should not, setting it in `pprint_kwargs` is always an option.

So, you say like it was in #250?

> What I am unsure about is the version check. I see that its not an optimal solution, but I think the feature might be worth the "price", so as long as this version check is not breaking anything it would be fine with it.

In any case, there is no difference in the results depending on any of these options, and no matter what type of suggested fix we use.

I don't think having such Python version check and having the dict (maybe unexpectedly) sorted is such a big deal, so I wonder a bit why we need to discuss this.

Atticus1806:
> What do you mean by manually set? And what option? You specifically refer to the sort_config option? It is either set to True (default because it was the old behavior) or False.

If `pprint_kwargs` already has `sort_dicts` the `setdefault` should not change anything right? I was referring to a comment from the review where it was discussed wether sorting both the same way should be default, with which I agree, because there is always the option to manually overwrite the default.

> I don't think having such Python version check and having the dict (maybe unexpectedly) sorted is such a big deal, so I wonder a bit why we need to discuss this.

I think here its fine, but we should be careful with using these checks, since they can quickly generate undesired behaviour when changing python versions. I agree though, here it should be fine and a sorted dict is no problem.

albertz (approved):

curufinwe (approved):

PR #252, opened 2022-04-19
DownloadAndPrepareTfDatasetJob

albertz:


datasets/tf_datasets.py
```diff
@@ -0,0 +1,117 @@
...
+        """
+        super().__init__()
+        self.dataset_name = dataset_name
+        self.max_simultaneous_downloads = max_simultaneous_downloads
+        self.max_workers = max_workers
+
+        self.out_data_dir = self.output_path("data_dir")
+
+    @classmethod
+    def hash(cls, kwargs):
```
JackTemaki, line 55: `hash` should be at the end, see https://github.com/rwth-i6/i6_core/blob/main/CONTRIBUTING.md

datasets/tf_datasets.py
```diff
@@ -0,0 +1,117 @@
...
+    Also maybe::
+
+        pip install datasets  # for Huggingface community datasets
+    """
+
+    def __init__(
+        self,
+        dataset_name: str,
+        *,
+        max_simultaneous_downloads: Optional[int] = None,
```
JackTemaki, line 34: I am not sure if it is a good idea to start mixing typing styles (python types vs docstring types). @curufinwe opinions on this?

datasets/tf_datasets.py
```diff
@@ -0,0 +1,117 @@
...
+    @classmethod
+    def hash(cls, kwargs):
+        # All other options except dataset_name are ignored, as they should not have an influence on the result.
+        d = {
+            "dataset_name": kwargs["dataset_name"],
+        }
+        return super().hash(d)
+
+    def tasks(self):
+        yield Task("run", resume="run", mini_task=True)
```
JackTemaki, line 63: The `resume` parameter does not make sense for mini-tasks, as it is only called when the job did not crash but goes "missing" from the SGE. The `localengine` which is used in case of `mini_task=True` has no support for `gs.STATE_INTERRUPTED`, which is the only case in which resume can be used.

datasets/tf_datasets.py
```diff
@@ -0,0 +1,117 @@
...
+        :param max_workers:
+            For download extractor (https://github.com/tensorflow/datasets/issues/3887)
+            and Apache Beam.
+        """
+        super().__init__()
+        self.dataset_name = dataset_name
+        self.max_simultaneous_downloads = max_simultaneous_downloads
+        self.max_workers = max_workers
+
+        self.out_data_dir = self.output_path("data_dir")
```
JackTemaki, line 52: I guess the directory is created automatically, but usually it makes sense to set `directory=True` so that it is created by Sisyphus and it is clear that this is a directory and not a file. But this is optional.

JackTemaki (changes requested):

datasets/tf_datasets.py
albertz: We used docstring types in RETURNN only because we wanted to have it Python 2 compatible. Also, I think PyCharm is the only tool which really makes use of them. Otherwise it's a bad idea to use docstring types and you always should use the standard type hints, where there are also other tools who can make use of them.

JackTemaki: It is clear that using standard types is better, but this is about if we want it consistent or not.

albertz: I would say that we also should change it everywhere else to standard types (but this can be done gradually), and even disallow any new code with docstring types.

datasets/tf_datasets.py
```diff
@@ -0,0 +1,117 @@
+"""
+https://www.tensorflow.org/datasets
+"""
+
```
michelwi, line 4: what is this? Is this supposed to be a docstring? If yes, can you please add documentation.

datasets/tf_datasets.py
```diff
@@ -0,0 +1,117 @@
...
+    ):
+        """
+        :param dataset_name:
+            https://www.tensorflow.org/datasets/overview
+            https://www.tensorflow.org/datasets/catalog/overview
+            https://www.tensorflow.org/datasets/community_catalog/huggingface
+        :param max_simultaneous_downloads: https://github.com/tensorflow/datasets/issues/3885
+        :param max_workers:
+            For download extractor (https://github.com/tensorflow/datasets/issues/3887)
+            and Apache Beam.
```
michelwi, line 45: A link to a github issue is **not** a documentation. Please provide a documentation.

michelwi (changes requested):
your documentation of this class can be improved

datasets/tf_datasets.py
```diff
@@ -0,0 +1,117 @@
...
+
+from typing import Optional
+from sisyphus import *
+
+
+class DownloadAndPrepareTfDatasetJob(Job):
+    """
+    https://www.tensorflow.org/datasets
+
+        pip install tensorflow-datasets
```
albertz, line 14:
```suggestion
      Install the dependencies:

        pip install tensorflow-datasets
```

datasets/tf_datasets.py
albertz:
```suggestion
        :param max_simultaneous_downloads: max simultaneous downloads
            (https://github.com/tensorflow/datasets/issues/3885)
        :param max_workers: max workers, for download extractor and Apache Beam
            (https://github.com/tensorflow/datasets/issues/3887)
```

datasets/tf_datasets.py
```diff
@@ -0,0 +1,117 @@
...
+        dataset_name: str,
+        *,
+        max_simultaneous_downloads: Optional[int] = None,
+        max_workers: Optional[int] = None,
+    ):
+        """
+        :param dataset_name:
+            https://www.tensorflow.org/datasets/overview
+            https://www.tensorflow.org/datasets/catalog/overview
+            https://www.tensorflow.org/datasets/community_catalog/huggingface
```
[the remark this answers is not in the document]
albertz: Well I totally disagree that this is a better documentation, as it is longer without adding any more information. But as you wish.

datasets/tf_datasets.py
albertz: Btw, why do you want to have the documentation here much longer than in all other modules (basically non-existing)?

michelwi (approved):
Thank you both for updating the docstrings.

I am in favor of changing the typing over time.

PR #258, opened 2022-04-22
Add write functions in `MixtureSet` class

mmz33:
The write functions are used but missing for some reason.

### DIFF
--- a/lib/rasr_cache.py
+++ b/lib/rasr_cache.py
@@ -834,12 +834,13 @@ class MixtureSet:
         return a
 
     # write routines
-    def write_str(self, s):
+    def write_str(self, s, enc="ascii"):
         """
         :param str s:
+        :param str enc:
         :rtype: int
         """
-        return self.f.write(pack("%ds" % len(s), s.encode("ascii")))
+        return self.f.write(pack("%ds" % len(s), s.encode(enc)))
 
     def write_char(self, i):
         """
--- a/lib/rasr_cache.py
+++ b/lib/rasr_cache.py
@@ -833,6 +833,58 @@ class MixtureSet:
         a.fromfile(self.f, size)
         return a
 
+    # write routines
+    def write_str(self, s):
+        """
+        :param str s:
+        :rtype: int
+        """
+        return self.f.write(pack("%ds" % len(s), s.encode("ascii")))
+
+    def write_char(self, i):
+        """
+        :param int i:
+        :rtype: int
+        """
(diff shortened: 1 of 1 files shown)


lib/rasr_cache.py
```diff
@@ -833,6 +833,58 @@ def read_v(self, size, a):
         a.fromfile(self.f, size)
         return a
 
+    # write routines
+    def write_str(self, s):
+        """
+        :param str s:
+        :rtype: int
+        """
+        return self.f.write(pack("%ds" % len(s), s.encode("ascii")))
```
michelwi, line 842:
```suggestion
    def write_str(self, s, enc="ascii"):
        """
        :param str s:
        :rtype: int
        """
        return self.f.write(pack("%ds" % len(s), s.encode(enc)))
```
make encoding a parameter like in the read function.

michelwi (changes requested):
in the `__init__` we have 
```
self.f = open(filename, "rb")
```

which is only closed upon deleting the `MixtureSet` object.
```
    def __del__(self):
        self.f.close()
```

now in your write method, `self.f` is overwritten by the new file
```
self.f = open(filename, "wb")
```
without closing the old file.

I understand this is not your code, but I think this should be fixed nevertheless.
Possible solutions I see:
* close the old `self.f` before reopening the new one
* close all files in the method they are opened (e.g. 
```
with open(filename, 'rb') as self.f:
    do_stuff()
```
* for writing open a local file and add a file_handle argument to all write functions
```
    def write(self, filename):
        f = open(filename, "wb")
        self.write_str(f, self.header + "t")

    def write_str(self, file, s):
        return file.write(pack("%ds" % len(s), s.encode("ascii")))
```

michelwi (approved):

christophmluscher:
do we have a Situation where we load a dataset, manipulate and save? Or is it more like: either load and then do something, or do something and save?
I might be cleaner if we split the current `__init__` into a `__init__` and `load` function?

michelwi:
> do we have a Situation where we load a dataset, manipulate and save? Or is it more like: either load and then do something, or do something and save? 

"dataset" -> "mixture". I am not sure what @mmz33 intends to do, but besides estimating mixtures and writing dummy mixtures I don't see anything one could do to create a mixture "from thin air" without loading one first - and there are already jobs for both tasks.

JackTemaki (approved):
okay now I understand this PR, those functions really just vanished...

PR #260, opened 2022-04-29
fix hash bug in cmllr flow

JackTemaki:
After discussing with @michelwi and the affected people @Marvin84 and @christophmluscher we came to the conclusion that it is better to not fix this issue "backwards compatible (example fix in https://github.com/rwth-i6/i6_core/tree/nick_cmllr_true_fix)", but just as is.

This will invalidate all existing HMM-GMM SAT training run graphs, which would have to be migrated before updating.

Due to the severity of this fix I would prefer if all reviewers would accept before merging.

### DIFF
--- a/sat/flow.py
+++ b/sat/flow.py
@@ -2,6 +2,8 @@ __all__ = ["add_cmllr_transform", "segment_clustering_flow"]
 
 import os
 
+from sisyphus.delayed_ops import DelayedFormat
+
 from i6_core.features import add_linear_transform
 from i6_core.rasr import FlowNetwork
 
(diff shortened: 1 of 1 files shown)


sat/flow.py
```diff
@@ -42,7 +44,13 @@ def add_cmllr_transform(
     multiply = net.add_node(
         "signal-matrix-multiplication-f32",
         "apply-cmllr-transform",
-        {"file": os.path.join(str(transform_dir), matrix_name)},
+        {
+            "file": DelayedFormat(
+                "{transform_dir}/{matrix_name}",
+                transform_dir=transform_dir,
+                matrix_name=matrix_name,
+            )
```
michelwi, line 52: if `transform_dir` is a Path, then you should be able to use 
```
transform_dir.join_right(matrix_name)
```
to simplify

JackTemaki: Okay this function is new to me...

JackTemaki: Okay, changed it and added docstrings (I think we should do this for every code we touch now)

michelwi (approved):

Marvin84 (approved):

christophmluscher (approved):

JackTemaki:
@curufinwe any last objection? If not I will merge this soon.

PR #261, opened 2022-05-02
convert {returnn,rasr}_{root,exe} to path

michelwi:
Based on recent discussions it seems that at many positions we allow `str` where we actually want to only support `Path`.
But oftentimes the default for these paths is taken from `sisyphus.global_settings` which does not support setting `Path` objects (and anyway is a bad place to put these defaults, but let this be a different discussion).

Here I propose to change the type of these arguments to `Optional[Path]` i.e. disallowing to give a `str`. If nothing is set, the usual `str` from `sisyphus.global_settings` is converted to a `Path` object.

Open questions:
* Should we keep allowing `str` as input (and then internally convert it to `Path`)?
* Should we disallow `str` also for other inputs, where we actually want to use `Path`? (e.g. `ReturnnTrainingFromFileJob.returnn_config`)
* With the changes I did, will all paths be registered as input and therefore verified by sisyphus?
* Some scorers also use some strings from `sisyphus.global_settings`, but only in the `run` method. I think it is pointless to convert into `Path` and then immediately call `get_path()`. Should we keep this for now? 
* Does this change hashes, when the member is now a `Path` instead of a `str`?


rasr/feature_scorer.py
```diff
@@ -83,7 +85,7 @@ def __init__(
         else:
             self.config.normalize_mixture_weights = False
         self.config.pymod_name = "returnn.SprintInterface"
-        self.config.pymod_path = os.path.join(tk.gs.RETURNN_ROOT, "..")
+        self.config.pymod_path = returnn_root.join_right("..")
```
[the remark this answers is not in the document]
christophmluscher:
```
ImportError: No module named returnn.SprintInterface
<?xml version="1.0" encoding="UTF-8"?>
<sprint>
  <critical-error component="acoustic-model-trainer.average-feature-scorer-activation.mixture-set">
    PythonTrainer: cannot import module 'returnn.SprintInterface'
    <system>
      No such file or directory
    </system>

    Creating stack trace (innermost first):
    #1  /work/tools/asr/rasr/20191102_generic/arch/linux-x86_64-standard/acoustic-model-trainer.linux-x86_64-standard(_ZNK4Core9Component13vErrorMessageENS0_9ErrorTypeEPKcP13__va_list_tag+0x85b) [0xd837bb]
    #2  /work/tools/asr/rasr/20191102_generic/arch/linux-x86_64-standard/acoustic-model-trainer.linux-x86_64-standard(_ZNK4Core9Component14vCriticalErrorEPKcP13__va_list_tag+0x19) [0xd82e19]
    #3  /work/tools/asr/rasr/20191102_generic/arch/linux-x86_64-standard/acoustic-model-trainer.linux-x86_64-standard(_ZN2Nn13PythonTrainerIfE19pythonCriticalErrorEPKcz+0xb9) [0x10c9659]
    #4  /work/tools/asr/rasr/20191102_generic/arch/linux-x86_64-standard/acoustic-model-trainer.linux-x86_64-standard(_ZN2Nn13PythonTrainerIfEC1ERKN4Core13ConfigurationE+0x35b) [0x10cdf9b]
    #5  /work/tools/asr/rasr/20191102_generic/arch/linux-x86_64-standard/acoustic-model-trainer.linux-x86_64-standard(_ZN2Nn20NeuralNetworkTrainerIfE25createUnsupervisedTrainerERKN4Core13ConfigurationE+0x126) [0x10c23c6]
    #6  /work/tools/asr/rasr/20191102_generic/arch/linux-x86_64-standard/acoustic-model-trainer.linux-x86_64-standard(_ZN2Nn20TrainerFeatureScorerC1ERKN4Core13ConfigurationENS1_3RefIKN2Mm10MixtureSetEEE+0x1f3) [0x1057313]
    #7  /work/tools/asr/rasr/20191102_generic/arch/linux-x86_64-standard/acoustic-model-trainer.linux-x86_64-standard(_ZN2Mm20FeatureScorerFactory14createInstanceIN2Nn20TrainerFeatureScorerENS_10MixtureSetEEEPNS_13FeatureScorerERKN4Core13ConfigurationENS7_3RefIKNS_18AbstractMixtureSetEEE+0x89) [0xfe5f89]
    #8  /work/tools/asr/rasr/20191102_generic/arch/linux-x86_64-standard/acoustic-model-trainer.linux-x86_64-standard(_ZNK2Mm7Module_19createFeatureScorerERKN4Core13ConfigurationENS1_3RefIKNS_18AbstractMixtureSetEEE+0x98) [0xa0b0d8]
    #9  /work/tools/asr/rasr/20191102_generic/arch/linux-x86_64-standard/acoustic-model-trainer.linux-x86_64-standard(_ZN6Speech30AverageFeatureScorerActivationC1ERKN4Core13ConfigurationEb+0x146) [0x7e46a6]
    #10  /work/tools/asr/rasr/20191102_generic/arch/linux-x86_64-standard/acoustic-model-trainer.linux-x86_64-standard(_ZN20AcousticModelTrainer39calculateAverageFeatureScorerActivationEv+0x50) [0x7ca310]
    #11  /work/tools/asr/rasr/20191102_generic/arch/linux-x86_64-standard/acoustic-model-trainer.linux-x86_64-standard(_ZN20AcousticModelTrainer4mainERKSt6vectorINSt7__cxx1112basic_stringIcSt11char_traitsIcESaIcEEESaIS6_EE+0x17d) [0x79c77d]
    #12  /work/tools/asr/rasr/20191102_generic/arch/linux-x86_64-standard/acoustic-model-trainer.linux-x86_64-standard(_ZN4Core11Application3runERKSt6vectorINSt7__cxx1112basic_stringIcSt11char_traitsIcESaIcEEESaIS7_EE+0x20) [0xd6ec00]
    #13  /work/tools/asr/rasr/20191102_generic/arch/linux-x86_64-standard/acoustic-model-trainer.linux-x86_64-standard(_ZN4Core11Application4mainEiPPc+0x60e) [0x7b4a6e]
    #14  /work/tools/asr/rasr/20191102_generic/arch/linux-x86_64-standard/acoustic-model-trainer.linux-x86_64-standard(main+0x38) [0x79be18]
    #15  /lib/x86_64-linux-gnu/libc.so.6(__libc_start_main+0xf0) [0x14eafdc87840]
    #16  /work/tools/asr/rasr/20191102_generic/arch/linux-x86_64-standard/acoustic-model-trainer.linux-x86_64-standard(_start+0x29) [0x7c5969]
  </critical-error>
  <critical-error component="acoustic-model-trainer.average-feature-scorer-activation.mixture-set">
    Terminating due to previous errors
  </critical-error>
  <warning component="acoustic-model-trainer">
    There are 1 left-over Python::Initializer instances. We uninit Python now.
  </warning>
<?xml version="1.0" encoding="UTF-8"?>

```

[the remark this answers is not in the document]
christophmluscher:
```suggestion
        self.config.pymod_path = returnn_root
```

christophmluscher (changes requested):

rasr/feature_scorer.py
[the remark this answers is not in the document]
michelwi: This is the old, python based feature scorer interface that has been deprecated in favor of the direct tf implementation in RASR. I haven't used it in a long time.

I hypothesize the change required comes from the refactoring of returnn that moved everything away from the main directory into meaningful sub-folders.

> hm I tested this and in my setup i have to drop the join_right completely. So only returnn_root. I use the GitCloneRepoJob to get RETURNN and use out_repository

Does this work with the old `os.path.join(tk.gs.RETURNN_ROOT, "..")` for you?

Maybe @JackTemaki knows. He said that all old module locations should still be accessible after the refactoring.

But anyway, this is - I think - orthogonal to this PR and this feature scorer would only be used in legacy setups.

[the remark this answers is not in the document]
JackTemaki: But I think we can anyway just put any solution that works. I do not think anybody will use i6_core with a RETURNN version that is older than 2 years.

JackTemaki:
> Open questions:
> 
>     * Should we keep allowing `str` as input (and then internally convert it to `Path`)?

I would vote for no, but....

> 
>     * Should we disallow `str` also for other inputs, where we actually want to use `Path`? (e.g. `ReturnnTrainingFromFileJob.returnn_config`)

...as much as I would like that, unfortunately is not possible to create a `Path` object with the same hash as a string. My recommendation would be to drop `str` from the docstring/typing, but that the job itself can still work with it.

> 
>     * With the changes I did, will all paths be registered as input and therefore verified by sisyphus?

I do not understand what you mean here...

> 
>     * Some scorers also use some strings from `sisyphus.global_settings`, but only in the `run` method. I think it is pointless to convert into `Path` and then immediately call `get_path()`. Should we keep this for now?

This is fine for now.

> 
>     * Does this change hashes, when the member is now a `Path` instead of a `str`?

if it is only the member no, if it is a parameter to `__init__`, yes.

michelwi:
> unfortunately is not possible to create a Path object with the same hash as a string. 

sure, this is what [hash_overwrite](https://github.com/rwth-i6/sisyphus/blob/ee4e69fafcb004547188f18abc1f296de2e4a2d0/sisyphus/job_path.py#L54) is for.

> >* With the changes I did, will all paths be registered as input and therefore verified by sisyphus?

> I do not understand what you mean here...

I mean the messages "Input missing" if the file does not exist.

JackTemaki:
> > unfortunately is not possible to create a Path object with the same hash as a string.
> 
> sure, this is what [hash_overwrite](https://github.com/rwth-i6/sisyphus/blob/ee4e69fafcb004547188f18abc1f296de2e4a2d0/sisyphus/job_path.py#L54) is for.

Unfortunately it is not:
```python
    def _sis_hash(self):
        assert not isinstance(self.creator, str)
        if self.hash_overwrite is None:
            creator = self.creator
            path = self.path
        else:
            overwrite = self.hash_overwrite
            assert_msg = "sis_hash for path must be str or tuple of length 2"
            if isinstance(overwrite, tuple):
                assert len(overwrite) == 2, assert_msg
                creator, path = overwrite
            else:
                assert isinstance(overwrite, str), assert_msg
                creator = None
                path = overwrite
        if hasattr(creator, '_sis_id'):
            creator = os.path.join(self.creator._sis_id(), gs.JOB_OUTPUT)
        return b'(Path, ' + tools.sis_hash_helper((creator, path)) + b')'
 ```
Check the return statement. All you can do is that `Path("abc") and Path("def", hash_overwrite="abc")` have the same hash. But just `"abc"` will have a different hash.

> 
> > > * With the changes I did, will all paths be registered as input and therefore verified by sisyphus?
> 
> > I do not understand what you mean here...
> 
> I mean the messages "Input missing" if the file does not exist.

I think yes.

**Comment by michelwi** [comment]
> Check the return statement. All you can do is that Path("abc") and Path("def", hash_overwrite="abc") have the same hash. But just "abc" will have a different hash.

Ah.. You are right. :/

> My recommendation would be to drop str from the docstring/typing, but that the job itself can still work with it.

You mean both for the specific roots/exes that are given and also for other paramters that should have been Path in the first place?

**Comment by christophmluscher** [comment]
> My recommendation would be to drop str from the docstring/typing, but that the job itself can still work with it.

I like this :)

**Comment by christophmluscher** [comment]
Maybe a bit redundant... :D


> Should we keep allowing str as input (and then internally convert it to Path)?

Yes. backward compatibility

> Should we disallow str also for other inputs, where we actually want to use Path? (e.g. ReturnnTrainingFromFileJob.returnn_config)

No. But encourage the use of Path in new setups! I ike the typing usggestion by @JackTemaki 

> With the changes I did, will all paths be registered as input and therefore verified by sisyphus?

I think so.

> Some scorers also use some strings from sisyphus.global_settings, but only in the run method. I think it is pointless to convert into Path and then immediately call get_path(). Should we keep this for now?

Yes. We might want to move non sisyphus parts away from global_settings/settings file. We could do these changes together.

> Does this change hashes, when the member is now a Path instead of a str?

If in `__init__` yes

File: tools/default_software.py [hunk 2]
```diff
@@ -0,0 +1,43 @@
...
+from typing import Optional
+
+from sisyphus import gs, Path
+
+import i6_core.tools as tools
+
+RETURNN_COMMIT = None
+
+
+def get_returnn_root(returnn_root: Optional[Path]) -> Path:
```
michelwi, line 13: I think the whole handling starts to get rather complicated, so lets move it to a separate function. I put it into a separate package, so that we can treat all similar cases (RASR, g2p, ..) consistently.

What do you think about
  1. the idea of creating such functions in a separate package?
  2. this specific function for returnn_root?

tools/default_software.py
```diff
@@ -0,0 +1,43 @@
...
+        logging.warning("Giving returnn_root as str is deprecated, use tk.Path instead")
+        return Path(returnn_root)
+    elif returnn_root is None:
+        if hasattr(gs, "RETURNN_ROOT"):
+            logging.warning(
+                "Using gs.RETURNN_ROOT is deprecated, please provide an explicit tk.Path object"
+            )
+            return Path(gs.RETURNN_ROOT)
+        else:
+            returnn_root = tools.CloneGitRepositoryJob(
```
michelwi, line 36: At first I thought we should add a warning also in this branch, that the current HEAD at the time of first running is used which leads to inconsistencies. But then maybe 
  1. this warning should be added to `CloneGitRepositoryJob` in general
  2. we would like to provide a means to set a "default commit" for the default returnn

So therefor we now have a `i6_core.tools.default_software.RETURNN_COMMIT` constant. A user could now in principle do 
```
import i6_core.tools.default_software as ds
ds.RETURNN_COMMIT = "abcdef"
```
this might even work well in 99% of the cases, but if anybody ever does
```
sis c config/here_I_set_RETURNN_COMMIT.py config/here_I_dont.py
tk.cleaner.clean_work_dir(mode="remove")
```
then all trainings in the second config will be deleted. So maybe we should not facilitate this error by letting the user provide a different commit than HEAD?

michelwi:
I have redesigned the functional interface. The logic is now 
* prefer the given value
  * if given value is `str` then convert to `tk.Path`
  * else return as is
* then check the `gs` for the appropriate string and return this as path
* as a last resort return a default path that for now points to the python used to start sisyphus.

This now already also incorporates and extends https://github.com/rwth-i6/i6_core/pull/317

util.py
```diff
@@ -301,3 +302,43 @@ def instanciate_delayed(o):
         for k in o:
             o[k] = instanciate_delayed(o[k])
     return o
+
+
+def get_executable_path(
+    path: tk.Path, gs_member_name: str, default_exec_path: Optional[tk.Path] = None
```
JackTemaki, line 308: `path` and `gs_member_name` are also `Optional` as it is fully intended that they can be `None`, also maybe this function here could get a docstring.

util.py
```diff
@@ -301,3 +302,43 @@ def instanciate_delayed(o):
...
+) -> tk.Path:
+    if path is not None:
+        if isinstance(path, tk.Path):
+            return path
+        elif isinstance(path, str):
```
JackTemaki, line 313: We could issue a warning here that it is not recommended/deprecated to use strings.

michelwi: yes.. and also a warning if the `gs_member_name` is used..?

util.py
michelwi: `path` is Optional now. I left the `gs_member_name` as currently I do not support _not_ giving it.

It would be easy to add this support, but for all paths with no defined `gs_member_name` we don't need this function in the first place

rasr/feature_scorer.py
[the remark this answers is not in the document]
michelwi: Is this comment still relevant? In https://github.com/rwth-i6/i6_core/pull/317 you also left the two dots.

JackTemaki:
It generally seems good, but I do not want to accept this before I could test this with my own setup. Unfortunately I currently have my own pending branches, with is a little bit annoying.

JackTemaki (approved):

util.py
```diff
@@ -301,3 +303,73 @@ def instanciate_delayed(o):
...
+
+        return tk.Path(getattr(gs, gs_member_name))
+    if default_exec_path is not None:
+        return default_exec_path
+    assert False, f"could not find executable for {gs_member_name}"
+
+
+def get_returnn_root(returnn_root: tk.Path) -> tk.Path:
+    """gets the path to the root folder of RETURNN"""
+    return get_executable_path(returnn_root, "RETURNN_ROOT")
```
michelwi, line 351: Should we provide a `default_returnn_root` if nothing is given? e.g.
```suggestion
    default_root = CloneGitRepositoryJob("https://github.com/rwth-i6/returnn").out_repository
    return get_executable_path(returnn_root, "RETURNN_ROOT", default_root)
```
I don't like it because:
  * the user looses the control over the version that is checked out
  * updating to a more recent version becomes painful / manual work
  * other users that rerun the pipeline might get different results because their HEAD differs

One could instead hardcode the current HEAD, but this might not be very future proof.

michelwi (commented):
I (verbally) approve the changes of @JackTemaki

util.py
JackTemaki: I think we should have no default, it is better if this is slightly more verbose. I (and I think Albert did something similar independently) have now a `default_tools.py` file, which has content like:

```python
RETURNN_ROOT = CloneGitRepositoryJob("https://github.com/rwth-i6/returnn", commit="45fad83c785a45fa4abfeebfed2e731dd96f960c").out_repository
RETURNN_ROOT.hash_overwrite = "LIBRISPEECH_DEFAULT_RETURNN_ROOT"
```
So yes there is additional overhead in passing this to all the jobs, but it offers:
 - updating without hash change
 - updating with hash change
 - running specific experiments with a different version but same hash
 - running specific experiments with a a different version and different hash

JackTemaki:
@christophmluscher

christophmluscher (approved):

JackTemaki:
Apparently some import was missing now: b580c99a0f033c2193c3b85133e8aa37d16b11d1

PR #262, opened 2022-05-03
add update method to ReturnnConfig

michelwi:
This PR adds an `ReturnnConfig.update` method that behaves similar to `dict.update`. e.g.
* if one config key contains nested structures, these are not merged, but overwritten
* only shallow copy is performed

This is useful If you have a base config and only small updates for individual experiments or for managing different kinds of config parameters in different parts of the code.

### DIFF
--- a/returnn/config.py
+++ b/returnn/config.py
@@ -131,6 +131,58 @@ class ReturnnConfig:
             return self.post_config[key]
         return self.config.get(key, default)
 
+    def update(self, other):
+        """
+        updates a ReturnnConfig with an other ReturnnConfig
+
+        :param ReturnnConfig other:
+        """
+        self.config.update(other.config)
+        self.post_config.update(other.post_config)
+        if other.staged_network_dict is not None:
+            self.staged_network_dict = other.staged_network_dict
+
+        if self.python_prolog is None:
+            self.python_prolog = other.python_prolog
+        elif isinstance(self.python_prolog, "str") and isinstance(
+            other.python_prolog, "str"
+        ):
+            self.python_prolog += other.python_prolog
+        else:
+            self.python_prolog = [self.python_prolog, other.python_prolog]
+        if self.python_prolog_hash is None:
+            self.python_prolog_hash = other.python_prolog_hash
+        if isinstance(self.python_prolog_hash, "str") and isinstance(
+            other.python_prolog_hash, "str"
+        ):
+            self.python_prolog_hash += other.python_prolog_hash
+        else:
+            self.python_prolog_hash = [
+                self.python_prolog_hash,
+                other.python_prolog_hash,
+            ]
+
+        if isinstance(self.python_epilog, "str") and isinstance(
+            other.python_epilog, "str"
+        ):
+            self.python_epilog += other.python_epilog
+        else:
+            self.python_epilog = [self.python_epilog, other.python_epilog]
+        if isinstance(self.python_epilog_hash, "str") and isinstance(
+            other.python_epilog_hash, "str"
+        ):
+            self.python_epilog_hash += other.python_epilog_hash
+        else:
+            self.python_epilog_hash = [
+                self.python_epilog_hash,
+                other.python_epilog_hash,
+            ]
+
+        self.sort_config = other.sort_config
+        self.pprint_kwargs.update(other.pprint_kwargs)
+        self.black_formatting = other.black_formatting
+        self.check_consistency()
+
     def _write_to_file(self, content, file_path):
         """
         write with optional black formatting


returnn/config.py
```diff
@@ -131,6 +131,58 @@ def get(self, key, default=None):
...
+        self.post_config.update(other.post_config)
+        if other.staged_network_dict is not None:
+            self.staged_network_dict = other.staged_network_dict
+
+        if self.python_prolog is None:
+            self.python_prolog = other.python_prolog
+        elif isinstance(self.python_prolog, "str") and isinstance(
+            other.python_prolog, "str"
+        ):
+            self.python_prolog += other.python_prolog
```
christophmluscher, line 150: maybe something like this? so that we dont end up with `import numpy as npimport tensorflow as tf` in the `returnn.config`. or is this not an issue?

```suggestion
            self.python_prolog += "\n"
            self.python_prolog += other.python_prolog
```

returnn/config.py
```diff
@@ -131,6 +131,58 @@ def get(self, key, default=None):
...
+        else:
+            self.python_prolog_hash = [
+                self.python_prolog_hash,
+                other.python_prolog_hash,
+            ]
+
+        if isinstance(self.python_epilog, "str") and isinstance(
+            other.python_epilog, "str"
+        ):
+            self.python_epilog += other.python_epilog
```
christophmluscher, line 168: same as `python_prolog`?

returnn/config.py
```diff
@@ -131,6 +131,58 @@ def get(self, key, default=None):
...
+            other.python_prolog_hash, "str"
+        ):
+            self.python_prolog_hash += other.python_prolog_hash
+        else:
+            self.python_prolog_hash = [
+                self.python_prolog_hash,
+                other.python_prolog_hash,
+            ]
+
+        if isinstance(self.python_epilog, "str") and isinstance(
```
christophmluscher, line 165: handling of `python_prolog` and `python_epilog` should be the same, right?

but code is different.

could this be moved to a function??

christophmluscher (changes requested):

returnn/config.py
michelwi: mh.. I think you are right. Also using newline here would be compatible to https://github.com/rwth-i6/i6_core/blob/48ec2dfd3d2dde9305892428f16c91f585d4aea8/returnn/config.py#L239-L240

lets do 
```suggestion
            self.python_prolog += "\n" + other.python_prolog
```

returnn/config.py
michelwi: > handling of python_prolog and python_epilog should be the same, right?

unfortunately the default argument of prolog is `None` while epilog is `""`. therefore the different handling.
https://github.com/rwth-i6/i6_core/blob/48ec2dfd3d2dde9305892428f16c91f585d4aea8/returnn/config.py#L78-L81

returnn/config.py
christophmluscher: yes thats cleaner

returnn/config.py
michelwi: As the default for epilog is `""` there will be the case here that either of them or both are empty. We would then insert unnecessary newlines.

I could extend this code further by checks if either is empty and then only take the other..

returnn/config.py
```diff
@@ -131,6 +131,58 @@ def get(self, key, default=None):
...
+            self.python_epilog = [self.python_epilog, other.python_epilog]
+        if isinstance(self.python_epilog_hash, "str") and isinstance(
+            other.python_epilog_hash, "str"
+        ):
+            self.python_epilog_hash += other.python_epilog_hash
+        else:
+            self.python_epilog_hash = [
+                self.python_epilog_hash,
+                other.python_epilog_hash,
+            ]
```
michelwi, line 179: maybe this is nicer in terms of code simplification and it handles the case of default arguments being either `None` or `""`:
```suggestion
        def join_code(my_code, other_code):
            if my_code is None or my_code == "":
                return other_code
            if other_code is None or other_code == "":
                return my_code
            if isinstance(my_code, "str") and isinstance(other_code, "str"):
                return my_code + "\n" + other_code
            return [my_code, other_code]
        
        self.python_prolog = join_code(self.python_prolog, other.python_prolog)
        self.python_prolog_hash = join_code(self.python_prolog_hash, other.python_prolog_hash)
        self.python_epilog = join_code(self.python_epilog, other.python_epilog)
        self.python_epilog_hash = join_code(self.python_epilog_hash, other.python_epilog_hash)

```

JackTemaki:
@michelwi I will review once you updated everything.

returnn/config.py
```diff
@@ -131,6 +131,40 @@ def get(self, key, default=None):
             return self.post_config[key]
         return self.config.get(key, default)
 
+    def update(self, other):
+        """
+        updates a ReturnnConfig with an other ReturnnConfig
```
JackTemaki, line 136: I think we should write here how each component is updated.

returnn/config.py
```diff
@@ -131,6 +131,40 @@ def get(self, key, default=None):
...
+
+        :param ReturnnConfig other:
+        """
+        self.config.update(other.config)
+        self.post_config.update(other.post_config)
+        if other.staged_network_dict is not None:
+            self.staged_network_dict = other.staged_network_dict
```
JackTemaki, line 143: I would prefer also `update` here instead of overwriting, this would be more consistent.

JackTemaki (changes requested):

returnn/config.py
michelwi: I specifically decided for overwriting the `staged_network_dict` instead of updating with the use-case of "entirely replacing the network architecture" in mind.

Assume you have `baseline_conformer_training_config` and want to do experiments with different layer dimension, or transformer, or linformer,.. then you might use 
```
baseline_conformer_training_config.update(my_network_config)
```
In the update case, if now the stage epoch numbers do not exactly coincide it is easy to accidentally mess up the pretraining and hard to intentionally design `my_network_config` to do it correctly.

In the overwrite case, if you expected "update behavior" you would most likely immediately notice your error as the epoch=0 stage would be missing and if you intentionally want to do the correct thing, then 
```
baseline_conformer_training_config.staged_network_dict.update(my_stage_config.staged_network_dict)
```
is only minimal overhead.

returnn/config.py
michelwi:
```suggestion
        updates a ReturnnConfig with an other ReturnnConfig:
          * config, post_config, and pprint_kwargs use dict.update
          * prolog, epilog, and hashes are concatenated
          * staged_network_dict, sort_config, and black_formatting are overwritten
```
pending our discussion about `staged_network_dict`

returnn/config.py
JackTemaki: Ok, I take your argument as valid :) Then please keep the full replacement.

JackTemaki (approved):

christophmluscher (approved):

PR #263, opened 2022-05-06
simplify FilterSegmentsByAlignmentConfidenceJob

michelwi:
Parameter `crp` is not really needed as it is only used to find the number of alignment log files.


michelwi:
A quick reminder to review, now that you have more time after ICASSP.

PR #264, opened 2022-05-06
add assert for DelayedBase in parse_python

JackTemaki:
A quick rescue operation for https://github.com/rwth-i6/i6_core/pull/245

### DIFF
--- a/returnn/config.py
+++ b/returnn/config.py
@@ -120,6 +120,7 @@ class ReturnnConfig:
                 self.python_epilog_hash = self.__parse_python(python_epilog)
             else:
                 self.python_epilog_hash = python_epilog
+        self.hash_full_python_code = hash_full_python_code
         self.sort_config = sort_config
         self.pprint_kwargs = pprint_kwargs or {}
         if sys.version_info[:2] >= (3, 8):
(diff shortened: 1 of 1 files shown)


returnn/config.py
```diff
@@ -235,6 +236,10 @@ def __parse_python(self, code, name=None):
         if isinstance(code, str):
             return code
         if isinstance(code, DelayedBase):
+            assert self.hash_full_python_code is False, (
```
michelwi, line 239: I think this this is only necessary for `tk.Variable`. `tk.Path` should not change the hash over time.
So we _could_ allow using Path objects for now.

On the other hand, we might want to change the representation of Path/Variable in the Manager and then it would break some hashes. So maybe disallowing all `DelayedBase` objects for now is a good idea.

Thinking of it.. `tk.Path.get()` in the hash will make the hash dependent of the full path.. :/

michelwi (approved):
I think we should make conscious decisions which part of the code is run in the manager and may not access the values in Path/Variable objects and which part only runs in the worker and may resolve them.
And then we should start making use of `check_is_worker` also in the recipes.

And then we really should sort out the problem of `get` vs `get_path` and `__str__` vs. `__repr__`

christophmluscher (approved):

PR #266, opened 2022-05-11
ReturnnConfig, docstring update, fix types

albertz:
Fixed order.

Fixed too long line.

Fixed type of hashes.

### DIFF
--- a/returnn/config.py
+++ b/returnn/config.py
@@ -88,14 +88,15 @@ class ReturnnConfig:
 
         :param dict config: dictionary of the RETURNN config variables that are hashed
         :param dict post_config: dictionary of the RETURNN config variables that are not hashed
+        :param None|dict[int, dict[str, Any]] staged_network_dict: dictionary of network dictionaries,
+            indexed by the desired starting epoch of the network stage.
+            By enabling this, an additional "networks" folder will be created next to the config location
         :param None|str|Callable|Class|tuple|list|dict python_prolog: str or structure containing str/callables/classes
             that should be pasted as code at the beginning of the config file
-        :param None|dict[int, dict[str, Any]] staged_network_dict: dictionary of network dictionaries, indexed by the desired starting epoch of the network stage
-            By enabling this, an additional "networks" folder will be created next to the config location
-        :param str|None python_prolog_hash: sets a specific hash for the python_prolog
+        :param None|Any python_prolog_hash: sets a specific hash for the python_prolog
         :param None|str|Callable|Class|tuple|list|dict python_epilog: str or structure containing
             str/callables/classes that should be pasted as code at the end of the config file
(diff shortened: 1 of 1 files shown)


returnn/config.py
```diff
@@ -88,14 +88,15 @@ def __init__(
...
         :param dict post_config: dictionary of the RETURNN config variables that are not hashed
+        :param None|dict[int, dict[str, Any]] staged_network_dict: dictionary of network dictionaries,
+            indexed by the desired starting epoch of the network stage.
+            By enabling this, an additional "networks" folder will be created next to the config location
         :param None|str|Callable|Class|tuple|list|dict python_prolog: str or structure containing str/callables/classes
             that should be pasted as code at the beginning of the config file
-        :param None|dict[int, dict[str, Any]] staged_network_dict: dictionary of network dictionaries, indexed by the desired starting epoch of the network stage
-            By enabling this, an additional "networks" folder will be created next to the config location
-        :param str|None python_prolog_hash: sets a specific hash for the python_prolog
+        :param None|Any python_prolog_hash: sets a specific hash for the python_prolog
```
michelwi, line 96: `Optional[str]` is the intended type for python_prolog_hash and python_epilog_hash.

michelwi (changes requested):

returnn/config.py
albertz: Why? This is not how it is used in the code. In the code, in the default case, it will be just the same as `python_prolog` and `python_epilog`.

JackTemaki:
```suggestion
        :param None|Any python_prolog_hash: set a specific hash (str) or any type of hashable objects to overwrite the default hashing for python_prolog
```

I think with a docstring like this the change is ok.

michelwi: Oh.. I thought that despite the implementation that may assign Any type to the internal variable, the user was supposed to only give str as an Input.
But I see that there might be use cases where e.g. the name of a function changes but the user would want to retain the hash so they would give the old function as hash.

Ok, then I agree with the change.

michelwi (approved):

returnn/config.py
```diff
@@ -88,14 +88,15 @@ def __init__(
...
         :param None|str|Callable|Class|tuple|list|dict python_epilog: str or structure containing
             str/callables/classes that should be pasted as code at the end of the config file
-        :param str|None python_epilog_hash: sets a specific hash for the python_epilog
+        :param None|Any python_epilog_hash: sets a specific hash for the python_epilog
```
JackTemaki, line 99:
```suggestion
        :param None|Any python_epilog_hash: set a specific hash (str) or any type of hashable objects
            to overwrite the default hashing for python_prolog
```

JackTemaki (approved):

PR #267, opened 2022-05-16
Add parameter to add output 'features' to several feature extraction functions

curufinwe:
Same as https://github.com/rwth-i6/i6_core/pull/144 , but now also for multiple other feature flows.


features/filterbank.py
```diff
@@ -59,6 +60,8 @@ def filterbank_flow(
     :param dict[str, Any]|None fft_options: parameter dict for :func:`fft_flow`
     :param bool apply_log: adds a logarithm before normalization
     :param bool add_epsilon: if a logarithm should be applied, add a small epsilon to prohibit zeros
+    :param bool add_features_output: Add the output port "features" when normalize is True. This should be set to True,
```
JackTemaki, line 63: I do not understand what this has to do with `normalize`... this parameter and the fix for `normalize=False` are somewhat independent or not?

JackTemaki (approved):

PR #268, opened 2022-05-18
add MakeJob

michelwi:
tested with 
```
import i6_core.tools as tools
from sisyphus import tk


def py():
    sctk_repo = tools.CloneGitRepositoryJob(
        url="https://github.com/usnistgov/SCTK.git"
    ).out_repository
    sctk_root = tools.MakeJob(
        folder=sctk_repo, make_sequence=["config", "all", "check", "install", "doc"]
    ).out_repository.join_right("bin")
    tk.register_output("sctk", sctk_root)
```

to compile SCTK. Maybe we can use it also to compile RASR?

### DIFF
--- a/tools/__init__.py
+++ b/tools/__init__.py
@@ -1,1 +1,2 @@
 from .git import *
+from .compile import *
--- a/tools/compile.py
+++ b/tools/compile.py
@@ -1,0 +1,60 @@
+__all__ = ["MakeJob"]
+
+import logging
+import subprocess as sp
+import shutil
+from tempfile import TemporaryDirectory
+from typing import Iterator, Optional, List
+
+from sisyphus import tk, gs, setup_path, Job, Task
+
+Path = setup_path(__package__)
+
+
+class MakeJob(Job):
+    """
+    Executes a sequence of make commands in a given folder
+    """
+
+    def __init__(
+        self,
+        folder: tk.Path,
+        make_sequence: Optional[List[str]] = None,
+        num_processes: int = 1,
+        output_folder_name: str = "repository",
+    ):
+        """
+
+        :param folder: folder in which the make commands are executed
+        :param make_sequence: list of options that are given to the make calls.
+            defaults to ["all"] i.e. "make all" is executed
+        :param num_processes: number of parallel running make processes
+        :param output_folder_name: name of the output path folder
+        """
+        self.folder = folder
+        self.make_sequence = make_sequence if make_sequence is not None else ["all"]
+        self.num_processes = num_processes
+
+        self.rqmt = {"cpu": num_processes}
+
+        self.out_repository = self.output_path(output_folder_name, True)
+
+    def tasks(self) -> Iterator[Task]:
+        yield Task("run", resume="run", rqmt=self.rqmt)
+
+    def run(self):
+        with TemporaryDirectory(prefix=gs.TMP_PREFIX) as temp_dir:
+            shutil.rmtree(temp_dir)
+            shutil.copytree(self.folder.get_path(), temp_dir, symlinks=True)
+
+            for command in self.make_sequence:
+                args = ["make"]
+                args.extend(command.split())
+                if "-j" not in args:
+                    args.extend(["-j", f"{self.num_processes}"])
+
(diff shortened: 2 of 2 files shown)


JackTemaki:
For many repos you need to call `./configure` first, so I think we should have a parameter for this as well. Then we could extend RASR to also have a proper `./configure` to set e.g. "i6" compile mode or anything like this. Otherwise we would need to have different branches all the time.

JackTemaki:
Sorry that I was not specific, I apologize for that. We also need to be able to pass an arbitrary list of arguments to `configure`.

Extreme example for ffmpeg:
```
./configure \
  --prefix="$HOME/ffmpeg_build" \
  --pkg-config-flags="--static" \
  --extra-cflags="-I$HOME/ffmpeg_build/include" \
  --extra-ldflags="-L$HOME/ffmpeg_build/lib" \
  --extra-libs="-lpthread -lm" \
  --ld="g++" \
  --bindir="$HOME/bin" \
  --enable-gpl \
  --enable-gnutls \
  --enable-libaom \
  --enable-libass \
  --enable-libfdk-aac \
  --enable-libfreetype \
  --enable-libmp3lame \
  --enable-libopus \
  --enable-libsvtav1 \
  --enable-libdav1d \
  --enable-libvorbis \
  --enable-libvpx \
  --enable-libx264 \
  --enable-libx265 \
  --enable-nonfree && \
```

JackTemaki:
@michelwi how urgent do you need the merge? I would test this job with FFMpeg, https://github.com/openitu/STL and RASR and then look if there is something missing

JackTemaki:
Ah right, https://github.com/openitu/STL  is `cmake` would we also support that here or make a separate job for this?

michelwi:
> @michelwi how urgent do you need the merge? I would test this job with FFMpeg, https://github.com/openitu/STL and RASR and then look if there is something missing

Not very urgent. If it'd take you 1 month to do the testing then I would help you to speed things up, otherwise I'm find with a few weeks delay.

> would we also support `cmake` here or make a separate job for this?

If the syntax and handling is very similar (i.e. adding a `make_command="make"` is suficcient) then we can support this here. 

Otherwise we could extend this to a "CallArbitraryCommandInFolderJob" and then support also `patch << somefile.txt` (apply some local changes before compilation), or `bazel` (compile tensorflow), etc..

JackTemaki:
Okay I added some ideas, please feel free to comment and refine

michelwi (commented):
I am wondering if `output_folder_name: Optional[str] = "repository"` and `link_outputs: Optional[Dict[str, str]] = None` are redundand and could be merged into a single parameter. e.g. `link_outputs` defaults to `{"repository": "."}`.
But then you do check and hardlink in some cases which would be much harder for joint parameter.

I don't have a better solution atm, just brainstorming.

michelwi:
another thing that occurred to me is, that some repositories require to call make separately in some subfolders. e.g. Kaldi
https://github.com/kaldi-asr/kaldi/blob/master/INSTALL. Can we / do we want to support that?

JackTemaki:
@michelwi @christophmluscher do we want to give this Job a higher default memory requirement? I am not sure how this affects the speed in compiling e.g. RASR. We could do this either as (unhashed) parameter or just in the `self.rqmt`. I would vote for 4 Gb.

michelwi:
> @michelwi @christophmluscher do we want to give this Job a higher default memory requirement? I am not sure how this affects the speed in compiling e.g. RASR. We could do this either as (unhashed) parameter or just in the `self.rqmt`. I would vote for 4 Gb.

Sure, we can do that. As the memory rqmt is highly dependent on the project, I vote (with low confidence) to have it as a parameter.

JackTemaki:
@michelwi I have to wait for some packages to be installed in the cluster to compile RASR via the Job, but afterwards everything should be ready.

JackTemaki (approved):

christophmluscher (approved):

michelwi (commented):
Mh.. seems "pull request authors cannot approve their own pull request", but I nevertheless approve the changes @JackTemaki made verbally.

PR #270, opened 2022-05-30
Extend ReturnnConfig for returnn_common usage

JackTemaki:
Extends the ReturnnConfig to allow using returnn_common both as "within manager" and "within job".
 For "within manager" the resulting py_str can be added as network stage, while otherwise we just pass elements to prolog/epilog. In this case we neither have "network" in the dict nor set `staged_network_dict`, so we have to remove the assert.

Also contains some additional docstrings and formatting.

related to: https://github.com/rwth-i6/i6_experiments/issues/63

### DIFF
--- a/returnn/config.py
+++ b/returnn/config.py
@@ -88,9 +88,11 @@ class ReturnnConfig:
 
         :param dict config: dictionary of the RETURNN config variables that are hashed
         :param dict post_config: dictionary of the RETURNN config variables that are not hashed
-        :param None|dict[int, dict[str, Any]] staged_network_dict: dictionary of network dictionaries,
+        :param None|dict[int, str|dict[str, Any]] staged_network_dict:
+            dictionary of network dictionaries or
+            returnn_common generated strings (as str variable from get_ext_net_dict_py_code_str),
             indexed by the desired starting epoch of the network stage.
-            By enabling this, an additional "networks" folder will be created next to the config location
+            By enabling this, an additional "networks" folder will be created next to the config location.
         :param None|str|Callable|Class|tuple|list|dict python_prolog: str or structure containing str/callables/classes
             that should be pasted as code at the beginning of the config file
         :param None|Any python_prolog_hash: set a specific hash (str) or any type of hashable objects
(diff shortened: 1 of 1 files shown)


returnn/config.py
```diff
@@ -328,7 +344,6 @@ def check_consistency(self):
             assert key not in self.post_config, (
                 "%s in post_config would overwrite existing entry in config" % key
             )
-        assert not (self.staged_network_dict and "network" in self.config)
```
michelwi, line 331: >  In this case we neither have "network" in the dict nor set staged_network_dict, so we have to remove the assert.

this assert should only break if both `staged_network_dict` and `network` are given. Giving neither should be fine as 
```
assert not (False and False) => assert True
```

Atticus1806: I dont see a reason both being not given to ever make sense so I agree, this should be changed. 
Does python have built in XOR for these types? Otherwise I guess we should go with
`assert (self.staged_network_dict and not "network" in self.config) or (not self.staged_network_dict and "network" in self.config)`

Whats possible (but maybe a bit too hidden) is: `assert (bool(self.staged_network_dict) ^ bool("network" in self.config))`

JackTemaki: But this is exactly what I wrote in the comment, there is a case now where both are not given. Not sure why this assert was removed in my code, Willi is right I can just add it again and everything should be fine.

Atticus1806 (approved):

returnn/config.py
```diff
@@ -88,9 +93,11 @@ def __init__(
 
         :param dict config: dictionary of the RETURNN config variables that are hashed
         :param dict post_config: dictionary of the RETURNN config variables that are not hashed
-        :param None|dict[int, dict[str, Any]] staged_network_dict: dictionary of network dictionaries,
+        :param None|dict[int, str|dict[str, Any]] staged_network_dict:
+            dictionary of network dictionaries or
+            returnn_common generated strings (as str variable from get_ext_net_dict_py_code_str),
```
michelwi, line 98: this comment is misleading.. sure the change now is to incorporate returnn_common, but in principle any string that somehow contains `network = something_valid_that_defines_a_network` can be given here independent of returnn_common.

Maybe we can update this docstring along the lines "or a str containing valid python code that defines a variable named `network`" to make the interface more clear. 
Feel free to add a reference to `get_ext_net_dict_py_code_str` as an example source of such a string.

PR #271, opened 2022-06-01
add ReturnnVocabFromBlissLexiconJob

JackTemaki:
Create a Returnn compatible picked vocabulary file based on the phoneme inventory of a bliss lexicon.

### DIFF
--- a/returnn/vocabulary.py
+++ b/returnn/vocabulary.py
@@ -6,7 +6,7 @@ from i6_core.lib import lexicon
 from i6_core.util import uopen
 
 
-class ReturnnVocabFromBlissLexiconJob(Job):
+class ReturnnVocabFromPhonemeInventory(Job):
     """
     Create a pickled vocab for RETURNN based on the phoneme-inventory of a bliss lexicon
 
@@ -21,9 +21,9 @@ class ReturnnVocabFromBlissLexiconJob(Job):
     tk.Variable out_vocab_size: integer variable containing the vocabulary size (``vocab_size``)
     """
 
-    def __init__(self, bliss_lexicon):
+    def __init__(self, bliss_lexicon: tk.Path):
         """
-        :param tk.Path bliss_lexicon: a bliss lexicon xml file
+        :param bliss_lexicon: a bliss lexicon xml file containg a phoneme inventory
         """
         self.bliss_lexicon = bliss_lexicon
 


christophmluscher (approved):

returnn/vocabulary.py
```diff
@@ -0,0 +1,43 @@
+import pickle
+
+from sisyphus import Job, Task, tk
+
+from i6_core.lib import lexicon
+from i6_core.util import uopen
+
+
+class ReturnnVocabFromBlissLexiconJob(Job):
```
michelwi, line 9: 1. Would this not better fit in the returnn/label package as this also is about a specific label set 
2. The name is misleading (at least to me). The first idea I get from "Vocab from Bliss Lexicon" would be that a word list is extracted. It is not obvious that a) the phoneme inventory is extracted and b) that it is returned in pickle format.
3. Do we always need a returnn vocab in this format (pickled dict from "some_key" to ID)? I have until now only used vocab files in plaintext (List of one "some_key") format e.g. for LM training.
4. Would there be other things/vocabs that we need in this pickled format?

This is just a brainstorming thought, not a specific advice:
maybe this job should be separated into a `DumpPhonemeInventoryFromBlissLexiconJob` that returns the phoneme inventory as a plain text file and a `PickleVocabFromTextJob` that takes an arbitrary plaintext vocab file and converts it to the pickled dict format.
This way we can pickle also BPE vocabs (Is this ever needed?) or train phonem LMs (I even did this recently. I did use a Pipeline job to generate the vocab file from a Lexicon). 

Are there any other vocab formats that we might need to consider?

michelwi: > We only have text/label which is independent mostly independent of RETURNN.

Ok, I take back 1.

> But it is "Returnn-Vocab" from Bliss Lexicon

Ok, considering this and the comments below I take back 2b) but 2a) still holds. (ReturnnVocab)FromBlissLexicon could still be referring to the words and not the phonemes.
You don't have to rename this Job just to please me (esp. as the docstring is clear), but I just want to consider what happens in the future if we have a Job that converts the words of a Bliss lexicon to a Returnn-Vocab

> This does only work for very special/legacy Datasets.

Haha, so exactly the datasets that I have been using all the time 🙃 

> The general problem is that many RETURNN datasets expect some different input files, but here I wanted to fulfill the requirements of the "Official" RETURNN vocab class.

Ok, I consider 3. and 4. sufficiently addressed.

> If you check https://github.com/rwth-i6/returnn/blob/master/returnn/datasets/util/vocabulary.py, you can see that only .pkl files or files that eval to python dictionaries are accepted.

Adding another discussion point:
5. Would we maybe want to output "files that eval to python dictionaries" instead of ".pkl files" to make it easier to inspect them?

JackTemaki: > > But it is "Returnn-Vocab" from Bliss Lexicon
> 
> Ok, considering this and the comments below I take back 2b) but 2a) still holds. (ReturnnVocab)FromBlissLexicon could still be referring to the words and not the phonemes. You don't have to rename this Job just to please me (esp. as the docstring is clear), but I just want to consider what happens in the future if we have a Job that converts the words of a Bliss lexicon to a Returnn-Vocab
> 

I understand the problem... So.... `ReturnnVocabFromPhonemeInventory`? 

> > If you check https://github.com/rwth-i6/returnn/blob/master/returnn/datasets/util/vocabulary.py, you can see that only .pkl files or files that eval to python dictionaries are accepted.
> 
> Adding another discussion point: 5. Would we maybe want to output "files that eval to python dictionaries" instead of ".pkl files" to make it easier to inspect them?

Maybe yes... I would introduce a `write_python_dict_to_file` helper or something like this then. I used pickle mostly because it is consistent to the BPE Job (which could of course also get a flag for a different output then).

michelwi (approved):

PR #272, opened 2022-06-02
ReturnnForwardJob

JackTemaki:
A job that does task=forward or task=eval, writing output data into HDF files.

### DIFF
--- a/returnn/forward.py
+++ b/returnn/forward.py
@@ -1,0 +1,213 @@
+__all__ = ["ReturnnForwardJob"]
+
+from sisyphus import *
+
+import copy
+import glob
+import os
+import shutil
+import subprocess as sp
+import tempfile
+
+from i6_core.returnn.config import ReturnnConfig
+from i6_core.returnn.training import Checkpoint
+import i6_core.util as util
+
+Path = setup_path(__package__)
+
+
+class ReturnnForwardJob(Job):
+    """
+    Run a RETURNN "forward" pass to HDF with a specified model checkpoint.
+    Also allows to run an "eval" task pass, which is similar to "forward" but treats all layers as in training mode,
+    which can be used to e.g. do cheating experiments.
+
+    Outputs:
+
+    Dict[tk.Path] out_hdf_files: Dictionary of all output HDF files that were requested by the key list `hdf_outputs`
+    tk.Path out_default_hdf: For forward (not eval) mode, this contains the default HDF file that is always written
+        by RETURNN, independent of other settings.
+    """
+
+    def __init__(
+        self,
+        model_checkpoint,
+        returnn_config,
+        hdf_outputs=None,
+        eval_mode=False,
+        *,  # args below are keyword only
+        log_verbosity=3,
+        device="gpu",
+        time_rqmt=4,
+        mem_rqmt=4,
+        cpu_rqmt=4,
+        returnn_python_exe=None,
+        returnn_root=None,
+    ):
+        """
+
+        :param Checkpoint model_checkpoint: Checkpoint object pointing to a stored RETURNN Tensorflow model
+        :param ReturnnConfig returnn_config: RETURNN config dict
+        :param dict returnn_post_config: RETURNN config dict (no hashing)
+        :param list[str] hdf_outputs: list of additional hdf output layer file names that the network generates (e.g. attention.hdf);
+          The hdf outputs have to be a valid subset or be equal to the hdf_dump_layers in the config.
+        :param bool eval_mode: run forward in eval mode, the default hdf is not available in this case and no search will be done.
+        :param int log_verbosity: RETURNN log verbosity
+        :param str device: RETURNN device, cpu or gpu
+        :param int time_rqmt: job time requirement
+        :param int mem_rqmt: job memory requirement
+        :param int cpu_rqmt: job cpu requirement
+        :param Path|str returnn_python_exe: path to the RETURNN executable (python binary or launch script)
+        :param Path|str returnn_root: path to the RETURNN src folder
+        """
+        self.returnn_python_exe = (
+            returnn_python_exe
+            if returnn_python_exe is not None
+            else gs.RETURNN_PYTHON_EXE
+        )
+        self.returnn_root = (
+            returnn_root if returnn_root is not None else gs.RETURNN_ROOT
+        )
+
+        self._model_checkpoint = model_checkpoint
+        self._returnn_config = returnn_config
+        self._eval_mode = eval_mode
+        self._log_verbosity = log_verbosity
+        self._device = device
+
+        self.out_returnn_config_file = self.output_path("returnn.config")
+
+        self.out_hdf_files = {}
+        hdf_outputs = hdf_outputs if hdf_outputs else []
+        for output in hdf_outputs:
+            self.out_hdf_files[output] = self.output_path(output)
+        if not eval_mode:
+            self.out_hdf_files["output.hdf"] = self.output_path("output.hdf")
+            self.out_default_hdf = self.out_hdf_files["output.hdf"]
+
+        self.rqmt = {
+            "gpu": 1 if device == "gpu" else 0,
+            "cpu": cpu_rqmt,
+            "mem": mem_rqmt,
+            "time": time_rqmt,
+        }
+
+    def tasks(self):
+        yield Task("create_files", mini_task=True)
+        yield Task("run", resume="run", rqmt=self.rqmt)
+
+    def create_files(self):
+        config = self.create_returnn_config(
+            model_checkpoint=self._model_checkpoint,
+            returnn_config=self._returnn_config,
+            eval_mode=self._eval_mode,
+            log_verbosity=self._log_verbosity,
+            device=self._device,
+        )
+        config.write(self.out_returnn_config_file.get_path())
+
+        cmd = [
+            tk.uncached_path(self.returnn_python_exe),
+            os.path.join(tk.uncached_path(self.returnn_root), "rnn.py"),
+            self.out_returnn_config_file.get_path(),
+        ]
+        util.create_executable("rnn.sh", cmd)
+
+        # check here if model actually exists
+        assert os.path.exists(
+            self._model_checkpoint.index_path.get_path()
+        ), "Provided model does not exists: %s" % str(self._model_checkpoint)
+
+    def run(self):
+        # run everything in a TempDir as writing HDFs can cause heavy load
+        with tempfile.TemporaryDirectory(prefix="work_") as d:
+            print("using temp-dir: %s" % d)
+            call = [
+                tk.uncached_path(self.returnn_python_exe),
+                os.path.join(tk.uncached_path(self.returnn_root), "rnn.py"),
+                self.out_returnn_config_file.get_path(),
+            ]
+
+            # stash a possible exception until we finished copying files from the temp work to the actual
+            # work folder to be able to examine the files
+            error = None
+            try:
+                sp.check_call(call, cwd=d)
+            except Exception as e:
+                print("Run crashed - copy temporary work folder")
+                error = e
+
+            # move log and tensorboard
+            shutil.move(os.path.join(d, "returnn.log"), "returnn.log")
+            tensorboard_dirs = glob.glob(os.path.join(d, "eval-*"))
+            for dir in tensorboard_dirs:
+                shutil.move(dir, os.path.basename(dir))
+
+            # move hdf outputs to output folder
+            for k, v in self.out_hdf_files.items():
+                try:
+                    shutil.move(os.path.join(d, k), v.get_path())
+                except Exception as e:
+                    if error is None:
+                        # if we had an error before it is expected to have missing hdf files
+                        # so we do not need to raise anything here then
+                        raise e
+
+            if error:
+                raise error
+
+    @classmethod
+    def create_returnn_config(
+        cls,
+        model_checkpoint,
+        returnn_config,
+        eval_mode,
+        log_verbosity,
+        device,
+        **kwargs,
+    ):
+        """
+
+        :param Checkpoint model_checkpoint:
+        :param ReturnnConfig returnn_config:
+        :param int log_verbosity:
+        :param str device:
+        :param kwargs:
+        :return:
+        """
+        assert device in ["gpu", "cpu"]
+        assert "network" in returnn_config.config
+
+        res = copy.deepcopy(returnn_config)
+
+        config = {"load": model_checkpoint}
+
+        post_config = {
+            "device": device,
+            "log": ["./returnn.log"],
+            "log_verbosity": log_verbosity,
+            "task": "eval" if eval_mode else "forward",
+        }
+
+        if not eval_mode:
+            post_config["forward_override_hdf_output"] = True
+            post_config["output_file"] = "output.hdf"
+
+        config.update(returnn_config.config)
+        post_config.update(returnn_config.post_config)
+
+        res.config = config
+        res.post_config = post_config
+        res.check_consistency()
+
+        return res
+
+    @classmethod
+    def hash(cls, kwargs):
+        d = {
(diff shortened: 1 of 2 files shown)


returnn/forward.py
```diff
@@ -0,0 +1,213 @@
...
+    ):
+        """
+
+        :param Checkpoint model_checkpoint: Checkpoint object pointing to a stored RETURNN Tensorflow model
+        :param ReturnnConfig returnn_config: RETURNN config dict
+        :param dict returnn_post_config: RETURNN config dict (no hashing)
+        :param list[str] hdf_outputs: list of additional hdf output layer file names that the network generates (e.g. attention.hdf);
+          The hdf outputs have to be a valid subset or be equal to the hdf_dump_layers in the config.
+        :param bool eval_mode: run forward in eval mode, the default hdf is not available in this case and no search will be done.
+        :param int log_verbosity: RETURNN log verbosity
```
michelwi, line 55: Is this not set in the ReturnnConfig object?

returnn/forward.py
```diff
@@ -0,0 +1,213 @@
...
+        model_checkpoint,
+        returnn_config,
+        hdf_outputs=None,
+        eval_mode=False,
+        *,  # args below are keyword only
+        log_verbosity=3,
+        device="gpu",
+        time_rqmt=4,
+        mem_rqmt=4,
+        cpu_rqmt=4,
```
michelwi, line 43: is this a reasonable default for a job that does not do CPU heavy load? 
```suggestion
        cpu_rqmt=1,
```

michelwi, line 43: is this a reasonable default for a job that does not do CPU heavy load? 
```suggestion
        cpu_rqmt=1,
```

returnn/forward.py
```diff
@@ -0,0 +1,213 @@
...
+        :param list[str] hdf_outputs: list of additional hdf output layer file names that the network generates (e.g. attention.hdf);
+          The hdf outputs have to be a valid subset or be equal to the hdf_dump_layers in the config.
+        :param bool eval_mode: run forward in eval mode, the default hdf is not available in this case and no search will be done.
+        :param int log_verbosity: RETURNN log verbosity
+        :param str device: RETURNN device, cpu or gpu
+        :param int time_rqmt: job time requirement
+        :param int mem_rqmt: job memory requirement
+        :param int cpu_rqmt: job cpu requirement
+        :param Path|str returnn_python_exe: path to the RETURNN executable (python binary or launch script)
+        :param Path|str returnn_root: path to the RETURNN src folder
```
michelwi, line 61:
```suggestion
        :param Path returnn_python_exe: path to the RETURNN executable (python binary or launch script)
        :param Path returnn_root: path to the RETURNN src folder
```
do not introduce already deprecated behavior.

returnn/forward.py
```diff
@@ -0,0 +1,213 @@
...
+            returnn_config=self._returnn_config,
+            eval_mode=self._eval_mode,
+            log_verbosity=self._log_verbosity,
+            device=self._device,
+        )
+        config.write(self.out_returnn_config_file.get_path())
+
+        cmd = [
+            tk.uncached_path(self.returnn_python_exe),
+            os.path.join(tk.uncached_path(self.returnn_root), "rnn.py"),
```
michelwi, line 111: avoid `tk.uncached_path`

michelwi, line 111: avoid `tk.uncached_path`

returnn/forward.py
```diff
@@ -0,0 +1,213 @@
...
+        util.create_executable("rnn.sh", cmd)
+
+        # check here if model actually exists
+        assert os.path.exists(
+            self._model_checkpoint.index_path.get_path()
+        ), "Provided model does not exists: %s" % str(self._model_checkpoint)
+
+    def run(self):
+        # run everything in a TempDir as writing HDFs can cause heavy load
+        with tempfile.TemporaryDirectory(prefix="work_") as d:
```
michelwi, line 123: does this guarantee that `d` is on the local drive? I think this will create a randomly named folder in the current working directory, i.e. `ReturnnForwardJob.somehash/work`. compare e.g. https://github.com/rwth-i6/i6_core/blob/48ec2dfd3d2dde9305892428f16c91f585d4aea8/features/extraction.py#L147 with gs.TMP_PREFIX defined as https://github.com/rwth-i6/sisyphus/blob/423b90086bced6eba98dfad184c2d13c7cdcb89a/sisyphus/global_settings.py#L272)

please verify your implementation, otherwise I suggest
```suggestion
        with tempfile.TemporaryDirectory(prefix=prefix=gs.TMP_PREFIX) as d:
```

michelwi, line 123: does this guarantee that `d` is on the local drive? I think this will create a randomly named folder in the current working directory, i.e. `ReturnnForwardJob.somehash/work`. compare e.g. https://github.com/rwth-i6/i6_core/blob/48ec2dfd3d2dde9305892428f16c91f585d4aea8/features/extraction.py#L147 with gs.TMP_PREFIX defined as https://github.com/rwth-i6/sisyphus/blob/423b90086bced6eba98dfad184c2d13c7cdcb89a/sisyphus/global_settings.py#L272)

please verify your implementation, otherwise I suggest
```suggestion
        with tempfile.TemporaryDirectory(prefix=gs.TMP_PREFIX) as d:
```

returnn/forward.py
```diff
@@ -0,0 +1,213 @@
...
+            print("using temp-dir: %s" % d)
+            call = [
+                tk.uncached_path(self.returnn_python_exe),
+                os.path.join(tk.uncached_path(self.returnn_root), "rnn.py"),
```
michelwi, line 127: avoid tk.uncached_path

michelwi, line 127: avoid tk.uncached_path

returnn/forward.py
```diff
@@ -0,0 +1,213 @@
...
+            shutil.move(os.path.join(d, "returnn.log"), "returnn.log")
+            tensorboard_dirs = glob.glob(os.path.join(d, "eval-*"))
+            for dir in tensorboard_dirs:
+                shutil.move(dir, os.path.basename(dir))
+
+            # move hdf outputs to output folder
+            for k, v in self.out_hdf_files.items():
+                try:
+                    shutil.move(os.path.join(d, k), v.get_path())
+                except Exception as e:
```
michelwi, line 150: `Exception` might be too broad, if you are maybe expecting a `FileNotFoundException`, then only drop those.

Are we even interested in these (partial) outputs if the job did crash? maybe just let it die..

michelwi, line 150: `Exception` might be too broad, if you are maybe expecting a `FileNotFoundException`, then only drop those.

returnn/forward.py
```diff
@@ -0,0 +1,213 @@
...
+
+        :param Checkpoint model_checkpoint:
+        :param ReturnnConfig returnn_config:
+        :param int log_verbosity:
+        :param str device:
+        :param kwargs:
+        :return:
+        """
+        assert device in ["gpu", "cpu"]
+        assert "network" in returnn_config.config
```
michelwi, line 179: is this still true, e.g. with staged_network_dict or get_network functions?

michelwi, line 179: is this still true, e.g. with staged_network_dict or get_network functions?

returnn/forward.py
```diff
@@ -0,0 +1,213 @@
...
+        if not eval_mode:
+            post_config["forward_override_hdf_output"] = True
+            post_config["output_file"] = "output.hdf"
+
+        config.update(returnn_config.config)
+        post_config.update(returnn_config.post_config)
+
+        res.config = config
+        res.post_config = post_config
+        res.check_consistency()
```
michelwi, line 201: maybe it would be more elegant to make use of the new `ReturnnConfig.update` method?

michelwi, line 201: maybe it would be more elegant to make use of the new `ReturnnConfig.update` method?

Also be aware about which config overwrites the other. Sometimes we might want the users choice to take precedence (maybe log_verbosity?), sometimes the jobs choice (e.g. Task, output file location, maybe log verbosity) 
Ok, actually nevermind, we want to keep nothing of the users parameters. You should invert the overwriting logic here.

returnn/forward.py
```diff
@@ -0,0 +1,213 @@
...
+
+    def __init__(
+        self,
+        model_checkpoint,
+        returnn_config,
+        hdf_outputs=None,
+        eval_mode=False,
+        *,  # args below are keyword only
+        log_verbosity=3,
+        device="gpu",
```
michelwi, line 40: is multigpu a thing we would consider here?

michelwi, line 40: is multigpu a thing we would consider here?

returnn/forward.py
```diff
@@ -0,0 +1,213 @@
...
+
+        res = copy.deepcopy(returnn_config)
+
+        config = {"load": model_checkpoint}
+
+        post_config = {
+            "device": device,
+            "log": ["./returnn.log"],
+            "log_verbosity": log_verbosity,
+            "task": "eval" if eval_mode else "forward",
```
michelwi, line 189: Would this imply that changing `eval_mode` does not influence the hash?

later the post_config is overwritten by the post_config of the input returnn_config. Could a user then overwrite this setting by accident and e.g. start a training?

In the training job, the task is set in the config.

michelwi, line 189: Would this imply that changing `eval_mode` does not influence the hash?

later the post_config is overwritten by the post_config of the input returnn_config. Could a user then overwrite this setting by accident and e.g. start a training?

returnn/forward.py
```diff
@@ -0,0 +1,213 @@
...
+                self.out_returnn_config_file.get_path(),
+            ]
+
+            # stash a possible exception until we finished copying files from the temp work to the actual
+            # work folder to be able to examine the files
+            error = None
+            try:
+                sp.check_call(call, cwd=d)
+            except Exception as e:
+                print("Run crashed - copy temporary work folder")
```
michelwi, line 137: you print "copy temporary work folder", but you never actually do...

michelwi, line 137: you print "copy temporary work folder", but you never actually do...

maybe you could remove the whole special handling of if-or-not an error occurred and just do a full copy of d to the work dir in the except block and raise any exception normally.
Then below assume everything went correct and copy only the files you need.

returnn/forward.py
```diff
@@ -0,0 +1,213 @@
...
+        )
+        self.returnn_root = (
+            returnn_root if returnn_root is not None else gs.RETURNN_ROOT
+        )
+
+        self._model_checkpoint = model_checkpoint
+        self._returnn_config = returnn_config
+        self._eval_mode = eval_mode
+        self._log_verbosity = log_verbosity
+        self._device = device
```
michelwi, line 76: why do all these guys start with an underscore?

michelwi, line 76: why do all these guys start with an underscore?

returnn/forward.py
christophmluscher: `log_verbosity` is not allowed in the `ReturnnConfig` if I remember correctly

returnn/forward.py
christophmluscher: this is forwarding through the network with TF. so I would say at least `2`. Although I am not sure what the limiting factor for `forward`/`eval` jobs are..

christophmluscher (changes requested):
I agree with the comments provided by @michelwi

returnn/forward.py
JackTemaki: If we do this, then should we drop `gs.` support completely?

returnn/forward.py
JackTemaki: I thought we could start making rather private members private, but I guess this is somewhat inconsistent then.

returnn/forward.py
JackTemaki: I hope we never get to the point where we need multi-gpu for forwarding... For large scale TTS you can use Sisyphus to split jobs.

returnn/forward.py
michelwi: "should" probably yes. "can" probably no.. see the whole `gs.RETURNN_ROOT` discussion.

There are some sisyphus settings that I think are fine to read, e.g. temp prefix for local caching - as long as there is no implementation for this in sisyphus directly.

returnn/forward.py
michelwi: This is true. Also in training the user can overwrite important parameters by specifying them in the `ReturnnConfig.config` dict and then get unspecified behavior.

But here I fear the danger that someone in the training phase wants to initialize using an existing model and sets `load = "/path/to/model"` in the training config. Then for forwarding the same training config is used and voila the checkpoint parameter is ignored. Maybe we can also assert that no `"load"` is specified in the given config.

returnn/forward.py
```diff
@@ -0,0 +1,204 @@
...
+        :param Checkpoint model_checkpoint:
+        :param ReturnnConfig returnn_config:
+        :param int log_verbosity:
+        :param str device:
+        :param kwargs:
+        :return:
+        """
+        assert device in ["gpu", "cpu"]
+        assert "task" not in returnn_config.config
+        assert "task" not in returnn_config.post_config
```
michelwi, line 168:
```suggestion
        assert "task" not in returnn_config.config
        assert "load" not in returnn_config.config
```
The post config will anyway be checked via `check_consistency`. I propose to also check for parameter "load"

returnn/forward.py
```diff
@@ -0,0 +1,213 @@
...
+        mem_rqmt=4,
+        cpu_rqmt=4,
+        returnn_python_exe=None,
+        returnn_root=None,
+    ):
+        """
+
+        :param Checkpoint model_checkpoint: Checkpoint object pointing to a stored RETURNN Tensorflow model
+        :param ReturnnConfig returnn_config: RETURNN config dict
+        :param dict returnn_post_config: RETURNN config dict (no hashing)
```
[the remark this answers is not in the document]
JackTemaki: I do not see this anymore....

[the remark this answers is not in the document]
michelwi: yes, I think my Github website bugged up. I also saw many duplicated comments from my last review and I think one comment to @christophmluscher previous question was shown as still pending although I am pretty sure I sent it earlier.
Please disregard any such bugs/duplications.

JackTemaki:
@Atticus1806 successfully used this job, so I would say it is ready for review.

JackTemaki:
@Atticus1806 successfully used this job, so I would say it is ready for review.

returnn/forward.py
```diff
@@ -0,0 +1,201 @@
...
+            except Exception as e:
+                print("Run crashed - copy temporary work folder as 'crash_dir'")
+                shutil.copytree(d, "crash_dir")
+                raise e
+
+            # move log and tensorboard
+            shutil.move(os.path.join(d, "returnn.log"), "returnn.log")
+            tensorboard_dirs = glob.glob(os.path.join(d, "eval-*"))
+            for dir in tensorboard_dirs:
+                shutil.move(dir, os.path.basename(dir))
```
michelwi, line 135: Just to clarify:
The intention here is to have the log and tensorboard dirs in the work folder of the job so we cal later look into it if needed?

michelwi (approved):

returnn/forward.py
JackTemaki: yes, although I already thought that we might make outputs out of them in the future...

returnn/forward.py
```diff
@@ -0,0 +1,201 @@
...
+
+        post_config = {
+            "device": device,
+            "log": ["./returnn.log"],
+            "log_verbosity": log_verbosity,
+        }
+
+        if not eval_mode:
+            post_config["forward_override_hdf_output"] = True
+            post_config["output_file"] = "output.hdf"
```
christophmluscher, line 182: didnt this cause problems with the file server?? because we are writing directly to network storage.

Atticus1806 (approved):

Atticus1806:
Maybe a slight suggestion (but not required for approval):
Import it in the `__init__.py`

PR #273, opened 2022-06-13
G2P extensions

SimBe195:
- Parallelization of ApplyG2PModelJob by splitting up the input word-list into equal parts and later merging the results
- Option to include all orthography variants of each lemma in BlissLexiconToG2PLexiconJob

### DIFF
--- a/g2p/apply.py
+++ b/g2p/apply.py
@@ -67,47 +67,39 @@ class ApplyG2PModelJob(Job):
         self.out_g2p_lexicon = self.output_path("g2p.lexicon")
         self.out_g2p_untranslated = self.output_path("g2p.untranslated")
 
-        if self.concurrent > 1:
-            self.out_g2p_lexicon_parts = [
-                f"g2p.lexicon.{i}"
-                for i in range(1, self.concurrent + 1)
-            ]
-            self.out_g2p_untranslated_parts = [
-                f"g2p.untranslated.{i}"
-                for i in range(1, self.concurrent + 1)
-            ]
-            num_digits = len(str(self.concurrent))
-            self.word_list_parts = [
-                f"words.{i:0{num_digits}d}" for i in range(1, self.concurrent + 1)
-            ]
+        self.out_g2p_lexicon_parts = [
+            f"g2p.lexicon.{i}"
+            for i in range(1, self.concurrent + 1)
+        ]
+        self.out_g2p_untranslated_parts = [
+            f"g2p.untranslated.{i}"
+            for i in range(1, self.concurrent + 1)
+        ]
+        num_digits = len(str(self.concurrent))
+        self.word_list_parts = [
+            f"words.{i:0{num_digits}d}" for i in range(1, self.concurrent + 1)
+        ]
 
         self.rqmt = {"cpu": 1, "mem": 1, "time": 2}
 
     def tasks(self):
-        if self.concurrent == 1:
-            yield Task("run", rqmt=self.rqmt)
-        else:
-            yield Task("create_files", mini_task=True)
-            yield Task(
-                "run", rqmt=self.rqmt, args=range(1, self.concurrent + 1)
-            )
-            yield Task("merge", mini_task=True)
+        yield Task("split_word_list", mini_task=True)
+        yield Task(
+            "run", rqmt=self.rqmt, args=range(1, self.concurrent + 1)
+        )
+        yield Task("merge", mini_task=True)
         if self.filter_empty_words:
             yield Task("filter", mini_task=True)
 
-    def create_files(self):
+    def split_word_list(self):
         self.sh(
             f"split --number=l/{self.concurrent} --numeric-suffixes=1 {self.word_list.get_path()} words."
         )
 
-    def run(self, task_id=None):
-        g2p_lexicon_path = self.out_g2p_lexicon.get_path()
-        g2p_untranslated_path = self.out_g2p_untranslated.get_path()
-        word_list_path = self.word_list.get_path()
-        if task_id:
-            g2p_lexicon_path = self.out_g2p_lexicon_parts[task_id - 1]
-            g2p_untranslated_path = self.out_g2p_untranslated_parts[task_id - 1]
-            word_list_path = self.word_list_parts[task_id - 1]
+    def run(self, task_id):
+        g2p_lexicon_path = self.out_g2p_lexicon_parts[task_id - 1]
+        g2p_untranslated_path = self.out_g2p_untranslated_parts[task_id - 1]
+        word_list_path = self.word_list_parts[task_id - 1]
 
         with uopen(g2p_lexicon_path, "wt") as out:
             with uopen(g2p_untranslated_path, "wt") as err:
--- a/g2p/apply.py
+++ b/g2p/apply.py
@@ -147,6 +147,5 @@ class ApplyG2PModelJob(Job):
     @classmethod
     def hash(cls, kwargs):
         kwargs_copy = dict(**kwargs)
-        if "concurrent" in kwargs_copy:
-            del kwargs_copy["concurrent"]
+        kwargs_copy.pop("concurrent", None)
         return super().hash(kwargs_copy)
--- a/g2p/apply.py
+++ b/g2p/apply.py
@@ -60,13 +60,6 @@ class ApplyG2PModelJob(Job):
         self.filter_empty_words = filter_empty_words
         self.concurrent = concurrent
 
-        self._g2p_lexicon_parts = [
-            f"g2p.lexicon.{i}" for i in range(1, self.concurrent + 1)
-        ]
-        self._g2p_untranslated_parts = [
-            f"g2p.untranslated.{i}" for i in range(1, self.concurrent + 1)
-        ]
-
         self.out_g2p_lexicon = self.output_path("g2p.lexicon")
         self.out_g2p_untranslated = self.output_path("g2p.untranslated")
 
@@ -91,8 +84,8 @@ class ApplyG2PModelJob(Job):
         )
 
     def run(self, task_id):
-        g2p_lexicon_path = self._g2p_lexicon_parts[task_id - 1]
-        g2p_untranslated_path = self._g2p_untranslated_parts[task_id - 1]
+        g2p_lexicon_path = f"g2p.lexicon.{task_id}"
+        g2p_untranslated_path = f"g2p.untranslated.{task_id}"
         num_digits = len(str(self.concurrent))
         word_list_path = f"words.{task_id:0{num_digits}d}"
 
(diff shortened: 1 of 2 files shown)


g2p/apply.py
```diff
@@ -16,7 +16,10 @@ class ApplyG2PModelJob(Job):
     Apply a trained G2P on a word list file
     """
 
-    __sis_hash_exclude__ = {"filter_empty_words": False}
+    __sis_hash_exclude__ = {
+        "filter_empty_words": False,
+        "concurrent": 1,
```
michelwi, line 21: Ideally the concurrency should not change the result and therefore be excluded from hash computation.

g2p/apply.py
```diff
@@ -56,20 +61,61 @@ def __init__(
         self.variants_number = variants_number
         self.word_list = word_list_file
         self.filter_empty_words = filter_empty_words
+        self.concurrent = concurrent
+
+        self.out_g2p_lexicon_parts = None
+        self.out_g2p_untranslated_parts = None
+        self.word_list_parts = None
```
michelwi, line 68: these are intermediate results, why do we need outputs for these?

g2p/apply.py
JackTemaki, line 21: I think one does not want to hash this value at all, so I think it is better to add a custom hash function and drop "concurrent" from the list of parameters. There should be other jobs where you can look at how this is typically done.

g2p/apply.py
```diff
@@ -56,20 +61,61 @@ def __init__(
         self.variants_number = variants_number
         self.word_list = word_list_file
         self.filter_empty_words = filter_empty_words
+        self.concurrent = concurrent
+
+        self.out_g2p_lexicon_parts = None
```
JackTemaki, line 66: The parts can be intermediate files just stored in `work` , there is no need to have this as output. Or do you want to use those explictely in your pipeline later on?

g2p/apply.py
```diff
@@ -56,20 +61,61 @@ def __init__(
...
     def tasks(self):
-        yield Task("run", rqmt=self.rqmt)
+        if self.concurrent == 1:
+            yield Task("run", rqmt=self.rqmt)
+        else:
+            yield Task("create_files", mini_task=True)
+            yield Task(
+                "run", rqmt=self.rqmt, args=range(1, self.concurrent + 1)
+            )
+            yield Task("merge", mini_task=True)
```
michelwi, line 97: do we need special handling for the case `concurrent==1`? Sure the split and merge tasks would be no-ops then, but in principle you could treat `concurrent==1` the same as "`concurrent==N` with N=1" (also in terms of filenames/logic within the run task)

This would make it more general / cleaner to add more functionality later.

g2p/apply.py
```diff
@@ -83,12 +129,22 @@ def run(self):
...
 
+    def merge(self):
+        g2p_lex_file_list = " ".join(map(tk.uncached_path, self.out_g2p_lexicon_parts))
+        self.sh(f"cat {g2p_lex_file_list} > {tk.uncached_path(self.out_g2p_lexicon)}")
+        g2p_untranslated_file_list = " ".join(
+            map(tk.uncached_path, self.out_g2p_untranslated_parts)
+        )
+        self.sh(
+            f"cat {g2p_untranslated_file_list} > {tk.uncached_path(self.out_g2p_untranslated)}"
+        )
```
michelwi, line 146: we want to get rid of `tk.uncached_path`. In your case it is always save to use `get_path` instead.

g2p/apply.py
```diff
@@ -83,12 +129,22 @@ def run(self):
...
                         "-a",
-                        self.word_list.get_path(),
+                        word_list_path,
                     ],
                     stdout=out,
                     stderr=err,
                 )
 
+    def merge(self):
+        g2p_lex_file_list = " ".join(map(tk.uncached_path, self.out_g2p_lexicon_parts))
```
JackTemaki, line 139: Please do not use `tk.uncached_path`, we want to get rid of this. Just use `variable.get_path()`.

But related to a statement to above, maybe you do not even need Paths here, but just regular strings to refer to intermediate files in the work folder.

michelwi (changes requested):
I request changes w.r.t. removal of `tk.uncached_path` but please also take other comments into consideration.

Your changes to convert seem fine to me. Next time maybe do separate PRs for unrelated changes. ;)

g2p/apply.py
```diff
@@ -83,12 +129,22 @@ def run(self):
...
+        self.sh(f"cat {g2p_lex_file_list} > {tk.uncached_path(self.out_g2p_lexicon)}")
```
JackTemaki, line 140: Please do not use `self.sh`, as you can have arbitrary code execution. You can use `f=open(...)` and then `subprocess.call(["cat"] + g2p_lex_file_list, stdout=f)`

g2p/convert.py
```diff
@@ -17,14 +17,24 @@ class BlissLexiconToG2PLexiconJob(Job):
...
-    def __init__(self, bliss_lexicon, include_pronunciation_variants=False):
+    __sis_hash_exclude = {
+        "include_orthography_variants": False,
+        "include_pronunciation_variants": False,
+    }
+
+    def __init__(
+        self,
+        bliss_lexicon,
+        include_orthography_variants=False,
```
JackTemaki, line 28: please add new parameters at the end, some people might use non-named arguments.

JackTemaki (changes requested):

g2p/apply.py
michelwi: >  You can use `f=open(...)` and then

better even:
```
with open(...) as f:
  subprocess.call(["cat"] + g2p_lex_file_list, stdout=f)
```

g2p/apply.py
SimBe195: Solved now by implementing a custom hash function.

g2p/apply.py
```diff
@@ -70,49 +67,41 @@ def __init__(
...
-            ]
-            self.out_g2p_untranslated_parts = [
-                self.output_path(f"g2p.untranslated.{i}")
-                for i in range(1, self.concurrent + 1)
-            ]
-            num_digits = len(str(self.concurrent))
-            self.word_list_parts = [
-                f"words.{i:0{num_digits}d}" for i in range(1, self.concurrent + 1)
-            ]
+        self.out_g2p_lexicon_parts = [
```
JackTemaki, line 70: Please use the `out_` prefix only for actual outputs, here those are just intermediate variables. I would also recommend to just declare them private, so `self._g2p_lexicon_parts`.

g2p/apply.py
```diff
@@ -56,20 +61,61 @@ def __init__(
...
+                "run", rqmt=self.rqmt, args=range(1, self.concurrent + 1)
+            )
+            yield Task("merge", mini_task=True)
         if self.filter_empty_words:
             yield Task("filter", mini_task=True)
 
-    def run(self):
-        with uopen(self.out_g2p_lexicon, "wt") as out:
-            with uopen(self.out_g2p_untranslated, "wt") as err:
+    def create_files(self):
```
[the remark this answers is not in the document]
SimBe195: Done.

g2p/apply.py
SimBe195: I generalized the code to exclude the special case handling of concurrent=1.

g2p/apply.py
SimBe195: Removed all instances of tk.uncached_path now.

g2p/apply.py
SimBe195: Added a custom hash function that does this now.

g2p/apply.py
SimBe195: No, they don't need to be outputs. Moved them to work now.

JackTemaki (changes requested):
Looks really good, just two small comments

g2p/apply.py
SimBe195: Yes, I changed it to do it in this way now.

g2p/apply.py
SimBe195: Changed all usages of `self.sh` to `subprocess.check_call` now.

g2p/convert.py
SimBe195: Done.

g2p/apply.py
```diff
@@ -56,20 +58,53 @@ def __init__(
         self.variants_number = variants_number
         self.word_list = word_list_file
         self.filter_empty_words = filter_empty_words
+        self.concurrent = concurrent
+
+        self.out_g2p_lexicon_parts = None
+        self.out_g2p_untranslated_parts = None
+        self.word_list_parts = None
```
michelwi: why do you define those as `None` first and then overwrite with the list later?
I think this can be deleted
```suggestion
```

g2p/apply.py
michelwi: as the names do not change, you might even not have these as members of the job at all, but create them as local variable in the respective task. Or refactor the loops to loop over integers and then open the file `f"g2p.lexicon.{i}"` instead of looping over the filenames if that is easier.

Also, now that these guys are not outputs any longer, they should - if you decide to keep them - be move above the definition of the outputs.

SimBe195: Done.

g2p/apply.py
```diff
@@ -101,3 +143,10 @@ def filter(self):
 
         os.remove(self.out_g2p_lexicon)
         os.rename(tmp_path, self.out_g2p_lexicon)
+
+    @classmethod
+    def hash(cls, kwargs):
+        kwargs_copy = dict(**kwargs)
+        if "concurrent" in kwargs_copy:
```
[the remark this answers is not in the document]
SimBe195: Ah yes, forgot about that, done now.

g2p/apply.py
SimBe195: Yeah, I missed these when generalizing the concurrent=1 case, removed the redundant definitions now.

g2p/apply.py
```diff
@@ -56,20 +58,46 @@ def __init__(
...
         self.word_list = word_list_file
         self.filter_empty_words = filter_empty_words
+        self.concurrent = concurrent
+
+        self._g2p_lexicon_parts = [
+            f"g2p.lexicon.{i}" for i in range(1, self.concurrent + 1)
+        ]
+        self._g2p_untranslated_parts = [
+            f"g2p.untranslated.{i}" for i in range(1, self.concurrent + 1)
+        ]
```
michelwi, line 68:
```suggestion
```

g2p/apply.py
```diff
@@ -56,20 +58,46 @@ def __init__(
...
+                f"--number=l/{self.concurrent}",
+                "--numeric-suffixes=1",
+                self.word_list.get_path(),
+                "words.",
+            ]
+        )
+
+    def run(self, task_id):
+        g2p_lexicon_path = self._g2p_lexicon_parts[task_id - 1]
+        g2p_untranslated_path = self._g2p_untranslated_parts[task_id - 1]
```
michelwi, line 95:
```suggestion
        g2p_lexicon_path = f"g2p.lexicon.{task_id}"
        g2p_untranslated_path = f"g2p.untranslated.{task_id}"
```

g2p/apply.py
```diff
@@ -83,12 +111,19 @@ def run(self):
...
                     stdout=out,
                     stderr=err,
                 )
 
+    def merge(self):
+        with uopen(self.out_g2p_lexicon, "wt") as f:
+            sp.check_call(["cat"] + self._g2p_lexicon_parts, stdout=f)
+
+        with uopen(self.out_g2p_untranslated, "wt") as f:
+            sp.check_call(["cat"] + self._g2p_untranslated_parts, stdout=f)
```
michelwi, line 125:
```suggestion
        with uopen(self.out_g2p_lexicon, "wt") as f:
            sp.check_call(["cat"] + [f"g2p.lexicon.{i}" for i in range(1, self.concurrent + 1)], stdout=f)

        with uopen(self.out_g2p_untranslated, "wt") as f:
            sp.check_call(["cat"] + [f"g2p.untranslated.{i}" for i in range(1, self.concurrent + 1)], stdout=f)
```

michelwi (approved):
Thanks, I think this PR is good to go and I would accept it as it is now.

I have added as "suggested change" an option to get rid of the `_g2p_lexicon_parts` and `_g2p_untranslated_parts` members that I find more elegant, but I leave it up to you if you want to add this.

PR #275, opened 2022-06-15
update set_executables to use tk.Path

JackTemaki:
The current `set_executables` function does not do what it should do (e.g. accepting paths from a MakeJob compiling RASR), so here are my suggested fixes for that.

### DIFF
--- a/rasr/crp.py
+++ b/rasr/crp.py
@@ -73,22 +73,26 @@ class CommonRasrParameters:
         """
         assert isinstance(rasr_binary_path, tk.Path)
         self.acoustic_model_trainer_exe = (
-            rasr_binary_path + f"acoustic-model-trainer.{rasr_arch}"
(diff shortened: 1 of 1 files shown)


christophmluscher (approved):

michelwi (commented):
This breaks backwards compatibility with str input. I do agree with that.
Another option would be to also allow str and emit a deprecation warning and cast to tk.Path in this method. What do you think?

Ah.. I didn't know that Path + str becomes DelayedAdd, good to know. In this case however, maybe you should be using the function Path.join_right(str) instead, because
1. It uses os.path.join internally and you join pathes here
2. The type in your case is changed to DelayedAdd and is no longer Path. This will give at least problems with type checking in the far future but maybe also functionality wise now (e.g. get vs get_path works differently for Path/Variable/DelayedBase)

PR #276, opened 2022-06-23
docstring changes and write exec for compile

JackTemaki:
Just some minor stuff I kept in my stashes for a long time.

### DIFF
--- a/returnn/training.py
+++ b/returnn/training.py
@@ -11,7 +11,6 @@ import shutil
 import subprocess as sp
 
 from sisyphus import *
-from sisyphus.delayed_ops import DelayedSlice
 
 import i6_core.util as util
 from .config import ReturnnConfig
--- a/returnn/training.py
+++ b/returnn/training.py
@@ -11,6 +11,7 @@ import shutil
 import subprocess as sp
 
 from sisyphus import *
+from sisyphus.delayed_ops import DelayedSlice
 
 import i6_core.util as util
 from .config import ReturnnConfig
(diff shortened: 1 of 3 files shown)


returnn/training.py
```diff
@@ -21,6 +22,8 @@
 class ReturnnModel:
     """
     Defines a RETURNN model as config, checkpoint meta file and epoch
+
+    This is deprecated, used Checkpoint instead.
```
michelwi, line 26:
```suggestion
    This is deprecated, use Checkpoint instead.
```
typo

returnn/training.py
```diff
@@ -11,6 +11,7 @@
 import subprocess as sp
 
 from sisyphus import *
+from sisyphus.delayed_ops import DelayedSlice
```
michelwi, line 14: where is this needed? You only change doc strings so far.
```suggestion
```

JackTemaki: ah yes, did not see that one. Not sure what I used that for...

michelwi (approved):

returnn/training.py
```diff
@@ -21,6 +21,8 @@
 class ReturnnModel:
     """
     Defines a RETURNN model as config, checkpoint meta file and epoch
+
+    This is deprecated, use Checkpoint instead.
```
Atticus1806, line 25:
```suggestion
    This is deprecated, use :class:`Checkpoint` instead.
```

Atticus1806 (commented):
For some reason my PyCharm does not find the declarations, still I think we should reference the classes with the macro here, maybe in more parts of the file.

Atticus1806 (approved):

Issue #274, opened 2022-06-15
Hash-bug in ReturnnSearchJob


JackTemaki:
#63 refactored the hashing for the `ReturnnSearchJob`, but I just noticed during hash tests for a completely unrelated pipeline that here was a flawed hashing introduced that nobody noticed. `Checkpoint.ckpt_path` is called during hash computation, which calls a `.get_path()`. Unfortunately, `cpkt_path` was not implemented with `DelayedSlice` at that time, and kept to not break old code too much. I now see that this was not the best idea to keep the compatibility while we were still in the "unstable" period.

All we would need to do is changing:
```
        config = {
            "load": model_checkpoint.ckpt_path,
            "search_output_file_format": output_mode,
            "need_data": False,
            "search_do_eval": 0,
        }
```
To 
```
        config = {
            "load": model_checkpoint,
            "search_output_file_format": output_mode,
            "need_data": False,
            "search_do_eval": 0,
        }
```

So I guess the only chance is to introduce a `ReturnnSearchJobV2`, because in this case expecting a graph migration from all users is not suitable anymore. I guess we can use this point to also refactor other parts of the ReturnnSearchJob.

Opinions please, as this will be the case where we define what will happen in the future when such things occur.

christophmluscher:
I prefer the first approach as well.

I find it hard to estimate  how much the `ReturnnSearchJob` is used and therefor how much people would be affected by breaking the hash..

JackTemaki:
Resolved via #278

PR #278, opened 2022-06-27
add Returnn V2 search job and deprecate old

JackTemaki:
Fix for #274 

As discussed we want to introduce a "V2" job and keep the old.
The new job will also have some updates besides fixing the hash bug which are:
 - requiring explicit tk.Paths for returnn root/exe
 - removing some unnecessary deepcopy operations
 - removing instances of `tk.uncached_path`

### DIFF
--- a/returnn/search.py
+++ b/returnn/search.py
@@ -11,6 +11,7 @@ import logging
 import os
 import shutil
 import subprocess as sp
+from typing import Any, Dict
 
 from sisyphus import *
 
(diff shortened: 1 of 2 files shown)


christophmluscher (approved):

returnn/search.py
```diff
@@ -32,50 +37,43 @@ def __init__(
...
         :param Checkpoint model_checkpoint:  TF model checkpoint. see `ReturnnTrainingJob`.
         :param ReturnnConfig returnn_config: object representing RETURNN config
+        :param tk.Path returnn_python_exe: path to the RETURNN executable (python binary or launch script)
+        :param tk.Path returnn_root: path to the RETURNN src folder
         :param str output_mode: "txt" or "py"
         :param int log_verbosity: RETURNN log verbosity
         :param str device: RETURNN device, cpu or gpu
         :param float|int time_rqmt: job time requirement in hours
         :param float|int mem_rqmt: job memory requirement in GB
         :param float|int cpu_rqmt: job cpu requirement in GB
```
michelwi, line 61:
```suggestion
        :param float time_rqmt: job time requirement in hours
        :param float mem_rqmt: job memory requirement in GB
        :param int cpu_rqmt: job cpu requirement
```
in typing `float` also include all integer types, so `float|int` is redundant.

returnn/search.py
```diff
@@ -32,50 +37,43 @@ def __init__(
...
         log_verbosity=3,
         device="gpu",
         time_rqmt=4,
         mem_rqmt=4,
         cpu_rqmt=2,
-        returnn_python_exe=None,
-        returnn_root=None,
     ):
         """
         :param dict[str] search_data: dataset used for search
```
michelwi, line 51:
```suggestion
        :param dict[str,Any] search_data: dataset used for search
```
dict typing takes two arguments. Would we capitalize the d of dict in docstring typing? Would we maybe want to switch to true type annotations if we're on it? (this is really no requirement for my approval)

returnn/search.py
```diff
@@ -9,20 +9,25 @@
...
 from i6_core.lib.corpus import Corpus
 import i6_core.util as util
+
 from .config import ReturnnConfig
 
 Path = setup_path(__package__)
 
 
-class ReturnnSearchJob(Job):
+class ReturnnSearchJob(_ReturnnSearchJob):
```
michelwi, line 26: Should we maybe override the docstring to explain why the class was deprecated?

returnn/search.py
```diff
@@ -137,17 +135,18 @@ def create_returnn_config(
         """
         assert device in ["gpu", "cpu"]
         original_config = returnn_config.config
+        res = copy.deepcopy(returnn_config)
         assert "network" in original_config
```
michelwi, line 139: is this assert still valid nowadays? e.g. in case of staged network dict or network definition in prolog/epilog.

returnn/search.py
JackTemaki: Actually using `typing` objects in the text-style docstrings is not really supported. We should use correct Python typing for new Jobs I guess...

JackTemaki:
> Nothing changed after my last review. But if you don't want to incorporate any changes, then I'm fine I guess..

I forgot the push....

michelwi (approved):
> I forgot the push....

Haha ok, now with more enthusiastic approval :D

Issue #279, opened 2022-06-27
Several problems with job: `corpus.CorpusReplaceOrthFromReferenceCorpus`


christophmluscher:
1. `CorpusReplaceOrthFromReferenceCorpus` -> `CorpusReplaceOrthFromReferenceCorpusJob`
2. typing of inputs missing
3. `corpus.Corpus` has an iterator over all segments this should be used for cleaner code
4. the assert in L56 has a variable named `s`. I cant find that variable

christophmluscher:
fixed by #280

PR #280, opened 2022-06-27
`CorpusReplaceOrthFromReferenceCorpus`

christophmluscher:
addresses the issues mentioned in #279

### DIFF
--- a/corpus/convert.py
+++ b/corpus/convert.py
@@ -44,7 +44,7 @@ class CorpusReplaceOrthFromReferenceCorpusJob(Job):
         for s in orth_c.segments():
             orth = s.orth
             tag = s.fullname()
-            orths[tag] = name
+            orths[tag] = orth
 
         c = corpus.Corpus()
         c.load(self.bliss_corpus.get_path())
--- a/corpus/convert.py
+++ b/corpus/convert.py
@@ -1,5 +1,5 @@
 __all__ = [
-    "CorpusReplaceOrthFromReferenceCorpusJob",
+    "CorpusReplaceOrthFromReferenceCorpus",
     "CorpusReplaceOrthFromTxtJob",
     "CorpusToStmJob",
     "CorpusToTextDictJob",
--- a/corpus/convert.py
+++ b/corpus/convert.py
@@ -51,9 +51,7 @@ class CorpusReplaceOrthFromReferenceCorpusJob(Job):
 
         for s in c.segments():
             tag in s.fullname()
-            assert tag in orths.keys(), (
-                    "Segment %s not found in reference corpus" % tag
-                    )
+            assert tag in orths.keys(), "Segment %s not found in reference corpus" % tag
             s.orth = orths[tag]
 
         c.dump(self.out_corpus.get_path())
--- a/corpus/convert.py
+++ b/corpus/convert.py
@@ -1,5 +1,5 @@
 __all__ = [
-    "CorpusReplaceOrthFromReferenceCorpus",
+    "CorpusReplaceOrthFromReferenceCorpusJob",
     "CorpusReplaceOrthFromTxtJob",
     "CorpusToStmJob",
     "CorpusToTextDictJob",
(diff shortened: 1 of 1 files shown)


Atticus1806 (approved):

corpus/convert.py
```diff
@@ -41,23 +41,18 @@ def run(self):
...
         orths = {}
-        for r in orth_c.all_recordings():
-            for i in range(len(r.segments)):
-                orth = r.segments[i].orth
-                tag = r.segments[i].fullname()
-                orths[tag] = orth
+        for s in orth_c.segments():
+            orth = s.orth
+            tag = s.fullname()
+            orths[tag] = name
```
michelwi, line 47: is the change `orth` -> `name` intentional? I don't think `name` is defined.
```suggestion
            orths[tag] = orth
```
the request changes is w.r.t. this comment.

corpus/convert.py
```diff
@@ -18,12 +18,12 @@
 Path = setup_path(__package__)
 
 
-class CorpusReplaceOrthFromReferenceCorpus(Job):
+class CorpusReplaceOrthFromReferenceCorpusJob(Job):
```
michelwi, line 21: Discussion question: the renaming will change the hash. I am not sure if only @Atticus1806 ever used this class, but if not, maybe we should keep the old name.

michelwi (dismissed):

corpus/convert.py
Atticus1806: nvm I had a different idea

corpus/convert.py
```diff
@@ -41,23 +41,18 @@ def run(self):
...
-                    "Segment %s not found in reference corpus" % s
-                )
-                orth = orths[tag]
-                r.segments[i].orth = orth
+        for s in c.segments():
+            tag in s.fullname()
+            assert tag in orths.keys(), "Segment %s not found in reference corpus" % tag
+            s.orth = orths[tag]
 
         c.dump(self.out_corpus.get_path())
```
Atticus1806, line 58:
```suggestion


class CorpusReplaceOrthFromReferenceCorpus(CorpusReplaceOrthFromReferenceCorpusJob):
    pass
    
```

Atticus1806: For my setup this does not yield new Jobs thus this should not break the hashes I think.

michelwi: yes this would work, but the missing "Job" in the name is only a cosmetic problem.

In my opinion, and I am open for discussion, keeping the cosmetic problem is a smaller evil than introducing additional classes.

some user who dislikes it could even do 
```
from i6_core.corpus.convert import CorpusReplaceOrthFromReferenceCorpus as CorpusReplaceOrthFromReferenceCorpusJob
```
and then happily continue their ways consistently.

Atticus1806: This works too, I am fine with both.

corpus/convert.py
christophmluscher: I would not introduce a second class. I would change the existing class name

michelwi:
> So whats the state on this? 

My requested changes have been addressed. 

Now the open question is how to handle the name (in parentheses I tried to collect peoples opinion on the solutions):
* keep the old name and live with inconsistency (me + @JackTemaki prefer this)
* rename the class in place breaking the hashes (@christophmluscher  suggested + @Atticus1806  approved - but @JackTemaki dislikes breaking old setups)
* duplicate classes (nobody likes this)

Anybody feels misrepresented or would like to change their opinion?

christophmluscher:
IMHO actual users should have the most impact on "cosmetic" changes that break hashes...

michelwi:
> IMHO actual users should have the most impact on "cosmetic" changes that break hashes...

Yes, If the actual users agree that they prefer to change the name at the cost of breaking their old setups, then I am ok with this.

corpus/convert.py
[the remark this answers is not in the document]
christophmluscher: @JackTemaki @Atticus1806 @timur-sch add `Job` :+1: or :-1: ?!

Atticus1806:
I would also be fine with just leaving it, tbh. I dont really care as long as we decide it soon. Maybe we should not create a precedent here breaking a hash for cosmetic reasons and leave it as is.

JackTemaki:
I still vote for leaving it as it is.

JackTemaki (approved):

Issue #281, opened 2022-06-27
`CorpusToStmJob` has unused input parameters


christophmluscher:
As far as I can see parameters `exclude_non_speech`, `remove_punctuation` and `fix_whitespace` are not used in the job, but they are set to `True`.
I am not sure how to fix this without changing the hash. If somebody has an idea, feel free to comment

michelwi:
The original implementation still had the functionality.
https://p.i6.informatik.rwth-aachen.de/diffusion/ASR/browse/master/corpus/__init__.py$159

maybe we can restore it?

christophmluscher:
addressed in #282 

@michelwi is something from the original implementation missing?! I dont think so.. so I would leave. either way, I think we should than do a new isse and PR

PR #282, opened 2022-06-27
Refactor CorpusToStmJob

christophmluscher:
addresses #281


corpus/convert.py
```diff
@@ -177,6 +192,30 @@ def run(self):
                     else segment.recording.name
                 )
                 segment_track = segment.track + 1 if segment.track else 1
+
+                orth = f" {segment.orth.strip()} "
```
michelwi, line 196: I see you add the spaces here because the following `replace_recursive` needs `" {token} "` even at the start of sentence. But then you should remove these artificial whitespaces also if `self.fix_whitespace` is false.

corpus/convert.py
```diff
@@ -177,6 +192,30 @@ def run(self):
...
+
+                if self.exclude_non_speech:
+                    for nst in self.non_speech_tokens:
+
+                        def replace_recursive(orthography, token):
+                            pos = orthography.find(f" {token} ")
+                            if pos == -1:
+                                return orthography
+                            else:
+                                orthography = orthography.replace(f" {token} ", "   ")
```
michelwi, line 206: why replace with three spaces instead of one?
```suggestion
                                orthography = orthography.replace(f" {token} ", " ")
```

corpus/convert.py
```diff
@@ -109,25 +114,33 @@ def run(self):
         c.dump(self.out_corpus.get_path())
 
 
-class CorpusToStmJob(Job):
+class CorpusToStmJob(_CorpusToStmJob):
+    pass
+
+
+class CorpusToStmJobV2(Job):
```
michelwi, line 121: I don't think we need to deprecate the old job. you can add the new parameters via sis_hash_exclude.

we don't even have to change the default `exclude_non_speech` and `remove_punctuation` from `True` to `False` because with empty tuples the for-loops below will do exactly nothing (i.e. default behavior of the original job)

The only question is, if we silently want to implement the fix_whitespace option. I think this is harmless and definitively worth it to avoid a V2 Job.

corpus/convert.py
```diff
@@ -109,25 +114,33 @@ def run(self):
...
-        exclude_non_speech=True,
-        remove_punctuation=True,
-        fix_whitespace=True,
-        name="",
-        tag_mapping=(),
+        bliss_corpus: Path,
+        exclude_non_speech: bool = False,
+        non_speech_tokens: Tuple[str, ...] = (),
+        remove_punctuation: bool = False,
+        punctuation_tokens: Tuple[str, ...] = (),
```
michelwi, line 132: do you really prefer a variable length tuple over a list with default None?
We could even give some default punctuation tokens here..

(this comment is not included in "request changes")

michelwi (changes requested):
Well, not "request" but "strongly suggest" changes.

corpus/convert.py
JackTemaki: If possible I would also avoid the V2 if were are not really having two different jobs with different behavior.

We certainly can write the hash function in a way to replicate old hashing.

corpus/convert.py
christophmluscher: I am not sure anymore.. not sure if this has a valid reason..

christophmluscher:
@JackTemaki @michelwi @albertz you guys have an idea with the hash test is breaking?

albertz:
> @JackTemaki @michelwi @albertz you guys have an idea with the hash test is breaking?

It's not about the hash. Your type info is invalid. E.g. `List[str, ...]`. I'm not even sure what that is supposed to mean. It contains ellipsis (`...`) which is not allowed.

christophmluscher:
> > @JackTemaki @michelwi @albertz you guys have an idea with the hash test is breaking?
> 
> It's not about the hash. Your type info is invalid. E.g. `List[str, ...]`. I'm not even sure what that is supposed to mean. It contains ellipsis (`...`) which is not allowed. Maybe you just mean `List[str]`? Or you mean `List[Any]`?

thanks. I somwhow missed that.. yeah `List[str]`

corpus/convert.py
```diff
@@ -172,6 +187,31 @@ def run(self):
...
+                                return orthography
+                            else:
+                                orthography = orthography.replace(f" {token} ", " ")
+                                return replace_recursive(orthography, token)
+
+                        orth = replace_recursive(orth, nst)
+
+                if self.remove_punctuation:
+                    for pt in self.punctuation_tokens:
+                        orth = orth.replace(pt, "")
```
michelwi, line 208: why do we need a recursive implementation for non-speech tokens but not for punctuation?

corpus/convert.py
```diff
@@ -109,29 +112,41 @@ class CorpusToStmJob(Job):
...
-        name="",
-        tag_mapping=(),
+        bliss_corpus: Path,
+        exclude_non_speech: bool = True,
+        non_speech_tokens: Optional[List[str]] = None,
+        remove_punctuation: bool = True,
+        punctuation_tokens: Optional[Union[str, List[str]]] = None,
+        fix_whitespace: bool = True,
+        name: str = "",
+        tag_mapping: Union[Tuple[()], Tuple[str, Dict[str, str]]] = (),
```
michelwi, line 126: The typing of tag_mapping is incorrect. The way it is currently implemented it should be
```suggestion
        tag_mapping: List[Tuple[Tuple[str, str, str], Dict[int, tk.Path]]] = (),
```
Where the 3-string tuple contains ("short name", "long name", "description") of each tag. and the `Dict[int, tk.Path]` is e.g. the output of a FilterSegments*Jobs https://github.com/rwth-i6/i6_core/blob/f267011280fe60c8ae23929eb6db77019ac59a1e/corpus/filter.py#L40-L43

Arguably this is not my finest implementation and we can revise this (now or in the future).

corpus/convert.py
christophmluscher: that is actually quite the good question.. not sure why I implemented this differently.. this shold work without recursion.. Oo

corpus/convert.py
christophmluscher: Originally this was the typing information: `tuple[str, dict[str, str]]` so this was wrong?

christophmluscher: ok thanks

christophmluscher:
@michelwi thanks for the explanation on typing

corpus/convert.py
```diff
@@ -109,29 +112,43 @@ class CorpusToStmJob(Job):
...
+        :param bliss_corpus: Path to Bliss corpus
+        :param exclude_non_speech: non speech tokens should be removed
+        :param non_speech_tokens: defines the list of non speech tokens
+        :param remove_punctuation: should punctuation be removed
+        :param punctuation_tokens: defines list/string of punctuation tokens
+        :param fix_whitespace: should white space be fixed
+        :param name: new corpus name
+        :param tag_mapping: 3-string tuple contains ("short name", "long name", "description") of each tag.
+            and the Dict[int, tk.Path] is e.g. the output of a FilterSegments*Jobs
+            https://github.com/rwth-i6/i6_core/blob/f267011280fe60c8ae23929eb6db77019ac59a1e/corpus/filter.py#L40-L43
```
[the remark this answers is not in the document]
christophmluscher: would be fine for me. but why?

corpus/convert.py
michelwi: maybe we need some test cases 😇

christophmluscher: hm yes maybe ^^

Chris assigns Chris :D

corpus/convert.py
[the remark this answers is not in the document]
michelwi: As a person using and programming these recipes on the command line, it feels always a a major disruption to switch to the browser.
Also in this case the link is a permalink to a specific version and bound to be out of date at some point.
Also the same information that is contained in the link can be given by the word "out_single_segment_files" actually, so I don't see any additional value.

michelwi (approved):
Ok now all my concerns are addressed.

christophmluscher:
as suggested added test case here

corpus/convert.py
michelwi: Well now we have a test case and it is failing because the expected output is 
```
116-288045-0000 1 116  0.00 10.65 <d0> AS I APPROACHED THE CITY I HEARD BELLS RINGING AND A LITTLE LATER I FOUND THE STREETS ASTIR WITH THRONGS OF WELL DRESSED PEOPLE IN FAMILY GROUPS WENDING THEIR WAY HITHER AND THITHER
;; LABEL "d0" "default0" "all other segments of category 0"
```
and the actual output is
```
116-288045-0000 1 116  0.00 10.65 <d0> AS I APPROACHED THE CITY I HEARD BELLS RINGING [noise] AND A LITTLE LATER I FOUND THE STREETS ASTIR WITH THRONGS OF WELL DRESSED PEOPLE IN FAMILY GROUPS WENDING THEIR WAY HITHER AND THITHER
;; LABEL "d0" "default0" "all other segments of category 0"
```
It seems the non recursive implementation does not capture two consecutive occurrences of `[noise]` in the corpus. maybe because the last space of the first match coincides with the first space of the last match? 
maybe we can use re.sub or similar to remedy the problem (e.g. define a match only if {nst} is surrounded by [whitespace, begin-, end-of sentence] but then only remove the token.. 

anyway.. _request change_ until test case works :)

tests/job_tests/corpus/files/test_job.corpus.stm
```diff
@@ -0,0 +1,2 @@
+116-288045-0000 1 116  0.00 10.65 <d0> AS I APPROACHED THE CITY I HEARD BELLS RINGING AND A LITTLE LATER I FOUND THE STREETS ASTIR WITH THRONGS OF WELL DRESSED PEOPLE IN FAMILY GROUPS WENDING THEIR WAY HITHER AND THITHER
+;; LABEL "d0" "default0" "all other segments of category 0"
```
michelwi, line 2: newline missing at end of file

michelwi (changes requested):

corpus/convert.py
christophmluscher: I would just re-add recursion

michelwi (approved):
And test works now :)

corpus/convert.py
```diff
@@ -109,29 +112,42 @@ class CorpusToStmJob(Job):
...
+
     def __init__(
         self,
-        bliss_corpus,
-        exclude_non_speech=True,
-        remove_punctuation=True,
-        fix_whitespace=True,
-        name="",
-        tag_mapping=(),
+        bliss_corpus: Path,
```
JackTemaki, line 119:
```suggestion
        bliss_corpus: Path,
        *,
```

JackTemaki: As you added new parameters that are not at the end, you should force the user to use named parameters (for all parameters except `bliss_corpus`), to not have accidentally wrong parameter assignment.

JackTemaki (changes requested):
Only small remark, otherwise good.

Atticus1806:
Nothing to add, agree with @JackTemaki , approval once that is processed.

christophmluscher:
@Atticus1806 ping :)

tests/job_tests/corpus/files/test_job.corpus.xml
```diff
@@ -0,0 +1,12 @@
+<?xml version="1.0" encoding="utf-8"?>
+<corpus name="job-testing">
+  <speaker-description name="116">
+    <gender>male</gender>
+  </speaker-description>
+  <recording name="116-288045-0000" audio="sample.wav">
+    <speaker name="116"/>
+    <segment name="116-288045-0000" start="0.0000" end="10.6500">
+      <orth> AS I APPROACHED [non-speech] THE CITY, I HEARD BELLS RINGING [noise] [noise] AND A    LITTLE LATER I FOUND THE STREETS ASTIR WITH THRONGS OF WELL DRESSED PEOPLE IN FAMILY GROUPS WENDING THEIR WAY HITHER AND THITHER!!! </orth>
```
Atticus1806, line 9: How important are the whitespace after the A? 
It does not matter for approval, but for the `CorpusReplaceOrthFromReferenceCorpus` test  these must not be there and if you "dont need them" we could maybe merge the test corpora? Not sure if this makes sense, or if each test should have its own files.

Atticus1806 (approved):

tests/job_tests/corpus/files/test_job.corpus.xml
michelwi: I think these whitespaces are there to explicitly test the fix whitespace option, so I think we should keep it. 
It is in general, I think, fine to have separate files per test case.

christophmluscher: the extra white space is there for a reason. to test the flag `fix_whitespace`

corpus/convert.py
```diff
@@ -172,6 +189,36 @@ def run(self):
                     else segment.recording.name
                 )
                 segment_track = segment.track + 1 if segment.track else 1
+
+                orth = f" {segment.orth.strip()} "
+
+                if self.exclude_non_speech:
+                    for nst in self.non_speech_tokens:
+
+                        def replace_recursive(orthography, token):
```
JackTemaki, line 198: can you change this to a classmethod/staticmethod? A definition in a loop does not seem nice.

JackTemaki (changes requested):

JackTemaki (approved):

PR #285, opened 2022-07-05
refactor code formatting for mfcc, gammatone, energy

JackTemaki:
Some style updates and incomplete docstrings. I did not update all files because of time issues.

The only relevant change is that `energy_flow` now correctly handles `samples_options` and `fft_options`.

### DIFF
--- a/features/gammatone.py
+++ b/features/gammatone.py
@@ -36,7 +36,7 @@ def gammatone_flow(
     minfreq: int = 100,
     maxfreq: int = 7500,
     channels: int = 68,
-    warp_freqbreak=None,
+    warp_freqbreak: Optional[int] = None,
     tempint_type: str = "hanning",
     tempint_shift: float = 0.01,
     tempint_length: float = 0.025,
--- a/features/gammatone.py
+++ b/features/gammatone.py
@@ -1,13 +1,17 @@
 __all__ = ["GammatoneJob", "gammatone_flow"]
 
 import copy
+from typing import Any, Dict, Optional
 
-from .common import *
-from .extraction import *
-import i6_core.rasr as rasr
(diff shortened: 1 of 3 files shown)


features/gammatone.py
```diff
@@ -29,27 +33,53 @@ def GammatoneJob(crp, gt_options=None, **kwargs):
 
 
 def gammatone_flow(
-    minfreq=100,
-    maxfreq=7500,
-    channels=68,
+    minfreq: int = 100,
+    maxfreq: int = 7500,
+    channels: int = 68,
     warp_freqbreak=None,
```
michelwi, line 39:
```suggestion
    warp_freqbreak: Optional[int] = None,
```
This should be of same type as minfreq/maxfreq

michelwi, line 39: according to https://github.com/rwth-i6/rasr/blob/907eec4f4e36c11153f6ab6b5dd7675116f909f6/src/Signal/GammaTone.cc#L228-L234 these can be of type float. But we do anyway always use ints here.. so no strong opinion on this one 🤷 

```suggestion
    minfreq: float = 100,
    maxfreq: float = 7500,
    channels: int = 68,
    warp_freqbreak: Optional[float] = None,
```

michelwi (approved):
Generally ok with the changes. I added typing suggestions inline. Choose at most one of them :)

christophmluscher (approved):

features/gammatone.py
JackTemaki: I think we usually treat frequencies as integers, so I would just leave it. If a person ever needs floats here that person would have so much knowledge about the pipeline that typing hints would not be a hindrance anyway :)

PR #286, opened 2022-07-05
refactor Sclite Job

JackTemaki:
Refactoring for the Sclite job, so that we can combine it with an explicit `MakeJob` for the new pipelines.

Also some other updates regarding `uncached_path` and docstrings.

Please have a close look at the new hashing function. Unfortunately due to the recent changes I have no fully finished GMM pipeline any more to test this job.

### DIFF
--- a/recognition/scoring.py
+++ b/recognition/scoring.py
@@ -71,7 +71,6 @@ class ScliteJob(Job):
         :param sort_files: sort ctm and stm before scoring
         :param additional_args: additional command line arguments passed to the Sclite binary call
         :param sctk_binary_path: set an explicit binary path.
-        :param hash_binary_path: if the binary path should influence the hash
         """
         self.set_vis_name("Sclite - %s" % ("CER" if cer else "WER"))
 
--- a/recognition/scoring.py
+++ b/recognition/scoring.py
@@ -12,6 +12,7 @@ import subprocess as sp
 import tempfile
 import collections
 import re
+from typing import List, Optional
 
 from sisyphus import *
 from i6_core.lib.corpus import *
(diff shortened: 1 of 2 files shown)


recognition/scoring.py
```diff
@@ -44,14 +45,42 @@ def run(self):
...
+
+    def __init__(
+        self,
+        ref: tk.Path,
+        hyp: tk.Path,
+        cer: bool = False,
+        sort_files: bool = False,
+        additional_args: Optional[List[str]] = None,
+        sctk_binary_path: Optional[tk.Path] = None,
+        hash_binary_path: bool = False,
```
michelwi, line 64: I don't like to introduce options to hash other parameters or not. Also I dislike that it is off by default.

I suggest instead to have 
```
__sis_hash_exclude__ = {"sctk_binary_path": None}
```
or equivalent implementation in hash function and then if None is given, we have the old behavior. and otherwise we use it always for the hash (similar to explicit returnn or rasr paths)

recognition/scoring.py
```diff
@@ -44,14 +45,42 @@ def run(self):
...
+
+    Outputs:
+        - out_report_dir: contains the report files with detailed scoring information
+        - out_*: the job also outputs many variables, please look in the init code for a list
+    """
+
+    def __init__(
+        self,
+        ref: tk.Path,
+        hyp: tk.Path,
```
michelwi, line 59: Is this compatible with the OptimizeAMandLMScaleJob ? I am not sure that this one inputs tk.Path objects
https://github.com/rwth-i6/i6_core/blob/8dab653e8c6d6e872b3b0be69ad3307186aeb915/recognition/optimize_parameters.py#L99-L102

JackTemaki: What about just using tk.Path in the optimizer?

recognition/scoring.py
JackTemaki: Got it

recognition/scoring.py
michelwi: the required changes should be part of this PR or this PR can only land after another PR addressing the changes in the optimizer job.

JackTemaki: but they are part of this PR!

michelwi (approved):

recognition/scoring.py
```diff
@@ -44,14 +45,42 @@ def run(self):
...
+        sctk_binary_path: Optional[tk.Path] = None,
+    ):
+        """
+        :param ref: reference stm text file
+        :param hyp: hypothesis ctm text file
+        :param cer: compute character error rate
+        :param sort_files: sort ctm and stm before scoring
+        :param additional_args: additional command line arguments passed to the Sclite binary call
+        :param sctk_binary_path: set an explicit binary path.
+        :param hash_binary_path: if the binary path should influence the hash
```
christophmluscher, line 74: remove?

christophmluscher (approved):
both comments do not warrant a request changes :D

recognition/scoring.py
christophmluscher, line 74:
```suggestion
```

PR #289, opened 2022-07-11
Add test case for CorpusToStmJob

christophmluscher:


christophmluscher:
should be merged after #282 

test case whould fail otherwise

tests/job_tests/corpus/files/test_job.corpus.xml
```diff
@@ -0,0 +1,12 @@
...
+  <speaker-description name="116">
+    <gender>male</gender>
+  </speaker-description>
+  <recording name="116-288045-0000" audio="sample.wav">
+    <speaker name="116"/>
+    <segment name="116-288045-0000" start="0.0000" end="10.6500">
+      <orth> AS I APPROACHED [non-speech] THE CITY, I HEARD BELLS RINGING [noise] [noise] AND A    LITTLE LATER I FOUND THE STREETS ASTIR WITH THRONGS OF WELL DRESSED PEOPLE IN FAMILY GROUPS WENDING THEIR WAY HITHER AND THITHER!!! </orth>
+    </segment>
+  </recording>
+</corpus>
```
michelwi, line 12: newline missing

michelwi:
maybe add this as part of #282 so we can see that the test is working before merging.

PR #290, opened 2022-07-13
Fix CorpusReplaceOrthFromReferenceCorpus

Atticus1806:
#280 introduced a bug, the tag was not actually changed in each iteration.

Tested in my pipeline, now works with the fix.

### DIFF
--- a/corpus/convert.py
+++ b/corpus/convert.py
@@ -50,7 +50,7 @@ class CorpusReplaceOrthFromReferenceCorpus(Job):
         c.load(self.bliss_corpus.get_path())
 
         for s in c.segments():
-            tag in s.fullname()
+            tag = s.fullname()
             assert tag in orths.keys(), "Segment %s not found in reference corpus" % tag
             s.orth = orths[tag]
 


Atticus1806:
So this is a first draft of a test case. This will break right now, because I would wait for #282 to go through and reuse the corpus file and add a "corrupt" one with a few changes. 
But feel free to comment already if there is anything obvious. The other part of the PR can also wait until then.

tests/job_tests/returnn/test_search.py
```diff
@@ -43,3 +44,16 @@ def test_search_bpe_to_words_nbest():
             open(bpe_to_words_job.out_word_search_results.get_path(), "rt").read()
         )
         assert reference_dict == job_dict
+
+
+def test_corpus_replace_orth_from_reference():
```
michelwi, line 49: I think these tests are organized per module. So the test for `i6_core.corpus.convert.CorpusReplaceOrthFromReferenceCorpus` should live in `tests/job_tests/corpus/test_convert.py`

tests/job_tests/returnn/test_search.py
```diff
@@ -43,3 +44,16 @@ def test_search_bpe_to_words_nbest():
...
+  from i6_core.corpus.convert import CorpusReplaceOrthFromReferenceCorpus
+  with tempfile.TemporaryDirectory() as tmpdir:
+    reference_corpus = Path("files/test_job.corpus.xml")
```
michelwi, line 52: Don't forget to also push these files. Now your test fails.

michelwi (commented):
Also check black for the test please.

tests/job_tests/returnn/files/test_job.corpus.xml
```diff
@@ -0,0 +1,12 @@
+<?xml version="1.0" encoding="utf-8"?>
+<corpus name="job-testing">
+  <speaker-description name="116">
+    <gender>male</gender>
+  </speaker-description>
+  <recording name="116-288045-0000" audio="sample.wav">
+    <speaker name="116"/>
+    <segment name="116-288045-0000" start="0.0000" end="10.6500">
+      <orth> AS I APPROACHED [non-speech] THE CITY, I HEARD BELLS RINGING [noise] [noise] AND A    LITTLE LATER I FOUND THE THIS DOES NOT BELONG HERE! BUT I WANT IT HERE! STREETS ASTIR WITH THRONGS OF WELL DRESSED PEOPLE IN FAMILY GROUPS WENDING THEIR WAY HITHER AND THITHER!!! </orth>
```
michelwi, line 9: this is purely a semantic problem, but I think you corrupted the original corpus and the corrupt corpus is with original transcription.

tests/job_tests/returnn/files/test_job.corpus.xml
```diff
@@ -0,0 +1,12 @@
...
+    </segment>
+  </recording>
+</corpus>
```
michelwi, line 12: newline missing at the end. I think this is what makes the test fail.

JackTemaki (approved):

PR #291, opened 2022-07-19
Do not filter out special lemmas.

sarahberanek:
Do not filter out special lemmas eg this silence lemma was filtered out after running the FilterJob:
<lemma special="silence"><orth/><phon>si</phon><synt/><eval/></lemma>

### DIFF
--- a/lexicon/conversion.py
+++ b/lexicon/conversion.py
@@ -70,7 +70,7 @@ class FilterLexiconByWordListJob(Job):
         for lemma in old_lexicon.findall("lemma"):
             if any(
                 transform(orth.text) in words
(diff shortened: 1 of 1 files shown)


michelwi (changes requested):

michelwi (approved):
Ok, but please also test my implementation. 🙃

sarahberanek:
> Ok, but please also test my implementation. upside_down_face

Jup will rerun the whole Lexicon creation from scratch now. Thanks

sarahberanek:
> Ok, but please also test my implementation. upside_down_face

Worked like a charm.

christophmluscher (approved):

Issue #254, opened 2022-04-19
move gs.RETURNN_PYTHON_EXE, gs.RETURNN_ROOT and maybe others


albertz:
sisyphus.global_settings was not meant to be used inside recipes. It's there to define sisyphus specific settings, e.g. timeouts, which engines to use or which environment should be set during job execution.

Workflow related things should be defined inside the recipes or config directory. If you want to define all external paths in one place you could create a file either inside the recipe folder or, especially if you don't want to cleanly share it across multiple recipes, inside your config folder. This file can then be imported in all relevant places. It would be possible to define the used path inside a tk.Path object there, or use the output of a job which sets everything up.

_Originally posted by @critias in https://github.com/rwth-i6/sisyphus/issues/82#issuecomment-1103097400_

michelwi:
> > i6_core does not need a settings file.
...
Instead we should simply use a different mechanism to provide a default value to returnn_root ... e.g. we can have i6_core.tools.default_software

> Isn't this a contradiction, or just rephrasing "settings file" as "default software" file?
Or do you say that it should not be allowed to overwrite the default? 

Yes, in my opinion the default should not be changed in general but rather overwritten by explicitly providing an argument to the job.

> why should you be able to overwrite the default env (via Sis global settings) but not this setting? 

No, also the env (I think you are referring to `gs.RETURNN_PYTHON_EXE` should not be set via Sis global settings but rather be the output of a `CreateVirtualenvJob` or a `tk.Path` object of which we can provide a general default but the user can (and should) provide their own via Job args.

Same as:
* `gs.RASR_ROOT` - should be `CompileRasrJob(CloneGitRepositoryJob(<RASR Repo>).out_repository)`
* `gs.RASR_ARCH` - can be string default inside the Job
* `gs.RASR_PYTHON_HOME` - should default to not set it if not given (?)
* `gs.G2P_PYTHON` - `CreateVirtualenvJob`
* `gs.SCTK_PATH` - `CloneGitRepositoryJob` (if it needs to be compiled maybe `CompileSctkJob`)
* `gs.KALDI_PATH` - `CloneGitRepositoryJob` -> `CompileKaldiJob`
* `gs.WARNING_NO_SENTENCEPIECE` - This is not needed. We don't use sentencepiece in the manager
* `gs.SUBWORD_NMT_PATH` - `CloneGitRepositoryJob`

[...]

PR #292, opened 2022-07-22
sctk_binary_path variable in Hub5ScoreJob

DanEnergetics:
The use of binary path variables in the sisyphus.global_settings might become deprecated in the future (as discussed in #254). This is a slight change of the Hub5ScoreJob similar to the ScliteJob in this direction. Moreover, some type annotations have been added to the __init__ signature.

### DIFF
--- a/recognition/scoring.py
+++ b/recognition/scoring.py
@@ -203,6 +203,7 @@ class ScliteJob(Job):
 
 class Hub5ScoreJob(Job):
     __sis_hash_exclude__ = {"sctk_binary_path": None}
+
     def __init__(
         self,
         ref: Union[tk.Path, str],
@@ -243,13 +244,14 @@ class Hub5ScoreJob(Job):
         yield Task("run", mini_task=True)
 
     def run(self, move_files=True):
-
         sctk_path = ""
         if self.sctk_binary_path is not None:
             sctk_path = self.sctk_binary_path.get_path()
         elif hasattr(gs, "SCTK_PATH"):
             sctk_path = gs.SCTK_PATH
-        hubscr_path = os.path.join(sctk_path, "hubscr.pl") # evaluates to just "hubscr.pl" if sctk_path is empty
+        hubscr_path = os.path.join(
+            sctk_path, "hubscr.pl"
+        )  # evaluates to just "hubscr.pl" if sctk_path is empty
 
         sctk_opt = ["-p", sctk_path] if sctk_path else []
 
--- a/recognition/scoring.py
+++ b/recognition/scoring.py
@@ -211,6 +211,12 @@ class Hub5ScoreJob(Job):
         hyp: tk.Path,
         sctk_binary_path: Optional[tk.Path] = None,
     ):
+        """
+        :param ref: reference stm text file
+        :param glm: text file containing mapping rules for scoring
+        :param hyp: hypothesis ctm text file
+        :param sctk_binary_path: set an explicit binary path.
+        """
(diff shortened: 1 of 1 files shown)


recognition/scoring.py
```diff
@@ -202,12 +202,21 @@ def calc_wer(self):
...
 
 class Hub5ScoreJob(Job):
-    def __init__(self, ref, glm, hyp):
+    __sis_hash_exclude__ = {"sctk_binary_path": None}
+
+    def __init__(
+        self,
+        ref: Union[tk.Path, str],
+        glm: Union[tk.Path, str],
+        hyp: Union[tk.Path, str],
```
michelwi, line 211:
```suggestion
        ref: tk.Path,
        glm: tk.Path,
        hyp: tk.Path,
```
We want to disallow using strings for inputs that are a file/folder in the file system.

Also `self.sctk_binary_path.get_path()` will not work on `str`.
Request changes w.r.t. this comment.

recognition/scoring.py
```diff
@@ -202,12 +202,21 @@ def calc_wer(self):
...
+        sctk_binary_path: Optional[tk.Path] = None,
+    ):
```
michelwi, line 213:
```suggestion
    ):
        """
        :param ref: reference stm text file
        :param glm: text file containing mapping rules for scoring
        :param hyp: hypothesis ctm text file
        :param sctk_binary_path: set an explicit binary path.
        """
```
If we are at it, let me add a doc string.

recognition/scoring.py
```diff
@@ -12,7 +12,7 @@
 import tempfile
 import collections
 import re
-from typing import List, Optional
+from typing import List, Optional, Union
```
michelwi, line 15:
```suggestion
from typing import List, Optional
```
If we disallow `str` below, then we don't need `Union`

michelwi (changes requested):

michelwi (approved):

DanEnergetics:
I added your suggested changes. I thought about removing str in the annotations, as well but wasn't sure, if in that case, we would also like to remove appearances of tk.uncached as suggested in #32 the way it's done in ScliteJob already. What's your opinion?

michelwi:
>  remove appearances of tk.uncached_path

sure, lets do it.

recognition/scoring.py
```diff
@@ -235,29 +250,33 @@ def tasks(self):
...
+            sctk_path, "hubscr.pl"
+        )  # evaluates to just "hubscr.pl" if sctk_path is empty
+
+        sctk_opt = ["-p", sctk_path] if sctk_path else []
 
         ref = self.ref
         try:
-            ref = shutil.copy(tk.uncached_path(ref), ".")
+            ref = shutil.copy(ref.get_path(), ".")
         except shutil.SameFileError:
```
christophmluscher, line 267: I am wondering: why are we copying ref and hyp?

christophmluscher: how can this be the same file?? `"."` should be the job work dir, correct? ergo this should not be the same file

christophmluscher (changes requested):

recognition/scoring.py
michelwi: > why are we copying ref and hyp?

dunno.

> how can this be the same file??

maybe during automatic LM scale optimization?

christophmluscher:
moving the copy of ref and hyp to a new issue #293

christophmluscher (approved):

PR #294, opened 2022-07-26
Add helper to get hdf content dict

Atticus1806:
Adds helper function to get the "data"/input value from a RETURNN written HDF file as dict indexed by sequence tag.

### DIFF
--- a/lib/hdf.py
+++ b/lib/hdf.py
@@ -1,0 +1,24 @@
+import h5py
+import numpy as np
+from typing import Dict
+
+
+def get_input_dict_from_returnn_hdf(hdf_file: h5py.File) -> Dict[str, np.ndarray]:
+    """
+    Generate dictionary containing the "data" value as ndarray indexed by the sequence tag
+
+    :param hdf_file: HDF file to extract data from
+    :return:
+    """
+    inputs = hdf_file["inputs"]
+    seq_tags = hdf_file["seqTags"]
+    lengths = hdf_file["seqLengths"]
+
+    output_dict = {}
+    offset = 0
+    for tag, length in zip(seq_tags, lengths):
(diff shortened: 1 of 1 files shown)


lib/hdf.py
```diff
@@ -0,0 +1,24 @@
+import h5py
+import numpy as np
+from typing import Dict
+
+
+def get_input_dict_from_returnn_hdf(hdf_file: h5py.File) -> Dict[str, np.ndarray]:
```
christophmluscher, line 6: Does is make sense to add the option for a Path Object?

michelwi: As this needs physical access to the underlaying file, this would anyway only be called in the worker, where we may have resolved all Path objects. I don't think it is that necessary.

Also all other library functions act directly on the file and not path objects.

christophmluscher (approved):

Atticus1806:
ping @michelwi @JackTemaki

michelwi (commented):
I have nothing for or against it.. only question: is this general enough that everybody will need this or should this rather be part of a personal setup?

Atticus1806:
So this will definitely be used by @JackTemaki , @timur-sch and me. About other potential usecases not sure, but e.g. if you want to quickly open/use the output of #297 then this would be the function to do it.

JackTemaki (approved):

christophmluscher:
IMHO if three people are using this then we can add this to `i6_core`

similar to the `*_flow` functions

since this is `i6_core` and not `i6_asr`, I would allow for TTS functionality ^^

christophmluscher:
if no vetos until Monday, I think we should merge this

@curufinwe

michelwi (approved):
> IMHO if three people are using this then we can add this to i6_core

Agreed.

PR #295, opened 2022-07-26
Add Blacklist option to ReturnnVocabFromPhonemeInventory

Atticus1806:
Adds the option to give a blacklist of excluded phonemes when construction the RETURNN vocab.


michelwi (commented):
I am wondering if it would make sense / be preferable to have the blacklist not as an `Iterable` but as a `tk.Path` file that is opened at runtime. This way it could be also a job output in the future.

Atticus1806:
I would still vouch for having the option to directly give the list here and not write it via Job first. I think @JackTemaki would not really care, what do you think @christophmluscher?

christophmluscher:
I would keep both options available.
1. already implemented
2. more flexibility

returnn/vocabulary.py
```diff
@@ -21,11 +22,19 @@ class ReturnnVocabFromPhonemeInventory(Job):
     tk.Variable out_vocab_size: integer variable containing the vocabulary size (``vocab_size``)
     """
 
-    def __init__(self, bliss_lexicon: tk.Path):
+    __sis_hash_exclude__ = {"blacklist": None}
+
+    def __init__(
+        self,
+        bliss_lexicon: tk.Path,
+        blacklist: Optional[Union[Iterable, tk.Path]] = None,
```
christophmluscher, line 30: `Iterable` allows `dict`. Is this possibility covered in the code?

christophmluscher: from code the phonemes need to be given as the values?? maybe extend doc string?

returnn/vocabulary.py
```diff
@@ -36,8 +45,16 @@ def tasks(self):
     def run(self):
         lex = lexicon.Lexicon()
         lex.load(self.bliss_lexicon.get_path())
-
-        vocab = {k: v for v, k in enumerate(lex.phonemes.keys())}
+        if isinstance(self.blacklist, tk.Path):
+            blacklist = uopen(self.blacklist.get_path()).readlines()
+            blacklist = [phoneme.strip() for phoneme in blacklist]
+        else:
+            blacklist = self.blacklist
```
christophmluscher, line 52: Check for `None` here and if `None` set `[]`

but this is personal preference

returnn/vocabulary.py
```diff
@@ -21,11 +22,19 @@ class ReturnnVocabFromPhonemeInventory(Job):
...
+    __sis_hash_exclude__ = {"blacklist": None}
+
+    def __init__(
+        self,
+        bliss_lexicon: tk.Path,
+        blacklist: Optional[Union[Iterable, tk.Path]] = None,
+    ):
         """
         :param bliss_lexicon: a bliss lexicon xml file containg a phoneme inventory
+        :param blacklist: Exclude phonemes in blacklist from vocab
```
christophmluscher, line 34: `if blacklist is Path` maybe specify file formatting?!

returnn/vocabulary.py
Atticus1806: Maybe this should be set to `Set` instead, was not sure whether something more generic would be nice, but I am not sure how the code behaves in these situations.
I think they would need to be given as keys.

returnn/vocabulary.py
```diff
@@ -36,8 +45,14 @@ def tasks(self):
     def run(self):
         lex = lexicon.Lexicon()
         lex.load(self.bliss_lexicon.get_path())
-
-        vocab = {k: v for v, k in enumerate(lex.phonemes.keys())}
+        if isinstance(self.blacklist, tk.Path):
+            blacklist = uopen(self.blacklist.get_path()).readlines()
+            blacklist = [phoneme.strip() for phoneme in blacklist]
```
michelwi, line 50: I don't like that in one case you have a set, in the other case you construct a list instead. As membership test in a set is faster, I would change this here to a set comprehension.

Not that I expect it to matter in practice as the blacklist is small.

```suggestion
            blacklist = {phoneme.strip() for phoneme in blacklist}
```

tests/job_tests/returnn/files/blacklist
```diff
@@ -0,0 +1,2 @@
+[SILENCE]
+HH
```
michelwi, line 2: maybe add newline character at the end of file?

returnn/vocabulary.py
```diff
@@ -36,8 +45,14 @@ def tasks(self):
...
+        elif isinstance(self.blacklist, Set):
+            blacklist = self.blacklist
+        else:
+            blacklist = []
```
michelwi, line 54: also here maybe switch to empty set to have consistent type of blacklist?
```suggestion
            blacklist = set()
```

returnn/vocabulary.py
michelwi: > I think they would need to be given as keys

yes, this is correct.
```
>>> a = {"foo": "bar"}
>>> "foo" in a
True
>>> "bar" in a
False
```

I think `Iterable` would be fine.. if someone inputs a dict as a blackLIST then they should check themselves how this is handled in that case imho.

If you revert to accepting `Iterable` then disregard my previous comment concerning "every blacklist should be `set`".

Atticus1806: I prefer enforcing set because this way its clear to the user what the input should be and as long as the user follows this they cannot do wrong. I see that not allowing a list for Blacklist is a bit off. Iterable would be fine to me too.

Atticus1806: For reading and empty set we should still use set I think anyways

michelwi (approved):

christophmluscher (approved):

PR #296, opened 2022-07-26
Add Job which dumps list[str] into text file.

sarahberanek:


text/processing.py
```diff
@@ -256,3 +263,28 @@ def run(self):
...
+
+
+class WriteToFileJob(Job):
+    """
+    Write a given list[str] into a text file, one entry per line
+    """
+
+    def __init__(
+        self,
+        list_to_dump,
```
JackTemaki, line 275: I think we do not have to be specific for lists here.

I would thus just call this parameter `content`, and then we can extend it to also support  `str` or sth. else later,
and name the job `WriteToTextFileJob`.

text/processing.py
```diff
@@ -256,3 +263,28 @@ def run(self):
...
+    ):
+        """
+        :param list[str] list_to_dump: input list of str which will be written into a text file
+        """
+        self.list_to_dump = list_to_dump
+
+        self.out_list = self.output_path("list.txt")
```
JackTemaki, line 282: I would suggest `out_file` then.

JackTemaki (changes requested):

christophmluscher (commented):
I would support @JackTemaki suggestions

christophmluscher (approved):

michelwi (approved):

text/processing.py
```diff
@@ -256,3 +263,28 @@ def run(self):
...
+        self.content = content
+
+        self.out_file = self.output_path("file.txt")
+
+    def tasks(self):
+        yield Task("run", mini_task=True)
+
+    def run(self):
+        with open(self.out_file.get_path(), "w") as f:
+            for line in self.content:
```
JackTemaki, line 289: I just noticed this breaks for type `str`, because then each character is written to a line, so the for loop should only be executed when it is of type list.

JackTemaki (changes requested):

text/processing.py
```diff
@@ -256,3 +263,28 @@ def run(self):
...
+                f.write(line + "\n")
```
michelwi, line 290: Ok, then let me suggest some handling of various types that we might want to support:
```suggestion
            if isinstance(self.content, str):
                f.write(self.content)
            elif isinstance(self.content, dict):
                for key, val in self.content.items():
                    f.write(f"{key}: {val}\n")
            elif isinstance(self.content, Iterable)
                for line in self.content:
                    f.write(f"{line}\n")
            else:
                raise NotImplementedError
```

add `from collections.abc import Iterable` on top of file.

sarahberanek: Like it. Done it.

JackTemaki (approved):

PR #297, opened 2022-07-26
ForwardJob allow network without Checkpoint/Params

Atticus1806:
Helper to run returnn feature extraction dumping the sequence lengths into hdf files.
Not tested yet, thus draft PR.

### DIFF
--- a/returnn/dataset.py
+++ b/returnn/dataset.py
@@ -149,7 +149,7 @@ def get_returnn_length_hdfs(
         returnn_root=returnn_root,
         hdf_outputs=hdf_outputs,
         mem_rqmt=mem_rqmt,
-        time_rqmt=time_rqmt
+        time_rqmt=time_rqmt,
     )
     if job_alias is not None:
         forward_job.add_alias(job_alias)
--- a/returnn/dataset.py
+++ b/returnn/dataset.py
@@ -1,15 +1,14 @@
-__all__ = ["ExtractDatasetMeanStddevJob", "get_returnn_length_hdfs"]
+__all__ = ["ExtractDatasetMeanStddevJob"]
 
 from sisyphus import *
 
 import os
 import shutil
 import subprocess
-from typing import Any, Dict, List, Optional
 
 import numpy
 
-from i6_core.returnn import ReturnnConfig, ReturnnForwardJob, Checkpoint
+from i6_core.returnn import ReturnnConfig
 from i6_core.util import create_executable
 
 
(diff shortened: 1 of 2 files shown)


returnn/forward.py
```diff
@@ -61,6 +62,8 @@ def __init__(
         :param cpu_rqmt: job cpu requirement
         """
         self.returnn_config = returnn_config
+        if model_checkpoint is None:
+            assert not eval_mode, "Eval requires a checkpoint"
```
michelwi, line 66: why? does task "eval" not work with random_model_init?

returnn/dataset.py
```diff
@@ -90,3 +91,64 @@ def run(self):
...
+    """
+    post_config = {"use_tensorflow": True}
+    config = {
+        "eval": dataset_dict,
+        "network": {
+            "output": {
+                "class": "length",
+                "axis": "T",
+                "from": f"data:{dataset_keys[0]}",
+            },
```
michelwi, line 123: do you really need a layer called "output" in the network, or could you just have the for loop run over all dataset_keys?

thinking more of it, this is probably tied to the eval mode not working and the forward mode always dumping the layer called "output". Maybe the returnn option `forward_output_layer` can be helpful?

returnn/forward.py
Atticus1806: eval requires to set an epoch, 0 needs to be set so RETURNN does not want to load a model and then it crashes in `init_network_from_config` line 1114.

returnn/dataset.py
Atticus1806: from my understanding the default output is required in forward, yes. Eval would not require this iI think. 

> forward_output_layer
I think this would make it possible to not have output as a layer in the network but we would still have "output.hdf" I think, so the benefit would be marginal.

Atticus1806:
This ran successfully in my pipeline and is ready for review now and can after review be merged with #297

michelwi (commented):
I accept the changes you did to `ReturnnForwardJob` but to the `get_returnn_length_hdf` function again the comment: is this general enough that everybody will need this or is this better suited in a personal setup/helper function file?

Atticus1806:
This function can be used to e.g. modify RASR alignments to match duration with RETURNN feature extraction (with center=False there might be an offset of 1 in both directions for sequences). As for the other helper this is our use case and the way we implemented it we thought to be general to properly fulfil the extraction of the returnn lengths, since in general you could use this for any data from which you cannot directly infer the final length RETURNN retrieves from the data.

christophmluscher:
I think the same reasoning applies for this PR: https://github.com/rwth-i6/i6_core/pull/294#issuecomment-1206256220

> IMHO if three people are using this then we can add this to i6_core
>
> similar to the *_flow functions
> 
> since this is i6_core and not i6_asr, I would allow for TTS functionality ^^

christophmluscher:
> the way we implemented it we thought to be general to properly fulfil the extraction of the returnn lengths

this should then fulfil the generality request

christophmluscher (approved):

returnn/dataset.py
```diff
@@ -1,14 +1,15 @@
-__all__ = ["ExtractDatasetMeanStddevJob"]
+__all__ = ["ExtractDatasetMeanStddevJob", "get_returnn_length_hdfs"]
```
michelwi, line 1: IIRC, we usually have helper functions in separate files than the job definitions. Would we want to move this to a separate file, e.g. `returnn/helpers.py`?

What is - in general - our take on jobs/helpers/pipeline code in i6_core/i6_experiments?

michelwi (commented):
> I think the same reasoning applies for this PR: https://github.com/rwth-i6/i6_core/pull/294#issuecomment-1206256220

Agreed, but see inline comment.

returnn/dataset.py
christophmluscher: Helper functions which are required by the job itself, are located in the same file: https://github.com/rwth-i6/i6_core/blob/main/features/filterbank.py

Helper functions which ease the usage of a job are in a separate file.

So I think these functions #294 and #297 both fall in the second category..

`i6_core`: jobs
`i6_experiments`: pipeline

helpers: depends?! -> helpers in `i6_core` seem to mainly deal with contruction of flow nodes and rasr configs (more or less). `i6_experiments` more varied. at least this is my feeling..  I am not sure where to exactly draw the line.. 

helpers for a job -> `i6_core`
helpers for a pipeline -> `i6_experiments`
????

michelwi: I have a very similar feeling and therefore my gut feeling would have placed this (#270) somewhere in i6_experiments/common. This is a week preference

#294 I am also ok with having this as an initial function in lib/hdf.py.. I think we already thought about creating such a lib..

I'll leave the final decision to you guys

Atticus1806: So I think lib/hdf.py will get a bit more content from @JackTemaki soon iirc. 
The other helper could ofc. be moved into a different file but I would argue its directly returnn dataset related and right now the file is not really long. Helpers can afterwards once files become too large easily be moved. Creating a helper file for a single function to me feels like making things too complicated (esp. as a general guideline). 
That being said I dont really mind either way, this might just be a case where it in the end doesnt really matter.

JackTemaki: I would say this is (In contrast to #294) very specific to a certain kind of pipeline. Lets change this PR here to only update the `ReturnnForwardJob` code, and put the rest in `i6_experiments/common`. Unfortunately it is still not clear how and where "Returnn" setup related code will be put and structured, but I will try to figure that out as soon as possible.

Atticus1806:
Okay moved the helper to the other PR in experiments (only the change with .config I found sensible enough to stay)

@christophmluscher you were right btw, initially splitting would have made sense :)

So this can then be merged now?

Atticus1806:
ping @christophmluscher @JackTemaki @michelwi :)

returnn/dataset.py
```diff
@@ -8,7 +8,7 @@
 
 import numpy
 
-from i6_core.returnn.config import ReturnnConfig
```
JackTemaki, line 11: Why this change?

returnn/forward.py
```diff
@@ -47,7 +47,8 @@ def __init__(
     ):
         """
 
-        :param model_checkpoint: Checkpoint object pointing to a stored RETURNN Tensorflow model
+        :param model_checkpoint: Checkpoint object pointing to a stored RETURNN Tensorflow model or None if network has
+          no parameters
```
JackTemaki, line 51: `... or should be randomly initialized`

I found out this week there are actually use cases for that... (e.g. profiling)

JackTemaki (changes requested):

returnn/dataset.py
Atticus1806: It got in as a change since we added the helper here, but when reverting the changes I found this to be better since we have the init file, but I can change it back if you want.

michelwi (approved):

JackTemaki (approved):

PR #299, opened 2022-08-01
Make blacklisted vocab consistent

Atticus1806:
Follow up to #295 , old code would just remove entries but not update the vocab order, thus the vocab (while length etc. was correct) would not have continuous values, e.g. {"AA": 1, "BB": 2, "DD": 4}


michelwi (commented):
The changes are good. Is the failing test because the test case was wrong and now you should correct it as well?

Atticus1806:
yep, the for the testcase i imported the vocab and used `del` by hand, but now I manually changed the indices.

michelwi (approved):

christophmluscher (approved):

JackTemaki (approved):

PR #301, opened 2022-08-02
FilterLexicon: keep lemma if synt/tok matches.

sarahberanek:
We don't want to filter out lemmata that match the synt tok from a wordlist/lm-vocabulary list.
Kept this as an option for backward compatibility and set the default to false.

And added doc string information.
Especially the part about where the case-sensitive parameter logic is inverse.

### DIFF
--- a/lexicon/conversion.py
+++ b/lexicon/conversion.py
@@ -42,12 +42,26 @@ class LexiconToWordListJob(Job):
 
 
 class FilterLexiconByWordListJob(Job):
-    def __init__(self, bliss_lexicon, word_list, case_sensitive=False):
+    """
+    Filter lemmata to given word list.
+    Warning: case_sensitive parameter does the opposite. Kept for backwards-compatibility.
+    """
+
+    def __init__(
+        self, bliss_lexicon, word_list, case_sensitive=False, check_synt_tok=False
+    ):
+        """
+        :param tk.Path bliss_lexicon: lexicon file to be handeled
+        :param tk.Path word_list: filter lexicon by this word list
+        :param bool case_sensitive: filter lemmata case-sensitive. Warning: parameter does the opposite.
+        :param bool check_synt_tok: keep also lemmata where the syntactic token matches word_list
+        """
         self.set_vis_name("Filter Lexicon by Word List")
 
         self.bliss_lexicon = bliss_lexicon
         self.word_list = word_list
         self.case_sensitive = case_sensitive
+        self.check_synt_tok = check_synt_tok
 
         self.out_bliss_lexicon = self.output_path(
             os.path.basename(tk.uncached_path(bliss_lexicon)), cached=True
@@ -72,6 +86,10 @@ class FilterLexiconByWordListJob(Job):
                 transform(orth.text) in words
                 or "special" in lemma.attrib
                 or (orth.text is not None and orth.text.startswith("["))
+                or (
+                    self.check_synt_tok
+                    and transform(lemma.find("./synt/tok").text) in words
+                )
                 for orth in lemma.findall("orth")
             ):
                 root.append(lemma)
--- a/lexicon/conversion.py
+++ b/lexicon/conversion.py
@@ -47,6 +47,8 @@ class FilterLexiconByWordListJob(Job):
     Warning: case_sensitive parameter does the opposite. Kept for backwards-compatibility.
     """
 
+    __sis_hash_exclude__ = {"check_synt_tok": False}
+
     def __init__(
         self, bliss_lexicon, word_list, case_sensitive=False, check_synt_tok=False
     ):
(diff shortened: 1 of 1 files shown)


lexicon/conversion.py
```diff
@@ -72,6 +86,10 @@ def run(self):
                 transform(orth.text) in words
                 or "special" in lemma.attrib
                 or (orth.text is not None and orth.text.startswith("["))
+                or (
+                    self.check_synt_tok
+                    and transform(lemma.find("./synt/tok").text) in words
```
michelwi, line 91: What if there are multiple tokens in the synt sequence, e.g.
```
  <lemma>
    <orth>AppTec</orth>
    <phon>æ p t ɛ k</phon>
    <synt>
        <tok> tech </tok>
        <tok> company </tok>
    </synt>
  </lemma>
```
I think in this case this line will only check the first token and not all of them. Maybe use
```suggestion
                    and all([transform(tok.text) in words for tok in lemma.findall("./synt/tok")])
```
so that the word is kept if all tokens are in the list.

lexicon/conversion.py
```diff
@@ -72,6 +86,10 @@ def run(self):
...
+                )
                 for orth in lemma.findall("orth")
             ):
```
michelwi, line 94: This code lives in an tuple comprehension that runs over all orth of the lemma i.e. is executed multiple times if there are multiple lemmata. A better place would be outside of this comprehension, i.e.

```suggestion
                transform(orth.text) in words
                or "special" in lemma.attrib
                or (orth.text is not None and orth.text.startswith("["))
                for orth in lemma.findall("orth")
            ) or (
                self.check_synt_tok
                and transform(lemma.find("./synt/tok").text) in words
            ):
```
((please test and black this handwritten suggestion 🙃 ))

at this point it might be easier to understand if we defined a function for all these conditions 🙈

lexicon/conversion.py
```diff
@@ -42,12 +42,26 @@ def run(self):
 
 
 class FilterLexiconByWordListJob(Job):
```
michelwi, line 44: Adding a new parameter to `__init__` changes the hashes of all existing jobs, even if the old behavior is the same. To prevent this, there is an option called `__sis_hash_exclude__`

```suggestion
class FilterLexiconByWordListJob(Job):
    __sis_hash_exclude__ = {"check_synt_tok" : False}
```

I request changes w.r.t. this comment.

lexicon/conversion.py
```diff
@@ -42,12 +42,26 @@ def run(self):
...
+
+    def __init__(
+        self, bliss_lexicon, word_list, case_sensitive=False, check_synt_tok=False
+    ):
+        """
+        :param tk.Path bliss_lexicon: lexicon file to be handeled
+        :param tk.Path word_list: filter lexicon by this word list
+        :param bool case_sensitive: filter lemmata case-sensitive. Warning: parameter does the opposite.
+        :param bool check_synt_tok: keep also lemmata where the syntactic token matches word_list
+        """
```
michelwi, line 58: Good job adding documentation, thanks! 👍

michelwi (changes requested):

lexicon/conversion.py
sarahberanek: Good point. I actually thought that there can only be one <tok> per lemma.

sarahberanek: Isnt it enough if ONE tok is in the list to have the lemma used in recognition ?

sarahberanek: eg exchange any with all? How is the calculation of lm probability done if there are multiple tokens in a lemma? (Not that I have seen it IRL though :sunglasses: )

michelwi: > Isnt it enough if ONE tok is in the list to have the lemma used in recognition ?

this depends on what the use case is. In case you want to filter out all words that are not in the LM, then I think all words should be there, because

> How is the calculation of lm probability done if there are multiple tokens in a lemma?

this is the product of the probabilities (or rather sum of log_prob) `p(tok1 | history) *  p(tok2 | history tok1) * ...`

sarahberanek: my question is if it would crash or become 0? log_prob is probably okay.
And if in this case we would like to remove the tok not on the lm and not the whole lemma, which would be happening now.

michelwi: > my question is if it would crash or become 0? log_prob is probably okay.

if e.g. tok2 didn't exist in the LM, then `p(tok2 | history tok1)` would be replaced with `p(<unk>| history tok1)`. If there is no unknown in LM, then the probability will probably be 0, i.e. log_prob = some_large_float. 

>  if in this case we would like to remove the tok not on the lm and not the whole lemma,

I don't have a strong argument here.. my gut feeling would be to remove it entirely and not to modify the lemma..

sarahberanek: Yes, i think I agree. Lets do that. As I said: I Have never seen a lemma with multiple tokens ever. :-)

michelwi (approved):

christophmluscher (approved):

PR #304, opened 2022-08-15
BlissChangeEncodingJob: allow no recover duration

vieting:
When I recently used this job to change the encoding for our switchboard corpus using the `BlissChangeEncodingJob`, it didn't work because `recover_duration` was automatically set to `True`. However, this requires that each recording in the corpus contains only a single segment which is not the case for switchboard. When checking selected converted files, the duration was exactly the same. I therefore propose to add an option to this job which allows disabling the duration recovering manually. A comment is added so users know about the risk of doing this.

### DIFF
--- a/audio/encoding.py
+++ b/audio/encoding.py
@@ -10,6 +10,8 @@ class BlissChangeEncodingJob(BlissFfmpegJob):
     Uses ffmpeg to convert all audio files of a bliss corpus (file format, encoding, channel layout)
     """
 
+    __sis_hash_exclude__ = {"recover_duration": None}
+
     def __init__(
         self,
         corpus_file,
@@ -20,6 +22,7 @@ class BlissChangeEncodingJob(BlissFfmpegJob):
         fixed_bitrate=None,
         force_num_channels=None,
         select_channels=None,
+        recover_duration=None,
         ffmpeg_binary=None,
         hash_binary=False,
     ):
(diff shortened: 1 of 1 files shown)


audio/encoding.py
```diff
@@ -20,6 +22,7 @@ def __init__(
         fixed_bitrate=None,
         force_num_channels=None,
         select_channels=None,
+        recover_duration=None,
```
michelwi, line 25: maybe add at the end of the argument list so that old setups that give positional args don't break

audio/encoding.py
```diff
@@ -38,6 +41,9 @@ def __init__(
         :param tuple(str)|None select_channels: tuple of (channel_layout, channel_name), see `ffmpeg -layouts`
             this is useful if the new encoding might have an effect on the duration, or if no duration was specified
             in the source corpus
+        :param bool|None recover_duration: This will open all files with "soundfile" and extract the length information.
+            There might be minimal differences when converting the encoding, so only set this to `False` if you're
+            willing to accept this risk.
```
michelwi, line 46: maybe add to the docstring explanation what happens if `None` is given (I can read the code but I don't understand why it is chosen that way)

> For all parameter holds that "None" means to use the ffmpeg defaults, which depend on the input file and the output format specified.

or is this also the case here?

JackTemaki (approved):

audio/encoding.py
vieting: `None` does not refer to the `ffmpeg` default but the old job default. This is not about an `ffmpeg` option after all. I added a sentence on this, please see if that is sufficient from your pov.

michelwi (approved):

vieting:
Thanks for the quick reviews! @michelwi @JackTemaki

PR #307, opened 2022-08-31
add tf-flow functions

JackTemaki:
As those will be used by many (Hybrid) Setups I think it makes sense to have this available in `i6_core` for everyone. There should be no other personal customization of those function needed. If there is an objection to this, we can also put the code under `i6_experiments/common/`, but as we already have all other flows in core I thought it just makes sense to also put this here.

### DIFF
--- a/returnn/flow.py
+++ b/returnn/flow.py
@@ -85,7 +85,9 @@ def make_precomputed_hybrid_tf_feature_flow(
     tf_flow.config[tf_fwd].loader.saved_model_file = tf_checkpoint
     if native_ops is not None:
         if isinstance(native_ops, list):
-            tf_flow.config[tf_fwd].loader.required_libraries = DelayedJoin(native_ops, ";")
+            tf_flow.config[tf_fwd].loader.required_libraries = DelayedJoin(
+                native_ops, ";"
+            )
         else:
             tf_flow.config[tf_fwd].loader.required_libraries = native_ops
 
--- a/returnn/flow.py
+++ b/returnn/flow.py
@@ -20,6 +20,7 @@ def make_precomputed_hybrid_tf_feature_flow(
 
     The resulting flow is a trivial:
 
+        <link from="<tf_fwd_input_name>" to="tf-fwd:input"/>
         <node name="tf-fwd" id="$(id)" filter="tensorflow-forward"/>
         <link from="tf-fwd:log-posteriors" to="network:features"/>
 
@@ -47,7 +48,7 @@ def make_precomputed_hybrid_tf_feature_flow(
     :param feature_tensor_name: name of the extern data ebtry to feed the features to
     :param output_tensor_name: the name of the output tensor, it is expected that "<name>/output_batch_major" exists
     :param native_ops: list of native op ".so" files to link
-    :param tf_fwd_input_name: Naming for the tf network input, usually no need to be changed
+    :param tf_fwd_input_name: naming for the tf network input, usually no need to be changed
     :return: tensorflow-forward node flow with output link and related config
     """
 
@@ -91,9 +92,9 @@ def add_tf_flow_to_base_flow(
     base_flow: rasr.FlowNetwork,
     tf_flow: rasr.FlowNetwork,
     tf_fwd_input_name: str = "tf-fwd-input",
-):
+) -> rasr.FlowNetwork:
     """
-    Integrate tf-fwd node into the regular flow network, passing the features into the input of the tf-flow net
+    Integrate tf-fwd node into the regular flow network, passing the features into the input of the tf-flow net.
 
     :param FlowNetwork base_flow:
     :param FlowNetwork tf_flow:
--- a/returnn/flow.py
+++ b/returnn/flow.py
@@ -9,7 +9,7 @@ from i6_core.returnn.training import Checkpoint
 def make_precomputed_hybrid_tf_feature_flow(
     tf_graph: tk.Path,
     tf_checkpoint: Checkpoint,
-    feature_tensor_name: str = "data",
+    extern_data_name: str = "data",
     output_layer_name: str = "output",
     native_ops: Optional[Union[tk.Path, List[tk.Path]]] = None,
     tf_fwd_input_name: str = "tf-fwd-input",
@@ -44,8 +44,8 @@ def make_precomputed_hybrid_tf_feature_flow(
 
     :param tf_graph: usually the output of a CompileTFGraphJob
     :param tf_checkpoint: the checkpoint to load the model from, e.g. from a ReturnnTrainingJob or similar
-    :param feature_tensor_name: name of the extern data entry to feed the features to
-    :param output_layer_name: the name of the output layer, it is expected that 
+    :param extern_data_name: name of the extern data entry to feed the features to
+    :param output_layer_name: the name of the output layer, it is expected that
         "<name>/output_batch_major" exists and returns log-probs.
     :param native_ops: list of native op ".so" files to link
     :param tf_fwd_input_name: naming for the tf network input, usually no need to be changed
@@ -67,11 +67,10 @@ def make_precomputed_hybrid_tf_feature_flow(
     tf_flow.config[
         tf_fwd
     ].input_map.info_0.tensor_name = (
-        f"extern_data/placeholders/{feature_tensor_name}/{feature_tensor_name}"
+        f"extern_data/placeholders/{extern_data_name}/{extern_data_name}"
     )
     tf_flow.config[tf_fwd].input_map.info_0.seq_length_tensor_name = (
-        f"extern_data/placeholders/"
-        f"{feature_tensor_name}/{feature_tensor_name}_dim0_size"
+        f"extern_data/placeholders/" f"{extern_data_name}/{extern_data_name}_dim0_size"
     )
 
     tf_flow.config[tf_fwd].output_map.info_0.param_name = "log-posteriors"
--- a/returnn/flow.py
+++ b/returnn/flow.py
@@ -1,5 +1,6 @@
 from sisyphus import tk
-from typing import Optional, List
+from sisyphus.delayed_ops import DelayedJoin
+from typing import Optional, List, Union
 
 from i6_core import rasr
 from i6_core.returnn.training import Checkpoint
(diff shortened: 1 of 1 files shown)


returnn/flow.py
```diff
@@ -0,0 +1,122 @@
...
+    tf_flow.config[tf_fwd].output_map.info_0.param_name = output_type
+    tf_flow.config[
+        tf_fwd
+    ].output_map.info_0.tensor_name = f"{output_tensor_name}/output_batch_major"
+
+    tf_flow.config[tf_fwd].loader.type = "meta"
+    tf_flow.config[tf_fwd].loader.meta_graph_file = tf_graph
+    tf_flow.config[tf_fwd].loader.saved_model_file = tf_checkpoint
+    if native_ops is not None:
+        tf_flow.config[tf_fwd].loader.required_libraries = native_ops
```
vieting, line 85: Does this work with a list? Right now I only see one case in the recipes where a list of ops is used and there, `":".join()` is applied to it before passing it.

vieting (commented):
Looks good to me, it is basically a cleaned up version of what I've been using quite a while I guess. Thanks for adding it here, makes sense to me.

I added mostly nitpicks. If you confirm that assigning a list to the `RasrConfig` works, the PR is fine for me.

vieting:
Yes, I saw the same. Thanks for checking!

But maybe if the current way doesn't work, that would be an issue of the `RasrConfig` rather than for this PR.

michelwi:
> > I added mostly nitpicks. If you confirm that assigning a list to the `RasrConfig` works, the PR is fine for me.
> 
> I am not sure, I already have seen different styles (e.g. comma separated, ":" separated), but the native code is actually just space separated:


According to the implementation of [`ParameterStringVector`](https://github.com/rwth-i6/rasr/blob/ef23a6aa03d5a6d816ae735ead43baf5d9c0548c/src/Core/Parameter.hh#L377), the separation delimiter is configurable and space-separated is the default (default delimiter is `""`, which is mapped to `" "`). 

For the parameter [`required-libraries`](https://github.com/rwth-i6/rasr/blob/4fe63be7ce49ccdaa9d60d6d7ecaeeb7af3abec1/src/Tensorflow/GraphLoader.cc#L19-L22) the separator symbol is `:`.

I have also seen comma-separated lists, e.g. as [action sequence](https://github.com/rwth-i6/rasr/blob/4fe63be7ce49ccdaa9d60d6d7ecaeeb7af3abec1/src/Tools/LatticeProcessor/LatticeProcessor.cc#L92-L95) for the lattice processing tool.

returnn/flow.py
michelwi: no, this will not work as the default stringification is space-separated-list but required_libraries requires colon-separated-list

christophmluscher:
> > > I added mostly nitpicks. If you confirm that assigning a list to the RasrConfig works, the PR is fine for me.
> >
> > I am not sure, I already have seen different styles (e.g. comma separated, ":" separated), but the native code is actually just space separated:
> 
> According to the implementation of [ParameterStringVector](https://github.com/rwth-i6/rasr/blob/ef23a6aa03d5a6d816ae735ead43baf5d9c0548c/src/Core/Parameter.hh#L377), the separation delimiter is configurable and space-separated is the default (default delimiter is "", which is mapped to " ").
> 
> For the parameter [required-libraries](https://github.com/rwth-i6/rasr/blob/4fe63be7ce49ccdaa9d60d6d7ecaeeb7af3abec1/src/Tensorflow/GraphLoader.cc#L19-L22) the separator symbol is :.
> 
> I have also seen comma-separated lists, e.g. as [action sequence](https://github.com/rwth-i6/rasr/blob/4fe63be7ce49ccdaa9d60d6d7ecaeeb7af3abec1/src/Tools/LatticeProcessor/LatticeProcessor.cc#L92-L95) for the lattice processing tool.

yes. @michelwi  is right. I remembered incorrectly. sry.

returnn/flow.py
```diff
@@ -0,0 +1,123 @@
...
```
christophmluscher, line 86:
```suggestion
        tf_flow.config[tf_fwd].loader.required_libraries = ":".join(native_ops) if isinstance(native_ops, list) else native_ops
```

michelwi: if `native_ops` is a list of `tk.Path` objects (e.g. from a `CompileNativeOpJob`), would then doing `str.join` on them convert to `str` and finally lead to hashing the absolute path?

christophmluscher: Hm good point.. does it make sense to solve this here? I think Nick mentioned that in RasrConfig might be better?
Or some delayed ops?

returnn/flow.py
```diff
@@ -0,0 +1,123 @@
...
+from i6_core.returnn.training import Checkpoint
+
+
+def make_precomputed_hybrid_tf_feature_flow(
+    tf_graph: tk.Path,
+    tf_checkpoint: Checkpoint,
+    output_type: str = "log-posteriors",
+    feature_tensor_name="data",
+    output_tensor_name: str = "output",
+    native_ops: Optional[List] = None,
```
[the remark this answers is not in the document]
JackTemaki: I would keep the type simple, if you only want one op then just use a list of length one.

JackTemaki:
mmm somehow the diff does not look correct, let me try to fix that.

returnn/flow.py
```diff
@@ -0,0 +1,123 @@
...
+        type               = meta
+
+        [flf-lattice-tool.network.recognizer.feature-extraction.tf-fwd.output-map.info-0]
+        param-name  = <output_type>
+        tensor-name = <output_tensor_name>/output_batch_major
+
+
+    :param tf_graph: usually the output of a CompileTFGraphJob
+    :param tf_checkpoint: the checkpoint to load the model from, e.g. a ReturnnTrainingJob or similar
+    :param output_type: type of the returned values, see RASR documentation for possible types
```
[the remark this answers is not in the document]
JackTemaki: If someone can add this this would be good, I do not know myself what there can be...

returnn/flow.py
```diff
@@ -0,0 +1,129 @@
...
+    :param feature_tensor_name: name of the extern data entry to feed the features to
+    :param output_tensor_name: the name of the output tensor, it is expected that "<name>/output_batch_major" exists
```
albertz, line 50: I would either:

- Keep the `output_tensor_name` param, but do not automatically add the `/output_batch_major`, but let the user add this explicitly. Then it is more generic. Further it should be documented here that RASR expects the shape [B,T,D].
- Or, rename the param, maybe to `output_layer_name`, to make it more clear that this is not the tensor name.

returnn/flow.py
albertz: `DelayedFormat` would also have worked, but maybe `DelayedJoin` is easier.

returnn/flow.py
JackTemaki: I would prefer the second approach (tensor is indeed a wrong name), as this function is not intended to be generic but very specific about the `precomputed-hybrid` module of RASR.
@christophmluscher what is your opinion on this?

returnn/flow.py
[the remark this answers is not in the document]
michelwi: I checked the RASR code and did not find any mention of an option that can take on the value `"log-posteriors"`. 

Pondering it further, I come to think that this is just some arbitrary name that mapps the flow-output-name to the specific tensor. i.e. we could also put the string "foobar" there and then the `tf-fwd` node had an output `foobar` that would yield the values from `<output_tensor_name>/output_batch_major`

```
        <link from="<tf_fwd_input_name>" to="tf-fwd:input"/>
        <node name="tf-fwd" id="$(id)" filter="tensorflow-forward"/>
        <link from="tf-fwd:foobar" to="network:features"/>
```
and
```
        [flf-lattice-tool.network.recognizer.feature-extraction.tf-fwd.output-map.info-0]
        param-name  = foobar
        tensor-name = my_output_layer_name/output_batch_major
```

Therefore we can safely hard-code the value to anything (e.g. `"log-posteriors"`) and remove the option `output_type` altogether, or 
```suggestion
    :param flow_output_name: name that will be used for the tf-fwd flow node output channel
```

returnn/flow.py
```diff
@@ -0,0 +1,129 @@
...
+        saved-model-file   = <tf_checkpoint>
+        type               = meta
+
+        [flf-lattice-tool.network.recognizer.feature-extraction.tf-fwd.output-map.info-0]
+        param-name  = <output_type>
+        tensor-name = <output_tensor_name>/output_batch_major
+
+
+    :param tf_graph: usually the output of a CompileTFGraphJob
+    :param tf_checkpoint: the checkpoint to load the model from, e.g. a ReturnnTrainingJob or similar
```
michelwi, line 47: Typo
```suggestion
    :param tf_checkpoint: the checkpoint to load the model from, e.g. from a ReturnnTrainingJob or similar
```

returnn/flow.py
michelwi: I, too, prefer the second approach.

returnn/flow.py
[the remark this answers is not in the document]
JackTemaki: So... you are telling me this parameter is fully cosmetic and never played any role, and RASR anyway does only support log-posteriors with NN-Precomputed-Hybrid?

returnn/flow.py
```diff
@@ -0,0 +1,128 @@
...
+from typing import Optional, List, Union
+
+from i6_core import rasr
+from i6_core.returnn.training import Checkpoint
+
+
+def make_precomputed_hybrid_tf_feature_flow(
+    tf_graph: tk.Path,
+    tf_checkpoint: Checkpoint,
+    feature_tensor_name: str = "data",
```
albertz, line 12: I think I would also rename this now, for consistency, and again correctness.
Maybe `extern_data_name`?

returnn/flow.py
```diff
@@ -0,0 +1,128 @@
...
+        [flf-lattice-tool.network.recognizer.feature-extraction.tf-fwd.output-map.info-0]
+        param-name  = <output_type>
+        tensor-name = <output_tensor_name>/output_batch_major
+
+
+    :param tf_graph: usually the output of a CompileTFGraphJob
+    :param tf_checkpoint: the checkpoint to load the model from, e.g. from a ReturnnTrainingJob or similar
+    :param feature_tensor_name: name of the extern data entry to feed the features to
+    :param output_layer_name: the name of the output layer containg log-probs,
+        it is expected that "<name>/output_batch_major" exists
```
michelwi, line 49:
```suggestion
    :param output_layer_name: the name of the output layer, it is expected that 
        "<name>/output_batch_major" exists and returns log-probs.
```
just a suggestion for better wording imho.

albertz (approved):

PR #308, opened 2022-09-05
Make rasr.FlowNetwork serialization deterministic.

curufinwe:
Serialization at the moment can differ between different runs of the python interpreter. This is somewhat inconvenient for AppTek as we have a workflow that packages models into a tar-ball and stores them under the hash of that archive. Having different hashes between different runs can result in different tar-balls for equivalent models.

### DIFF
--- a/rasr/flow.py
+++ b/rasr/flow.py
@@ -358,16 +358,15 @@ class FlowNetwork:
 
         for name in self.__compute_node_order():
             attr = self.nodes[name]
-            nstr = {"name": name}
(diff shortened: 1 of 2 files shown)


Marvin84 (approved):

michelwi (approved):

SimBe195 (approved):

albertz:
I assume this was only a problem with very old Python versions, right (Python <=3.5)? Since Python 3.6, it should be deterministic.

PR #312, opened 2022-09-22
CompileTFGraph add config as output

JackTemaki:
This is good for checking what you actually compiled, which is not visible otherwise if you use work folder cleaning.

### DIFF
--- a/returnn/compile.py
+++ b/returnn/compile.py
@@ -91,6 +91,7 @@ class CompileTFGraphJob(Job):
 
         else:
             returnn_config_path = self.returnn_config
+            shutil.copy(self.returnn_config, self.out_returnn_config.get_path())
 
         args = [
             tk.uncached_path(self.returnn_python_exe),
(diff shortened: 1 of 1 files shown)


Atticus1806 (approved):

returnn/compile.py
```diff
@@ -82,9 +83,10 @@ def tasks(self):
...
         if isinstance(self.returnn_config, tk.Path):
             returnn_config_path = self.returnn_config.get_path()
+            shutil.copy(returnn_config_path, self.out_returnn_config.get_path())
 
         elif isinstance(self.returnn_config, ReturnnConfig):
-            returnn_config_path = "returnn.config"
+            returnn_config_path = self.out_returnn_config.get_path()
             self.returnn_config.write(returnn_config_path)
 
         else:
```
michelwi, line 92: the else case would be `self.returnn_config` is a string. I think we also need a copy for this case (similar to line 86)

michelwi: Missing is that the file located at the "path as string" be copied to `self.out_returnn_config.get_path()`.

I mean independent of whether `self.returnn_config` is a `tk.Path` object or a `str` representing a path, the same action should be taken. And for the `tk.Path` case you do a copy so I thought in the `str` case a copy is missing.

JackTemaki: Sorry I looked at wrong code, will fix it now

michelwi (approved):

PR #313, opened 2022-09-22
add SpeakerLabelHDFFromBlissJob

JackTemaki:
I had a local copy of the `SimpleHDFWriter` before, but this is really not clean. So now I decided to just import it from RETURNN, either directly if installed in the env or via a root path that is linked dynamically.

### DIFF
--- a/returnn/dataset.py
+++ b/returnn/dataset.py
@@ -3,13 +3,16 @@ __all__ = ["ExtractDatasetMeanStddevJob"]
 from sisyphus import *
 
 import os
+import pickle
 import shutil
 import subprocess
 
 import numpy
 
 from i6_core.returnn.config import ReturnnConfig
-from i6_core.util import create_executable
+from i6_core.lib import corpus
+from i6_core.lib.hdf import get_returnn_simple_hdf_writer
+from i6_core.util import create_executable, uopen
 
 
 class ExtractDatasetMeanStddevJob(Job):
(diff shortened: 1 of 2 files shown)


Atticus1806 (approved):

returnn/dataset.py
```diff
@@ -90,3 +93,56 @@ def run(self):
...
+        )
+        hdf_writer = SimpleHDFWriter(
+            self.out_speaker_hdf.get_path(), dim=num_speakers, ndim=1
+        )
+
+        for recording in bliss.all_recordings():
+            for segment in recording.segments:
+                speaker_name = segment.speaker_name or recording.speaker_name
+                speaker_index = index_by_speaker[speaker_name]
+                segment_name = "/".join([bliss.name, recording.name, segment.name])
```
michelwi, line 143: what if there are sub-corpora? Better use the existing fullname function

```suggestion
                segment_name = segment.fullname()
```

returnn/dataset.py
```diff
@@ -90,3 +93,56 @@ def run(self):
...
             self.out_std_dev.set(numpy.sqrt(total_var))
+
+
+class SpeakerLabelHDFFromBlissJob(Job):
+    """
+    Extract speakers from a bliss corpus and create an HDF file with a speaker index
+    matching the speaker entry in the corpus speakers for each segment
+    """
+
+    def __init__(self, bliss_corpus, returnn_root=None):
```
christophmluscher, line 104: typing

returnn/dataset.py
```diff
@@ -90,3 +93,56 @@ def run(self):
...
+        """
+        :param bliss_corpus: bliss XML corpus where the speakers and segments are taken from
+        :param returnn_root: used to import SimpleHDFWriter from a specific path if not installed in the worker env
+        """
+        self.bliss_corpus = bliss_corpus
+        self.returnn_root = returnn_root
+        self.out_speaker_hdf = self.output_path("speaker_labels.hdf")
```
christophmluscher, line 111: add: `cached=True`?

returnn/dataset.py
```diff
@@ -90,3 +93,56 @@ def run(self):
...
+        self.out_num_speakers = self.output_var("num_speakers")
+        self.out_speaker_dict = self.output_path("speaker_dict.pkl")
```
christophmluscher, line 113: add: `cached=True`?

christophmluscher (commented):
more general question maybe: should all jobs use a temp file/dir?? should sisyphus handle this?

returnn/dataset.py
JackTemaki: As we discussed somewhere else the Sisyphus internal caching only works for rasr related things, and those here are not. So I think there is no need. HDF caching will be done by RETURNN anyways.

returnn/dataset.py
JackTemaki: Ah yes sure.

returnn/dataset.py
JackTemaki: good catch

JackTemaki:
> more general question maybe: should all jobs use a temp file/dir?? should sisyphus handle this?

In this case it is writing so fast that it makes no difference if we write to the NFS or first to a tempdir and then copy, but yes we can discuss this. I would not let sisyphus handle this per default, because then you have no intermediate work files if there is a crash.

michelwi (approved):

christophmluscher:
> > more general question maybe: should all jobs use a temp file/dir?? should sisyphus handle this?
> 
> In this case it is writing so fast that **_I think_** it makes no difference if we write to the NFS or first to a tempdir and then copy, but yes we can discuss this. I would not let sisyphus handle this per default, because then you have no intermediate work files if there is a crash.

I agree that this is a more general discussion

christophmluscher (approved):

PR #314, opened 2022-09-22
Reports

Atticus1806:
WIP PR to enable (mailed) reports for experiments similar to what MT already has/had. 

Even though this is still in development feel free to comment on stuff.


Atticus1806:
So I had some more looks at the old code and I think with proper input and some typing this looks fine.
Also added an example report, but this will most likely be in some private user folder but for now if someone wants to test it.

I will most likely do some more changes on monday and then I will mark this as ready I think, at least for a 1.0.

Atticus1806:
Okay lets for now start discussing details on this via review. 
Open question:
Should the example report format stay there? Should it be moved somewhere? Is this entirely user dependent (so in i6_experiments/user/...)?

report/report.py
```diff
@@ -0,0 +1,213 @@
...
+class ReportResultsJob(Job):
+    """
+    Job to report results either via mail and/or just condensed in a file
+    """
+
+    def __init__(
+        self,
+        name: str,
+        report: _Report_Type,
+        report_dir: tk.Path = None,
```
christophmluscher, line 23: `Optional[tk.Path]`?

report/report.py
```diff
@@ -0,0 +1,213 @@
...
+        report_format: Callable[[_Report_Type], str] = None,
```
christophmluscher, line 24: `Optional[tk.Path]`?

report/report.py
```diff
@@ -0,0 +1,213 @@
...
+import sys
+import os
+import getpass
+import pprint
+import gzip
+from typing import Dict, Union, Callable
+import time
+import json
+
+_Report_Type = Dict[str, Union[tk.Path, any]]
```
christophmluscher, line 11: `Any`?

and import from typing? same as `Dict`?!

report/report.py
```diff
@@ -0,0 +1,213 @@
...
+        report["config_path"] = self.config_path
+        report["config"] = self.config
+        report["settings"] = self.settings
+        report["sis_command_line"] = self.sis_command_line
+        report["cwd"] = self.cwd
+
+        if self.report_format is None:
+            self.report_format = pprint.pformat
+
+        with gzip.open(str(self.out_report), "w") as f:
```
christophmluscher, line 90: `self.out_report.get_path()`

report/report.py
```diff
@@ -0,0 +1,213 @@
...
+        :param (report: dict) -> List[str] report_format:
+        :param recipe_path:
+        """
+        self.name = name
+        self.report = report
+        self.report_dir = report_dir
+        self.report_format = report_format
+        self.recipe_path = recipe_path
+
+        self.out_report = self.output_path("report.gz")
```
christophmluscher, line 41: idea: maybe the job could have a flag to either compress the report file or not..

report/report.py
```diff
@@ -0,0 +1,213 @@
...
+                "zcat {out_report} | mail -s 'Report finished: {name} {config_path}' {mail_address}"
+            )
+
+    def tasks(self):
+        yield Task("run", mini_task=True)
+
+
+def gmm_example_report_format(report: Dict[str, tk.Variable]) -> str:
+    """
+    Example report format for a GMM evaluated on dev-clean and dev-other
```
christophmluscher, line 116:
```suggestion
    Example report format for a GMM evaluated on dev-clean and dev-other of the LibrSpeech dataset
```

michelwi (changes requested):
I am a bit unsure about the tradeoff generalizability vs. nice reports you are doing here.

In my expectation, a Job like this should have two purposes
a) notify me via mail that a specific output is finished so that I can take further action.
b) create a nice table that collects the results

We could maybe even split the functionality: one job that creates a Report and a second job that sends mail notifications (and optionally attaches the output)

To create the report we should discuss the interface.. I don't like the current one with a dict that contains many mandatory keys that are ignored later for printing.
Do we want to hardcode a 2D table? Or a nD table with 0<=n<=2?
Then have 0, 1, or 2 vectors with table headings and an Array with values?
Maybe even support for automatic ascii table formatting using |---+---| etc..?

report/report.py
```diff
@@ -0,0 +1,51 @@
...
+        report = {
+            "output": self.output.get_path(),
+            "user": user,
+            "name": self.name,
+            "sis_command_line": self.sis_command_line,
+            "cwd": self.cwd,
+        }
+
+        report = eval(repr(report))
+        pprint.pprint(report)
```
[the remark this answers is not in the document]
Atticus1806: I think this was also outdated from an earlier draft right?

report/report.py
```diff
@@ -0,0 +1,213 @@
...
+        if self.report_dir:
+            report_file_name = "%s.%s.%s.gz" % (
+                user,
+                self.config_path.replace("/", "_"),
+                self.name,
+            )
+            report_path = os.path.join(self.report_dir, report_file_name)
+            print("Write report to %s" % report_path)
+            with gzip.open(report_path, "w") as f:
+                f.write(json.dumps(report).encode() + b"\n")
```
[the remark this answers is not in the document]
Atticus1806: So this should be removed? I agree looking at it again this looks weird.

report/report.py
```diff
@@ -0,0 +1,213 @@
...
+        if self.mail_address:
+            self.sh(
+                "zcat {out_report} | mail -s 'Report finished: {name} {config_path}' {mail_address}"
+            )
+
+    def tasks(self):
+        yield Task("run", mini_task=True)
+
+
+def gmm_example_report_format(report: Dict[str, tk.Variable]) -> str:
```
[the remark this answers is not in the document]
Atticus1806: Okay. Would you want this to be user specific or somewhere maybe in `common/baselines/librispeech`?

report/report.py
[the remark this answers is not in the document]
Atticus1806: Yeah this is where things (might) get ugly. In the end the way the job was structured from where I copied it it really depends on what your `report_format` function is doing. The usual input would maybe be something like `Union[tk.Variable, tk.Path, str]` or similar, but you could think of arbitrary things here. 
Actually this is where also other information like the keys come in, if the report makes use of them they are required.
The old jobs had a bunch of different formats and even more than one ReportJob (thats why it was named simple beforehand). The more I think about it maybe I will just remove a lot of stuff I deem irrelevant and then make more restrictions on the format of the report_format.

>Then if Variable is used, I don't see any more where get() is called.

This cost me like 2 hours and its hidden in the call of `str()` and `repr()` that are done within the reports (or by pprint).
https://github.com/rwth-i6/sisyphus/issues/104 this might even break things. But this just as a sidenote.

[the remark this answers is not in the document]
Atticus1806: But I think this is something that the report_format should handle, since this might be really user specific what the input is. So we could completely split it and say the `ReportJob` only gets a string it mails (then its just `MailJob`?) but I like that global information like the `config_path` and `sis_command_line` are actually set by the job since otherwise thats something the user would always have to deal with. And I am not sure if passing a string would require a job to produce this list since otherwise I think we would just end up with a lot of "UnifinishedVariable" in the report.

report/report.py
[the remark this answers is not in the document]
michelwi: This specific function gives a report for the GMMSystem class so I would put it close to that. But we can discuss this.

report/report.py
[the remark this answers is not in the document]
michelwi: yes I would say these lines should be removed. The report should only ever be written to the output folder.

Atticus1806:
If I didn't miss anything all comments should be addressed in one way or the other. I again changed the structure of the job a bit more liberal this time to make it more simple. 
The main idea now is: 
There are a few "basic" informations like date, user, name etc. that are set by the job in addition to what the user gives into the job. The report format is now mandatory and should be able to handle str and tk.AbstractPath inputs. Any kwargs can be passed by job kwargs. 
Again this is a more radical change so feel free to discuss changes. 
The example will be moved/removed to i6_experiments once this PR is "done" as a last step, this is just to keep an example (and potential changes to it after changes in the job) easily accessible for now.

report/report.py
```diff
@@ -0,0 +1,178 @@
...
+        :param compress: Whether to compress the report or not
+        """
+        self.name = name
+        self.report = report
+        self.report_format = report_format
+        self.report_format_kwargs = report_format_kwargs
+        self.compress = compress
+
+        self.sis_command_line = sys.argv
+        self.mail_address = getattr(gs, "MAIL_ADDRESS", None)
```
michelwi, line 40: I didn't find any `MAIL_ADDRESS` in `sisyphus/global_settings.py`, is this defined somewhere else or did you make it up?

We want to get away from defining recipe-specific settings in the sisyphus settings. especially if they change how a job behaves. So maybe make the mail address an input to the job?

Actually, I grow more and more fond of the Idea of a dedicated `MailJob`.

report/report.py
```diff
@@ -0,0 +1,178 @@
...
+
+        if self.compress:
+            with gzip.open(self.out_report.get_path(), "w") as f:
+                f.write(
+                    self.report_format(report, **self.report_format_kwargs).encode()
+                    + b"\n"
+                )
+        else:
+            with open(self.out_report.get_path(), "w") as f:
+                f.write(self.report_format(report, **self.report_format_kwargs) + "\n")
```
michelwi, line 67: Is it intentional that one branch encodes the string and the other doesn't? Is it required by `gzip.open` vs `open`?
Or is it a problem of specifying mode "wt" vs "wb" explicitly?

michelwi (commented):
> The example will be moved/removed to i6_experiments once this PR is "done" as a last step, this is just to keep an example (and potential changes to it after changes in the job) easily accessible for now.

agreed.

> The report format is now mandatory and should be able to handle str and tk.AbstractPath inputs.

I am not overly happy with the current interface between the `report` input and the `report_format` function. Maybe we can discuss this once other people have voiced their opinion.

report/report.py
Atticus1806: For me this is part of my `settings.py` ever since I got my first "copy" of a sis setup, I thought this would be standard :sweat_smile: 
But this can be solved without problems I think. It just calls `getpass.getuser()` so this could be a default in a parameter.
I am still not sure how a `MailJob` would interact with just sisyphus variables etc. but I will write a quick draft of a job as proof of concept and push it here so we have something to compare to.

report/report.py
Atticus1806: Tbh. I dont know. The binary variant is copied from the old code and the non compressed version was added by me. But from my understanding (which might be off by a longshot, I dont have much experience with this) I think "wb" expects some kind of binary input and does not convert "internally".

report/report.py
Atticus1806: Also just having a `MailJob` would not hurt. So even if we later decide to keep this interface like it is right now, we can just also add a `MailJob`.

Atticus1806:
So I added a MailJob @michelwi. Is this how you intended it? If the "wait_on" is not used it produces "Unfinished Variables" and I think this is integral to the job so I think its fine to set inside the job (and not have the user have to think of setting it).
My code right now would now look like this (examplatory):

```python
waits = list(scores.values())
content = gmm_example_report_format(scores)
report = MailJob(subject=alias_prefix, content=content, mail_address=getpass.getuser(), wait_on=waits)
tk.register_output(f"reports/{alias_prefix}", report.out_content)
```
Tested it with both compress on and off. 
This would work for me, so the other Job could be removed (or kept if people prefer to have another option, both fine with me)

report/report.py
michelwi: see https://docs.python.org/3/library/functions.html#open
> [mode] defaults to 'r' which means open for reading in text mode.

and https://docs.python.org/3/library/gzip.html
> The mode argument can be any of 'r', 'rb', 'a', 'ab', 'w', 'wb', 'x' or 'xb' for binary mode, or 'rt', 'at', 'wt', or 'xt' for text mode. The default is 'rb'.

so 'w' for builtin `open` defaults to text mode and 'w' for `gzip.open` defaults to binary. I suggest to decide which one you want and then use 'wb' or 'wt' consistently.

report/report.py
```diff
@@ -0,0 +1,99 @@
...
+        yield Task("run", mini_task=True)
+
+    def run(self):
+
+        if self.report_template:
+            report = self.report_template(self.report_values)
+        elif callable(self.report_values):
+            report = str(self.report_values())
+        else:
+            report = pprint.pformat(self.report_values, width=140)
```
christophmluscher, line 44: I am not sure about this if...elif...else clause.

why should the `report_values` be callable?? it is only of type `Dict`?!

the check on `report_template` could be restricted to None or even a callable..

report/report.py
```diff
@@ -0,0 +1,99 @@
...
+
+        :param report_values:
+        :param report_template:
+        :param compress:
+        """
+        self.report_values = report_values
+        self.report_template = report_template
+        self.compress = compress
+
+        self.out_report = self.output_path("report.gz" if self.compress else "report")
```
christophmluscher, line 32:
```suggestion
        self.out_report = self.output_path("report.txt.gz" if self.compress else "report.txt")
```

report/report.py
```diff
@@ -0,0 +1,99 @@
...
+    def run(self):
+
+        if self.report_template:
+            report = self.report_template(self.report_values)
+        elif callable(self.report_values):
+            report = str(self.report_values())
+        else:
+            report = pprint.pformat(self.report_values, width=140)
+
+        if self.compress:
```
christophmluscher, line 46: just as a suggestion:

`fopen = gzip.open if self.compress else open`

report/report.py
Atticus1806: The typing was wrong. So this can be callable (similar to how sisyphus register report has it as a possibilty) , for example some function that returns some string (interpretable) output. You cannot pass the string directly because you dont want to generate the string at graph construction different reasons.
This is quite close to just passing a format function but since sisyphus supported it we decided to also do this here.

report/report.py
```diff
@@ -0,0 +1,98 @@
...
+            "report.txt.gz" if self.compress else "report.txt"
+        )
+
+    def tasks(self):
+        yield Task("run", mini_task=True)
+
+    def run(self):
+
+        if self.report_template:
+            report = self.report_template(self.report_values)
```
dthulke, line 42: I think it would be nice if `report_values` and `report_template` follows the same interface as `register_report` in sisyphus. This allows to reuse existing "sisyphus reports" with these jobs to send mails.
In sisyphus report_template is not a callable but a format string (see https://github.com/rwth-i6/sisyphus/blob/032239c4f473742b96178f9614521324ac0472f7/sisyphus/graph.py#L187).

Atticus1806: I would like to support both, but added the option if the type of the `report_template` is str to be used as format string.

report/report.py
```diff
@@ -0,0 +1,101 @@
...
+    def run(self):
+
+        if self.subject is None:
+            subject = f"Output {str(self.result)} is finished"
+        else:
+            subject = self.subject
+
+        if self.send_contents:
+            value = self.sh(
+                f"zcat -f {self.result.get_path()} | mail -s '{subject}' {self.mail_address}"
```
JackTemaki, line 94: We should avoid using `self.sh` here in case this ever gets part of something automatic, as it allows for random shell execution.

Some like (this is random code):
```python
p1 = subprocess.Popen(["zcat", "-f", self.result.get_path()], stdout=subprocess.PIPE)
value = subprocess.check_output(["mail", "-s", subject, self.mail_address], stdin=p1.stdout)
```

JackTemaki (changes requested):

report/report.py
Atticus1806: Updated it, but didnt test it yet. Does this also work like this when send_contents is False? Haven't used much of subprocess yet, so just tried to "copy" it from the other case.

report/report.py
```diff
@@ -0,0 +1,101 @@
+from sisyphus import *
```
michelwi, line 1: Can you add an `__all__` entry s.t. `from report import *` does not in fact import everything.

report/report.py
michelwi: not directly, but with the newer `subprocess.run` method you can directly provide a string for stdin. It will internally use `Popen.communicate`.

Atticus1806:
Okay applied the comments and changes. So again requesting for feedback or approval.

report/report.py
```diff
@@ -0,0 +1,106 @@
...
+
+        if self.send_contents:
+            p1 = subprocess.Popen(
+                ["zcat", "-f", self.result.get_path()], stdout=subprocess.PIPE
+            )
+            value = subprocess.check_output(
+                ["mail", "-s", subject, self.mail_address], stdin=p1.stdout
+            )
+        else:
+            out = subprocess.run(
```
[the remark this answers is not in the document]
michelwi: > was input="" important here?

the `mail` command expects the body of the mail to be piped in from stdin. To avoid creating another process only to pipe empty string to stdin we switched to run.

But actually in the first draft, here the body was not empty but the subject line. i.e. `input=subject`
Maybe we want to have this again?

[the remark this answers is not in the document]
Atticus1806: Thinking about it, sending empty emails is weird to me. So I changed it back.

Atticus1806:
Comments are addressed now

JackTemaki (approved):
If everything works as expected it looks good to me

michelwi (approved):

PR #315, opened 2022-09-28
CompileTFGraphJob add epoch parameter

JackTemaki:
Related to:
https://github.com/rwth-i6/returnn/pull/1133

To be able to compile a specific epoch for networks that might change with every epoch.

### DIFF
--- a/returnn/compile.py
+++ b/returnn/compile.py
@@ -21,7 +21,7 @@ class CompileTFGraphJob(Job):
 
     """
 
-    __sis_hash_exclude__ = {"device": None}
+    __sis_hash_exclude__ = {"device": None, "epoch": None}
 
     def __init__(
         self,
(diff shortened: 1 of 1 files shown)


returnn/compile.py
```diff
@@ -42,6 +43,7 @@ def __init__(
         :param int train:
         :param int eval:
         :param int search:
+        :param int epoch:
```
michelwi, line 46:
```suggestion
        :param int epoch: compile a specific epoch for networks that might change with every epoch
```

michelwi (approved):
Let's have a docstring. Otherwise ok.

returnn/compile.py
christophmluscher: `None` should also be in the typing?

Issue #306, opened 2022-08-24
Hash-bug in ReturnnComputePriorJob


michelwi:
The same problem as discussed in https://github.com/rwth-i6/i6_core/issues/274 happens also in the ReturnnComputePriorJob:
during config creation the `checkpoint.ckpt_path` is used instead of using only the `checkpoint`.

https://github.com/rwth-i6/i6_core/blob/63a6d691f3fc8a270f87ee656c1437fc4ed01b59/returnn/extract_prior.py#L218

Should we do a V2 Job here as well?
Are there any other improvements we should tackle while were at it?

JackTemaki:
I vote for V2 as well, I do not see any sense in treating this differently.

JackTemaki:
I currently do not have a pipeline that uses this, but @christophmluscher  and @vieting have, so maybe they have ideas for other improvements. As I am on vacation the next two weeks, you have my approval to just fix this job in the same way as the ReturnnSearchJob.

vieting:
I don't have setups using this job, so either way of handling is fine for me. Agree though, that there's no reason to treat it differently than #274 if setups exist, e.g. by @christophmluscher.

christophmluscher:
I have a setup with this job. I can do the PR and also add a test.

I would also vote for V2. Ill check for improvements as well.

@Marvin84 are you using this job as well?

Marvin84:
> @Marvin84 are you using this job as well?

No, I have my own prior computation job which can be used also context-dependent case.

JackTemaki:
Solved with #316

PR #316, opened 2022-10-03
ReturnnComputePriorJob refactoring

christophmluscher:
* fix hash bug via introducing ReturnnComputePriorJobV2
* some refactoring
* doc

addresses #306


returnn/extract_prior.py
```diff
@@ -72,52 +78,55 @@ def run(self):
...
-        log_verbosity=3,
-        device="gpu",
-        time_rqmt=4,
-        mem_rqmt=4,
-        cpu_rqmt=2,
-        returnn_python_exe=None,
-        returnn_root=None,
+        log_verbosity: int = 3,
+        device: str = "gpu",
+        time_rqmt: Union[float, int] = 4,
```
michelwi, line 104:
```suggestion
        time_rqmt: float = 4,
```
float implies int

returnn/extract_prior.py
```diff
@@ -72,52 +78,55 @@ def run(self):
...
-            else gs.RETURNN_PYTHON_EXE
-        )
-
-        self.returnn_root = (
-            returnn_root if returnn_root is not None else gs.RETURNN_ROOT
-        )
+        self.returnn_python_exe = returnn_python_exe
+        self.returnn_root = returnn_root
 
         self.returnn_config = ReturnnComputePriorJob.create_returnn_config(**kwargs)
```
michelwi, line 131:
```suggestion
        self.returnn_config = ReturnnComputePriorJobV2.create_returnn_config(**kwargs)
```
I request changes w.r.t. this comment.
If you don't specify V2 here, I think you will still have the hash bug.

michelwi (changes requested):

returnn/extract_prior.py
christophmluscher: > ```suggestion
>         self.returnn_config = ReturnnComputePriorJobV2.create_returnn_config(**kwargs)
> ```
> I request changes w.r.t. this comment.
> If you don't specify V2 here, I think you will still have the hash bug.

Good catch 👍

returnn/extract_prior.py
```diff
@@ -72,54 +78,57 @@ def run(self):
...
-        time_rqmt=4,
-        mem_rqmt=4,
-        cpu_rqmt=2,
-        returnn_python_exe=None,
-        returnn_root=None,
+        log_verbosity: int = 3,
+        device: str = "gpu",
+        time_rqmt: float = 4,
+        mem_rqmt: Union[float, int] = 4,
+        cpu_rqmt: Union[float, int] = 2,
```
JackTemaki, line 106: This is only int...

JackTemaki (approved):
Only one small comment, the rest "should" be fine...

returnn/extract_prior.py
christophmluscher: true xD

albertz (approved):

michelwi (approved):

JackTemaki (changes requested):
same remark as @michelwi

christophmluscher:
> "returnn_config":  ReturnnComputePriorJobV2.create_returnn_config(**kwargs)

done

> also what about ReturnnRasrComputePriorJob?

here a new version would also be necessary, correct?

michelwi:
> > also what about ReturnnRasrComputePriorJob?
> 
> here a new version would also be necessary, correct?

Yes. But probably not much code changes necessay, only inherit from V2 Job.

> new PR?? that would speed things up for this job.. not sure

Ok.

returnn/extract_prior.py
```diff
@@ -72,54 +78,57 @@ def run(self):
...
+        :param returnn_python_exe: path to the RETURNN executable (python binary or launch script)
+        :param returnn_root: path to the RETURNN src folder
         """
         assert isinstance(returnn_config, ReturnnConfig)
         kwargs = locals()
         del kwargs["self"]
 
         self.model_checkpoint = model_checkpoint
 
-        self.returnn_python_exe = (
```
JackTemaki, line 112: If we remove this, we also need to remove the default `None`, otherwise the job will crash.

michelwi: if we incorporate https://github.com/rwth-i6/i6_core/pull/261 then we could leave default `None` which then falls back to the system python (or some virtual env that is used to run the manager)

returnn/extract_prior.py
```diff
@@ -114,12 +132,11 @@ def __init__(
             if returnn_python_exe is not None
             else gs.RETURNN_PYTHON_EXE
```
JackTemaki, line 133: both `gs.` calls need to be wrapped with `tk.Path`, otherwise later code will not work.

JackTemaki (changes requested):

JackTemaki:
> I thought it was a good feature to _not_ have fallbacks on `gs.RETURNN_ROOT` and `gs.RETURNN_PYTHON_EXE`. Any reason you added them back?

Good point, as this is V2 we could remove those completely.

michelwi (approved):
I think its good now.

JackTemaki (approved):

PR #317, opened 2022-10-03
returnn root as option

christophmluscher:
* rm gs.RETURNN_ROOT
* set returnn root as param

### DIFF
--- a/rasr/feature_scorer.py
+++ b/rasr/feature_scorer.py
@@ -87,7 +87,7 @@ class ReturnnScorer(FeatureScorer):
         self.config.pymod_path = (
             returnn_root
             if returnn_root is not None
-            else os.path.join(gs.RETURNN_ROOT, "..")
+            else tk.Path(gs.RETURNN_ROOT).join_right("..")
         )
         self.config.pymod_config = StringWrapper(
             "epoch:%d,action:forward,configfile:%s"
--- a/rasr/feature_scorer.py
+++ b/rasr/feature_scorer.py
@@ -71,6 +71,7 @@ class ReturnnScorer(FeatureScorer):
         mixture_scale=1.0,
         prior_scale=1.0,
         prior_file=None,
+        returnn_root=None,
     ):
         super().__init__()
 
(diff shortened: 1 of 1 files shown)


michelwi (commented):
We should not break old setups by removing the `gs.RETURNN_ROOT`

JackTemaki:
Yes, just add a fallback to `gs.RETURNN_ROOT` with the ".." if it is none.

JackTemaki (changes requested):
Yes, just add a fallback to `gs.RETURNN_ROOT` with the ".." if it is none.

Atticus1806:
We also have similar behavior e.g. in `ReturnnTrainingJob` 
https://github.com/rwth-i6/i6_core/blob/6d3614706826dd5789756042b6d829baacf78553/returnn/training.py#L125-L132

Also while at it, since its often a minimal change you could also add typing for the class init, since e.g. for me these types are not all clear. Technically this would be a different PR but I feel this is small enough to add a slightly bigger (documentation) change.

rasr/feature_scorer.py
```diff
@@ -83,7 +84,11 @@ def __init__(
         else:
             self.config.normalize_mixture_weights = False
         self.config.pymod_name = "returnn.SprintInterface"
-        self.config.pymod_path = os.path.join(tk.gs.RETURNN_ROOT, "..")
+        self.config.pymod_path = (
+            returnn_root
+            if returnn_root is not None
+            else os.path.join(gs.RETURNN_ROOT, "..")
```
michelwi, line 90: A) can we make it a tk.Path object also in the else branch?
B) would we also need to add the .. if a returnn_root was given by the user?
C) do you need to add something below to handle tk.Path vs. str?

christophmluscher: A) yes
B) the RETURNN dir structure changed at some point.. 
```
returnn
  - code
```
now
```
returnn
   - returnn
     - code
```
so I think for the current setup, which I tested you do not need it.. I did not test the old setup..

C) could you clarify where exactly??

D) I added typing but this somehow broke some tests.. this would need investigation

JackTemaki: B) Old setup means older than 2 year old RETURNN. 
So if people set this by hand, I am quite certain they use a new version.

JackTemaki: And what was the issue with the typing?

michelwi: C) sorry I was just lazy to check. Probably nothing needs to be done after second thought.

B) are we now implicitly assuming that gs.RETURNN_ROOT implies old folder structure and giving directly implies new folder structure?
Or would the new structure also work with the two dots as in gs.RETURNN_ROOT?

christophmluscher: > And what was the issue with the typing?

I added typing and some tests failed because import failed. still not sure why.. 

B) I think it should work but I am not sure.. I did not test

michelwi: Ok. concerning A) I still suggest
```suggestion
            else tk.Path(os.path.join(gs.RETURNN_ROOT, ".."))
```
or
```suggestion
            else tk.Path(gs.RETURNN_ROOT).join_right("..")
```

JackTemaki (approved):

michelwi (approved):

PR #319, opened 2022-10-05
GroupJob

albertz:
I used this in my setup to explicitly set some dependencies (and to trigger the execution of some jobs).
It has very much the same use cases as `tf.group` or `tf.control_dependencies`.

### DIFF
--- a/util.py
+++ b/util.py
@@ -1,3 +1,4 @@
+from typing import Optional
 import gzip
 import os
 import shutil
@@ -301,3 +302,36 @@ def instanciate_delayed(o):
         for k in o:
             o[k] = instanciate_delayed(o[k])
     return o
+
+
+class GroupJob(Job):
+    """
+    Like `tf.group`. Depends on `inputs`.
+    It is itself a no-op.
+    Because Sisyphus needs an output, it creates a dummy output.
+    Alternatively, you can also specify the output, which is then symlinked.
+    Thus, it is also like `tf.identity` with `control_dependencies`.
+    """
+
+    def __init__(self, inputs, *, output: Optional[tk.Path] = None):
+        """
+        :param inputs: any dependencies. can be any nested structure. should contain some other paths
+        :param output: if given, this path is symlinked to the output
+        """
+        super(GroupJob, self).__init__()
+        self.inputs = inputs
+        self._output = output
+        self.output = self.output_path(
+            os.path.basename(output.path) if output else "dummy.txt"
+        )
+
+    def tasks(self):
+        """tasks"""
+        yield Task("run", resume="run", mini_task=True)
+
+    def run(self):
+        """run"""
+        if self._output:
+            os.symlink(self._output.get_path(), self.output.get_path())
+        else:
+            open(self.output.get_path(), "wt").close()


michelwi (commented):
This Job should not be necessary. Maybe tell us where sisyphus fails to discover a dependency on its own and we can figure out how to fix it.

albertz:
> Maybe tell us where sisyphus fails to discover a dependency on its own and we can figure out how to fix it.

Sisyphus does not fail. There is no dependency. But I want to artificially introduce a dependency, and for that, I use this job. Exactly the same as `tf.group` or `tf.control_dependencies`.

Atticus1806:
What do you use this dependency for? To register it as an output? To add it as an input to some job? To `.hold()` this?

albertz:
> Finished pipeline or not, you would anyway need one register_output call.

Yes I have, but only on the final output of my pipeline (sth like `final_result_summary`). This `GroupJob` comes somewhere in between.

JackTemaki:
I would be fine with accepting it if you rename the job to `DummyJob`, and change the docstring away from Tensorflow specific terminology to the actual use cases (being a replacement for a not-yet existing job to trigger execution of jobs without using tk.register_output explicitly and collecting multiple inputs to be used as single dependency via `job.add_input`).

christophmluscher (commented):
I agree with @JackTemaki comment and I was wondering, is this the correct place for the job?! since in `util.py`, so far there are no Jobs. I have no really strong feeling about this but just want to make sure this is the right place.

albertz:
> rename the job to `DummyJob`

I thought a name like `GroupJob` would help all those people who are familiar with TensorFlow as they now `tf.group`. I feel a name like `DummyJob` is less helpful because it does not really tell anything. But if you prefer that, fine.

> actual use cases

I still think there might be other potential use cases despite what I currently use it for. E.g. if the user wants to trigger the execution of some jobs without using `register_output`.

> is this the correct place for the job?

I thought about this.

Actually, I think all the other code in `util.py` would rather belong to `lib/*.py`, as `lib` was the place for random utilities used by jobs, but `lib` does not define jobs itself.

But probably we don't want to change that now.

We could have a separate Python module. Bot how to name it? `util_jobs.py` or so would be very inconsistent to all the other Python module names.
Or `misc.py` or so, but then it's very unclear what the difference between `util.py` and `misc.py` is (just from the names), so that is also not good.

Maybe just a separate Python module only for this module, like `dummy.py` or `dummy_job.py`. But not sure if this is justified.

So, all those options look worse than just adding it to `util.py`. But maybe you have a better idea?

JackTemaki:
> E.g. if the user wants to trigger the execution of some jobs without using `register_output`

This is still an ill formulation. It is simply not possible to trigger execution without registering somewhere. You can insert the dummy job as stand-in for an actual job (thus the name), and call `register_output` at a later point. This is the first of the two use cases I specified as what should be in the docstring. So again the use cases are:

 - DummyJob as stand-in for triggering jobs without having a finished pipeline (your use case)
 - merging multiple paths dependencies into a single path dependency for execution order control (unlikely scenario but still okay to do).

Adding it to `util.py` is fine, as all other solutions are worse, and I also agree with that some things should be moved to `lib`.

albertz:
> > E.g. if the user wants to trigger the execution of some jobs without using `register_output`
> 
> This is still an ill formulation. It is simply not possible to trigger execution without registering somewhere.

With "without using `register_output`", I meant directly on the inputs to the `GroupJob`. This is what I wanted to avoid in the first place.

Of course I don't mean that you would not have any `register_output` anywhere. That does not make sense. I never ever said anything like that. Sure you would register somewhere, something, which directly or indirectly then causes the `GroupJob` to run, and thus then triggers the inputs of `GroupJob` to run, exactly as I want.

christophmluscher:
> But maybe you have a better idea?

I was thinking along the line of `pipeline.py`. since this job is basically about pipeline management... but not 100% sure

> I also agree with that some things should be moved to lib.

yes but that would be a different PR :)

JackTemaki:
> > But maybe you have a better idea?
> 
> I was thinking along the line of `pipeline.py` or `graph.py`. since this job is basically about pipeline management or sisyphus graph manipulation... but not 100% sure

I would be ok with `pipeline.py`.

JackTemaki:
I am closing this now due to inactivity

PR #321, opened 2022-10-05
Add apply spm

Nintendo-Wei:
Add a job to apply sentencepiece model to lexicon file


text/label/sentencepiece/apply.py
```diff
@@ -0,0 +1,91 @@
...
+                for t in eval:
+                    lm_tokens.add(t)
+
+        lm_tokens = list(lm_tokens)
+        spm = sentencepiece.SentencePieceProcessor(model_file=self.spm_model.get_path())
+        bpe_tokens = []
+        for t in lm_tokens:
+            bpe_tokens.append(spm.encode(t, out_type=str))
+
+        w2b = {w: b for w, b in zip(lm_tokens, bpe_tokens)}
```
JackTemaki, line 70: As this is just copied from the BPE code maybe we can refactor it to only exist once?

JackTemaki:
@Nintendo-Wei any motivation to complete this? I still think this would a nice addition, so maybe then somebody else (e.g. myself) can finish.

Nintendo-Wei:
Sorry for the delay, I somehow forgot...

text/label/subword_utils.py
```diff
@@ -0,0 +1,51 @@
...
+    for l in lexicon.lemmata:
+        if l.special is None and len(l.orth) > 0:
+            if not l.synt and len(l.eval) == 0:
+                o = l.orth[0]
+                l.synt = w2s[o]
+                l.eval.append([o])
+            if l.synt:
+                l.synt = sum([w2s[token] for token in l.synt], [])
+            if len(l.eval) > 0:
+                l.eval = [sum([w2s[t] for t in token_sequence], []) for token_sequence in l.eval]
```
michelwi, line 48: as you don't use `elif` these `if`'s will be triggered by the `l.synt` you have set above. Is this intended? Especially for the synt the w2s transformation is applied twice, right?

text/label/subword_utils.py
```diff
@@ -0,0 +1,51 @@
...
+
+    elem = lexicon.to_xml()
+    return ET.ElementTree(elem)
```
michelwi, line 51: Why convert to ElementTree? The Lexicon Object also has a write method that could be used by the subsequent code.

text/label/sentencepiece/apply.py
```diff
@@ -0,0 +1,54 @@
...
+
+try:
+    import sentencepiece
+except ImportError:
+    if not hasattr(gs, "WARNING_NO_SENTENCEPIECE") or gs.WARNING_NO_SENTENCEPIECE is True:
+        logging.warning(
+            "The package 'sentencepiece' is not installed in the manager python env. Please make "
+            "sure it is installed in the python environment running the Sisyphus worker. To "
+            "suppress this warning set 'WARNING_NO_SENTENCEPIECE=False' in the settings.py"
+        )
```
michelwi, line 22: Do we really need this?

text/label/subword_utils.py
Nintendo-Wei: Thanks for the good catch! I find that a bit strange now too. This is copied from the line 83-92 in [text/label/subword_nmt/apply.py](https://github.com/rwth-i6/i6_core/blob/main/text/label/subword_nmt/apply.py), maybe @JackTemaki could comment on that?

text/label/subword_utils.py
Nintendo-Wei: I didn't see a write function in https://github.com/rwth-i6/i6_core/blob/main/lib/lexicon.py. And it is originally also used this way: https://github.com/rwth-i6/i6_core/blob/main/text/label/subword_nmt/apply.py#L95

text/label/sentencepiece/apply.py
Nintendo-Wei: It's also in https://github.com/rwth-i6/i6_core/blob/main/text/label/sentencepiece/train.py to make sure the sentencepiece package is installed?

JackTemaki:
Closing this now as this is not needed right now, this can be covered in a future re-work together with BPE.

PR #322, opened 2022-10-06
returnn.search, missing __all__ entries

albertz:


christophmluscher (approved):
PR is fine. Maybe we should split the `search.py`?! The file is getting quite large and is not only about search anymore, but search post-processing as well. but probably better adressed in a different PR?! @JackTemaki

JackTemaki:
I think as it works on the specific return output format this is okay in search. Moving would also not work anymore without Job placeholders in the file here, so lets just keep it this way.

JackTemaki (approved):

PR #325, opened 2022-10-12
ReturnnSearchJobV2, support gzipped out

albertz:
When you add `.gz` to the filename, RETURNN will automatically gzip the output (just with the most recent version of RETURNN).

So in the `ReturnnSearchJobV2`, it needs to add `.gz` to the filename to enable that.

I'm not sure whether we maybe want to enable it always? Then we maybe want to check all other relevant jobs whether they already support that. Many already do.

For example, `SearchBPEtoWordsJob`, `SearchWordsToCTMJob`, `SearchTakeBestJob` etc all do because they use `util.uopen` which will automatically do the right thing.

RETURNN `calculate-word-error-rate.py` and thus `ReturnnComputeWERJob` also does the right thing already.

Not sure if anything else is missing?

So if you think we always should enable that, I would remove the option and just always enable it.

### DIFF
--- a/returnn/search.py
+++ b/returnn/search.py
@@ -91,7 +91,7 @@ class ReturnnSearchJobV2(Job):
         self.out_returnn_config_file = self.output_path("returnn.config")
 
         self.out_search_file = self.output_path(
-            f"search_out" + (f".{output_mode}.gz" if output_gzip else "")
+            "search_out" + (f".{output_mode}.gz" if output_gzip else "")
         )
 
         self.returnn_config = ReturnnSearchJobV2.create_returnn_config(**kwargs)
--- a/returnn/search.py
+++ b/returnn/search.py
@@ -59,6 +59,7 @@ class ReturnnSearchJobV2(Job):
         returnn_root: tk.Path,
         *,
         output_mode: str = "py",
+        output_gzip: bool = False,
         log_verbosity: int = 3,
         device: str = "gpu",
         time_rqmt: float = 4,
@@ -72,6 +73,7 @@ class ReturnnSearchJobV2(Job):
         :param tk.Path returnn_python_exe: path to the RETURNN executable (python binary or launch script)
         :param tk.Path returnn_root: path to the RETURNN src folder
         :param str output_mode: "txt" or "py"
+        :param bool output_gzip: if true, gzip the output file
         :param int log_verbosity: RETURNN log verbosity
         :param str device: RETURNN device, cpu or gpu
         :param time_rqmt: job time requirement in hours
@@ -88,7 +90,9 @@ class ReturnnSearchJobV2(Job):
 
         self.out_returnn_config_file = self.output_path("returnn.config")
 
-        self.out_search_file = self.output_path("search_out")
+        self.out_search_file = self.output_path(
+            f"search_out.{output_mode}." + (".gz" if output_gzip else "")
+        )
 
         self.returnn_config = ReturnnSearchJobV2.create_returnn_config(**kwargs)
         self.returnn_config.post_config["search_output_file"] = self.out_search_file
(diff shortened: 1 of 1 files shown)


albertz:
Hm I wonder, for any existing completed `ReturnnSearchJobV2` dirs, as this PR changes the output filename, what happens? It would probably complain that some output is missing of a complete job or so? Is this problematic?

JackTemaki (changes requested):
Changing the output name unconditionally will break the job. Adding `.gz` to the original name based on the new parameter is not a problem though.

albertz:
> Changing the output name unconditionally will break the job.

I'm not exactly sure I understand what you mean.

"Changing the output name unconditionally" - to what?

"will break the job" - in what way? You mean existing job dirs, what I described? But new runs will just run fine, right? So is this really a problem?

> Adding `.gz` to the original name based on the new parameter is not a problem though.

But this would unconditionally change the output name, or not? Or you mean conditioned on such parameter, and the parameter is False by default?

I also don't really understand what your statement is now regarding my question. My question: should we just enable it always or not, i.e. having an option for it? You did not really answer this.

returnn/search.py
```diff
@@ -88,7 +90,9 @@ def __init__(
 
         self.out_returnn_config_file = self.output_path("returnn.config")
 
-        self.out_search_file = self.output_path("search_out")
+        self.out_search_file = self.output_path(
+            f"search_out.{output_mode}." + (".gz" if output_gzip else "")
```
christophmluscher, line 94:
```suggestion
            f"search_out.{output_mode}" + (".gz" if output_gzip else "")
```

JackTemaki:
>  So is this really a problem?

It will break existing graphs

> Or you mean conditioned on such parameter, and the parameter is False by default?

Correct, this is possible to change.

>  should we just enable it always or not, i.e. having an option for it? You did not really answer this.

I did. Changing the default file to .gz breaks compatibility, as any name change breaks the graph. If a name change is hidden behind a new flag it is ok.

returnn/search.py
christophmluscher: also: ```search_output``` to ```search_output.{txt,py}{,.gz}```

this change should already change the hashes of the following jobs

christophmluscher: Is V2 already heavy in use? Or can we change this although breaking hashes?

christophmluscher, line 94:
```suggestion
            f"search_out" + (".{output_mode}.gz" if output_gzip else "")
```

I think this is what I would suggest

albertz: I think it is. I changed this back.

albertz: Ah I see. Ok I did that.

albertz:
@JackTemaki What changes did you actually request? Is it ok now?

JackTemaki:
> > > should we just enable it always or not, i.e. having an option for it? You did not really answer this.
> > 
> > 
> > I did. Changing the default file to .gz breaks compatibility, as any name change breaks the graph. If a name change is hidden behind a new flag it is ok.
> 
> You still did not really answer it. I assume your answer is: we must have an option for it? Or not? I'm still not exactly sure whether you are saying that or not.

I just used the words "parameter" or "flag" instead of option, but I said exactly this.

JackTemaki (approved):

albertz:
> I just used the words "parameter" or "flag" instead of option, but I said exactly this.

No you said it breaks the graph. But my question was whether we should have such parameter/flag/option, or not. It was not clear to me.

returnn/search.py
```diff
@@ -88,7 +90,9 @@ def __init__(
 
         self.out_returnn_config_file = self.output_path("returnn.config")
 
-        self.out_search_file = self.output_path("search_out")
+        self.out_search_file = self.output_path(
+            f"search_out" + (f".{output_mode}.gz" if output_gzip else "")
```
christophmluscher, line 94:
```suggestion
            "search_out" + (f".{output_mode}.gz" if output_gzip else "")
```

no f-string required but not something critical ^^

christophmluscher (approved):

PR #326, opened 2022-10-12
ReturnnSearchJobV2, fix hash when output_gzip is enabled

albertz:
Followup to #325. I forgot that we have a custom `hash` in this job.

### DIFF
--- a/returnn/search.py
+++ b/returnn/search.py
@@ -201,6 +201,8 @@ class ReturnnSearchJobV2(Job):
             "returnn_python_exe": kwargs["returnn_python_exe"],
             "returnn_root": kwargs["returnn_root"],
         }
+        if kwargs.get("output_gzip", False) is not False:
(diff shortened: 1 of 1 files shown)


JackTemaki (approved):
I was actually fine with not having it hashed, on the contrary the problem would have been bigger in the case that we would not have a custom hash (Then you need  to set `_sis_hash_exclude` to not alter old hashes). But this is maybe a little bit nicer.

albertz:
In my setup, it crashed now exactly as what I assumed would happen when there were existing job dirs because their hash did not change after I changed the option.
```
FileNotFoundError: [Errno 2] No such file or directory: '/u/zeyer/setups/combined/2021-05-31/work/i6_core/returnn/search/ReturnnSearchJobV2.mwvCMAMLM8AP/output/search_out.py.gz'
```

christophmluscher (approved):

PR #327, opened 2022-10-19
add BlissToPcmHDFJob

michelwi:
Can be useful to do returnn-based feature extraction and on the fly data perturbation in returnn using HDFDataset.

### DIFF
--- a/corpus/convert.py
+++ b/corpus/convert.py
@@ -4,18 +4,22 @@ __all__ = [
     "CorpusToStmJob",
     "CorpusToTextDictJob",
     "CorpusToTxtJob",
+    "CorpusToRawWavHDFJob",
 ]
 
 import gzip
 import itertools
+import numpy
 import pprint
 import re
+import wave
 
 from typing import Dict, List, Optional, Tuple, Union
 
 from sisyphus import *
 
 from i6_core.lib import corpus
+from i6_core.lib.hdf import get_returnn_simple_hdf_writer
 from i6_core.util import uopen
 
 Path = setup_path(__package__)
@@ -326,3 +330,69 @@ class CorpusToTxtJob(Job):
                     segment.fullname() in segments_whitelist
                 ):
                     f.write(segment.orth + "\n")
+
+
+class CorpusToRawWavHDFJob(Job):
+    """
+    Gets audio files from a Bliss corpus and stores them as HDF file
+    compatible with the returnn HDFDataset
+    """
+
+    def __init__(
+        self,
+        bliss_corpus: tk.Path,
+        segment_file: Optional[tk.Path] = None,
+        returnn_root: Optional[tk.Path] = None,
+    ):
+        """
+
+        :param bliss_corpus: Bliss corpus to read segments and audio files from
+        :param segment_file: segment file that lists allowed segments
+        :param returnn_root: returnn repository
+        """
+        self.set_vis_name("Dump audio to HDF")
+
+        self.bliss_corpus = bliss_corpus
+        self.segment_file = segment_file
+        self.returnn_root = returnn_root
+
+        self.out_hdf = self.output_path("audio.hdf")
+
+    def tasks(self):
+        yield Task("run", mini_task=True)
+
+    def run(self):
+        SimpleHDFWriter = get_returnn_simple_hdf_writer(self.returnn_root.get_path())
+
+        c = corpus.Corpus()
+        c.load(self.bliss_corpus.get_path())
+
+        if self.segment_file:
+            with uopen(self.segment_file, "rt") as f:
+                segments_whitelist = set(
+                    l.strip() for l in f.readlines() if len(l.strip()) > 0
+                )
+        else:
+            segments_whitelist = None
+
+        out_hdf = SimpleHDFWriter(filename=self.out_hdf, dim=1)
+
+        for segment in c.segments():
+            if (not segments_whitelist) or (segment.fullname() in segments_whitelist):
+                audio_file = segment.recording.audio
+                audio = wave.open(audio_file)
+                audio.setpos(int(segment.start * audio.getframerate()))
+                data = audio.readframes(
+                    int((segment.end - segment.start) * audio.getframerate())
+                )
+                audio.close()
+                data_s16 = numpy.frombuffer(data, dtype=numpy.int16)
+                data_f32 = data_s16.astype(numpy.float32)
+
+                out_hdf.insert_batch(
+                    inputs=data_f32.reshape(1, -1, 1),
+                    seq_len=[data_f32.shape[0]],
+                    seq_tag=[segment.fullname()],
+                )
+
+        out_hdf.close()
--- a/corpus/convert.py
+++ b/corpus/convert.py
@@ -4,22 +4,18 @@ __all__ = [
     "CorpusToStmJob",
     "CorpusToTextDictJob",
     "CorpusToTxtJob",
-    "CorpusToRawWavHDFJob",
 ]
 
 import gzip
 import itertools
-import numpy
 import pprint
 import re
-import wave
 
 from typing import Dict, List, Optional, Tuple, Union
 
 from sisyphus import *
 
 from i6_core.lib import corpus
-from i6_core.lib.hdf import get_returnn_simple_hdf_writer
 from i6_core.util import uopen
 
 Path = setup_path(__package__)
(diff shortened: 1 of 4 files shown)


corpus/convert.py
```diff
@@ -326,3 +330,69 @@ def run(self):
                     segment.fullname() in segments_whitelist
                 ):
                     f.write(segment.orth + "\n")
+
+
+class CorpusToRawWavHDFJob(Job):
+    """
+    Gets audio files from a Bliss corpus and stores them as HDF file
+    compatible with the returnn HDFDataset
```
michelwi, line 338: Well, the simple hdf writer claims this, I did not test compatibility with returnn yet.

corpus/convert.py
```diff
@@ -326,3 +330,69 @@ def run(self):
                     segment.fullname() in segments_whitelist
                 ):
                     f.write(segment.orth + "\n")
+
+
+class CorpusToRawWavHDFJob(Job):
```
michelwi, line 335: Discussion question:
is corpus/convert.py the right place? The Job does not convert the corpus (transcript) as the other jobs in this file, but it does convert the collection of audio files (i.e. the corpus?)

corpus/convert.py
```diff
@@ -326,3 +330,69 @@ def run(self):
...
+            if (not segments_whitelist) or (segment.fullname() in segments_whitelist):
+                audio_file = segment.recording.audio
+                audio = wave.open(audio_file)
+                audio.setpos(int(segment.start * audio.getframerate()))
+                data = audio.readframes(
+                    int((segment.end - segment.start) * audio.getframerate())
+                )
+                audio.close()
+                data_s16 = numpy.frombuffer(data, dtype=numpy.int16)
+                data_f32 = data_s16.astype(numpy.float32)
```
michelwi, line 390: We always have features as float. but storing as signed integer would save 50% of storage space. 

Are integer features supported on returnn side?
Do we want to save space on disk and do the conversion in the network?

michelwi (commented):
some discussion points inline.

corpus/convert.py
JackTemaki: It seems RETURNN is just passing the data type as is without modification. So if you store numpy.int16 it will feed it this way to the tensorflow graph, mostly like resulting in an error. I already discussed with @christophmluscher that we might want to have a "pre_process" parameter for the HDF Dataset just like for the OggZip.

corpus/convert.py
JackTemaki: As the SimpleHDFWriter comes from RETURNN itself, it should hold the claim, if not this is a RETURNN bug.

corpus/convert.py
michelwi: what if the first layer in the feature extraction would to the casting?

JackTemaki: "Probably" fine...

corpus/convert.py
JackTemaki: what about `returnn.hdf`? As this uses the returnn backend to create an HDF file it might fit better there.

vieting: > what about `returnn.hdf`? As this uses the returnn backend to create an HDF file it might fit better there.

I second this, would also be analog to similar jobs for the `OggZipDataset`.

corpus/convert.py
```diff
@@ -326,3 +330,69 @@ def run(self):
...
+        self.segment_file = segment_file
+        self.returnn_root = returnn_root
+
+        self.out_hdf = self.output_path("audio.hdf")
+
+    def tasks(self):
+        yield Task("run", mini_task=True)
+
+    def run(self):
+        SimpleHDFWriter = get_returnn_simple_hdf_writer(self.returnn_root.get_path())
```
vieting, line 365: This doesn't seem to work if `returnn_root` is `None`.

corpus/convert.py
```diff
@@ -326,3 +330,69 @@ def run(self):
...
+                )
+        else:
+            segments_whitelist = None
+
+        out_hdf = SimpleHDFWriter(filename=self.out_hdf, dim=1)
+
+        for segment in c.segments():
+            if (not segments_whitelist) or (segment.fullname() in segments_whitelist):
+                audio_file = segment.recording.audio
+                audio = wave.open(audio_file)
```
vieting, line 383: I think at other places, we rather use `soundfile` which is also in the `requirements.txt`. This would be more consistent.

corpus/convert.py
michelwi: you are right.. the main problem is that the function `get_returnn_simple_hdf_writer` wants a str instead of a path. Maybe we want to change that?

otherwise:
```suggestion
        returnn_root = None if self.returnn_root is None else self.returnn_root.get_path()
        SimpleHDFWriter = get_returnn_simple_hdf_writer(returnn_root)
```

corpus/convert.py
michelwi: good point. I'll try to rewrite this.

vieting: Not sure how much of the processing time is consumed by loading, but you might also rewrite the code a bit to load the audio only once for each recording and then iterate over the segments. In case you have a corpus with many segments per recording, this might be beneficial.

vieting: That's true. I think the first case is more likely, but I'm fine with keeping the logic as is, too.

corpus/convert.py
JackTemaki: > Maybe we want to change that?

I used str because it is an internal function...

corpus/convert.py
JackTemaki, line 335: I did not notice it before, but `RawWav` is not an optimal term. Because if we use `soundfile` instead of `wave` (very good suggestion from @vieting) we can load "any" data. And then we should name the job `CorpusToPcmHDF`, as we are storing PCM data.

michelwi:
tests fail because `libsoundfile.so` is missing :/
it does run locally though

corpus/convert.py
michelwi: @JackTemaki and I discussed offline and came to the conclusion that we want to have s16 as the default data type (and conversion to float would be done inside the network) to save storage space.
If someone wanted to have float, there is a parameter to override it.

JackTemaki:
> tests fail because `libsoundfile.so` is missing :/ it does run locally though

You can try adding `run: sudo apt-get install -y libsndfile1` (or similar not sure about the package name)

michelwi:
I think I have addressed all comments so far, please check again.

returnn/hdf.py
```diff
@@ -195,3 +199,79 @@ def create_files(self):
...
+                    segment.fullname() in segments_whitelist
+                ):
+                    audio.seek(int(segment.start * audio.samplerate))
+                    data = audio.read(
+                        int((segment.end - segment.start) * audio.samplerate),
+                        dtype=self.output_dtype,
+                    )
+                    out_hdf.insert_batch(
+                        inputs=data.reshape(1, -1, 1),
+                        seq_len=[data.shape[0]],
```
JackTemaki, line 271: Just to confirm, `data` is [T,1]? I had cases where I assumed this but suddenly it was [1,T], this is why I ask.

JackTemaki: And how will this behave for multi-channel audio, e.g. Stereo?

michelwi: `data` at this point is a one-dimensional numpy array, i.e. [T]
After I reshape it, it is written to HDF as [B=1, T, F=1].

>And how will this behave for multi-channel audio, e.g. Stereo?

I did not consider that. From the documentation: for stereo audio, `data` will be [T, 2] so that the reshaping will result in [B=1, T=2xT, F=1] which is wrong.

What should we do instead? average both channels? always pick channel 0?

JackTemaki: For now write an assert for the channel count to be one (I guess you can get that from the `audio` object). I think we should not delay this PR for this use case yet that needs extra discussion.

returnn/hdf.py
```diff
@@ -195,3 +199,79 @@ def create_files(self):
...
+                    l.strip() for l in f.readlines() if len(l.strip()) > 0
+                )
+        else:
+            segments_whitelist = None
+
+        out_hdf = SimpleHDFWriter(filename=self.out_hdf, dim=1)
+
+        for recording in c.all_recordings():
+            audio_file = recording.audio
+            audio = sf.SoundFile(audio_file)
```
michelwi, line 258:
```suggestion
            audio = sf.SoundFile(audio_file)
            assert audio.channels == 1, "Multichannel audio not yet implemented"
```

JackTemaki (approved):

returnn/hdf.py
```diff
@@ -195,3 +199,80 @@ def create_files(self):
             f.write(
                 '<?xml version="1.0" ?>\n<network><out name="features" /></network>'
             )
+
+
+class BlissToPcmHDFJob(Job):
+    """
+    Gets audio files from a Bliss corpus and stores them as HDF file
+    compatible with the returnn HDFDataset
```
vieting, line 207:
```suggestion
    compatible with the RETURNN HDFDataset
```

returnn/hdf.py
```diff
@@ -195,3 +199,80 @@ def create_files(self):
...
+        segment_file: Optional[tk.Path] = None,
+        output_dtype: str = "int16",
+        returnn_root: Optional[tk.Path] = None,
+    ):
+        """
+
+        :param bliss_corpus: Bliss corpus to read segments and audio files from
+        :param segment_file: segment file that lists allowed segments
+        :param output_dtype: dtype that should be written in the hdf (supports float64, float32, int32, int16)
+        :param returnn_root: returnn repository
```
vieting, line 222:
```suggestion
        :param returnn_root: RETURNN repository
```

returnn/hdf.py
```diff
@@ -195,3 +199,80 @@ def create_files(self):
...
+        self.bliss_corpus = bliss_corpus
+        self.segment_file = segment_file
+        assert output_dtype in ["float64", "float32", "int32", "int16"]
+        self.output_dtype = output_dtype
+        self.returnn_root = returnn_root
+
+        self.out_hdf = self.output_path("audio.hdf")
+
+    def tasks(self):
+        yield Task("run", mini_task=True)
```
vieting, line 235: Are you sure this is always suitable as a mini task even for large corpora?

vieting (approved):
Just nitpicks

returnn/hdf.py
michelwi: I am always dumping HDFs with ~10h of audio. This takes ~5 min.

If one were to dump the whole of LibriSpeech into one file, then one certainly would need a cluster node.

Maybe we should make it always a full task, even for the 5 min case.

Atticus1806: I am not sure where, but I remember seeing this in some jobs as an internal flag, where the size of the corpus is checked and depending on size mini_task is either set or not. But I cant recall where right now.

michelwi: This would require resolving a Path object (the corpus) in the manager. I do not want to do that.

vieting: Yes, then making it a full task seems reasonable to me.

Atticus1806: So I found this: 
https://github.com/rwth-i6/i6_core/blob/e904b981a8fce46caaf449f252a2c4f2a00e44ad/datasets/huggingface.py#L37
there it is set by the user manually. I am not sure if this makes sense here, otherwise full task makes sense to me.
Most likely what I saw was dependent on some user set input parameter (like size or so), but I am not 100% sure anymore.

vieting: The `CorpusObject` from `meta/system.py` has an attribute `duration`, this could also be related. But here, the input is a `tk.Path`.

vieting (approved):

PR #328, opened 2022-10-20
add TASK parameter to flow in AdvancedTreeSearchJob

JackTemaki:
It was found that the AdvancedTreeSearchJob is currently not adding the `TASK` parameter to the flow, thus crashing when the cache manager module is not existing in RASR. All other jobs (Alignment, Mixtures) are adding the parameter within the `create_files` Job-Task, so we also should do it here.

### DIFF
--- a/recognition/advanced_tree_search.py
+++ b/recognition/advanced_tree_search.py
@@ -226,6 +226,7 @@ class AdvancedTreeSearchJob(rasr.RasrCommand, Job):
 
     def create_files(self):
         self.write_config(self.config, self.post_config, "recognition.config")
+        self.feature_flow.add_param("TASK")
         self.feature_flow.write_to_file("feature.flow")
         util.write_paths_to_file(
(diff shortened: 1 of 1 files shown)


michelwi (approved):
I find it very sad that this does not influence the hash.

Will adding a param here add it to the original flow (because `self.feature_flow` is a reference to the original one)? Can this cause problems down the setup?
Mh.. the hash check seems to answer "no"..

JackTemaki:
> I find it very sad that this does not influence the hash.
> 
> Will adding a param here add it to the original flow (because `self.feature_flow` is a reference to the original one)? Can this cause problems down the setup? Mh.. the hash check seems to answer "no"..

Maybe doing a copy would be nicer, but as this edit is in the task (will be executed by the worker) there is no problem w.r.t. to the manager.

christophmluscher (approved):

PR #332, opened 2022-10-31
Add GetBestEpochJob and GetBestTFCheckpointJob

JackTemaki:
Jobs to determine the best epoch(s) of a RETURNN training and to save TF models.

I personally do not like the usage of `eval`, maybe someone has a better idea that is not overly complicated, RETURNN does also use `eval`.

### DIFF
--- a/returnn/training.py
+++ b/returnn/training.py
@@ -573,3 +573,177 @@ class ReturnnTrainingFromFileJob(Job):
         }
 
         return super().hash(d)
+
+
+class GetBestEpochJob(Job):
+    """
+    Provided a RETURNN model directory and a score key, finds the best epoch.
+    The sorting is lower=better, so to acces the model with the highest values use negative index values (e.g. -1 for
+    the model with the highest score)
+
+    """
+
+    def __init__(
+        self, model_dir: tk.Path, learning_rates: tk.Path, key: str, index: int = 0
+    ):
+        """
+        :param model_dir: model_dir output from a RETURNNTrainingJob
+        :param learning_rates: learning_rates output from a RETURNNTrainingJob
+        :param key: a key from the learning rate file that is used to sort the models,
+            e.g. "dev_score_output/output_prob"
+        :param index: index of the sorted list to access, 0 for the lowest, -1 for the highest score
+        """
+        self.model_dir = model_dir
+        self.learning_rates = learning_rates
+        self.index = index
+        self.out_epoch = self.output_var("epoch")
+        self.key = key
+
+    def run(self):
+        # this has to be defined in order for "eval" to work
+        def EpochData(learningRate, error):
+            return {"learning_rate": learningRate, "error": error}
+
+        with open(self.learning_rates.get_path(), "rt") as f:
+            text = f.read()
+
+        data = eval(text, {"inf": 1e99, "EpochData": EpochData})
+
+        epochs = list(sorted(data.keys()))
+
+        # some epochs might not have all keys, we require that the last entry has the key, other epochs which
+        # do not have it might be ignored
+        available_keys = data[epochs[-1]]["error"]
+        if self.key not in available_keys:
+            raise KeyError(
+                f"{self.key} is not available in the provided learning_rates file f{self.learning_rates.get_path()}"
+            )
+
+        scores = [
+            (epoch, data[epoch]["error"][self.key])
+            for epoch in epochs
+            if self.key in data[epoch]["error"]
+        ]
+        sorted_scores = list(sorted(scores, key=lambda x: x[1]))
+
+        self.out_epoch.set(sorted_scores[self.index][0])
+
+    def tasks(self):
+        yield Task("run", mini_task=True)
+
+
+class GetBestTFCheckpointJob(GetBestEpochJob):
+    """
+    Returns the best checkpoint given a training model dir and a learning-rates file
+    The best checkpoint will be HARD-linked if possible, so that no space is wasted but also the model not
+    deleted in case that the training folder is removed.
+    """
+
+    def __init__(
+        self, model_dir: tk.Path, learning_rates: tk.Path, key: str, index: int = 0
+    ):
+        """
+
+        :param Path model_dir: model_dir output from a RETURNNTrainingJob
+        :param Path learning_rates: learning_rates output from a RETURNNTrainingJob
+        :param int index: index of the sorted list to access, 0 for the lowest, -1 for the highest score
+        :param str key: a key from the learning rate file that is used to sort the models
+            e.g. "dev_score_output/output_prob"
+        """
+        super().__init__(model_dir, learning_rates, key, index)
+        self._out_model_dir = self.output_path("model", directory=True)
+
+        # Note: checkpoint.index (without epoch number) is only allowed be cause we are symlinking
+        # to a file that contains an epoch number, as RETURNN will resolve symlinks automatically and
+        # then check for the epoch number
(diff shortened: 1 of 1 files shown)


albertz:
> I personally do not like the usage of `eval`

Note that unpickling is just equivalent to `eval` in expressiveness - it can run arbitrary Python code, by design (i.e. this is not a bug of unpickling but really designed this way). So, the only difference between `eval` and unpickling is the serialization format: pure Python code in one case, and some binary format in the other case.

You might argue both the pickle format and Python code are bad as representation formats. Maybe, but that's just the formats we use often. 

There is also `ast.literal_eval` but I think it would not work together with `EpochData`.

returnn/training.py
```diff
@@ -573,3 +573,179 @@ def hash(cls, kwargs):
...
+        :param Path model_dir: model_dir output from a RETURNNTrainingJob
+        :param Path learning_rates: learning_rates output from a RETURNNTrainingJob
+        :param int index: index of the sorted list to access, 0 for the lowest, -1 for the highest score
+        :param str key: a key from the learning rate file that is used to sort the models
+            e.g. "dev_score_output/output_prob"
+        """
+        super().__init__(model_dir, learning_rates, key, index)
+        self._out_model_dir = self.output_path("model", directory=True)
+
+        # Note: checkpoint.index (without epoch number) is only allowed be cause we are symlinking
```
michelwi, line 658:
```suggestion
        # Note: checkpoint.index (without epoch number) is only allowed because we are symlinking
```
typo

returnn/training.py
```diff
@@ -573,3 +573,179 @@ def hash(cls, kwargs):
...
+    def __init__(
+        self, model_dir: tk.Path, learning_rates: tk.Path, key: str, index: int = 0
+    ):
+        """
+
+        :param Path model_dir: model_dir output from a RETURNNTrainingJob
+        :param Path learning_rates: learning_rates output from a RETURNNTrainingJob
+        :param int index: index of the sorted list to access, 0 for the lowest, -1 for the highest score
+        :param str key: a key from the learning rate file that is used to sort the models
+            e.g. "dev_score_output/output_prob"
```
michelwi, line 653:
```suggestion
        :param str key: a key from the learning rate file that is used to sort the models
            e.g. "dev_score_output/output_prob"
        :param int index: index of the sorted list to access, 0 for the lowest, -1 for the highest score
```
change order to definition in constructor

albertz:
Note that in my pipeline, I do this different than having such jobs here:

- I do recog on some predefined epochs (e.g. via keep_epochs, or some other custom scheme), as soon as they are available
- And I additionally check the scores (learning rate file) for best N epochs for every existing key on dev and do recog on them as well.

This also means that the number of models where recog is performed on is variable. Sometimes it's just as much as keep_epochs, sometimes it's more. When there are multiple keys, the number of selected epochs is also variable. 

I also would not do recog twice for an epoch.

This must be via the `Job.update` mechanism, and selecting these epochs must actually run in the Sis manager process.

I think both these properties would not work with your job here.

- You cannot have a variable amount of selected epochs. You must have fixed that in your pipeline.
- How does it work together with predefined epochs (keep_epochs)? I assume it would perform recog twice, if it is a predefined epoch and also an epoch selected by the job.

Anyway, that's not an argument that these jobs should not be merged, or have no use. I just want to point that out. I will probably also prepare a PR soon with code which does it as I just described.

returnn/training.py
```diff
@@ -573,3 +573,179 @@ def hash(cls, kwargs):
...
+    def tasks(self):
+        yield Task("run", mini_task=True)
+
+    def run(self):
+        super().run()
+
+        try:
+            os.link(
+                os.path.join(
+                    self.model_dir.get_path(), "epoch.%.3d.index" % self.out_epoch.get()
```
michelwi, line 672: do you think hardcoding `"epoch.%.3d" % epoch` is general enough? 
What if we have a pre-train epoch or someone renamed the default name in their custom training job?

michelwi (commented):
I did not find a simpler solution than to write a custom learning_rates file parser. So the eval is fine with me. If you are concerned, you can maybe check if the `learning_rates` file is as expected before you run eval on it..

returnn/training.py
```diff
@@ -573,3 +573,179 @@ def hash(cls, kwargs):
...
+        :param int index: index of the sorted list to access, 0 for the lowest, -1 for the highest score
+        :param str key: a key from the learning rate file that is used to sort the models
+            e.g. "dev_score_output/output_prob"
+        """
+        super().__init__(model_dir, learning_rates, key, index)
+        self._out_model_dir = self.output_path("model", directory=True)
+
+        # Note: checkpoint.index (without epoch number) is only allowed be cause we are symlinking
+        # to a file that contains an epoch number, as RETURNN will resolve symlinks automatically and
+        # then check for the epoch number
```
albertz, line 660: I would reformulate that. There is no reason that RETURNN must behave this way. It's just that way for historical reasons. You could easily also change RETURNN if you would like to.

returnn/training.py
```diff
@@ -573,3 +573,179 @@ def hash(cls, kwargs):
...
 
         return super().hash(d)
+
+
+class GetBestEpochJob(Job):
+    """
+    Provided a RETURNN model directory and a score key, finds the best epoch.
+    The sorting is lower=better, so to acces the model with the highest values use negative index values (e.g. -1 for
+    the model with the highest score)
+
```
albertz, line 583: I would maybe add that "score" in RETURNN terminology actually (usually) refers to a loss, such as CE or FER.

albertz (approved):
Looks fine, despite my comments.

returnn/training.py
albertz: Ah yes I forgot about pretraining. We will probably not have that anymore in newer configs (because via `get_network` mechanism, you can also do pretraining, but models checkpoints are always called the same). But it would be nice to support them as well here. I would maybe add an argument `last_pretrain_epoch=0` or so.

JackTemaki: "Modern" pre-training logic does not alter the name any more... And I would say the first person who actually has a custom training job can add a new parameter if needed...

JackTemaki: Ah we answered simultaneously :) I think we can add such a parameter if actually needed. I think those jobs will be used for rather new setups. Also I never had the case (yet) in the models using the old pre-training logic went that far that the checkpoint was actually considered for "best selection".

returnn/training.py
JackTemaki: As long as this is not actually merged into RETURNN, I want to keep this as reminder for myself. If you do not want to have this text I can just link the issue.

> There is no reason that RETURNN must behave this way. It's just that way for historical reasons. You could easily also change RETURNN if you would like to.

But I also have to say I do not see how this text is in any way conflicting with those statements.

returnn/training.py
michelwi: ok, then I think we can keep it this way

michelwi (approved):

returnn/training.py
albertz: All what you write are technical internal details, only true for current RETURNN, dependent on actual somewhat buggy or strange behavior of RETURNN, and will likely become wrong in the future. From your comment, it reads as if you would always expect RETURNN to behave this way, or a reader of the comment could get this impression.

I would formulate it more in a way like: "Currently RETURNN has the flaw that it parses the epoch number out of the checkpoint filename. This still works if it is a symlink to such a filename. This will likely change in the future but we anyway want to support current and older RETURNN versions."

JackTemaki: Yes I already wrote a new text

PR #333, opened 2022-11-01
ReturnnTrainingJob, multi_node_slots option

albertz:
Depends on https://github.com/rwth-i6/sisyphus/pull/114.

### DIFF
--- a/returnn/training.py
+++ b/returnn/training.py
@@ -193,9 +193,15 @@ class ReturnnTrainingJob(Job):
 
         if (self.horovod_num_processes or 1) > (self.multi_node_slots or 1):
             assert self.horovod_num_processes % (self.multi_node_slots or 1) == 0
-            self.rqmt["cpu"] *= self.horovod_num_processes // (self.multi_node_slots or 1)
-            self.rqmt["gpu"] *= self.horovod_num_processes // (self.multi_node_slots or 1)
-            self.rqmt["mem"] *= self.horovod_num_processes // (self.multi_node_slots or 1)
+            self.rqmt["cpu"] *= self.horovod_num_processes // (
+                self.multi_node_slots or 1
+            )
+            self.rqmt["gpu"] *= self.horovod_num_processes // (
+                self.multi_node_slots or 1
+            )
+            self.rqmt["mem"] *= self.horovod_num_processes // (
+                self.multi_node_slots or 1
+            )
 
     def _get_run_cmd(self):
         run_cmd = [
--- a/returnn/training.py
+++ b/returnn/training.py
@@ -185,16 +185,25 @@ class ReturnnTrainingJob(Job):
         }
 
         if self.pe_num_processes:
-            assert self.horovod_num_processes and self.horovod_num_processes >= self.pe_num_processes
+            assert self.horovod_num_processes
+            assert self.horovod_num_processes >= self.pe_num_processes
             assert self.horovod_num_processes % self.pe_num_processes == 0
         if (self.horovod_num_processes or 1) > (self.pe_num_processes or 1):
             assert self.horovod_num_processes % (self.pe_num_processes or 1) == 0
-            self.rqmt["cpu"] *= self.horovod_num_processes // (self.pe_num_processes or 1)
-            self.rqmt["gpu"] *= self.horovod_num_processes // (self.pe_num_processes or 1)
-            self.rqmt["mem"] *= self.horovod_num_processes // (self.pe_num_processes or 1)
+            self.rqmt["cpu"] *= self.horovod_num_processes // (
+                self.pe_num_processes or 1
+            )
+            self.rqmt["gpu"] *= self.horovod_num_processes // (
+                self.pe_num_processes or 1
+            )
+            self.rqmt["mem"] *= self.horovod_num_processes // (
+                self.pe_num_processes or 1
+            )
 
         if self.pe_num_processes:
-            self.rqmt.setdefault("qsub_args", []).extend(["-pe", "mpi", str(self.pe_num_processes)])
+            self.rqmt.setdefault("qsub_args", []).extend(
+                ["-pe", "mpi", str(self.pe_num_processes)]
+            )
 
     def _get_run_cmd(self):
         run_cmd = [
--- a/returnn/training.py
+++ b/returnn/training.py
@@ -97,7 +97,7 @@ class ReturnnTrainingJob(Job):
         mem_rqmt=4,
         cpu_rqmt=2,
         horovod_num_processes=None,
-        parallel_tasks=None,
+        multi_node_slots=None,
         returnn_python_exe=None,
         returnn_root=None,
     ):
@@ -116,7 +116,7 @@ class ReturnnTrainingJob(Job):
         :param int|float mem_rqmt:
         :param int cpu_rqmt:
         :param int horovod_num_processes: If used without pe_num_processes, then single node, otherwise multi node.
-        :param int parallel_tasks: enables SGE parallel environment, i.e. adds `-pe mpi N` to the `qsub` args.
+        :param int multi_node_slots: multi-node multi-GPU training. See Sisyphus rqmt documentation.
             Currently only with Horovod,
             and horovod_num_processes should be set as well, usually to the same value.
             See https://returnn.readthedocs.io/en/latest/advanced/multi_gpu.html.
@@ -137,7 +137,7 @@ class ReturnnTrainingJob(Job):
             returnn_root if returnn_root is not None else gs.RETURNN_ROOT
         )
         self.horovod_num_processes = horovod_num_processes
-        self.parallel_tasks = parallel_tasks
+        self.multi_node_slots = multi_node_slots
         self.returnn_config = ReturnnTrainingJob.create_returnn_config(**kwargs)
 
         stored_epochs = list(range(save_interval, num_epochs, save_interval)) + [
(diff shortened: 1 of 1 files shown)


returnn/training.py
```diff
@@ -177,10 +184,26 @@ def __init__(
             "time": time_rqmt,
         }
 
-        if self.use_horovod:
-            self.rqmt["cpu"] *= self.horovod_num_processes
-            self.rqmt["gpu"] *= self.horovod_num_processes
-            self.rqmt["mem"] *= self.horovod_num_processes
+        if self.pe_num_processes:
```
JackTemaki, line 187: There should be an assert that checks if Sisyphus is actually configured for the SGE, and not Slurm or local execution.

albertz: I'm not sure we really have access to the engine at this point?

Maybe this check should be elsewhere? E.g. in the engine, assert that there are no `qsub` args?

Or maybe `rqmt` should be extended by an extra `pe` flag or so?

I'm a bit afraid that every solution will have drawbacks and not really be generic and later cause problems.

michelwi: well we might want to do multi-node multi-gpu training also at places where Slurm is used for job management.

so maybe 
* assert that with pe_num_processes > 1 not local engine is configured
* add handling for slurm (`rqmt["sbatch_args"]`) if slurm is configured.

(Ok, Slurm can be added in another PR)

returnn/training.py
```diff
@@ -132,6 +138,7 @@ def __init__(
         )
         self.use_horovod = True if (horovod_num_processes is not None) else False
```
michelwi, line 139: I think we don't need this member any more.

returnn/training.py
```diff
@@ -177,10 +184,26 @@ def __init__(
...
+            )
+            self.rqmt["gpu"] *= self.horovod_num_processes // (
+                self.pe_num_processes or 1
+            )
+            self.rqmt["mem"] *= self.horovod_num_processes // (
+                self.pe_num_processes or 1
+            )
+
+        if self.pe_num_processes:
+            self.rqmt.setdefault("qsub_args", []).extend(
```
michelwi, line 204: maybe we can add a new requirement category to sisyphus (e.g. rqmt["pe"]) and then each engine would implement it accordingly? 

@critias

returnn/training.py
albertz: I have no idea how this works in Slurm. `sbatch_args` is specific to Slurm?

So in SGE, parallel execution means that the job is executed with the same rqmt on multiple nodes (but I think there are many possibilities on the exact scheme of that, or any constraints for it). Does Slurm have the same, or is the logic different there?

If it has the same, and this is maybe similar in most engines, we can add some `pe_num_procs` or so to `rqmt`, and then each engine should properly check and handle that.

To your specific suggestions: I think at this point we don't have access to the engine, or at least I'm not sure. Even if we have, I'm not sure we should make a job dependent on the engine, or whether the job should be generic and put such information into rqmt and then the engine should handle it, or throw an exception.

returnn/training.py
albertz: Yep, this part here is very SGE specific, which is bad. Other engines could maybe assert that no `qsub_args` are set in rqmt, but then it would only work with SGE.

returnn/training.py
```diff
@@ -252,6 +275,19 @@ def _relink(src, dst):
         os.link(src, dst)
 
     def run(self):
+        if self.pe_num_processes:
+            # Some useful debugging:
+            if "PE_HOSTFILE" in os.environ:
```
albertz, line 280: Note: Also this is SGE specific. Not sure if this is a problem though. Maybe I just should extend the comment.

returnn/training.py
albertz: Ok, but as I argued, I don't think checking this here and making the job engine-dependent is a good idea. So you disagree?

JackTemaki: But we can also add a helper function which has a more general coverage, like checking that `engine()` is actually returning an `EngineSelector` and otherwise directly testing for the return value, someting like:
```python
def engine_is_sge():
    engine = gs.engine()
    if instance(engine, EngineSelector):
        engine = gs.engine().get_used_engine_by_rqmt({})
    return isinstance(engine, SonOfGridEngine)
```

michelwi: > Ok, but as I argued, I don't think checking this here and making the job engine-dependent is a good idea. So you disagree?

I do agree with @albertz here, the job schould be general and backend agnostic.

albertz: I'm not talking about making such check more generic. My argument is that the job should better be completely independent from the engine, not check anything about the engine. Rather, it should set some generic flag to `rqmt`, and then any engine should check later whether that flag can be handled and otherwise throw an exception.

JackTemaki: > So you disagree?

No, the code can also be in the `Task` itself. I am just saying we should not have this without checks.

returnn/training.py
```diff
@@ -189,7 +211,7 @@ def _get_run_cmd(self):
             self.out_returnn_config_file.get_path(),
         ]
 
-        if self.use_horovod:
+        if self.horovod_num_processes:
             run_cmd = [
```
michelwi, line 215: Does the `run_cmd` not need information from the `os.environ["PE_HOSTFILE"]` to decide to which nodes to connect?

I remember that when running multi-gpu multi-node training manually, I provided the option `-H localhost,<other_host_ip>`. 

Ah, I see there is also the option `--hostfile`.

returnn/training.py
JackTemaki: > My argument is that the job should better be completely independent from the engine

Totally agree as well. Then I am not sure what we are discussing here, it was your proposal to have it SGE specific. Then lets move this discussion over to Sisyphus on the engine independent "multi-node" parameter to add.

albertz: So it seems like Slurm `sbatch` is somehow similar. Usually `mpirun` should know about the hosts automatically. SGE somehow tells `mpirun`, and I assume Slurm does the same.

I'm not so sure about the Slurm terminology. A Slurm node (`-N`) is a separate computer/hostname, as in SGE? SGE PE does not require that they are run on separate SGE nodes but they could also get scheduled on the same SGE node, wherever there are free resources. So Slurm "task" (`-n`) is actually like SGE PE?

So, how should we call the option in rqmt? `num_parallel_tasks`? And then the Slurm engine and SGE engine would handle that, and other engines would throw an exception.

albertz: > it was your proposal to have it SGE specific.

No? I never said that. You argued for that ("There should be an assert that checks if Sisyphus is actually configured for the SGE"), I argued all the time against that, that we should not check or access the engine here.

returnn/training.py
michelwi: Ok, thanks for clarification. Yes, I was running manually on just two nodes outside of a cluster (and outside of sisyphus)

albertz: I added a comment.

returnn/training.py
JackTemaki: Okay, again from the beginning:

 - I am against having SGE specific code here at all. You added code and even a flag here that is SGE specific
 - I thought if you put SGE specific code here, there needs to be at least a check that SGE is actually used, this is all I argued.
 - Also yes, checking for sbatch_args and qsub_args in the Task itself makes sense.
 
But adding an independent rqmt flag (e.g. called just "nodes" or so) would make sense as well, instead of having SGE specific code here in the Training job.

albertz: > even a flag here that is SGE specific

It's not really SGE specific. Or maybe only the name. But I actually did not know, I thought maybe it's called the same for other engines. I asked already, what other name do you suggest?

I'm not working on a Sis PR where the flag is called `parallel_tasks` in the `rqmt`. Probably the argument here in the job then also should be renamed accordingly.

> checking ... in the Task itself

What do you mean by that? Which code exactly? In the `Task` class? Why there? So far we are discussing here about either the job (`ReturnnTrainingJob`) or the engine (e.g. SGE `SonOfGridEngine`).

albertz: Sis PR here: https://github.com/rwth-i6/sisyphus/pull/114

albertz: Ok done.

returnn/training.py
albertz: Done. https://github.com/rwth-i6/sisyphus/pull/114

albertz:
Ok, https://github.com/rwth-i6/sisyphus/pull/114 is merged, so this here is ready?

michelwi (approved):

albertz:
Can someone review?

returnn/training.py
```diff
@@ -177,10 +183,25 @@ def __init__(
...
+            assert self.horovod_num_processes >= self.multi_node_slots
+            assert self.horovod_num_processes % self.multi_node_slots == 0
+            self.rqmt["multi_node_slots"] = self.multi_node_slots
+
+        if (self.horovod_num_processes or 1) > (self.multi_node_slots or 1):
+            assert self.horovod_num_processes % (self.multi_node_slots or 1) == 0
+            self.rqmt["cpu"] *= self.horovod_num_processes // (
+                self.multi_node_slots or 1
+            )
+            self.rqmt["gpu"] *= self.horovod_num_processes // (
```
JackTemaki, line 199: Is this expect to have non-fitting ratios? So what should happen if `self.horovod_num_processes` is 3 but `self.multi_node_slots` is 2? Does that even make sense? I assume you can not get 2 gpus on node and 1 on the other. And in this case we now get 1 GPU requested per node right?

albertz: There is an assert for it above.

JackTemaki: ah did not see that

JackTemaki (approved):

returnn/training.py
michelwi: >I assume you can not get 2 gpus on node and 1 on the other. 

Technically, when you use horovod_num_processes==multi_node_slots==3 then two slots can be on the same node. But each would get the indicated rqmt individually.

albertz: Note that we don't really have any specific logic at all for this. We just pass on these numbers to SGE `-pe` and `mpirun` respectively. I assume there are countless of possible configurations on how exactly it would then do the distribution.

PR #334, opened 2022-11-02
Option to mark special steps in Monophone Training

Atticus1806:
Extend align_and_accumulate_sequence to manually define steps that are kept.
Can be done in the training args via e.g.:  "keep_steps": [0, 1]
Also adds some documentation

### DIFF
--- a/meta/mm_sequence.py
+++ b/meta/mm_sequence.py
@@ -7,7 +7,7 @@ __all__ = [
     "first_split_acc_then_align_split_acc",
 ]
 import os
-from typing import Optional, Dict, List
+from typing import Optional, Union, List, Tuple
 
 from i6_core.mm.alignment import AlignmentJob, AMScoresFromAlignmentLogJob
 from i6_core.mm.flow import FlowNetwork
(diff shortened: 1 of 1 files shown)


Atticus1806:
Is there a proper way to test this without the other PR being merged?

michelwi (commented):
generally ok.

I just wonder if it would be more intuitive interface to change the existing `mark_accumulate` and `mark_align` to be of type `Union[bool, List[int]]` and then `True` marks all, List of int marks only the selected and `False` marks none.
this way the user does not have to remember what the order and number of accumulates and aligns is.

Atticus1806:
This was my first intuition, but left me with one big question: For accumulate would the numbering be "per align" or in total? 
Since the structure is: align, n x accum, align, n x accum... 
So would 1 mark every first accum or just the (global) first?

michelwi:
> This was my first intuition, but left me with one big question: For accumulate would the numbering be "per align" or in total? Since the structure is: align, n x accum, align, n x accum... So would 1 mark every first accum or just the (global) first?

If you set the option `mark_accumulate=True`, then the last accumulate of each align iteration is marked. So for `mark_accumulate=[3,7]` I would define to mark the last accumulate of the third and seventh iteration.

This would then not allow to mark some intermediate accumulation result. Would this be something that you would need?

Atticus1806:
>This would then not allow to mark some intermediate accumulation result. Would this be something that you would need?

Yes I was specifically looking into marking the first step but decided to try to make it more general, so yes I would need that.

michelwi:
> Yes I was specifically looking into marking the first step but decided to try to make it more general, so yes I would need that.

Ok , then instead of 

> change the existing mark_accumulate and mark_align to be of type Union[bool, List[int]] and then True marks all, List of int marks only the selected and False marks none.

mark_accumulate would be `List[List[int]]` with the shape (n_align, n_accum) ?

Atticus1806:
So in general this would be someting we could do. It would work for me. But I am wondering, for example in my case, whether its a bit tedious to have to have a list of shape (75, 1) in my case just to mark the first accumulate. If you say though this is the best generalization then I would go with it. Maybe opinions from @JackTemaki @christophmluscher?
For align this ofc. is easy there its just of shape (n_align,).

michelwi:
How about `List[Tuple[int,int]]` where each tuple indexes one accumulation to be marked. e.g. `[(3,0),(7,0)]` would index the first accumulate of the fourth and eighth align each?

Atticus1806:
Sounds good, implemented and tested it now, I like the style.

JackTemaki (approved):

michelwi (approved):

PR #335, opened 2022-11-08
Add and remove speaker Tags

Atticus1806:
Adds two jobs which can either add or remove speaker tags from the given bliss.
We need it for some exps and figured this might be something general enough for core.

### DIFF
--- a/corpus/convert.py
+++ b/corpus/convert.py
@@ -1,4 +1,6 @@
 __all__ = [
+    "CorpusAddSpeakerTagsFromMappingJob",
+    "CorpusRemoveSpeakerTagsJob",
     "CorpusReplaceOrthFromReferenceCorpus",
     "CorpusReplaceOrthFromTxtJob",
     "CorpusToStmJob",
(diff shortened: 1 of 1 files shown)


corpus/convert.py
```diff
@@ -21,6 +25,74 @@
...
+            mapping = pickle.load(f)  # type: Dict
+
+        bliss.speakers = collections.OrderedDict()
+        for id in set(mapping.values()):
+            speaker = corpus.Speaker()
+            speaker.name = str(id)
+            bliss.add_speaker(speaker)
+        for recording in bliss.all_recordings():
+            for segment in recording.segments:
+                segment.speaker_name = mapping[segment.fullname()]
```
michelwi, line 61: would we maybe want to have a "default_speaker" that is used if one segment is not in the mapping?

michelwi (approved):

corpus/convert.py
Atticus1806: how would you choose this? Just take "a random one"? 
If so I think this should not be hidden but rather be a flag, maybe even given as id? 
Smth like: `default_speaker_id: Optional[int] = None`
and then 
`segment.speaker_name = mapping[segment.fullname()] if segment.fullname() in mapping else self.default_speaker_id`
maybe?

michelwi: more like create a dummy speaker here in the job. 

Or maybe we could just not assign a speaker if the mapping is empty.

But then I think it would be fair to expect that all segments are defined in the mapping and that the job crashes if one segment is missing.. 


What I want to say is: I approve of your PR, but did you think about what happens when `mapping[segment.fullname()]` does not exist?  (and "then it crashes" is a valid answer xD)

Atticus1806: I guess its a design choice I dont have a strong opinion about. For me I would assume its always well defined so crashing is fine, but I can also see this default flag. 
Creating a dummy speaking would be something I would personally not like since that would hide potential errors in the mapping.

Atticus1806: We can leave it like this but since I am working on this Job rn, if you say an "add_dummy" flag or dummy integer would be something you would like to have in this job I would just add it right now. But else we can just leave it to the person who wants it when they want it.

michelwi: I also don't have a strong opinion and am fine with the current implementation

JackTemaki (approved):

PR #338, opened 2022-11-16
added fairseq manifest creation job

LabChameleon:
A job that computes a manifest file for a given directory of audio files. These manifests are commonly used for Fairseq experiments.

@vieting @christophmluscher

### DIFF
--- a/fairseq/manifest.py
+++ b/fairseq/manifest.py
@@ -75,9 +75,7 @@ class CreateManifestJob(Job):
 
     def run(self, task_id):
         with open("%d.tsv" % task_id, "w") as f:
-            for path in glob.iglob(
-                self.audio_dir_paths[task_id - 1].get_path() + "/*", recursive=True
-            ):
+            for path in glob.iglob(self.audio_dir_paths[task_id - 1].get_path() + "/*", recursive=True):
                 frames = soundfile.info(path).frames
                 rel_path = os.path.relpath(path, self.common_dir)
                 f.write("{}\t{}\n".format(rel_path, frames))
@@ -115,9 +113,7 @@ class SplitTrainCvDataJob(Job):
         rand = random.Random(self.seed)
         with open(self.tsv_file, "r") as f_in, open(
             os.path.join(self.out_manifest_path, "train.tsv"), "w"
-        ) as f_train, open(
-            os.path.join(self.out_manifest_path, "valid.tsv"), "w"
-        ) as f_valid:
+        ) as f_train, open(os.path.join(self.out_manifest_path, "valid.tsv"), "w") as f_valid:
             for line in f_in:
                 # common path
                 if "\t" not in line:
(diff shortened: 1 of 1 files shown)


vieting:
Thanks @michelwi for your extensive comments! Regarding the possibility to split it up into multiple steps including some using the `PipelineJob`: this can most likely be done and would be more consistent with the sisyphus guideline of breaking things into small units. @dierkes-j is currently focusing on making his pipelines available in the main branches while making sure that they are still working, i.e., hashes stay consistent. This would not work with your suggestion I guess. If we say this job should not be merged here because it is too specific (which there are valid arguments for), I guess the job will stay in i6_experiments and we should consider this option for future setups.

fairseq/manifest.py
```diff
@@ -0,0 +1,217 @@
...
+                ]
+            )
+            upsampling_factors = np.linalg.solve(w_mat, b_vec)
+            # assert all entries larger than 1 because we only want to
+            # upsample and never downsample
+            assert min(
+                upsampling_factors >= 1
+            )
+
+            upsampling_factors = np.insert(upsampling_factors, max_length_index, 1.0)
```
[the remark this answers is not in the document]
Judyxujj: I finally understood the math here. So in [paper](https://arxiv.org/abs/2006.13979) , it is suggested to upsample the low resource language by `p_l**alpha`, p_l is the calculated by dividing the total number of hours of language l by total number of hours if all languages. For example the language probability is [p1, p2, p3, p4] and alpha is 0.5, after upsampling the probability would be  [p1**0.5, p2**0.5, p3**0.5, p4**0.5], the upsample factor is [p1**0.5/p1, p2**0.5/p2, p3**0.5/p3, p4**0.5/p4]. But here we don't want to upsample the largest language, so the upsample factor is normalised by the minimum. So it would be` min = [p1**0.5/p1, p2**0.5/p2, p3**0.5/p3, p4**0.5/p4]` and final result is` [(p1**0.5/p1)/min, (p2**0.5/p2)/min, (p3**0.5/p3)/min, (p4**0.5/p4)/min]`

Judyxujj:
I have split the job into multiple small jobs based on the suggestion from @michelwi

fairseq/manifest.py
```diff
@@ -0,0 +1,236 @@
...
+
+        self.out_tsv_file = self.output_path("train.tsv")
+
+        self.rqmt = {"time": time_rqmt, "mem": mem_rqmt, "cpu": cpu_rqmt}
+
+    def tasks(self):
+        yield Task("run", rqmt=self.rqmt)
+
+    def run(self):
+        assert self.alpha is not None and self.alpha < 1
```
curufinwe, line 169: this assertion should be done in the constructor

fairseq/manifest.py
```diff
@@ -0,0 +1,236 @@
...
+        all_train_data = defaultdict(list)
+        with open(self.train_tsv_file, "r") as f_in, open(
+            self.out_tsv_file, "w"
+        ) as f_out:
+            while True:
```
curufinwe, line 174: see above

curufinwe (changes requested):

curufinwe (approved):

Judyxujj (approved):

PR #339, opened 2022-11-16
added fairseq training job

dierkes-j:
A job for performing trainings with Fairseq.

@vieting @christophmluscher


fairseq/training.py
```diff
@@ -0,0 +1,434 @@
...
+        ).get("max_epoch", None):
+            run_cmd += ["optimization.max_epoch=" + str(self.max_epoch)]
+
+        if self.use_cache_manager:
+            run_cmd += ["task.data=" + self.out_cached_audio_manifest.get_path()]
+
+        if self.fairseq_root is not None:
+            sys.path.insert(0, self.fairseq_root)
+            hydra_train_entry = self.fairseq_root + "fairseq_cli/hydra_train.py"
+            run_cmd.insert(0, tk.uncached_path(hydra_train_entry))
```
michelwi, line 418: shun `tk.uncached_path`. Once you decide to only use `tk.Path` you can get `get_path()` to convert to `str`

fairseq/training.py
```diff
@@ -0,0 +1,434 @@
...
+                                epoch = int(split[split.index("epoch") + 1])
+                                score = float(split[split.index("valid_loss") + 1])
+                                error = 1 - float(
+                                    split[split.index("valid_accuracy") + 1]
+                                )
+                            except ValueError:
+                                continue
+                            valid_score[epoch] = score
+                            valid_error[epoch] = error
+                            i += 1
```
michelwi, line 313: Are you changing the loop variable inside the loop? How C of you.. xD

fairseq/training.py
```diff
@@ -0,0 +1,434 @@
...
+                k,
+            )
+            for k in stored_epochs
+            if k in self.keep_epochs
+        }
+        self.out_cached_audio_manifest = self.output_path(
+            "cached_audio_manifest", directory=True
+        )
+        self.out_plot_se = self.output_path("score_and_error.png")
+        self.out_plot_lr = self.output_path("learning_rate.png")
```
michelwi, line 178: how about using a vector graphics format rather than raster graphics?

michelwi (changes requested):
some suggestions inline.

My request changes explicitly only refers to the use of `str` instead of `tk.Path` and the use of `tk.uncached_path`.

fairseq/training.py
```diff
@@ -0,0 +1,434 @@
...
+
+        for cur in os.walk(directory):
+            dir_path = cur[0]
+            files = cur[2]
+            if "hydra_train.log" in files:
+                with open(f"{dir_path}/hydra_train.log", "r") as f:
+                    lines = f.readlines()
+                    for i in range(len(lines) - 1):
+                        line = lines[i]
+                        if 'begin validation on "valid" subset' in line:
```
nishthajain1611, line 299: it might be better to change this from
`if 'begin validation on "valid" subset' in line: `  to->  `if 'begin validation on' in line:`
because the name of the valid subset can be anything, eg 'dev'

fairseq/training.py
```diff
@@ -0,0 +1,434 @@
...
+                            split = re.sub('[{,":]', "", lines[i + 1]).split(" ")
+                            #  Exception handling for cases in which accuracy and loss 
+                            #  are not logged in next line
+                            try:
+                                epoch = int(split[split.index("epoch") + 1])
+                                score = float(split[split.index("valid_loss") + 1])
```
nishthajain1611, line 305: Similarly, it can be "dev_loss" based on the name of the validation set. ie if validation set is "valid"then it is "valid_loss" but if the name is "dev" then it is "dev_loss"

fairseq/training.py
```diff
@@ -0,0 +1,434 @@
...
+                                error = 1 - float(
+                                    split[split.index("valid_accuracy") + 1]
```
nishthajain1611, line 307: for my use case,ie audio-finetuning, there is no "valid_accuracy" or "train_accuracy"

here is an example of the dev log file line:
[2022-11-02 11:38:35,221][dev][INFO] - {"epoch": 501, "dev_loss": "90.382", "dev_ntokens": "608.174", "dev_nsentences": "4.01144", "dev_nll_loss": "0.596", "dev_uer": "9.178", "dev_wer": "22.85", "dev_raw_wer": "22.85", "dev_wps": "5192.6", "dev_wpb": "608.2", "dev_bsz": "4", "dev_num_updates": "20000", "dev_best_wer": "22.85"}

and an example of the train log line:
[2022-11-02 11:39:28,709][train][INFO] - {"epoch": 501, "train_loss": "174.778", "train_ntokens": "20434.2", "train_nsentences": "125.719", "train_nll_loss": "1.075", "train_wps": "2207", "train_ups": "0.11", "train_wpb": "20434.2", "train_bsz": "125.7", "train_num_updates": "20000", "train_lr": "2.5e-06", "train_gnorm": "141.885", "train_loss_scale": "8", "train_train_wall": "137", "train_gb_free": "6.1", "train_wall": "0"}

michelwi: Well, in this case the try-except block will run into the `ValueError` case and then silently deliver empty arrays for `epoch`, `score`, and `error`. So I guess the job will work but the plots will just be empty.

We could do separate try-except blocks for loss and accuracy so you can have the loss plot even when no accuracy is reported.

fairseq/training.py
michelwi: @nishthajain1611 fyi, the rectangle with the +/- sign lets you insert code suggestions like this one to make more clear what you want to change (and one can directly apply it if one wants.)
```suggestion
                        if 'begin validation on' in line:
```

fairseq/training.py
michelwi: good point. If the line contains a printed dict as @nishthajain1611 indicated below, then you could maybe try to parse it like
```
epoch_dict = eval(line[line.index("{"):])
epoch = epoch_dict[epoch]
losses = {k:v for k,v in epoch_dict.items() if k.endswith("_loss")}
erros = {k:v for k,v in epoch_dict.items() if k.endswith("_error")}
```

fairseq/training.py
```diff
@@ -0,0 +1,434 @@
...
+class FairseqHydraConfig:
+    """
+    An object that manages a Fairseq hydra config (inspired by the ReturnnConfig).
+    """
+
+    def __init__(self, config_dict, *, yaml_prefix=""):
+        """
+        :param dict config_dict: Contains the information which is needed for fairseq-hydra-train. Will be converted and dumped into a .yaml
+        :param str yaml_prefix: Prefix which should be written to the beginning of the config, for example "# @package _group_"
+        """
```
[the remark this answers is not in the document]
Judyxujj: as far as I understood, the yaml_prefix here just aims to add an @package directive as required here `https://hydra.cc/docs/upgrades/0.11_to_1.0/adding_a_package_directive/#adding-an-package-directive`. It is different from the prolog or epilog in returnn @michelwi 
example of hydra config
```
# @package _group_
checkpoint:
    no_epoch_checkpoints: false
    restore_file: checkpoint_last.pt
    save_dir: output/checkpoints
    save_interval: 5
common:
    fp16: true
    log_format: json
    log_interval: 200
criterion:
    _name: wav2vec
    infonce: true
    log_keys:
    - prob_perplexity
    - code_perplexity
```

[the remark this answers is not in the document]
michelwi: No. Actually in this case if only ever one "@ package" directive should be used, then maybe remove the (arbitrary) prefix and instead introduce a `package_name` parameter. 

then in the write function replace 
```
 config_yaml = self.yaml_prefix + "\n" + config_yaml
```
with 
```
 config_yaml = f"# @package {self.package_name}\n" + config_yaml
```

But I have no hard feelings about this point..

fairseq/training.py
```diff
@@ -0,0 +1,434 @@
...
+            Use None for each save_interval-th epoch
+        :param int|float time_rqmt: Overall time requirements
+        :param int|float mem_rqmt: Memory requirements (per GPU)
+        :param int cpu_rqmt: Required number of CPUs (per GPU)
+        :param int gpu_rqmt: Number of required GPUs
+        :param Path|str python_exe: File path to the executable for running python
+        :param Path|str fairseq_hydra_exe: File path to the python executable for running fairseq-hydra-train.
+            (usually in the same folder as python exe '.../bin/')
+        :param Path|str fairseq_root: File path to the fairseq git for alternative call of fairseq-hydra-train
+            (no need to install fairseq here)
```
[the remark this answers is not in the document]
Judyxujj: I think we do not need both ways here, I would prefer specifying the fairseq_root since we don't need to install the fairseq @michelwi which way would you suggest?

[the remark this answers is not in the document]
michelwi: yes, your suggestion sounds reasonable. This way we can use a CloneGitRepositoryJob.

fairseq/training.py
```diff
@@ -0,0 +1,434 @@
...
+        ), "save_interval has to be set explicitly or via fairseq_hydra_config"
+        max_epoch_config = fairseq_hydra_config.config_dict.get("optimization", {}).get(
+            "max_epoch", None
+        )
+        if max_epoch is not None and max_epoch_config is not None:
+            logging.warning("'max_epoch' {}: {}".format(warning_text, max_epoch))
+        self.max_epoch = max_epoch or max_epoch_config
+        assert (
+            self.max_epoch is not None
+        ), "max_epoch has to be set explicitly or via fairseq_hydra_config"
```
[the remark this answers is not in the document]
Judyxujj: I think the `save_interval` and `max_epoch` are already set in fairseq_hydra_config, here if `save_interval` and `max_epoch` are given as input, then they will overwrite the previous values in fairseq_hydra_config. Would it be better to add the `post_config` as ReturnnConfig for FairseqHydraConfig and put these two values in the `post_config`?

fairseq/training.py
```diff
@@ -0,0 +1,434 @@
...
+                    ) as cached_audio_manifest_file:
+                        cached_audio_manifest_file.write("\n".join(manifest_lines))
+
+        my_env = os.environ
+        if self.fairseq_root is not None:
+            my_env["PYTHONPATH"] = self.fairseq_root
+        sp.check_call(self._get_run_cmd(), env=my_env)
+
+    def plot(self):
+        directory = "./outputs"
```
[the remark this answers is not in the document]
Judyxujj: Yes, the outputs folder will be generated automatically, but actually the this folder is for logging outputs. As in Returnn training job, the loggings are also saved in the training job work folder. But here the folder name `outputs` is very confusing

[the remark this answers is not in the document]
michelwi: ok. Some jobs do make their logs an output. But we don't have to do it. I am fine either way

fairseq/training.py
Judyxujj: @nishthajain1611  do you have two separate log files?

Judyxujj:
I have modified this PR based on the previous comments from @michelwi , now the plotting should be more general. Also,   post config and check_consistency function are now added to the FairseqHydraConfig.  The arguments `save_interval` and `max_epoch` are added to the post config, so they won't be hashed. I still have issues with the black formatting checks, I use the exactly the same black version `black==22.3.0`, but I get different results as in checks.

fairseq/training.py
```diff
@@ -88,69 +148,46 @@ def __init__(
...
             if fairseq_python_exe is not None
             else getattr(gs, "FAIRSEQ_PYTHON_EXE", None)
         )
-        self.fairseq_hydra_exe = fairseq_hydra_exe
         self.fairseq_root = fairseq_root
-        # We assume that only one of the two possible entry points is given as an input
-        assert (self.fairseq_root is not None) ^ (self.fairseq_hydra_exe is not None)
+        assert self.fairseq_root is not None
         if self.fairseq_root is not None:
             assert self.fairseq_python_exe is not None
```
michelwi:
```suggestion
```
assertion and if condition are always true based on code above

fairseq/training.py
```diff
@@ -71,15 +132,14 @@ def __init__(
...
+        max_epoch=1,
+        save_interval=1,
         keep_epochs=None,
         time_rqmt=4,
         mem_rqmt=4,
         cpu_rqmt=2,
         gpu_rqmt=1,
         fairseq_python_exe=None,
-        fairseq_hydra_exe=None,
         fairseq_root=None,
```
michelwi, line 143: below you assert that `fairseq_root is not None`, but here the default is None. Do you suggest that there should be no default then? In this case just remove the default. (Because of the `*,` above it should be allowed to not give any default here without reordering the arguments.
```suggestion
        fairseq_root,
```

fairseq/training.py
```diff
@@ -88,69 +148,46 @@ def __init__(
...
-            range(self.save_interval, self.max_epoch, self.save_interval)
-        ) + [self.max_epoch]
-        if keep_epochs is None:
-            self.keep_epochs = set(stored_epochs)
-        else:
-            self.keep_epochs = set(keep_epochs)
         self.fairseq_python_exe = (
             fairseq_python_exe
             if fairseq_python_exe is not None
             else getattr(gs, "FAIRSEQ_PYTHON_EXE", None)
```
michelwi, line 183: I would not encourage getting the path from the global settings. Maybe use system Python as a default?

fairseq/training.py
```diff
@@ -168,14 +208,13 @@ def __init__(
                 self.output_path("checkpoints/checkpoint{}.pt".format(k)),
                 k,
             )
-            for k in stored_epochs
-            if k in self.keep_epochs
+            for k in self.keep_epochs
         }
         self.out_cached_audio_manifest = self.output_path(
             "cached_audio_manifest", directory=True
         )
```
michelwi: I just noticed that this output will always exist, but the file is only written if `use_cache_manager` is True. This is not nice. could we do one of the following
1. only create the output if we use cache manager
```suggestion
        if use_cache_manager:
            self.out_cached_audio_manifest = self.output_path(
                "cached_audio_manifest", directory=True
            )
```
2. always create the file, but copy over the default manifest if we do not use the cache manager
3. never have this as an output but store it as a regular file in the work folder.

fairseq/training.py
```diff
@@ -400,35 +474,40 @@ def _get_run_cmd(self):
...
+            )
+        ]
+        run_cmd += [
+            "optimization.max_epoch="
+            + str(
+                self.fairseq_hydra_config.config_dict.get("optimization").get(
+                    "max_epoch"
+                )
+            )
+        ]
```
michelwi, line 492: do we still need this if it is anyway guaranteed to exist in the config file?

michelwi (commented):
Some more comments. In some instances I might be wrong, when there are some specific constraints on the contents of fairseq configs (e.g. they are always nested dicts of depths 2).
If this is the case, then please disregard my comments that go in that direction.

fairseq/training.py
```diff
@@ -5,56 +5,117 @@
...
-        self.yaml_prefix = yaml_prefix
+        self.post_config_dict = post_config_dict if post_config_dict is not None else {}
+        self.package_name = package_name
+        self.check_consistency()
+
+    @staticmethod
+    def update_nested_dict(dict1, dict2):
+        for g in dict2:
+            if g in dict1:
+                dict1[g].update(dict2[g])
```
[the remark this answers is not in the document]
Judyxujj: The wav2vec 2 model only has maximum depth of 2 for the nested dict. But in case of the future usage for other models, it would be safer if we also consider nested dict with maximum depth of more than 2.  I have updated this function

fairseq/training.py
```diff
@@ -5,56 +5,117 @@
...
+        for group in self.config_dict:
+            if isinstance(self.config_dict[group], dict):
+                for key in disallowed_in_config:
+                    assert self.config_dict[group].get(key) is None, (
+                        "please define %s only as parameter in the post_config_dict"
+                        % key
+                    )
+
+    def _sis_hash(self):
+        h = {"fairseq_hydra_config": self.config_dict}
```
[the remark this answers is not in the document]
Judyxujj: Yes, I think the package name should not be included since it would not affect the defined model

Judyxujj:
Thanks @michelwi for the useful comments, I have made some modifications based on the comments.

fairseq/training.py
```diff
@@ -0,0 +1,502 @@
...
+    def update_nested_dict(dict1, dict2):
+        for k, v in dict2.items():
+            if isinstance(v, collections.abc.Mapping):
+                dict1[k] = update_nested_dict(dict1.get(k, {}), v)
+            else:
+                dict1[k] = v
+        return dict1
+
+    def write(self, path: str):
+        path_corrected_config = self.config_dict.copy()
```
vieting, line 51: The name `path_corrected_config` does not yet seem to fit well here, right?

fairseq/training.py
```diff
@@ -0,0 +1,502 @@
...
+        FairseqHydraConfig.update_nested_dict(
+            path_corrected_config, self.post_config_dict
+        )
+
+        # recursively go through config dictionary to get all sisyphus paths inplace
+        path_corrected_config = util.instanciate_delayed(path_corrected_config)
+
+        config_yaml = yaml.dump(path_corrected_config)
+        # "# @package _group_" was written at the beginning in the example .yaml from fairseq:
```
vieting, line 60: I guess this comment might be removed, what do you think?

fairseq/training.py
```diff
@@ -0,0 +1,502 @@
...
+        for k, v in dict2.items():
+            if isinstance(v, collections.abc.Mapping):
+                dict1[k] = update_nested_dict(dict1.get(k, {}), v)
+            else:
+                dict1[k] = v
+        return dict1
+
+    def write(self, path: str):
+        path_corrected_config = self.config_dict.copy()
+        FairseqHydraConfig.update_nested_dict(
```
vieting, line 52: I'd prefer to use the returned value to improve the readability and make clear which variable changes.
```suggestion
        path_corrected_config = FairseqHydraConfig.update_nested_dict(
```

fairseq/training.py
```diff
@@ -0,0 +1,502 @@
...
+        """
+        assert isinstance(config_dict, dict)
+        assert isinstance(post_config_dict, dict) or post_config_dict is None
+        self.config_dict = config_dict
+        self.post_config_dict = post_config_dict if post_config_dict is not None else {}
+        self.package_name = package_name
+        self.check_consistency()
+
+    @staticmethod
+    def update_nested_dict(dict1, dict2):
```
vieting, line 42: Should we make this a general helper function? I don't think it's really related to the `FairseqHydraConfig`.

fairseq/training.py
```diff
@@ -0,0 +1,502 @@
...
+        fairseq_python_exe=None,
+        fairseq_root,
+        use_cache_manager=True,
+        zipped_audio_dir=None,
+    ):
+        """
+        :param FairseqHydraConfig fairseq_hydra_config:
+        :param list command_line_args: Additional command line arguments (starting with "--*"),
+            to configure the Fairseq-hydra task
+        :param int max_epoch: maximum number of epochs to run. Note that this value IS currently HASHED.
```
vieting, line 157: This is not hashed anymore, since it is moved to the post config, right?

fairseq/training.py
```diff
@@ -0,0 +1,502 @@
...
+            file.write(config_yaml)
+
+    def update(self, other):
+        """
+        updates a FairseqHydraConfig with another FairseqHydraConfig:
+          * config_dict, post_config_dict use dict.update
+        :param FairseqHydraConfig other:
+        """
+        FairseqHydraConfig.update_nested_dict(self.config_dict, other.config_dict)
+        FairseqHydraConfig.update_nested_dict(
```
vieting, line 73: I'd prefer to use the returned value to improve the readability and make clear which variable changes, see above.

vieting (changes requested):
Thanks for updating this! I added some small comments.

fairseq/training.py
Judyxujj: Yes, it is not very related to FairseqHydraConfig, so I use the static method here. But we could make this a general helper function, where would you suggest to put the general helper function to?

vieting: I don't have any strong feelings here. You could e.g. just make it a function outside the class.

michelwi: I would propose `util.py`

michelwi: but actually.. there must already be an existing implementation in the internet.. I would guess we are not the first to want to implement nested dict update..

michelwi: >  there must already be an existing implementation in the internet

Haha, I did a little research, found a nice implementation and noticed that you indeed found and used the same. Good job, nevermind my previous comment xD

Judyxujj: ah yes, I have searched in stack flow and found this implementation. I have put this function in util now xD

fairseq/training.py
```diff
@@ -0,0 +1,490 @@
...
+        """
+        updates a FairseqHydraConfig with another FairseqHydraConfig:
+          * config_dict, post_config_dict use dict.update
+        :param FairseqHydraConfig other:
+        """
+        self.config_dict = util.update_nested_dict(self.config_dict, other.config_dict)
+        self.post_config_dict = util.update_nested_dict(
+            self.post_config_dict, other.post_config_dict
+        )
+        self.package_name = other.package_name
```
michelwi, line 64:
```suggestion
        if other.package_name != "":
            self.package_name = other.package_name
```
proposal: update only if new name is not empty.

fairseq/training.py
```diff
@@ -0,0 +1,490 @@
...
+
+        fig, axs = plt.subplots(2, sharex=True, figsize=(12, 9))
+
+        color_index = 0
+        for k in train_loss.keys():
+            axs[0].plot(
+                train_loss[k].keys(),
+                train_loss[k].values(),
+                "o-",
+                color=colors[color_index],
```
michelwi, line 404: would you need some failsafe in case the color_index is larger than the defined colors? e.g.
```suggestion
                color=colors[color_index % len(colors)],
```

michelwi (commented):
found one more problem with the recent updates

fairseq/training.py
```diff
@@ -0,0 +1,490 @@
...
+        :param list[int]|set[int]|None keep_epochs: specify which checkpoints are kept in self.out_models.
+            Use None for each save_interval-th epoch
+        :param int|float time_rqmt: Overall time requirements
+        :param int|float mem_rqmt: Memory requirements (per GPU)
+        :param int cpu_rqmt: Required number of CPUs (per GPU)
+        :param int gpu_rqmt: Number of required GPUs
+        :param tk.Path fairseq_python_exe: File path to the executable for running python
+        :param tk.Path fairseq_root: File path to the fairseq git for alternative call of fairseq-hydra-train
+            (no need to install fairseq here)
+        :param bool use_cache_manager: enables caching of data given in the manifest with the i6 cache manager
```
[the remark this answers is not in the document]
vieting: We added it because other jobs that do caching use either RASR or RETURNN, both of which support caching. With fairseq, this is different. I'm not 100% sure about the details, but most likely it will directly read the files from python and certainly it does not know about the i6 cache manager. That's why we cached the files at the beginning of the `run` task and use an updated manifest with the cached paths when starting the fairseq training.

fairseq/training.py
```diff
@@ -0,0 +1,490 @@
...
+    def create_fairseq_hydra_config(
+        cls, fairseq_hydra_config, max_epoch, save_interval, **kwargs
+    ):
+        res = copy.deepcopy(fairseq_hydra_config)
+        config_dict = {}
+        post_config_dict = {
+            "optimization": {"max_epoch": max_epoch},
+            "checkpoint": {"save_interval": save_interval},
+        }
+        res.update(FairseqHydraConfig(config_dict, post_config_dict))
```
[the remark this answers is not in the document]
Judyxujj: ohh, yes. Nice catch!

fairseq/training.py
```diff
@@ -0,0 +1,491 @@
...
+
+        fig, axs = plt.subplots(2, sharex=True, figsize=(12, 9))
+
+        color_index = 0
+        for k in train_loss.keys():
+            axs[0].plot(
+                train_loss[k].keys(),
+                train_loss[k].values(),
+                "o-",
+                color=colors[color_index % len(colors)],
```
michelwi, line 405: The same comment applies to many other places here (415, 429, 439, from a quick glance) I think you should do it consistently for all.

Judyxujj: the rest are all updated now

fairseq/training.py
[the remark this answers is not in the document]
dthulke: Couldn't this be fixed by replacing the explicit calls to cf with `gs.file_caching` (and rename this arg to something like use_file_caching). Then whatever caching implementation the user specified in it's settings.py is used.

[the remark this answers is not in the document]
vieting: I'm not very familiar with `gs.file_caching`. In the case here, we have a text file with all audio paths inside that will be read by fairseq. I'm not sure if this would be handled by sisyphus settings.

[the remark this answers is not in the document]
michelwi: The function `gs.file_caching` takes as input a `str` (the original location) and returns a `str` (the cached location). In between the function does the caching. [Default implementation](https://github.com/rwth-i6/sisyphus/blob/e07c227fbcbcb2e6e5b6ad03ce5ff87c30a90026/sisyphus/global_settings.py#L107-L117) is to just return the original string.
It is invoked whenever one calls [`tk.Path.get_cached_path`](https://github.com/rwth-i6/sisyphus/blob/e07c227fbcbcb2e6e5b6ad03ce5ff87c30a90026/sisyphus/job_path.py#L188).

I think one could replace all calls to ```sp.check_output(["cf", something]).strip().decode("utf8")``` with ```gs.file_caching(something)``` except maybe the bundle file, which is a specialty of the i6 cache manager.


To clarify my point: I to not insist on changing it to be i6 independent, I merely start the discussion if we maybe think this is more elegant..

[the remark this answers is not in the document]
vieting: Clearly this is not a very elegant solution. However, fairseq reads the audio files from a txt file. So we cache the audios and modify the txt file to contain the cached paths. It is not clear to me how sisyphus could interfere in this process of fairseq reading the filenames from the txt file. Or how we could modify the contents of the txt file such that fairseq would automatically use caching while reading.

fairseq/training.py
```diff
@@ -0,0 +1,491 @@
...
+                        print(f"Cache manager: Error occurred for files in {name}")
+                        raise
+
+                    with open(cached_audio_fn) as local_bundle:
+                        bundle_lines = list(local_bundle.readlines())
+                        manifest_lines[0] = os.path.commonpath(bundle_lines)
+                        manifest_lines[1:] = map(
+                            lambda line: os.path.relpath(line, manifest_lines[0]),
+                            manifest_lines[1:],
+                        )
```
michelwi, line 279: @vieting this is my vision how to replace the specific cache manager call (for the bundle file) with the more general sisyphus file_caching call (for each file individually).
(pls. do not just use this diff, it is written in a very small window without actual testing 🙃 )
```suggestion
                    cached_bundle_lines = [gs.file_caching(l) for l in bundle_lines]
                    manifest_lines[0] = os.path.commonpath(cached_bundle_lines )
                    manifest_lines[1:] = map(
                        lambda line: os.path.relpath(line, manifest_lines[0]),
                        manifest_lines[1:],
                    )
```
Note that in some versions of the file_caching function flying around, no actual caching is done, but only `'cf <str> '` is returned.. this would not work here, as fairseq does not interpret these strings. instead the function should directly use the cache manager (either via `sp.check_call(['cf', str])` or by importing the cachemanager python module and then directly using its api)

fairseq/training.py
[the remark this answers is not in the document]
michelwi: I don't have a much more elegant solution. However the solution I propose (see other comment below in run function) would be more generic as it would allow other places to insert their own version of caching.

fairseq/training.py
vieting: That would be more general, yes. If @Judyxujj could test this, that would be nice. But then, we should also generalize this for a cached zip file below (currently line 290).

fairseq/training.py
```diff
@@ -0,0 +1,491 @@
...
+        fairseq_hydra_config,
+        *,  # args below are keyword only
+        command_line_args=None,
+        max_epoch=1,
+        save_interval=1,
+        keep_epochs=None,
+        time_rqmt=4,
+        mem_rqmt=4,
+        cpu_rqmt=2,
+        gpu_rqmt=1,
```
vieting, line 136: Do we maybe also want to allow setting `gpu_mem_rqmt`? This would be helpful for our slurm engine.

fairseq/training.py
JackTemaki: From what I understand this will not work properly because currently we use this `cf <>` syntax as string in file caching, instead of performing the actual caching. This was very specific for the RASR configs that support this, or jobs that used "shell" execution (which is kind of deprecated now). I am not sure how we can properly fix this design issue in our recipes and Sisyphus itself...

JackTemaki: Unfortunately not at all, because then we would write the `/var/tmp` path in the rasr config files in the `create_files` step, which would break nearly all setups...

michelwi: that is true.. we would have to merge all "create_config" and "run" steps.. :(

ok, what are other solutions? 
* Keep the i6-specific option in this job and everybody else (i.e. AppTek) will just ignore it?
* An i6-specific wrapper that should be passed to `fairseq_python_exe` that correctly resolves `cf <>` entries in the yaml file?

Judyxujj:
@michelwi @JackTemaki @vieting I have now changed the cache to enable i6-specific caching via `"cf"` or general caching via `gs.file_caching`, could you please check again and see if you have other suggestions?

fairseq/training.py
```diff
@@ -469,10 +448,10 @@ def _get_run_cmd(self):
         run_cmd += self.command_line_args
         run_cmd += ["checkpoint.save_dir=" + self.out_checkpoint_dir.get_path()]
 
-        if self.use_cache_manager:
+        if self.cache_manager:
             run_cmd += ["task.data=" + self.out_cached_audio_manifest.get_path()]
 
-        run_cmd.insert(0, self.fairseq_root.get_path() + "fairseq_cli/hydra_train.py")
+        run_cmd.insert(0, os.path.join(self.fairseq_root.get_path(), "fairseq_cli/hydra_train.py"))
```
michelwi, line 454:
```suggestion
        run_cmd.insert(0, os.path.join(self.fairseq_root.get_path(), "fairseq_cli", "hydra_train.py"))
```
if you already use `os.path.join` then we can have it join all the components of the path.

fairseq/training.py
```diff
@@ -469,10 +448,10 @@ def _get_run_cmd(self):
         run_cmd += self.command_line_args
         run_cmd += ["checkpoint.save_dir=" + self.out_checkpoint_dir.get_path()]
 
-        if self.use_cache_manager:
+        if self.cache_manager:
```
michelwi, line 451:
```suggestion
        if self.cache_manager is not None:
```

fairseq/training.py
```diff
@@ -154,7 +153,7 @@ def __init__(
         :param tk.Path fairseq_python_exe: File path to the executable for running python
         :param tk.Path fairseq_root: File path to the fairseq git for alternative call of fairseq-hydra-train
             (no need to install fairseq here)
-        :param bool use_cache_manager: enables caching of data given in the manifest with the i6 cache manager
+        :param str cache_manager: enables caching of data given in the manifest with cache manager
```
michelwi, line 156: I would explicitly list the available options
```suggestion
        :param str cache_manager: if given, enables caching of data given in the manifest with cache manager
            possible values: None: no caching, "i6": use the cache manager, "general": apply gs.file_caching
```

maybe also add an assert that only these three options are given.

michelwi (commented):
I like your changes, only a few minor comments from my side.

michelwi (approved):

curufinwe (approved):
Looks OK to me. Some small things could be changed, but they are not critical. @Judyxujj if you have time try to fix them, if not we can also merge as is.

fairseq/training.py
```diff
@@ -0,0 +1,472 @@
...
+        :param list[int]|set[int]|None keep_epochs: specify which checkpoints are kept in self.out_models.
+            Use None for each save_interval-th epoch
+        :param int|float time_rqmt: Overall time requirements
+        :param int|float mem_rqmt: Memory requirements (per GPU)
+        :param int cpu_rqmt: Required number of CPUs (per GPU)
+        :param int gpu_rqmt: Number of required GPUs
+        :param tk.Path fairseq_python_exe: File path to the executable for running python
+        :param tk.Path fairseq_root: File path to the fairseq git for alternative call of fairseq-hydra-train
+            (no need to install fairseq here)
+        :param str cache_manager: if given, enables caching of data given in the manifest with cache manager
```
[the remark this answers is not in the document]
Judyxujj: I now define enum CacheManagerType for the cache manager type

vieting (approved):
Thanks @Judyxujj!

PR #340, opened 2022-11-21
Idea to add a tmp dir option for all RasrCommand Jobs

christophmluscher:
The idea is to allow the user to specify a temporary dir in which the rasr job will be executed. This allows the job to be executed locally and not directly on a network file system.
This can be set job specific or globally via `settings.py`. At the moment this is hardcoded for EstimateMixturesJob.

- Do you think this is the correct place to add this?
- Should this be globally set? Hard-coded?

Would like to get feedback on this.

### DIFF
--- a/mm/mixtures.py
+++ b/mm/mixtures.py
@@ -387,6 +387,7 @@ class EstimateMixturesJob(MergeMixturesJob):
         self.concurrent = crp.concurrent
 
         self._old_mixtures = old_mixtures
+        self.use_tmp_dir = True
 
         self.out_log_file = self.log_file_output_path("accumulate", crp, True)
 
@@ -425,7 +426,10 @@ class EstimateMixturesJob(MergeMixturesJob):
 
     def accumulate(self, task_id):
         self.run_script(
-            task_id, self.out_log_file[task_id], "./accumulate.sh", use_tmp_dir=True
+            task_id,
+            self.out_log_file[task_id],
+            "./accumulate.sh",
+            use_tmp_dir=self.use_tmp_dir,
         )
 
     def delete_accumulators(self):
--- a/mm/mixtures.py
+++ b/mm/mixtures.py
@@ -11,6 +11,7 @@ import shutil
 import stat
 import struct
 import tempfile
+from typing import Dict, Optional, Union
 
 from sisyphus import *
 
@@ -360,15 +361,15 @@ class LinearAlignmentJob(MergeMixturesJob):
 class EstimateMixturesJob(MergeMixturesJob):
     def __init__(
         self,
-        crp,
-        old_mixtures,
-        feature_flow,
-        alignment,
-        split_first=True,
-        keep_accumulators=False,
-        extra_merge_args=None,
-        extra_config=None,
-        extra_post_config=None,
+        crp: rasr.CommonRasrParameters,
+        old_mixtures: tk.Path,
+        feature_flow: Union[str, tk.Path, rasr.FlagDependentFlowAttribute],
+        alignment: Union[str, tk.Path, rasr.FlagDependentFlowAttribute],
+        split_first: bool = True,
+        keep_accumulators: bool = False,
+        extra_merge_args: Optional[Dict] = None,
+        extra_config: Optional[rasr.RasrConfig] = None,
+        extra_post_config: Optional[rasr.RasrConfig] = None,
     ):
         self.set_vis_name("Split Mixtures" if split_first else "Accumulate Mixtures")
 
@@ -423,7 +424,9 @@ class EstimateMixturesJob(MergeMixturesJob):
         self.write_run_script(self.exe, "accumulate-mixtures.config", "accumulate.sh")
 
     def accumulate(self, task_id):
-        self.run_script(task_id, self.out_log_file[task_id], "./accumulate.sh")
+        self.run_script(
+            task_id, self.out_log_file[task_id], "./accumulate.sh", use_tmp_dir=True
+        )
 
     def delete_accumulators(self):
         for i in range(1, self.concurrent + 1):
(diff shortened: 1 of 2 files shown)


michelwi:
I don't like to introduce new global settings when we are actively trying to get rid of those.

The running in a tmp dir part is good.

Alternatively or additionally we could again try to use the cache manager (if available) also to write back outputs via the built-in RASR module.

christophmluscher:
> I don't like to introduce new global settings when we are actively trying to get rid of those.

Maybe I misunderstood, but `RASR_ROOT` and so on should not be in the `settings.py` since the are filepaths, but something like `file_caching` should and I would consider this flag similar to `file_caching`

> Alternatively or additionally we could again try to use the cache manager (if available) also to write back outputs via the built-in RASR module.

I considered this but this would take a lot more effort IMHO, since the adaptation to each individual job would be more extensiv or we would need to integrate this on sisyphus side... I already tried a few things but so far not really to my satisfaction...

rasr/command.py
```diff
@@ -140,7 +179,7 @@ def run_cmd(self, cmd, args=None, retries=2):
         for t in range(retries + 1):
             try:
                 self.cleanup_before_run(cmd, t, *args)
-                sp.check_call([cmd] + [str(arg) for arg in args])
+                sp.check_call([cmd] + [str(arg) for arg in args], cwd=cwd)
```
Atticus1806, line 182: Just a sidenote: I think sp is imported via `from i6_core.util import *` which might be a bit hidden, maybe we should do the import manually here aswell.

rasr/command.py
```diff
@@ -118,19 +123,53 @@ def select_exe(cls, specific_exe, default_exe_name):
...
         args = [] if args is None else args
         tmp_log_file = remove_suffix(
             os.path.basename(tk.uncached_path(log_file)), ".gz"
         )
-        self.run_cmd(cmd, [task_id, tmp_log_file] + args, retries)
+
+        if use_tmp_dir or (
+            hasattr(gs, "JOB_RUN_RASR_TMP_DIR") and gs.JOB_RUN_RASR_TMP_DIR
+        ):
+            with tempfile.TemporaryDirectory(prefix=gs.TMP_PREFIX) as tmp_dir:
```
Atticus1806, line 143: Should this `TMP_PREFIX` be from gs or should this maybe be handled differently? How is it done in other cases where tmp dirs are constructed

rasr/command.py
```diff
@@ -118,19 +123,53 @@ def select_exe(cls, specific_exe, default_exe_name):
...
+                print("using temp-dir: %s" % tmp_dir)
+                try:
+                    self.run_cmd(cmd, [task_id, tmp_log_file] + args, retries, tmp_dir)
+                    file_names = os.listdir(tmp_dir)
+                    for fn in file_names:
+                        shutil.move("%s/%s" % (tmp_dir, fn), fn)
```
Atticus1806, line 149: Does this string always yield proper directory names regarding / etc.? This is more of a beauty aspect, that the name is rather e.g. logs/log_1 and not logslog_1

michelwi:
> > I don't like to introduce new global settings when we are actively trying to get rid of those.
> 
> Maybe I misunderstood, but `RASR_ROOT` and so on should not be in the `settings.py` since the are filepaths, but something like `file_caching` should and I would consider this flag similar to `file_caching`

As far as I understand it, the global_settings are for sisyphus (i.e. the workflow software) and not for settings of the recipe. Also all available settings are defined in `sisyphus/global_settings.py` and I would like to move away from us defining new settings inside the recipes and likewise relying on them being set somewhere so they can be used.

`file_caching` is a setting (feature) of sisyphus and has a clear interface in global_settings.

Besides, when you can set it directly through the function, what is the need for a global setting?

mm/mixtures.py
```diff
@@ -423,7 +424,9 @@ def create_files(self):
         self.write_run_script(self.exe, "accumulate-mixtures.config", "accumulate.sh")
 
     def accumulate(self, task_id):
-        self.run_script(task_id, self.out_log_file[task_id], "./accumulate.sh")
+        self.run_script(
+            task_id, self.out_log_file[task_id], "./accumulate.sh", use_tmp_dir=True
```
michelwi, line 428: Other options could be to expose `use_tmp_dir` in the args of the job.

Or (imho good solution) to add a member `self.use_tmp_dir` in the `__init__` that is always (True/False) but that can be overridden in the pipeline for specific jobs.

rasr/command.py
michelwi: yes, this is well defined https://github.com/rwth-i6/sisyphus/blob/cc58196b0c410a6d98ba7c1d79ac2cd1eb48ae39/sisyphus/global_settings.py#L273

In other cases we do it similarly. (Mostly because I advocated for it xD)

JackTemaki:
I agree with @michelwi that we can just not have the global settings parameter. The job in question has this fixed now, and we can enable this for all other jobs that need this as well. I think there is no need for a global configuration. If we really need it in can be on the job-level as parameter.

rasr/command.py
christophmluscher: I think this should be ok. but I did not test this. @Atticus1806 you ran this, does this cause problems for you?

michelwi: we could to `os.path.join` instead of `"%s/%s" %`, which would then be more general over other os.. but internally this is anyway what's happening.

Or are you afraid of double `"//"`? I don't think it would be a problem.

christophmluscher:
> Would you like to enable this also for other jobs than `EstimateMixturesJob`?

In principle yes, but wanted to maybe first do a bit of testing before enabling for more jobs, especially I want to see the effect on filesystem load.

Jobs were I want to enable:
`AlignmentJob`
`AdvTreeSearch`

michelwi (approved):
ok

JackTemaki (approved):

PR #342, opened 2022-11-30
Rework of BPE jobs for LM pipeline

JackTemaki:
This one was a little bit tricky, so here the summary:

 - There are different kinds of "vocabulary" files that have different content or structure. A vocabulary in the sense of RETURNN is a `bpe_token -> index` mapping, while for the subword-nmt software this is a `bpe_token -> count` file. Within the bpe training I am now creating an artificial `bpe_token -> count` file (counts are all 100), so that the vocabulary parameter can be set to not cause invalid merges when applying bpe on unseen data via the apply job. 
 
If we want true counts in the future we can have an extra job for that, where we also supply a corpus to get the counts on.

The second change is that the apply job operates in a temp-dir and also allows for gzipped output (for LM stuff the data sizes can be quite large).

I hope I added sufficient documentation to not cause confusion.

### DIFF
--- a/text/label/subword_nmt/apply.py
+++ b/text/label/subword_nmt/apply.py
@@ -118,6 +118,7 @@ class ApplyBPEToTextJob(Job):
         bpe_vocab: Optional[tk.Path] = None,
         subword_nmt_repo: Optional[tk.Path] = None,
         gzip_output: bool = False,
+        mini_task=True,
     ):
         """
         :param text_file: words text file to convert to bpe
@@ -126,6 +127,7 @@ class ApplyBPEToTextJob(Job):
             use e.g. ReturnnTrainBpeJob.out_bpe_dummy_count_vocab
         :param subword_nmt_repo: subword nmt repository path. see also `CloneGitRepositoryJob`
         :param gzip_output: use gzip on the output text
+        :param mini_task: if the Job should run locally, e.g. only a small (<1M lines) text should be processed
         """
         self.text_file = text_file
         self.bpe_codes = bpe_codes
@@ -141,8 +143,14 @@ class ApplyBPEToTextJob(Job):
             "words_to_bpe.txt.gz" if gzip_output else "words_to_bpe.txt"
         )
 
+        self.mini_task = mini_task
+        self.rqmt = {"cpu": 1, "mem": 2, "time": 2}
+
     def tasks(self):
-        yield Task("run", mini_task=True)
+        if self.mini_task:
+            yield Task("run", mini_task=True)
+        else:
+            yield Task("run", rqmt=self.rqmt)
 
     def run(self):
         with tempfile.TemporaryDirectory(prefix=gs.TMP_PREFIX) as tmp:
@@ -175,3 +183,8 @@ class ApplyBPEToTextJob(Job):
                     sp.call(["gzip"], stdin=fin, stdout=fout)
             else:
                 shutil.copy(tmp_outfile, self.out_bpe_text.get_path())
+
+    @classmethod
+    def hash(cls, parsed_args):
+        del parsed_args["mini_task"]
+        return super().hash(parsed_args)
--- a/text/label/subword_nmt/apply.py
+++ b/text/label/subword_nmt/apply.py
@@ -123,7 +123,7 @@ class ApplyBPEToTextJob(Job):
         :param text_file: words text file to convert to bpe
         :param bpe_codes: bpe codes file, e.g. ReturnnTrainBpeJob.out_bpe_codes
         :param bpe_vocab: if provided, then merge operations that produce OOV are reverted,
-            use e.g. ReturnnTrainBpeJob.out_bpe_apply_vocab
+            use e.g. ReturnnTrainBpeJob.out_bpe_dummy_count_vocab
         :param subword_nmt_repo: subword nmt repository path. see also `CloneGitRepositoryJob`
         :param gzip_output: use gzip on the output text
         """
--- a/text/label/subword_nmt/apply.py
+++ b/text/label/subword_nmt/apply.py
@@ -2,7 +2,9 @@ __all__ = ["ApplyBPEModelToLexiconJob", "ApplyBPEToTextJob"]
 
 import subprocess as sp
 import os
+import shutil
 import sys
+import tempfile
 import xml.etree.ElementTree as ET
 
 from sisyphus import *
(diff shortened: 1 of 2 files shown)


text/label/subword_nmt/apply.py
```diff
@@ -106,39 +108,68 @@ class ApplyBPEToTextJob(Job):
...
         """
         :param Path text_file: words text file to convert to bpe
-        :param Path bpe_codes: bpe codes file
-        :param Path|None bpe_vocab: if provided, then merge operations that produce OOV are reverted
-        :param Path|str|None subword_nmt_repo: subword nmt repository path. see also `CloneGitRepositoryJob`
+        :param Path bpe_codes: bpe codes file, e.g. ReturnnTrainBpeJob.out_bpe_codes
+        :param Path|None bpe_vocab: if provided, then merge operations that produce OOV are reverted,
+            use e.g. ReturnnTrainBpeJob.out_bpe_apply_vocab
+        :param Path|None subword_nmt_repo: subword nmt repository path. see also `CloneGitRepositoryJob`
         """
```
Atticus1806, line 127: even though its trivial, comment for `gzip_output` please.

Atticus1806:
Else looks fine.

text/label/subword_nmt/apply.py
JackTemaki: Done

text/label/subword_nmt/apply.py
```diff
@@ -106,39 +109,69 @@ class ApplyBPEToTextJob(Job):
...
-            cmd += ["--vocabulary", self.bpe_vocab.get_path()]
-
-        util.create_executable("apply_bpe.sh", cmd)
-        sp.run(cmd, check=True)
+        with tempfile.TemporaryDirectory(prefix=gs.TMP_PREFIX) as tmp:
+            input_file = self.text_file.get_path()
+            tmp_infile = os.path.join(tmp, "in_text.txt")
+            tmp_outfile = os.path.join(tmp, "out_text.txt")
+            with util.uopen(tmp_infile, "wt") as out:
+                sp.call(["zcat", "-f", input_file], stdout=out)
```
michelwi, line 153: has this any benefit over 
```suggestion
            shutil.copy(input_file, tmp_infile)
```

text/label/subword_nmt/train.py
```diff
@@ -95,33 +95,51 @@ class ReturnnTrainBpeJob(Job):
...
+    def __init__(
+        self,
+        text_file: tk.Path,
+        bpe_size: int,
+        unk_label: str = "UNK",
+        subword_nmt_repo: Optional[tk.Path] = None,
+    ):
         """
-        :param Path text_file: corpus text file
+        :param text_file: corpus text file, only accepts uncompressed text
```
michelwi, line 120: why does it only support uncompressed text files? Below `util.uopen` is used to open the file which should handle gzipped files correctly.

michelwi (commented):
Why is there a problem when using the RETURNN-style vocab for subword-nmt?
When doing the application the counts _should_ not matter as only the list of tokens is relevant. So either supplying
```
token1 100
token2 100
token3 100
```
or 
```
token1 0
token2 1
token3 2
```
should be the same?

text/label/subword_nmt/apply.py
JackTemaki: Yes, the tmp_infile needs to be uncompressed. So if it is plain the zcat will basically be a copy, if not it will uncompress on the way.

text/label/subword_nmt/train.py
JackTemaki: Yes this is a leftover because I first added this comment, and then added the automatic uncompressing beforehand. So this can be removed now.

JackTemaki:
> Why is there a problem when using the RETURNN-style vocab for subword-nmt? When doing the application the counts _should_ not matter as only the list of tokens is relevant. So either supplying
> 
> ```
> token1 100
> token2 100
> token3 100
> ```
> 
> or
> 
> ```
> token1 0
> token2 1
> token3 2
> ```
> 
> should be the same?

The apply_bpe.py has a function to limit based on a vocabulary threshold, so it is not the same and the counts to matter in that case. Also the RETURNN vocab is not in 

> ```
> token1 0
> token2 1
> token3 2
> ```

format but like:
```
 {
   "token1": 0.
   "token2": 1,
   "token3": 2,
 }
 ```
 
 which is not accepted by apply_bpe.py.

michelwi:
> The apply_bpe.py has a function to limit based on a vocabulary threshold, 

Ok, but this can be turned off, right? (Edit: It is off by default, so no filtering happens)


> Also the RETURNN vocab is not in [...] format but like: [...] which is not accepted by apply_bpe.py.

Ok, I agree we need a different input here. But I am a bit concerned about your way of estimating it.
1. the value 100 is a bit arbitrary. How about `0`, `1`, or `param_fake_count` instead?
2. you are adding another output to the job, so the signature changes based on the version of the recipes. I would try to avoid this. So maybe hide this behind a parameter `compute_bpe_apply_vocab`? But then this parameter will be obsolet, once we have a proper job to estimate the counts. -> rather implement the correct counting method now? 
3. Is the bpe_vocab (in principle) tied to the bpe codes file or would it in any case be sensible to use different vocabs with different codes? Then this would be an even stronger argument for separating the vocab estimation and the codes estimation.

JackTemaki:
> the value 100 is a bit arbitrary

To be honest here I went with a "like Kazuki did it"....

> rather implement the correct counting method now?

I found it strange that a single BpeTrain job does not give all outputs that are needed to actually call the apply job. Yes, I could add a count job and put this into the pipeline, or even create a count file based on the input of the train job. The later one is maybe the best solution.

> you are adding another output to the job, so the signature changes based on the version of the recipes

But adding outputs is no severe issue at all, is it? Yes, you could then write a pipeline where suddenly files are missing, but then you can simply re-run the job. Also apply bpe is not really the most used job. I tried it for the first time now in an actual pipeline and nothing worked, thus this PR...

michelwi:
> But adding outputs is no severe issue at all, is it? Yes, you could then write a pipeline where suddenly files are missing, but then you can simply re-run the job. 

yes, thats why I was only commenting to have a discussion rather than "request changes"

> Also apply bpe is not really the most used job. I tried it for the first time now in an actual pipeline and nothing worked, thus this PR...

This is a strong indicator that this hasn't been used at all.. so then it would be fine to change it.

JackTemaki:
> > But adding outputs is no severe issue at all, is it? Yes, you could then write a pipeline where suddenly files are missing, but then you can simply re-run the job.
> 
> yes, thats why I was only commenting to have a discussion rather than "request changes"
> 
> > Also apply bpe is not really the most used job. I tried it for the first time now in an actual pipeline and nothing worked, thus this PR...
> 
> This is a strong indicator that this hasn't been used at all.. so then it would be fine to change it.

Okay so adding the output is fine, having the counts with 100 is not. So I just calculate the actual counts on the provided data? Would that be okay?

michelwi:
> Okay so adding the output is fine, having the counts with 100 is not. So I just calculate the actual counts on the provided data? Would that be okay?

Setting some dummy value is also fine, I am just putting to the discussion why 100? (Well, ok, Kazuki did it). 
If it is ok with you, I would propose to change it to 0 or 1 (or maybe even better -1 so that it is not confused with actual counts)

This should not matter if the count based filtering is deactivated:
cf. https://github.com/rwth-i6/subword-nmt/blob/6ba4515d684393496502b79188be13af9cad66e2/apply_bpe.py#L255-L259, `None` is default threshold.

JackTemaki:
Maybe also:
```python
 self.out_bpe_fake_count_vocab = self.output_path("bpe.fake_count.vocab.txt")
 ```
 then?

michelwi:
I like `dummy_count` 👍

Atticus1806 (approved):

PR #343, opened 2022-11-30
allow DelayedBase in CodeWrapper

JackTemaki:


Atticus1806:
Is there any chance this was/is being misused somehow already so that this breaks smth? Else approved.

JackTemaki:
I do not think so, whenever `CodeWrapper` does not return a string there is an error.

Atticus1806 (approved):

mmz33 (approved):

michelwi:
I remember - and please correct me when I am wrong - that the original idea for `DelayedBase` and derived objects was to have the `__repr__` function be save to call in the Manager as it would return something like `"<Path /path/to/object>"` that cannot be used further, whereas `__str__` returns a usable representation (e.g. `self.get_path()` or `self.get()`) and should only be used by the worker.

Now I am wondering if we explicitly handle `DelayedBase` here, should we also adhere to the same principles and rather have a `__str__` function while the `__repr__` makes use of `repr(self.code)`?

JackTemaki:
> I remember - and please correct me when I am wrong - that the original idea for `DelayedBase` and derived objects was to have the `__repr__` function be save to call in the Manager as it would return something like `"<Path /path/to/object>"` that cannot be used further, whereas `__str__` returns a usable representation (e.g. `self.get_path()` or `self.get()`) and should only be used by the worker.
> 
> Now I am wondering if we explicitly handle `DelayedBase` here, should we also adhere to the same principles and rather have a `__str__` function while the `__repr__` makes use of `repr(self.code)`?

I am not sure if I can completely follow. We use `__repr__` so that there is no additional quotation but everything is pasted "as is". The Problem is just that `__repr__`  per definition has to return a string object, thus `self.code` could only be a string. But if I want to use Path objects in that string, I have to use `DelayedBase`, which is now not a string type. This is why `.get()` has to be called before. I do not think that `repr(delayed_base_object)` would be better here... as the call should anyway only occur in the worker but never in the manager.

`repr(self.code)` would lead to wrong behavior, because then we get an additional quotation that we do not want to have.

PR #345, opened 2022-12-02
Add AverageTFCheckpointsJob

JackTemaki:
I added some more people for the review just that you are aware that this jobs exists now. Depending on our learning rate scheduling settings this can give some boost in model performance basically for free.

### DIFF
--- a/returnn/training.py
+++ b/returnn/training.py
@@ -3,6 +3,9 @@ __all__ = [
     "Checkpoint",
     "ReturnnTrainingJob",
     "ReturnnTrainingFromFileJob",
+    "GetBestEpochJob",
+    "GetBestTFCheckpointJob",
+    "AverageCheckpointsJob",
 ]
 
 import copy
@@ -10,6 +13,7 @@ import sys
 import os
 import shutil
 import subprocess as sp
+from typing import List, Union
 
 from sisyphus import *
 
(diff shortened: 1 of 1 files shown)


returnn/training.py
```diff
@@ -788,3 +792,63 @@ def run(self):
...
+        """
+        self.model_dir = model_dir
+        self.epochs = epochs
+        self.returnn_python_exe = returnn_python_exe
+        self.returnn_root = returnn_root
+
+        self._out_model_dir = self.output_path("model", directory=True)
+        self.out_checkpoint = Checkpoint(self.output_path("model/average.index"))
+
+        self.rqmt = {"cpu": 1, "time": 0.5, "mem": max(2 * len(epochs), 4)}
```
michelwi, line 824: isn't `2*len(epochs)` always larger or equal to `4` -- assuming that you provide at least 2 epochs. 
I would rather specify a maximum memory requirement (if that is possible). e.g.
```suggestion
        self.rqmt = {"cpu": 1, "time": 0.5, "mem": min(2 * len(epochs), 16)}
```

Or is the mem rqmt strictly linear at 2GB per epoch s.t. we cannot merge arbitrary many epochs together?

returnn/training.py
```diff
@@ -788,3 +792,63 @@ def run(self):
...
+        max_epoch = max(epochs)
+
+        # we are writing a checkpoint with the maximum epoch index in the file name because Returnn
+        # resolves symlinks and reads the name to determine the "checkpoint epoch"
+        out_path = os.path.join(self._out_model_dir.get_path(), "epoch.%3d" % max_epoch)
+        args = [
+            self.returnn_python_exe.get_path(),
+            os.path.join(self.returnn_root.get_path(), "tools/tf_avg_checkpoints.py"),
+            "--checkpoints",
+            ",".join([str(epoch) for epoch in epochs]),
```
michelwi, line 840: is it correct that no leading zeros are needed here?
from https://github.com/rwth-i6/returnn/blob/2c0bf3666e721b86d843f2ef54cd416dfde20566/tools/tf_avg_checkpoints.py#L76-L83 it seems to me that they would be necessary

returnn/training.py
JackTemaki: But you are right that I can drop the max, as no one will ever run with less than 2.

michelwi: I see.. then maybe the implementation of the script has to be altered if memory turns out to be a problem in the future.

There is still the check_engine_limits function to reduce the mem request.

mmz33 (approved):

Issue #346, opened 2022-12-05
Issues with temp-dir in RASR jobs


JackTemaki:
It seems that https://github.com/rwth-i6/i6_core/pull/340 does not always work as intended. I now manually pushed a commit that disables using temp-dirs so that the old behavior is kept. @Atticus1806 and I will do more tests to figure out what exactly goes wrong and needs to be fixed.

JackTemaki:
Fixed with #347

PR #347, opened 2022-12-06
Tmp Dir: Introduce Copy List

Atticus1806:
Handles #346 
This PR introduces a copy list for using a tmp dir with `run_script`. The reason for that is that this was the only way we could find to handle concurrent jobs which write files into the work folder without causing race conditions through this writing.
Since we manually need to enable using tmp dir usage, adding a copy list to those jobs which run with more than one concurrent job in the array should not be a problem.
Currently testing if works and approved can merge later.

### DIFF
--- a/mm/mixtures.py
+++ b/mm/mixtures.py
@@ -387,7 +387,8 @@ class EstimateMixturesJob(MergeMixturesJob):
         self.concurrent = crp.concurrent
 
         self._old_mixtures = old_mixtures
-        self.use_tmp_dir = False
+        self.use_tmp_dir = True
+        self.copy_list = ["accumulate-mixtures.config", "accumulate.sh", "alignment.flow"]
 
         self.out_log_file = self.log_file_output_path("accumulate", crp, True)
 
@@ -430,6 +431,7 @@ class EstimateMixturesJob(MergeMixturesJob):
             self.out_log_file[task_id],
             "./accumulate.sh",
             use_tmp_dir=self.use_tmp_dir,
+            copy_tmp_ls=self.copy_list if self.use_tmp_dir else None
         )
 
     def delete_accumulators(self):
--- a/rasr/command.py
+++ b/rasr/command.py
@@ -2,7 +2,6 @@ __all__ = ["RasrCommand"]
 
 import logging
 import shutil
-import subprocess as sp
 import tempfile
 import time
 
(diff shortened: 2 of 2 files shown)


rasr/command.py
```diff
@@ -132,17 +131,30 @@ def run_script(
...
         args = [] if args is None else args
         tmp_log_file = remove_suffix(
             os.path.basename(tk.uncached_path(log_file)), ".gz"
         )
 
-        if use_tmp_dir:
-            date_time_cur = time.strftime("%y%m%d-%H%M%S", time.localtime())
+        work_dir = os.path.abspath(os.curdir)
+        if use_tmp_dir or (
+            hasattr(gs, "JOB_RUN_RASR_TMP_DIR") and gs.JOB_RUN_RASR_TMP_DIR
```
christophmluscher, line 143: We agreed to remove the global parameter.

rasr/command.py
```diff
@@ -151,7 +163,8 @@ def run_script(
                     print(
                         "'%s' crashed - copy temporary work folder as 'crash_dir'" % cmd
                     )
-                    shutil.copytree(tmp_dir, "crash_dir_" + date_time_cur)
+                    if "crash_dir_%d" % task_id not in os.listdir(work_dir):
```
christophmluscher, line 166: Only copy when it is not in the work dir seems a bit risky. I would prefer the version with the task id and the datetime.

mm/mixtures.py
```diff
@@ -387,7 +387,12 @@ def __init__(
         self.concurrent = crp.concurrent
 
         self._old_mixtures = old_mixtures
-        self.use_tmp_dir = False
+        self.use_tmp_dir = True
+        self.copy_list = [
```
JackTemaki, line 391: better call it `tmp_dir_copy_list' so that it is clear that those are related.

rasr/command.py
JackTemaki, line 143: Please do not re-introduce the already remove `gs` parameter.

JackTemaki: Also you removed `date_time_cur`

rasr/command.py
```diff
@@ -151,7 +163,8 @@ def run_script(
...
+                        shutil.copytree(tmp_dir, "crash_dir_%d" % task_id)
```
JackTemaki, line 167: also add `date_time_cur` here again

JackTemaki (changes requested):

Atticus1806:
oh yeah something went wrong when stashing changes, I will redo the things.

JackTemaki (changes requested):
You are now missing the `task_id` in the crash folder name

JackTemaki (approved):

rasr/command.py
```diff
@@ -2,7 +2,6 @@
 
 import logging
 import shutil
-import subprocess as sp
```
christophmluscher, line 5: I added this to make the subprocess import clearer, since this gets imported via 15: `from i6_core.util import *`

christophmluscher (approved):
I would like to keep the import, maybe we can remove the `import *` at some point. But not something blocking merging IMHO

rasr/command.py
Atticus1806: another thing that got lost during stashing...

Atticus1806: guess gotta be more careful next time

rasr/command.py
```diff
@@ -132,17 +132,29 @@ def run_script(
...
                 try:
+                    if copy_tmp_ls is None:
+                        assert (
+                            task_id == 1
+                        ), "Concurrent Jobs need a list of files to copy due to race conditions"
+                        file_names = os.listdir(work_dir)
+                    else:
+                        file_names = copy_tmp_ls
+                    for fn in file_names:
+                        print(fn)
```
michelwi, line 156: please remove this print if it is a remainder of debugging or add a more verbose message

rasr/command.py
```diff
@@ -132,17 +132,29 @@ def run_script(
...
+                        shutil.copy(os.path.join(work_dir, fn), "%s/%s" % (tmp_dir, fn))
                     self.run_cmd(cmd, [task_id, tmp_log_file] + args, retries, tmp_dir)
                     file_names = os.listdir(tmp_dir)
                     for fn in file_names:
```
michelwi, line 160: do we need a similar logic also here? E.g. only copy files back that are _not_ in the `copy_tmp_ls`.

PR #348, opened 2022-12-06
Tmp Dir: Check if file needs to be copied

Atticus1806:
Follow up to #347 
Check if the file that is moved back into the work dir was copied beforehand, then it does not need to be copied. Reduces fileserver operations and possibly prevents weird race condition when multiple copies happen.


Atticus1806:
also removes a print that somehow was left over during debugging

michelwi (approved):

JackTemaki (approved):

PR #350, opened 2022-12-08
Switchboard LDC Speakers, Fisher Transcriptions and corpus corrections

JackTemaki:
This is the third and hopefully final PR for the switchboard corpus. (at least for i6_core, all the pipeline stuff will come later in experiments/common)

It adds:
 - LDC Speaker job to get the correct 520 speakers as intended by the switchboard corpus design
 - Fisher transcription job to create the training data for LMs (count or neural)
 -  Lowercase option and an "empty recording" ignore option for the broken recording in the original LDC data. I also made both the default now because this his how we should consistently work with it.

### DIFF
--- a/datasets/switchboard.py
+++ b/datasets/switchboard.py
@@ -756,3 +756,52 @@ class CreateSwitchboardE2EBlissCorpusJob(Job):
             segment.orth = orth
 
         c.dump(self.out_e2e_corpus.get_path())
+
+
+class CreateFisherTranscriptionsJob(Job):
+    """
+    Create the compressed text data based on the fisher transcriptions which can be used for LM training
+
+    Part 1: https://catalog.ldc.upenn.edu/LDC2004T19
+    Part 2: https://catalog.ldc.upenn.edu/LDC2005T19
+    """
+
+    def __init__(
+        self,
+        fisher_transcriptions1_folder: tk.Path,
+        fisher_transcriptions2_folder: tk.Path,
+    ):
+        """
+        :param fisher_transcriptions1_folder: path to unpacked LDC2004T19.tgz, usually named fe_03_p1_tran
+        :param fisher_transcriptions2_folder: path to unpacked LDC2005T19.tgz, usually named fe_03_p2_tran
+        """
+        self.fsh_trans1_folder = fisher_transcriptions1_folder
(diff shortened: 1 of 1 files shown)


datasets/switchboard.py
```diff
@@ -160,24 +160,89 @@ def run(self):
...
+class CreateLDCSwitchboardSpeakerListJob(Job):
+    """
+    This creates the speaker list according to the conversation and speaker table
+    from the LDC documentation: https://catalog.ldc.upenn.edu/docs/LDC97S62
+
+    The resulting file contains 520 speakers in the format of:
+        <speaker_id> <gender> <recording>
+    """
+
+    def __init__(self, caller_tab_file, conv_tab_file):
```
michelwi, line 172: pls add docstring as to what files should be input here (if it is not immediately obvious)

michelwi (approved):

JackTemaki:
@mmz33 @Marvin84 @vieting  ping!

datasets/switchboard.py
```diff
@@ -215,6 +284,8 @@ def run(self):
         for rec_name, segs in sorted(rec_to_segs.items()):
             recording = corpus.Recording()
             recording.name = rec_name
+            if self.skip_empty_ldc_file and rec_name == "sw02167B":
```
Marvin84, line 287: wouldn't it be better to just do this check before starting the loop and in case `skip_empty_ldc_file` is `True` just to eliminate it from the dictionary? Maybe not a relevant gain, but still better than doing the check multiple times within the loop.

datasets/switchboard.py
```diff
@@ -286,7 +358,11 @@ def _filter_orth(orth):
         out = " ".join(filtered_orth)
 
         # replace &
-        out = out.replace("AT&T's", "AT and T")
+        # for AT&T's we drop the 's as t's is not in the lexicon
```
Marvin84, line 361: Is this the only case we have?

Marvin84 (approved):
You might want to consider a small change, but it is not essential.

datasets/switchboard.py
JackTemaki: Yes this is the only case we have.

PR #351, opened 2022-12-19
Add LmIndexVocabularyFromLexiconJob

JackTemaki:
This is a job which creates an index vocab file for LM training based on a given lexicon. This job was missing to run any kind of consistent word-based Neural-LM training with our pipelines.

### DIFF
--- a/lm/vocabulary.py
+++ b/lm/vocabulary.py
@@ -1,6 +1,125 @@
-from sisyphus import Job, Task
+from sisyphus import Job, Task, tk
+
+from dataclasses import dataclass
+from typing import Optional, Union
 
 from i6_core.lib.lm import Lm
+from i6_core.lib.lexicon import Lexicon
+from i6_core.util import uopen
+
+
+@dataclass(frozen=True)
+class LmIndexVocabulary:
+    vocab: tk.Path
+    vocab_size: tk.Variable
+    unknown_token: Union[tk.Variable, str]
+
+
+class LmIndexVocabularyFromLexiconJob(Job):
+    """
+    Computes a <word>: <index> vocabulary file from a bliss lexicon for Word-Level LM training
+
+    Sentence begin/end will point to index 0, unknown to index 1.
+    Both are taking directly from the lexicon via the "special" marking:
+      - <lemma special="sentence-begin"> -> index 0
+      - <lemma special="sentence-end"> -> index 0
+      - <lemma special="unknown"> -> index 1
+
+    If <synt> tokens are provided in a lemma, they will be used instead of <orth>
+
+    CAUTION:
+    Be aware of: https://github.com/rwth-i6/returnn/issues/1245 when using Returnn's LmDataset
+    """
+
+    def __init__(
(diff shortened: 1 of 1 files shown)


lm/vocabulary.py
```diff
@@ -1,6 +1,125 @@
...
+      - <lemma special="unknown"> -> index 1
+
+    If <synt> tokens are provided in a lemma, they will be used instead of <orth>
+
+    CAUTION:
+    Be aware of: https://github.com/rwth-i6/returnn/issues/1245 when using Returnn's LmDataset
+    """
+
+    def __init__(
+        self, bliss_lexicon: tk.Path, count_ordering_text: Optional[tk.Path] = None
```
michelwi, line 35: pls add docstring. It was not immediately clear to me, what the `count_ordering_text` is used for.

lm/vocabulary.py
```diff
@@ -1,6 +1,125 @@
...
+        if self.count_ordering_text is not None:
+            word_set = set(wordlist)
+            word_counts = {}
+            for w in wordlist:
+                word_counts[w] = 0
+            with uopen(self.count_ordering_text, "rt") as f:
+                for line in f.readlines():
+                    for word in line.strip().split(" "):
+                        if word in word_set:
+                            word_counts[word] += 1
```
michelwi, line 102: `word_set` is not needed here, as the keys of a dict already are a set (with the same inplications for runtime cost)

```suggestion
            word_counts = {}
            for w in wordlist:
                word_counts[w] = 0
            with uopen(self.count_ordering_text, "rt") as f:
                for line in f.readlines():
                    for word in line.strip().split(" "):
                        if word in word_counts:
                            word_counts[word] += 1
```

michelwi: but this comment got me thinking: should the `wordlist` not rather be a `wordset`? I get it, that you would like to impose some kind of ordering and in the absence of a text to get word counts from you would rely on the order of appearance in the original lexicon. 
But what if there are homonyms (multiple lemmata with same orth or synth) in the lexicon? Then the same word would appear twice in the list.

Maybe the best solution would be to have a dict (OrderedDict) to begin with and during the `for lemma in lex.lemmata` loop the counts are initialized as 0. Then if `self.count_ordering_text is not None` the counts are updated and the `wordlist` is created as you have below and else ```wordlist = word_counts.keys()``` relying on the OrderedDict to keep the order of appearance.

michelwi (changes requested):
I request changes w.r.t. the possibility of duplicate words in the vocab.

lm/vocabulary.py
JackTemaki: But this will remove duplicates or not?
```python
            for w in wordlist:
                word_counts[w] = 0
                ```

michelwi: yes, but only in the case that a `self.count_ordering_text` is given, otherwise this part of the code is not used and there could be duplicates.

JackTemaki: maybe like this?

michelwi (approved):

Nintendo-Wei (approved):

PR #352, opened 2023-01-19
Adds multi_channel_strategy 'pick_first' to BlissToPcmHDFJob

michelwi:
This adds multi-channel audio file handling to BlissToPcmHDFJob.
currently only one strategy is implemented: to select the waveform from the first channel.

Default behavior stays to assert for single channel audio file

### DIFF
--- a/returnn/hdf.py
+++ b/returnn/hdf.py
@@ -207,11 +207,14 @@ class BlissToPcmHDFJob(Job):
     compatible with the RETURNN HDFDataset
     """
 
+    __sis_hash_exclude__ = {"multi_channel_strategy": "none"}
+
     def __init__(
         self,
         bliss_corpus: tk.Path,
         segment_file: Optional[tk.Path] = None,
         output_dtype: str = "int16",
+        multi_channel_strategy: str = "none",
         returnn_root: Optional[tk.Path] = None,
     ):
         """
@@ -219,14 +222,23 @@ class BlissToPcmHDFJob(Job):
         :param bliss_corpus: Bliss corpus to read segments and audio files from
         :param segment_file: segment file that lists allowed segments
         :param output_dtype: dtype that should be written in the hdf (supports float64, float32, int32, int16)
+        :param multi_channel_strategy: defines what should happen to multi-channel audio files.
+            Currently implemented are:
+            "pick_first": Takes audio from first channel
+            "none": assume only one channel
         :param returnn_root: RETURNN repository
         """
         self.set_vis_name("Dump audio to HDF")
         assert output_dtype in ["float64", "float32", "int32", "int16"]
+        assert multi_channel_strategy in [
+            "pick_first",
+            "none",
+        ], f"multi_channel_strategy {multi_channel_strategy} is not implemented"
 
         self.bliss_corpus = bliss_corpus
         self.segment_file = segment_file
         self.output_dtype = output_dtype
+        self.multi_channel_strategy = multi_channel_strategy
         self.returnn_root = returnn_root
         self.rqmt = {}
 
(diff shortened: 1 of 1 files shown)


returnn/hdf.py
```diff
@@ -207,26 +207,38 @@ class BlissToPcmHDFJob(Job):
...
     """
 
+    __sis_hash_exclude__ = {"multi_channel_strategy": "none"}
+
     def __init__(
         self,
         bliss_corpus: tk.Path,
         segment_file: Optional[tk.Path] = None,
         output_dtype: str = "int16",
+        multi_channel_strategy: str = "none",
```
vieting, line 217: Wouldn't it be nicer to have the actual `None` as default? Also, setting an integer to pick the channel seems quite straight-forward. More advanced things like channel-averaging might be specified via strings.

michelwi: > Wouldn't it be nicer to have the actual None as default? 

I was thinking this would be "an existing strategy to do mothing" instead of "no strategy", but I am open to have `None` as default (or change it to `"assert_single_channel"`).

> Also, setting an integer to pick the channel seems quite straight-forward. More advanced things like channel-averaging might be specified via strings.

I would like to avoid very elaborate multi-typing, e.g.
```
multi_channel_strategy: Optional[Union[str,int]] = None,
```

The "most correct" thing would be to have
```
class strategy(enum.Enum):
  ASSERT_SINGLE_CHANNEL = enum.auto()
  PICK_NTH = enum.auto()

[...]
        multi_channel_strategy: strategy = strategy.ASSERT_SINGLE_CHANNEL,
        multi_channel_options: Dict[str,Any] = None
```
where in the case `PICK_NTH` we can select the n in the options, and e.g. for averaging we can have arithmetic or geometric, or define some weights...

But idk, if you want to go as far now..

vieting: Well, I don't find the multi-typing you described so terrible. But I'm fine with leaving it as a string as well. I agree we shouldn't overcomplicate as long as only one case is needed.

JackTemaki: The strategy can even be an object holding the options itself, I think I would prefer that solution. So it can be a dataclass inherited from an "empty" class instead of an enum.

so something like (defined maybe within the job):
```python
class Strategy():
    pass
    
 class AssertSingleChannel(Strategy):
     pass
 
 @dataclass(frozen=True)
 class PickNth(Strategy):
     channel: int
 ```

Comment (michelwi) [reply returnn/hdf.py]: I tried your approach now. I was unsure about how to have an abstract baseclass `Strategy` as a type that itself never should be given as an argument. (And because of the "else" both `Strategy` and `AssertSingleChannel` were redundant in all but name.)

So now instead the `BaseStrategy` means no handling i.e. assert single channel and the only derived strategy `PickNth` is a dataclass.

**Review by JackTemaki — APPROVED** [review]

File: returnn/hdf.py [hunk 2]
```diff
@@ -207,18 +208,32 @@ class BlissToPcmHDFJob(Job):
...
     """
 
+    class BaseStrategy:
+        pass
+
+    @dataclass(frozen=True)
+    class PickNth(BaseStrategy):
+        channel: int
+
+    __sis_hash_exclude__ = {"multi_channel_strategy": "none"}
```
michelwi, line 218: this needs to be changed as well. But would this work, or would a different BaseStrategy instance be a problem?
```suggestion
    __sis_hash_exclude__ = {"multi_channel_strategy": BaseStrategy()}
```

michelwi: I request changes on my own pr.. 🤷 
@JackTemaki can we revert this merge, as this job is used in production but the sis hash exclude does not work I.e. hashes will be different.

JackTemaki: I think this would not work:

```python
class Demo():
    pass
    
exclude = {"x": Demo()}

def f(x = Demo()):
    if x != exclude["x"]:
        print("would include")
    else:
        print("would not include")
 ```
 yields "would include", and this is the check sisyphus does.

JackTemaki: So the same example with:
```python
class Demo():
    def __eq__(self, other):
        return type(other) == Demo
        ```
works.

PR #353, opened 2023-01-24
feat: Show highest finished training epoch in sis overview

NeoLegends:
This PR adds a small indicator showing RETURNN training progress. It shows the highest finished training epoch and the number of total epochs. I tried to keep it minimal.

Looks like this when run (here on a subclass):
image

My understanding is that sisyphus regularly calls `Job.info()`, so the counter _should_ update itself. Please correct me if I'm wrong.

I'm open for suggestions for visual improvements. Looking forward to y'all's reviews!

### DIFF
--- a/returnn/training.py
+++ b/returnn/training.py
@@ -239,21 +239,27 @@ class ReturnnTrainingJob(Job):
         return run_cmd
 
     def info(self):
-        def try_load_lr_log(file_path: str):
+        def try_load_lr_log(file_path: str) -> Optional[dict]:
             # Used in parsing the learning rates
             @dataclass
             class EpochData:
                 learningRate: float
                 error: Dict[str, float]
 
-            with open(file_path, "rt") as file:
-                return eval(file.read().strip())
+            try:
+                with open(file_path, "rt") as file:
+                    return eval(file.read().strip())
+            except FileNotFoundError:
+                return None
 
         lr_file = os.path.join(
             self._sis_path(gs.JOB_WORK_DIR),
             self.returnn_config.get("learning_rate_file", "learning_rates"),
         )
         epochs = try_load_lr_log(lr_file)
+
+        if epochs is None:
+            return None
 
         if not isinstance(epochs, dict):
             raise TypeError(
--- a/returnn/training.py
+++ b/returnn/training.py
@@ -9,11 +9,12 @@ __all__ = [
 ]
 
 import copy
+from dataclasses import dataclass
 import sys
 import os
 import shutil
 import subprocess as sp
-from typing import List, Union
+from typing import Dict, List, Optional, Union
 
 from sisyphus import *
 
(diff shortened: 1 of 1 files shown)


NeoLegends:
Ah, the impl as it is is too slow to be merged. It's going through all checkpoints, which is too much FS work.

NeoLegends:
According to @JackTemaki the sis manager doesn't periodically refresh, but only actually when a job state changes (which, in pipelines w/ only training jobs, doesn't happen often). Therefore this is way less useful than I imagined.

Is there perhaps another (clean) way to trigger the sis manager to update a job's `.info()`?

michelwi:
> Or parsing a logfile instead of going through the checkpoints.

you could parse the `self.out_learning_rates` file. This will also include all checkpoints even when `cleanup_old_models` or `save_interval` is set

michelwi:
> According to @JackTemaki the sis manager doesn't periodically refresh, but only actually when a job state changes (which, in pipelines w/ only training jobs, doesn't happen often). Therefore this is way less useful than I imagined.

You could always implement a new parameter in sisyphus global_settings.py e.g. `JOBINFO_REFRESH_INTERVAL` and extend [this ](https://github.com/rwth-i6/sisyphus/blob/cc58196b0c410a6d98ba7c1d79ac2cd1eb48ae39/sisyphus/manager.py#L588-L596) block to auto update after `JOBINFO_REFRESH_INTERVAL` many seconds even if not state update was found.

NeoLegends:
Parsing the log file seems fast enough. I'm now doing that.

With this done I think we can discuss merging this PR. Finding a solution to the periodic refresh seems like a different task (and a different PR).

What I could think of for the meanwhile is indicating that the found epoch is the lower bound. E.g. instead of logging `ep 73/500` log `ep 73+/600`.

NeoLegends:
> You could always implement a new parameter in sisyphus global_settings.py e.g. JOBINFO_REFRESH_INTERVAL and extend [this ](https://github.com/rwth-i6/sisyphus/blob/cc58196b0c410a6d98ba7c1d79ac2cd1eb48ae39/sisyphus/manager.py#L588-L596)block to auto update after JOBINFO_REFRESH_INTERVAL many seconds even if not state update was found.

That would be an idea.

I wonder if people would want a sisyphus that printed a new job summary (w/ potentially unchanging info) every 5mins or so though. It seems rather clean to me the current sisyphus _does not_ bloat the console if nothing has changed. Erasing previously written output from the console could be an option, this might lead down a rabbit hole though?

returnn/training.py
```diff
@@ -237,6 +238,48 @@ def _get_run_cmd(self):
...
+                except FileNotFoundError:
+                    pass
+
+            return None
+
+        lr_during_training = os.path.join(
+            self._sis_path(gs.JOB_WORK_DIR),
+            self.returnn_config.get("learning_rate_file", "learning_rates"),
+        )
+        lr_after_training = tk.uncached_path(self.out_learning_rates)
```
Atticus1806, line 262: This should not be required, since the LR file will stay in the work folder even if training is finished.

Atticus1806: also don't use `uncached_path` but instead `.get_path()`

michelwi: > This should not be required, since the LR file will stay in the work folder even if training is finished.

There exists an option in sisyphus to clean up the work folder after the training is finished. If someone had this activated, then the `lr_after_training` would be necessary.

But +1 for 
```suggestion
        lr_after_training = self.out_learning_rates.get_path()
```

JackTemaki: > > This should not be required, since the LR file will stay in the work folder even if training is finished.
> 
> There exists an option in sisyphus to clean up the work folder after the training is finished. If someone had this activated, then the `lr_after_training` would be necessary.

To my understanding `info` will only ever be called when the job is running, so if it is finished and cleaned, we do not need this function anyway.

returnn/training.py
```diff
@@ -237,6 +238,37 @@ def _get_run_cmd(self):
 
         return run_cmd
 
+    def info(self):
+        def try_load_lr_log(file_path: str):
```
michelwi, line 242: you have now removed the "try" part from the `try_load_lr_log` function completely. (If you keep it this way, remove the try part from the name).

But I am wondering if in some cases this will fail (e.g. training just startet so no LR file will be present) and if the manager then crashes..?

NeoLegends: Yeah leaving the check in is probably best.

christophmluscher (approved):

michelwi (approved):

PR #355, opened 2023-02-13
rename E2E to SpokenForm

JackTemaki:
No one is using this Job yet, so we want to rename it before it is too late.

### DIFF
--- a/datasets/switchboard.py
+++ b/datasets/switchboard.py
@@ -700,11 +700,11 @@ class CreateRT03sCTSCorpusJob(Job):
         shutil.copy(glm_file, self.out_glm.get_path())
 
 
-class CreateSwitchboardE2EBlissCorpusJob(Job):
+class CreateSwitchboardSpokenFormBlissCorpusJob(Job):
     """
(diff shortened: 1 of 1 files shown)


christophmluscher (approved):

mmz33 (approved):

datasets/switchboard.py
```diff
@@ -700,11 +700,11 @@ def run(self):
         shutil.copy(glm_file, self.out_glm.get_path())
 
 
-class CreateSwitchboardE2EBlissCorpusJob(Job):
+class CreateSwitchboardSpokenFormBlissCorpusJob(Job):
     """
     Creates a special E2E version of switchboard-1 used for e.g. BPE or Sentencepiece based models.
```
dthulke, line 705: Shouldn't this be changed, too?
```suggestion
    Creates a special spoken form version of switchboard-1 used for e.g. BPE or Sentencepiece based models.
```

Also `self.out_e2e_corpus` -> `self.out_spoken_form_corpus `

PR #357, opened 2023-02-15
add all_unkown flag to FilterCorpusRemoveUnknownWordSegmentsJob

JackTemaki:
The current job only removed segments where ALL words are unknown. But of course we also want to have the possibility to filter sequence where at least one word is unknown, thus this added flag.


michelwi (approved):
I am not particularly happy with the name of the parameter (as it is not self explanatory). But I have no better idea, so approve 🤷

mmz33 (approved):

PR #361, opened 2023-02-17
add workflow to trigger bitbucket pipeline

michelwi:
The general plan is the following
* this workflow triggers a pipeline to be run on bitbucket 
* we pass the current branch name to bitbucket
* bitbucket clones the branch of the pr instead of main
* this workflow waits while bitbucket runs the pipeline
* this workflow returns result of pipeline

General TODOS
* all done

Security concerns:
* the authentication token is stored inside a github secret that only owners(?) od i6_core have read access to
* the contents of the secret are redacted from all logs
* users could create malicious pull requests changing the workflow of i6_core wich accesses the token but
a) the token only allows to start/stop/read existing pipelines
b) github does not automatically run workflows for first time contributors

### DIFF
--- a/.github/workflows/apptek_hashes.yml
+++ b/.github/workflows/apptek_hashes.yml
@@ -4,7 +4,7 @@ env:
   auth_token: "-H 'Authorization: Bearer ${{ secrets.APPTEK_BITBUCKET_AUTHENTICATION }}'"
   get_header: "-X GET -s -H 'Accept: application/json'"
   post_header: "-X POST -s -H 'Content-Type: application/json'"
-  post_content: '{"target": {"ref_type": "branch", "ref_name": "willi-i6_core-integration", "type": "pipeline_ref_target", "selector": {"type": "custom", "pattern": "hashes-only"}}, "variables": [{"key": "i6_core_branch", "value": "'
+  post_content: '{"target": {"ref_type": "branch", "ref_name": "willi-i6_core-integration", "type": "pipeline_ref_target", "selector": {"type": "custom", "pattern": "hashes+tests"}}, "variables": [{"key": "i6_core_branch", "value": "'
 on:
   push:
     branches:
--- a/.github/workflows/apptek_hashes.yml
+++ b/.github/workflows/apptek_hashes.yml
@@ -1,4 +1,12 @@
 name: AppTek hashes
+env:
+  api_url: "https://api.bitbucket.org/2.0/repositories/omnifluent/apptek_asr/pipelines/"
+  auth_token: "-H 'Authorization: Bearer ${{ secrets.APPTEK_BITBUCKET_AUTHENTICATION }}'"
+  get_header: "-X GET -s -H 'Accept: application/json'"
+  post_header: "-X POST -s -H 'Content-Type: application/json'"
+  post_content: '{"target": {"ref_type": "branch", "ref_name": "willi-i6_core-integration", "type": "pipeline_ref_target", "selector": {"type": "custom", "pattern": "hashes-only"}}, "variables": [{"key": "i6_core_branch", "value": "'${GITHUB_HEAD_REF}'"}]}'
+
+
 on:
   push:
     branches:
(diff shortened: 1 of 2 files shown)


JackTemaki (approved):
So from what I understand is that this is triggering the bitbucket pipelines, meaning no computation is done on the Github side and we are just polling the result of the bitbucket pipeline every 10 seconds. I assume that this is also basically a cheap op.

As you did all the relevant tests I do not see why this should not be merged. 

Of course we now have the problem that we can not see why the change broke something, but we can see how to deal with this once it happens and it is not directly visible why.

michelwi:
> @JackTemaki marked this pull request as ready for review

I will keep it as draft because it depends on the associated PR in Bitbucket (B-PR) and this will be still changes needed after B-PR has landed.

> Of course we now have the problem that we can not see why the change broke something,

yes, this is intended to reduce the possibility to reverse-engineer our pipelines. I might add a message like "The AppTek Pipeline <ID here> has indicated some problems. Please check with wmichel@apptek.com (or ebeck, or ..) to find out what is wrong"

.github/workflows/apptek_hashes.yml
```diff
@@ -0,0 +1,49 @@
...
+        python-version: 3.8
+    - name: Start Bitbucket Pipeline
+      run: |
+        curl -X POST -s -u ${{ secrets.MICHEL_BITBUCKET_AUTHENTICATION }} \
+            -H 'Content-Type: application/json' \
+            https://api.bitbucket.org/2.0/repositories/omnifluent/apptek_asr/pipelines/ \
+            -d '{"target": {"ref_type": "branch", "ref_name": "willi-i6_core-integration",
+                 "type": "pipeline_ref_target", "selector": {"type": "custom", "pattern": "hashes-only"}},
+                 "variables": [{"key": "i6_core_branch", "value": "'${GITHUB_HEAD_REF}'"}]}' |\
+        python -c "import sys,json; print(json.load(sys.stdin)['uuid'])" |\
```
curufinwe, line 24: You can do json processing the linux way with https://stedolan.github.io/jq/ , no need for python. Also: if you are loading python you could do the postprocessing there and not in sed.

.github/workflows/apptek_hashes.yml
```diff
@@ -0,0 +1,49 @@
...
+        sed 's/{/%7B/' | sed 's/}/%7D/' > pipeline_uuid.txt
+    - name: Wait for Results
+      run: |
+        while [ COMPLETED != $(\
+            curl -X GET -s -u ${{ secrets.MICHEL_BITBUCKET_AUTHENTICATION }} \
```
curufinwe, line 29: there is a lot of repetition in the curl commands, could we put the flags that are always used in env variables (e.g. $CURL_FLAGS) and reference them here?

curufinwe (changes requested):

.github/workflows/apptek_hashes.yml
michelwi: I wasn't sure if jq was available.. but then I could also just install it in the pipeline :D

.github/workflows/apptek_hashes.yml
michelwi: I have tried to unify it in a sensible way now.

.github/workflows/apptek_hashes.yml
curufinwe: please do

g2p/apply.py
```diff
@@ -28,6 +28,7 @@ def __init__(
         g2p_python=None,
         filter_empty_words=False,
         concurrent=1,
+        nonce="this value is not needed",
```
curufinwe, line 31: I guess you added this to make sure the pipeline fails.

michelwi: correct. I was testing the error message that point the user to us to check what is wrong with the test. (See the failed Pipeline for the current version of the error Message)

michelwi: and also to prevent people from merging this PR before our internal PR landed

.github/workflows/apptek_hashes.yml
michelwi: done

michelwi:
All prerequisites are in place now to land this PR.

I have another change open that should reduce test time on AppTek side from 7 to 5 minutes, but that will happen transparently for i6_core.

curufinwe (approved):

PR #363, opened 2023-02-23
fix ffmpeg all_recordings()

Paubaquero:


michelwi (approved):
The original code missed all recordings that were located in subcorpora. This change should still give the same results in case of corpora without any subcorpus.

Plus the `run_recover_duration` function is already using `all_recordings()` -> more consistent.

PR #366, opened 2023-02-27
add default branch for apptek workflow

michelwi:
As the environment variable `${GITHUB_HEAD_REF}` is only set during a PR, therefore currently the workflow breaks after the merge on the main branch. As a quick fix, I add the branch name "main" as a fallback value.

This will break if we rename main or if somebody wants to run a manual test for a specific branch outside of a pull request.


JackTemaki (approved):

JackTemaki:
I will bypass regulations for this PR so that we can directly see the result.

PR #369, opened 2023-03-03
SearchWordsToCTMJob, empty_seq_label option

### LINKED ISSUE #368: SearchWordsToCTMJob: better handling of empty seqs, causes sclite to fail otherwise with "... not synchronized"
I see this in the ctm:
```
4515-11057-0052 1 0.000000 0.000000 HE 0.99
4515-11057-0052 1 0.000000 0.000000 WENT 0.99
4515-11057-0052 1 0.000000 0.000000 PRESENTLY 0.99
;; dev-other/4515-11057-0053/4515-11057-0053 (0.000000-0.000000)
;; dev-other/4515-11057-0054/4515-11057-0054 (0.000000-0.000000)
4515-11057-0054 1 0.000000 0.000000 WHERE 0.99
4515-11057-0054 1 0.000000 0.000000 HE 0.99
4515-11057-0054 1 0.000000 0.000000 MARRIED 0.99
```
So 4515-11057-0053 is empty here.

Then later sclite fails with:
```
hyp file '4515-11057-0054' and ref file '4515-11057-0053' not synchronized
sclite: Alignment failed.  Exiting.
```

@curufinwe said that RASR has a workaround that it always adds some dummy label in case it would be empty otherwise. Btw, what label specifically? Does it matter? Just `<EMPTY>` or so?

Should we add such logic to `SearchWordsToCTMJob` as well?

What label would we use then?

Or is there a more explicit way to tell sclite that this is an empty sequence?

I assume this should be a new option, to not just cause different behavior than before? And by default disabled to not change hashes?

albertz:
Fix #368.

Note that I'm unsure about a reasonable default. I think we have a conflict between:

- Keeping hash the same.
- Keeping behavior the same for a given hash.
- Reasonable default which probably would be used by everyone.

Note: I personally would just change and fix the behavior, maybe without even adding an option, as any such ctm files were anyway broken for sclite.

### DIFF
--- a/returnn/search.py
+++ b/returnn/search.py
@@ -375,9 +375,6 @@ class SearchWordsToCTMJob(Job):
         :param Path recog_words_file: search output file from RETURNN
         :param Path bliss_corpus: bliss xml corpus
         :param bool filter_tags: if set to True, tags such as [noise] will be filtered out
-        :param empty_seq_label: if set, this label will be used for empty sequences,
-            otherwise empty sequences do not have any entries
-            (but an empty seq usually causes sclite to fail, so you probably always want to set this)
         """
         self.recog_words_file = recog_words_file
         self.bliss_corpus = bliss_corpus
--- a/returnn/search.py
+++ b/returnn/search.py
@@ -15,7 +15,7 @@ import logging
 import os
 import shutil
 import subprocess as sp
-from typing import Any, Union, Set, Dict
+from typing import Optional, Any, Union, Set, Dict
 
 from sisyphus import *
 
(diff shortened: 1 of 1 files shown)


JackTemaki:
If it is clear that the results are not influenced for the case that no empty hypothesis exists, and in every case where an empty sequence exists the following Sclite would crash, I would be fine with changing the default behavior but keeping the hash the same.

albertz:
> If it is clear that the results are not influenced for the case that no empty hypothesis exists,

Yes, see the code.


> and in every case where an empty sequence exists the following Sclite would crash

I don't really know but I think so. At least that's what I'm getting now, and by this change, it runs through. Maybe someone can verify?

michelwi:
> > and in every case where an empty sequence exists the following Sclite would crash
> 
> I don't really know but I think so. At least that's what I'm getting now, and by this change, it runs through. Maybe someone can verify?

Of cause there are ways around this (e.g. add a pipeline job to add some tokens, or pass the (I believe) `-m hyp` option ((not recommended, this truncates the reference to the set of available hypotheses and leads to false WERs))) but I believe sclite does not handle empty hypotheses.
So I agree with @JackTemaki

mmz33 (approved):

albertz:
Ok I removed the option, and just always enabled this, using the hardcoded label `<empty-sequence>`.
I also added a comment.

Atticus1806:
Always enabling implies that empty sequences are „correct“ output and not some part of the pipeline failing right? I am wondering if this can hide errors.

Maybe we can dump all sequences where this is applied to a separate output file and if one wants to check that they could check if the file is empty.

returnn/search.py
```diff
@@ -15,7 +15,7 @@
 import os
 import shutil
 import subprocess as sp
-from typing import Any, Union, Set, Dict
+from typing import Optional, Any, Union, Set, Dict
```
michelwi, line 18:
```suggestion
from typing import Any, Union, Set, Dict
```
import not needed any more

albertz:
> Always enabling implies that empty sequences are „correct“ output and not some part of the pipeline failing right? I am wondering if this can hide errors.

Most (all?) our models allow for an empty sequence as output. But if all your output is empty, you would also just get 100% WER, so not really hidden.


> in Sclite we then show a substitution instead of a deletion error

That is right, but I think we don't really have any alternative here, except of a crash like currently.

Usually, when this happens, the model is often not so good, and WERs are anyway high. When you look into the pra file, you also see this special label there.

JackTemaki (approved):

PR #375, opened 2023-03-10
DumpHDFJob: write dump config into explicit file

JackTemaki:
I had issues passing the string directly to the script when using this code via ReturnnRasrDumpHDFJob. Writing the dict into an explicit file and passing it helped.

### DIFF
--- a/returnn/hdf.py
+++ b/returnn/hdf.py
@@ -80,7 +80,10 @@ class ReturnnDumpHDFJob(Job):
         data = self.data
         if isinstance(data, dict):
             instanciate_delayed(data)
-            data = str(data)
+            with open("dataset.config", "wt") as dataset_file:
+                dataset_file.write("#!rnn.py\n")
+                dataset_file.write("train = %s\n" % str(data))
+            data = "dataset.config"
         elif isinstance(data, tk.Path):
             data = data.get_path()
 
--- a/returnn/hdf.py
+++ b/returnn/hdf.py
@@ -181,7 +181,10 @@ class ReturnnRasrDumpHDFJob(ReturnnDumpHDFJob):
         self.alignment = alignment
         self.rasr_exe = rasr.RasrCommand.select_exe(crp.nn_trainer_exe, "nn-trainer")
         self.feature_flow = ReturnnRasrTrainingJob.create_flow(feature_flow, alignment)
-        (self.rasr_config, self.rasr_post_config,) = ReturnnRasrTrainingJob.create_config(
+        (
+            self.rasr_config,
+            self.rasr_post_config,
+        ) = ReturnnRasrTrainingJob.create_config(
             crp=crp,
             alignment=alignment,
             num_classes=num_classes,
--- a/returnn/hdf.py
+++ b/returnn/hdf.py
@@ -74,18 +74,27 @@ class ReturnnDumpHDFJob(Job):
         self.out_hdf = self.output_path("data.hdf")
 
     def tasks(self):
+        if isinstance(self.data, (dict, str)):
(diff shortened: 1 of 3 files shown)


returnn/hdf.py
```diff
@@ -75,7 +75,10 @@ def run(self):
         data = self.data
         if isinstance(data, dict):
             instanciate_delayed(data)
-            data = str(data)
+            with open("dataset.config", "wt") as dataset_file:
```
michelwi, line 78: If you say it works if we write just `train = ...` in a config file, 
then we can do it..

I was just now a bit hesitant that we have a config file in the work folder that only exists after the job has started. But the alternatives are 
* have a Task "write_config" that does only exist if data is not a path
* make the config an output of the job

I'll leave this comment here for a while and if nobody else got better solutions then we keep it.

Atticus1806:
Whats the status on this? So the current version works for me.

returnn/hdf.py
```diff
@@ -75,7 +75,10 @@ def run(self):
...
+                dataset_file.write("#!rnn.py\n")
+                dataset_file.write("train = %s\n" % str(data))
+            data = "dataset.config"
         elif isinstance(data, tk.Path):
             data = data.get_path()
```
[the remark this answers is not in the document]
michelwi: what makes you think this would not work? 

This code lives in the run function i.e. is executed in the worker. Therefore we can call `data.get_path()`. After that `data` is a str containing the location of the config file. I think providing the path to a config file is supported cf. https://github.com/rwth-i6/returnn/blob/9b8928676bdde9af93809346d6abf829d428c8b9/tools/hdf_dump.py#L102-L104
-> should work.

michelwi (commented):
> Whats the status on this? So the current version works for me.

There are still some open comments from my side that I find not yet fully addressed. 
* that if the input of the job is a `str`, it is not written to file but should be
* weather or not the config should be an output or if we want a separate task to write it

Atticus1806:
Also pinging @curufinwe since I am not sure about @michelwi 's availability.

christophmluscher (approved):

Atticus1806:
ping @JackTemaki @michelwi @vieting @curufinwe

michelwi (approved):

returnn/hdf.py
```diff
@@ -74,15 +74,27 @@ def __init__(
...
-            instanciate_delayed(data)
-            data = str(data)
-        elif isinstance(data, tk.Path):
-            data = data.get_path()
+        instanciate_delayed(data)
+        data = str(data)
+        with open("dataset.config", "wt") as dataset_file:
+            dataset_file.write("#!rnn.py\n")
+            dataset_file.write("train = %s\n" % str(data))
+        self.data = "dataset.config"
```
JackTemaki, line 91: It surprises me that this works, `self` assignments in Tasks should not be persistent.

michelwi: I think in run the original self.data (str, dict) is used and the config file is only for the users information. We can remove this line.
```suggestion
```

returnn/hdf.py
```diff
@@ -74,15 +74,27 @@ def __init__(
...
+
+    def run(self):
+        if isinstance(self.data, tk.Path):
+            data = self.data.get_path()
+        else:
+            data = self.data
```
JackTemaki, line 97: we could just set `data = "dataset.config"` here, or not?

michelwi: Yes

Atticus1806:
ping @JackTemaki

JackTemaki:
> ping @JackTemaki

"approve"

PR #378, opened 2023-03-14
Add data key as parameter to ExtractDatasetMeanStddevJob

mmz33:
The default key used by `tools/dump-dataset.py` is "data" which is used as default param here.


JackTemaki (approved):

JackTemaki:
@mmz33 you need to assign more than one reviewer

michelwi (approved):

JackTemaki:
@michelwi thanks!

PR #379, opened 2023-03-14
Add clone submodules parameter to CloneGitRepositoryJob

mmz33:
Add a flag to clone submodules if needed. In my case, I had to use repos from `returnn/extern`.

### DIFF
--- a/tools/git.py
+++ b/tools/git.py
@@ -42,11 +42,11 @@ class CloneGitRepositoryJob(Job):
         logging.info("running command: %s" % " ".join(args))
         sp.run(args, check=True)
 
-        if self.clone_submodules:
-            args = ["git", "submodule", "update", "--init", "--recursive"]
-            sp.run(args, cwd=repository_dir, check=True)
-
         if self.commit is not None:
             args = ["git", "checkout", self.commit]
             logging.info("running command: %s" % " ".join(args))
+            sp.run(args, cwd=repository_dir, check=True)
+
+        if self.clone_submodules:
+            args = ["git", "submodule", "update", "--init", "--recursive"]
             sp.run(args, cwd=repository_dir, check=True)
--- a/tools/git.py
+++ b/tools/git.py
@@ -2,6 +2,7 @@ __all__ = ["CloneGitRepositoryJob"]
 
 import logging
 import subprocess as sp
+from typing import Optional
 
 from sisyphus import *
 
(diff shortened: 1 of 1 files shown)


michelwi (commented):
I am wondering if it would be better to have the submodule init call after the specific commit checkout call.

What would happen if some submodules were updated between `commit` and `HEAD` in either case?

mmz33:
@michelwi Ah I thought that the commits are independent. So what do you suggest?

michelwi:
> @michelwi Ah I thought that the commits are independent. So what do you suggest?

I genuinely don't know how/if this works. I imagined that e.g. at commit `123` the repo is
```
repo/
  submodule1
  submodule2@version1
  own_module
```
and then it is later updated to `HEAD`
```
repo/
  submodule2@version2
  own_module
```

Now in the current implementation if you run `CloneGitRepositoryJob("repo", commit="123", clone_submodules=True)`
first the `HEAD` is checked out, then the submodules are initialized (i.e. `submodule2@version2`) and then the specific commit `123` is checked out.
My question is: will `submodule1` now also be prepared? And will `submodule2`'s version be moved to `version1`?

mmz33:
ok so this can be fixed by doing the submodule cloning at the end after checking out the commit. this should work i think.

tools/git.py
```diff
@@ -38,6 +41,11 @@ def run(self):
         args += [repository_dir]
         logging.info("running command: %s" % " ".join(args))
         sp.run(args, check=True)
+
+        if self.clone_submodules:
```
christophmluscher, line 45: IMHO it would be easier to directly call:

`git clone --recurse-submodules`

christophmluscher: although this might run into the same issue as mentioned by Willi

michelwi: I did some research and the same flag also exists for checkout

> ```git checkout --recurse-submodules```
Using --recurse-submodules will update the content of all initialized submodules according to the commit recorded in the superproject.
If local modifications in a submodule would be overwritten the checkout will fail unless -f is used.

so you could change the PR in a way that if the flag `clone_submodules` is set, then `--recurse-submodules` is added both to clone and checkout.

JackTemaki (approved):
I verified with @mmz33 locally that this works.

JackTemaki:
@Atticus1806 thanks a lot

Atticus1806 (approved):

PR #384, opened 2023-03-17
Implement AddEowPhonemesToLexiconJob

SimBe195:
Implements new `AddEowPhonemesToLexiconJob` which adds eow-phonemes to a lexicon. Such an extended lexicon is useful for recognition with the upcoming GenericSeq2Seq decoder.

### DIFF
--- a/lexicon/modification.py
+++ b/lexicon/modification.py
@@ -3,7 +3,7 @@ __all__ = ["WriteLexiconJob", "MergeLexiconJob", "AddBoundaryMarkerToLexiconJob"
 import copy
 from collections import OrderedDict, defaultdict
 import itertools
-from typing import Generator, List, Set
+from typing import Generator, List, Optional, Set
 
 from sisyphus import Job, Task, tk
 
@@ -189,15 +189,21 @@ class MergeLexiconJob(Job):
 
 
 class AddEowPhonemesToLexiconJob(Job):
-    def __init__(self, bliss_lexicon: tk.Path):
+    def __init__(self, bliss_lexicon: tk.Path, nonword_phones: Optional[List] = None, boundary_marker: str = "#"):
         """
         Extends phoneme set of a lexicon by additional end-of-word (eow) versions
         of all regular phonemes. Modifies lemmata to use the new eow-version
         of the final phoneme in each pronunciation.
 
-        :param tk.Path bliss_lexicon: Base lexicon to be modified.
+        :param bliss_lexicon: Base lexicon to be modified.
+        :param nonword_phones: List of nonword-phones for which no eow-versions will be added, e.g. [noise].
+                               Phonemes that occur in special lemmata are found automatically and do not need
+                               to be specified here.
+        :param boundary_marker: String that is appended to phoneme symbols to mark eow-version.
         """
         self.bliss_lexicon = bliss_lexicon
+        self.nonword_phones = nonword_phones or []
+        self.boundary_marker = boundary_marker
 
         self.out_lexicon = self.output_path("lexicon.xml")
 
@@ -219,13 +225,14 @@ class AddEowPhonemesToLexiconJob(Job):
                 continue
             for phon in lemma.phon:
                 special_phonemes.update(phon.split())
+        special_phonemes.update(self.nonword_phones)
 
         # Add all phonemes to out_lexicon and create list of eow-phonemes
         eow_phonemes = []
         for phoneme, variation in in_lex.phonemes.items():
             out_lex.add_phoneme(phoneme, variation)
             if phoneme not in special_phonemes:
-                eow_phonemes.append((phoneme + "#", variation))
+                eow_phonemes.append((phoneme + self.boundary_marker, variation))
 
         # Add eow-phonemes to out_lexicon
         for eow_phoneme, variation in eow_phonemes:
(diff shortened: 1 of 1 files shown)


lexicon/modification.py
```diff
@@ -185,3 +186,56 @@ def run(self):
...
+
+    def run(self):
+        in_lex = lexicon.Lexicon()
+        in_lex.load(self.bliss_lexicon.get_path())
+
+        out_lex = lexicon.Lexicon()
+
+        # Identify all 'special' phonemes in the given lexicon. A phoneme is
+        # deemed 'special' if it appears in the pronunciation of a special lemma.
+        # This should identify e.g. 'silence', 'blank', 'unknown' etc.
```
michelwi, line 215: how about `[noise]` and similar?

lexicon/modification.py
```diff
@@ -185,3 +186,56 @@ def run(self):
...
+                continue
+            for phon in lemma.phon:
+                special_phonemes.update(phon.split())
+
+        # Add all phonemes to out_lexicon and create list of eow-phonemes
+        eow_phonemes = []
+        for phoneme, variation in in_lex.phonemes.items():
+            out_lex.add_phoneme(phoneme, variation)
+            if phoneme not in special_phonemes:
+                eow_phonemes.append((phoneme + "#", variation))
```
michelwi, line 228: would it make sense to make the EOW marker `"#"` configurable by the user?

SimBe195: Sure, why not.

lexicon/modification.py
SimBe195: Yes, fixed now by adding a `nonword_phones` parameter to the job where such phonemes can be specified.

lexicon/modification.py
```diff
@@ -1,10 +1,11 @@
-__all__ = ["WriteLexiconJob", "MergeLexiconJob"]
+__all__ = ["WriteLexiconJob", "MergeLexiconJob", "AddBoundaryMarkerToLexiconJob"]
```
JackTemaki, line 1: Does not match the job name

JackTemaki (changes requested):

lexicon/modification.py
```diff
@@ -185,3 +186,63 @@ def run(self):
...
+            if phoneme not in special_phonemes:
+                eow_phonemes.append((phoneme + self.boundary_marker, variation))
+
+        # Add eow-phonemes to out_lexicon
+        for eow_phoneme, variation in eow_phonemes:
+            out_lex.add_phoneme(eow_phoneme, variation)
+
+        # Modify lemmata to include eow-phoneme
+        for lemma in in_lex.lemmata:
+            if lemma.special is None and lemma.phon is not None:
```
michelwi, line 243: And you also need to check, that the (last?) phone of the lemma is not in `special_phones` (or more precise `self.nonword_phones`) as here I believe you add the boundary marker to the noise lemma.

SimBe195: Yes that's true. Fixed now.

lexicon/modification.py
```diff
@@ -210,6 +210,32 @@ def __init__(self, bliss_lexicon: tk.Path, nonword_phones: Optional[List] = None
...
+        :param pron: Original pronunciation as a string containing phonemes separated by whitespaces
+        :param special_phonemes: Set of special phonemes that should be skipped over.
+        :return: Modified pronunciation
+        """
+        phoneme_list = pronunciation.split()
+
+        for i, phoneme in reversed(list(enumerate(phoneme_list))):
+            if phoneme in special_phonemes:
+                continue
+            phoneme_list[i] = self._eow_phoneme(phoneme)
```
michelwi, line 235:
```suggestion
            phoneme_list[i] = self._eow_phoneme(phoneme)
            break
```
I think you are currently adding eow marker to all non-special phonemes. 
Fix this by ending the for-loop after first match.

lexicon/modification.py
```diff
@@ -185,3 +186,89 @@ def run(self):
...
+        return phoneme + self.boundary_marker
+
+    def _modify_pronunciation(self, pronunciation: str, special_phonemes: Set[str]) -> str:
+        """
+        Find the rightmost phoneme in the pronunciation that is not special and replace it by its
+        EOW version. Might do nothing if all phonemes are special.
+
+        Example: "AA BB [noise]" -> "AA BB# [noise]"
+
+        :param pron: Original pronunciation as a string containing phonemes separated by whitespaces
```
Atticus1806, line 226:
```suggestion
        :param pronunciation: Original pronunciation as a string containing phonemes separated by whitespaces
```

lexicon/modification.py
```diff
@@ -185,3 +186,89 @@ def run(self):
...
+        self.boundary_marker = boundary_marker
+
+        self.out_lexicon = self.output_path("lexicon.xml")
+
+    def tasks(self):
+        yield Task("run", mini_task=True)
+
+    def _eow_phoneme(self, phoneme: str) -> str:
+        """
+        Creates the EOW version of a given phoneme.
```
Atticus1806, line 215:
```suggestion
        Creates the eow-version of a given phoneme.
```

Atticus1806 (approved):
Looks good. Only minor doc stuff.

lexicon/modification.py
```diff
@@ -185,3 +186,89 @@ def run(self):
...
+        """
+        return phoneme + self.boundary_marker
+
+    def _modify_pronunciation(self, pronunciation: str, special_phonemes: Set[str]) -> str:
+        """
+        Find the rightmost phoneme in the pronunciation that is not special and replace it by its
+        EOW version. Might do nothing if all phonemes are special.
```
christophmluscher, line 222:
```suggestion
        eow-version. Might do nothing if all phonemes are special.
```

JackTemaki (approved):

PR #385, opened 2023-03-24
remove duplicate pronunciations from g2p output

michelwi:
In a pipeline that processes G2P outputs sometimes the same pronunciation can be assigned to the same orth multiple times.
```
WTI   0.0    0.8053393333333334     d ʌ b l j u t i aɪ 
WTI   0.0    0.938586       d ʌ b l j u t i aɪ
```
this leads to a lemma having the same pronunciation twice.

Making `oov_words[orth]` a set will mitigate that. 

I think it is save to always do so, as we would never want the same pronunciation twice in a lemma.


sarahberanek (approved):

michelwi:
A ping to the honored reviewers of i6_core.

JackTemaki (approved):

PR #381, opened 2023-03-16
add verify step to apptek hash test

michelwi:
This PR adds a verify step to the AppTek test pipeline so that if the trigger fails, we do not waste build minutes on waiting for a result that never comes.

PR #386, opened 2023-03-30
Add parameter to the AM config for using corrected FSA during viterbi alignment

Marvin84:
Adding the logic for using the applicator of type `corrected` during Viterbi alignment.

### DIFF
--- a/am/config.py
+++ b/am/config.py
@@ -1,7 +1,9 @@
-__all__ = ["acoustic_model_config", "adjust_am_config_for_corrected_applicator"]
+__all__ = ["acoustic_model_config", "get_align_config_and_crp_for_corrected_applicator"]
 
+import copy
 from dataclasses import dataclass
 from typing import List, Literal, Tuple, Union, Optional
+
 import i6_core.rasr as rasr
 
 from sisyphus import tk
@@ -105,17 +107,29 @@ def acoustic_model_config(
     return config
 
 
-def adjust_am_config_for_corrected_applicator(config: rasr.RasrConfig):
+def get_align_config_and_crp_for_corrected_applicator(crp: rasr.CommonRasrParameters, exit_penalty: float = 0.0):
     """
     Set the correct type of applicator, default is "legacy". Moreover, set exit penalities to zero
     For a given word sequence the exit penalty is constant with respect to the max/sum
     :param config:
     :return:
     """
-    config.tdp.applicator_type = "corrected"
+
+    align_crp = rasr.CommonRasrParameters(base=crp)
+    align_crp.acoustic_model_config.tdp.applicator_type = "corrected"
     transition_types = ["*", "silence"]
-    if config.tdp.tying_type == "global-and-nonword":
+    if align_crp.acoustic_model_config.tdp.tying_type == "global-and-nonword":
         for nw in [0, 1]:
             transition_types.append(f"nonword-{nw}")
     for t in transition_types:
-        config.tdp[t].exit = 0.0
+        align_crp.acoustic_model_config.tdp[t].exit = exit_penalty
+
+    align_crp.acoustic_model_config.fix_allophone_context_at_word_boundaries = True
+    align_crp.acoustic_model_config.transducer_builder_filter_out_invalid_allophones = True
+
+    pre_pattern = "acoustic-model-trainer.aligning-feature-extractor.feature-extraction.alignment.allophone-state-graph-builder.orthographic-parser"
+    extra_config = rasr.RasrConfig()
+    extra_config[f"{pre_pattern}.normalize-lemma-sequence-scores"] = False
+    extra_config[f"{pre_pattern}.allow-for-silence-repetitions"] = False
+
+    return align_crp, extra_config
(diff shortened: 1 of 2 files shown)


meta/mm_sequence.py
```diff
@@ -117,8 +119,23 @@ def update_rqmt(rqmt, extra):
...
+                    from i6_core.rasr.crp import CommonRasrParameters
+                    import copy
+
+                    align_crp = CommonRasrParameters(base=crp)
+                    adjust_am_config_for_corrected_applicator(align_crp.acoustic_model_config)
+                    align_crp.acoustic_model_config = copy.deepcopy(crp.acoustic_model_config)
+                    align_crp.acoustic_model_config["*"].allow_for_silence_repetitions = False
+                    align_crp.acoustic_model_config["*"].normalize_lemma_sequence_scores = False
+                    align_crp.acoustic_model_config["*"].fix_allophone_context_at_word_boundaries = True
+                    align_crp.acoustic_model_config["*"].transducer_builder_filter_out_invalid_allophones = True
```
michelwi, line 133: can we find and use the full identifier?

am/config.py
```diff
@@ -103,3 +103,19 @@ def acoustic_model_config(
         config.phonology.future_length = phon_future_length
 
     return config
+
+
+def adjust_am_config_for_corrected_applicator(config: rasr.RasrConfig):
+    """
+    Set the correct type of applicator, default is "legacy". Moreover, set exit penalities to zero
+    For a given word sequence the exit penalty is constant with respect to the max/sum
```
michelwi, line 111: In case of optional silence, the silence exit penalty does make a difference, right?

Anyway, I would rather not set the exit TDP to 0.0 here, because then they might be different from what the user is expecting when he specifies other values in the crp. 
If you think they absolutely always should be zero, I would propose an assert statement rather than a silent overwrite.

michelwi:
The AppTek pipeline cannot run since this PR is from a fork and not a branch. (cf. https://github.com/rwth-i6/i6_core/pull/381)
I therefore cancelled that pipeline to save us some of the build hours.

am/config.py
Marvin84: > In case of optional silence, the silence exit penalty does make a difference, right?

This should be defined on the lemma level for both silence and speech with proper normalization. For now we assume that optional silence and speech on lemma level get probability 1.0, so no penalty. The only transition weight is then forward/loop.

> Anyway, I would rather not set the exit TDP to 0.0 here, because then they might be different from what the user is expecting when he specifies other values in the crp.
Here we are only touching a copy of the crp for specific alignment job and not modifying the original crp.

michelwi: > > In case of optional silence, the silence exit penalty does make a difference, right?
>
> This should be defined on the lemma level for both silence and speech with proper normalization. 

I do not disagree with that, however the proper support to define this on lemma level does not exist yet, right?

> For now we assume that optional silence and speech on lemma level get probability 1.0, so no penalty. The only transition weight is then forward/loop.

Then this assumption would limit us in practice as it removes one degree of freedom.

> Here we are only touching a copy of the crp for specific alignment job and not modifying the original crp.

Ok, but same argument for this specific job.

meta/mm_sequence.py
Marvin84: Currently, this is already under a very long identifier, not sure if it is a good idea to make it even longer. I can check if you think this is necessary

Marvin84: I see, thanks for the suggestion and the hint. I will check this out and submit the correction,

meta/mm_sequence.py
```diff
@@ -117,8 +119,28 @@ def update_rqmt(rqmt, extra):
 
         for a_idx, action in enumerate(action_sequence):
             if action.startswith("align"):
+                if use_corrected_applicator:
+                    from i6_core.am.config import adjust_am_config_for_corrected_applicator
```
JackTemaki, line 123: Imports can be moved to the top

JackTemaki (changes requested):
Only a small thing

Marvin84:
@michelwi @christophmluscher could you check this out? I would like to merge this ASAP. I introduced now something that can be used also for full-sum training, and is not coupled with only GMM viterbi iterations.

meta/mm_sequence.py
```diff
@@ -117,8 +120,13 @@ def update_rqmt(rqmt, extra):
 
         for a_idx, action in enumerate(action_sequence):
             if action.startswith("align"):
+                if use_corrected_applicator:
+                    align_crp, align_extra_config = get_align_config_and_crp_for_corrected_applicator(crp)
+                    align_extra_args["extra_config"] = align_extra_config
```
JackTemaki, line 125: make sure any user decision here does not get lost, so that the user config has the priority

JackTemaki (changes requested):

christophmluscher (approved):

am/config.py
```diff
@@ -103,3 +105,31 @@ def acoustic_model_config(
         config.phonology.future_length = phon_future_length
 
     return config
+
+
+def get_align_config_and_crp_for_corrected_applicator(crp: rasr.CommonRasrParameters, exit_penalty: float = 0.0):
```
JackTemaki, line 110: If you want you can also add return value types.

JackTemaki (approved):
I can not say much about it on a principle level, but from the code it seems fine, so this is good to go (see small remark).

am/config.py
Marvin84: done

PR #387, opened 2023-03-31
HDF dump job for RASR alignment caches

Marvin84:
This is a job that allows dumping alignments from RASR caches in HDF, using a user-defined state-tying. It only supports a list of alignment caches and will output corresponding HDF files. There is no support for alignment `bundle`.

### DIFF
--- a/returnn/hdf.py
+++ b/returnn/hdf.py
@@ -1,16 +1,18 @@
 __all__ = ["ReturnnDumpHDFJob", "ReturnnRasrDumpHDFJob", "BlissToPcmHDFJob"]
 
 from dataclasses import dataclass
+import numpy as np
 import os
 import shutil
 import soundfile as sf
 import subprocess as sp
 import tempfile
-from typing import Optional
+from typing import List, Optional
 
 from .rasr_training import ReturnnRasrTrainingJob
 from i6_core.lib import corpus
 from i6_core.lib.hdf import get_returnn_simple_hdf_writer
+from i6_core.lib.rasr_cache import FileArchive
 import i6_core.rasr as rasr
 from i6_core.util import instanciate_delayed, uopen
 from i6_core import util
@@ -293,35 +295,34 @@ class RasrAlignmentDumpHDFJob(Job):
         alignment_caches: List[tk.Path],
         allophone_file: tk.Path,
         state_tying_file: tk.Path,
+        data_type: type = np.uint16,
         returnn_root: Optional[tk.Path] = None,
     ):
         """
-        :param list[Path]: alignment_caches
-        :param Path: allophone_file
-        :param Path: state_tying_file
-        :param Optional[Path] returnn_root: file path to the RETURNN repository root folder
+        :param alignment_caches: e.g. output of an AlignmentJob
+        :param allophone_file
+        :param state_tying_file: e.g. output of a DumpStateTyingJob
+        :param returnn_root: file path to the RETURNN repository root folder
         """
         self.alignment_caches = alignment_caches
         self.allophone_file = allophone_file
         self.state_tying_file = state_tying_file
         self.out_hdf_files = [self.output_path(f"data.hdf.{d}", cached=False) for d in range(len(alignment_caches))]
         self.returnn_root = returnn_root
+        self.data_type = data_type
         self.rqmt = {"cpu": 1, "mem": 8, "time": 0.5}
 
     def tasks(self):
-        yield Task("run", resume="run", rqmt=self.rqmt, args=range(1, (len(self.alignment_caches) + 1)))
+        yield Task("run", rqmt=self.rqmt, args=range(1, (len(self.alignment_caches) + 1)))
 
     def run(self, task_id):
-        from i6_core.lib.rasr_cache import FileArchive
         import numpy as np
 
         state_tying = dict(
             (k, int(v)) for l in open(self.state_tying_file.get_path()) for k, v in [l.strip().split()[0:2]]
         )
 
-        alignment_cache = FileArchive(
-            self.alignment_caches[min(task_id - 1, len(self.alignment_caches) - 1)].get_path()
-        )
+        alignment_cache = FileArchive(self.alignment_caches[task_id - 1].get_path())
         alignment_cache.setAllophones(self.allophone_file.get_path())
 
         returnn_root = None if self.returnn_root is None else self.returnn_root.get_path()
@@ -341,7 +342,7 @@ class RasrAlignmentDumpHDFJob(Job):
             for allophone in alignmentStates:
                 targets.append(state_tying[allophone])
 
-            data = np.array(targets).astype("int32")
+            data = np.array(targets).astype(np.dtype(self.data_type))
             out_hdf.insert_batch(
                 inputs=data.reshape(1, -1, 1),
                 seq_len=[data.shape[0]],
--- a/returnn/hdf.py
+++ b/returnn/hdf.py
@@ -302,6 +302,7 @@ class RasrAlignmentDumpHDFJob(Job):
         :param alignment_caches: e.g. output of an AlignmentJob
         :param allophone_file: e.g. output of a StoreAllophonesJob
         :param state_tying_file: e.g. output of a DumpStateTyingJob
+        :param data_type: type that is used to store the data
         :param returnn_root: file path to the RETURNN repository root folder
         """
         self.alignment_caches = alignment_caches
--- a/returnn/hdf.py
+++ b/returnn/hdf.py
@@ -300,7 +300,7 @@ class RasrAlignmentDumpHDFJob(Job):
     ):
         """
         :param alignment_caches: e.g. output of an AlignmentJob
-        :param allophone_file
+        :param allophone_file: e.g. output of a StoreAllophonesJob
         :param state_tying_file: e.g. output of a DumpStateTyingJob
         :param returnn_root: file path to the RETURNN repository root folder
         """
@@ -316,8 +316,6 @@ class RasrAlignmentDumpHDFJob(Job):
         yield Task("run", rqmt=self.rqmt, args=range(1, (len(self.alignment_caches) + 1)))
 
     def run(self, task_id):
-        import numpy as np
-
         state_tying = dict(
             (k, int(v)) for l in open(self.state_tying_file.get_path()) for k, v in [l.strip().split()[0:2]]
         )
--- a/returnn/hdf.py
+++ b/returnn/hdf.py
@@ -1,4 +1,4 @@
-__all__ = ["ReturnnDumpHDFJob", "ReturnnRasrDumpHDFJob", "BlissToPcmHDFJob"]
+__all__ = ["ReturnnDumpHDFJob", "ReturnnRasrDumpHDFJob", "BlissToPcmHDFJob", "RasrAlignmentDumpHDFJob"]
 
 from dataclasses import dataclass
 import numpy as np
--- a/returnn/hdf.py
+++ b/returnn/hdf.py
@@ -281,3 +281,71 @@ class BlissToPcmHDFJob(Job):
             audio.close()
 
         out_hdf.close()
+
+
+class RasrAlignmentDumpHDFJob(Job):
+    """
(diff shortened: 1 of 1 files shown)


JackTemaki:
@Marvin84 as @michelwi already mentioned in the last PR, please do not create PRs from your own repo, but push the branches to the i6 github to avoid issues with the tests.

returnn/hdf.py
```diff
@@ -281,3 +281,71 @@ def run(self):
...
+            # alignment
+            targets = []
+            alignment = alignment_cache.read(file, "align")
+            alignmentStates = ["%s.%d" % (alignment_cache.allophones[t[1]], t[2]) for t in alignment]
+            for allophone in alignmentStates:
+                targets.append(state_tying[allophone])
+
+            data = np.array(targets).astype("int32")
+            out_hdf.insert_batch(
+                inputs=data.reshape(1, -1, 1),
```
JackTemaki, line 346: Will it cause problems that the format is not sparse? For my speaker tags I used squeeze but this is ugly. Not sure if I did this because there were limitations in the HDFWriter, but if possible we should definitely output it sparse, as our networks expect it to be sparse as far as I know.

JackTemaki (changes requested):

returnn/hdf.py
```diff
@@ -281,3 +281,71 @@ def run(self):
...
+    def run(self, task_id):
+        from i6_core.lib.rasr_cache import FileArchive
+        import numpy as np
+
+        state_tying = dict(
+            (k, int(v)) for l in open(self.state_tying_file.get_path()) for k, v in [l.strip().split()[0:2]]
+        )
+
+        alignment_cache = FileArchive(
+            self.alignment_caches[min(task_id - 1, len(self.alignment_caches) - 1)].get_path()
```
michelwi, line 323: why the minimum here? I think `task_id - 1` should always be correct. And even if not, then we dump the same file twice to different outputs which is probably not expected, so maybe rather fail with a KeyError here.

returnn/hdf.py
```diff
@@ -281,3 +281,71 @@ def run(self):
...
+            seq_name = info.name
+
+            # alignment
+            targets = []
+            alignment = alignment_cache.read(file, "align")
+            alignmentStates = ["%s.%d" % (alignment_cache.allophones[t[1]], t[2]) for t in alignment]
+            for allophone in alignmentStates:
+                targets.append(state_tying[allophone])
+
+            data = np.array(targets).astype("int32")
```
michelwi, line 344: I know returnn expects int32 by default, but you can specify that the dtype is int16 as well. And saving the data as int16 will save 50% of the storage space (and will most likely be enough for the number of outputs that we usually use).

There can be cases (compatibility to old setups, or dense triphone cart) where we still need int32, so maybe we make it configurable.

I propose int16 as default purely based on consistency to https://github.com/rwth-i6/i6_core/blob/07b40104764da1ae2b3209b2fd69964e4497e977/returnn/hdf.py#L214

returnn/hdf.py
```diff
@@ -281,3 +281,71 @@ def run(self):
...
+
+
+class RasrAlignmentDumpHDFJob(Job):
+    """
+    This Job reads Rasr alignment caches and dump them in hdf files.
+    """
+
+    def __init__(
+        self,
+        alignment_caches: List[tk.Path],
```
michelwi, line 293: missing
```
from typing import List
```
breaks the tests.

michelwi (changes requested):

returnn/hdf.py
JackTemaki: but when do you use HDFs in RASR? Do we have that in a setup?

JackTemaki: And @michelwi, should it then here not be better uint16?

michelwi: > The problem here is not returnn but rasr. for feature scorers that rely on tensorflow node one prepares the data as int32. I personally do not want to need to redefine this detail.

Exactly this is an argument to make it configurable. 
We can discuss what should be the default type. I tend to agree with @JackTemaki on uint16.

Thinking about it: If you talk about "tensorflow node", you mean the flow node? But in rasr flow you do have generic type convert nodes, so this could still be used. Concerning the trade-off "having N more lines in your config" vs "double storage usage for the data" I would always decide in favor of storage.

Marvin84: This is the case for the factored hybrid that needs to feed the context. You prepare the data in the feature scorer in rasr and you need to define the type. 
There is also a more important problem. I for example use dense tying and this does not fit in 16 bits (~(42**3)*6). This is also the reason why I still cannot integrate my triphone model in rasr search.
Please consider also this edge cases (factored hybrid) when you decide for the default. This should be more than 16 bits.

Marvin84: > but when do you use HDFs in RASR? Do we have that in a setup?

See above response. If you need the label as context that is fed.

michelwi: > I for example use dense tying and this does not fit in 16 bits (~(42**3)*6). 

yes, definitively should be configurable then!

returnn/hdf.py
Marvin84: Oh yes, this was from another logic, where it was possible to have less alignment caches than hdf files. But here we do not need this.

returnn/hdf.py
Marvin84: I did not still test this actually for training. Maybe it is worth first test it before deciding

returnn/hdf.py
```diff
@@ -281,3 +281,71 @@ def run(self):
...
+        """
+        self.alignment_caches = alignment_caches
+        self.allophone_file = allophone_file
+        self.state_tying_file = state_tying_file
+        self.out_hdf_files = [self.output_path(f"data.hdf.{d}", cached=False) for d in range(len(alignment_caches))]
+        self.returnn_root = returnn_root
+        self.rqmt = {"cpu": 1, "mem": 8, "time": 0.5}
+
+    def tasks(self):
+        yield Task("run", resume="run", rqmt=self.rqmt, args=range(1, (len(self.alignment_caches) + 1)))
```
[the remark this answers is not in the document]
Marvin84: this is the run argument, and should run from task 1 to n, but the list index is task-1

returnn/hdf.py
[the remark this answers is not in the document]
JackTemaki: This is a one time job where the cache manager would not really give a benefit. Also we never properly implemented this for "within python" jobs, as `file_caching` only returns the modified string. So before we have a universal solution for this I would not add that here.

But still thanks for considering it!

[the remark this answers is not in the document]
NeoLegends: Hm, I was thinking about the individual read accesses the FileArchive class will do while reading the file, and if cache manager access would be easier on the FS than that.

But I guess then no need.

returnn/hdf.py
```diff
@@ -293,35 +295,34 @@ def __init__(
...
     ):
         """
-        :param list[Path]: alignment_caches
-        :param Path: allophone_file
-        :param Path: state_tying_file
-        :param Optional[Path] returnn_root: file path to the RETURNN repository root folder
+        :param alignment_caches: e.g. output of an AlignmentJob
+        :param allophone_file
+        :param state_tying_file: e.g. output of a DumpStateTyingJob
+        :param returnn_root: file path to the RETURNN repository root folder
```
michelwi, line 305:
```suggestion
        :param allophone_file: e.g. output of a StoreAllophonesJob
        :param state_tying_file: e.g. output of a DumpStateTyingJob
        :param data_type: type that is used to store the data
        :param returnn_root: file path to the RETURNN repository root folder
```
added new data_type to docstring.

michelwi (commented):
Thanks. Only minor nitpicks from my side remaining.

returnn/hdf.py
```diff
@@ -281,3 +283,68 @@ def run(self):
...
+        alignment_caches: List[tk.Path],
+        allophone_file: tk.Path,
+        state_tying_file: tk.Path,
+        data_type: type = np.uint16,
+        returnn_root: Optional[tk.Path] = None,
+    ):
+        """
+        :param alignment_caches: e.g. output of an AlignmentJob
+        :param allophone_file: e.g. output of a StoreAllophonesJob
+        :param state_tying_file: e.g. output of a DumpStateTyingJob
```
michelwi, line 304:
```suggestion
        :param state_tying_file: e.g. output of a DumpStateTyingJob
        :param data_type: type that is used to store the data
```
The added `:param data_type:` from my last comment was missing in the last commit, so here again :P

michelwi (approved):

returnn/hdf.py
```diff
@@ -281,3 +283,69 @@ def run(self):
...
+        self.state_tying_file = state_tying_file
+        self.out_hdf_files = [self.output_path(f"data.hdf.{d}", cached=False) for d in range(len(alignment_caches))]
+        self.returnn_root = returnn_root
+        self.data_type = data_type
+        self.rqmt = {"cpu": 1, "mem": 8, "time": 0.5}
+
+    def tasks(self):
+        yield Task("run", rqmt=self.rqmt, args=range(1, (len(self.alignment_caches) + 1)))
+
+    def run(self, task_id):
```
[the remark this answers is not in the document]
NeoLegends: The SimpleHDFWriter actually already writes through a temporary file, at least as per its docs.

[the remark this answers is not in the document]
Marvin84: My understanding was that this is handled by the HDFWriter?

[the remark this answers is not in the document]
christophmluscher: ok perfect :)

returnn/hdf.py
```diff
@@ -281,3 +283,69 @@ def run(self):
...
+        state_tying = dict(
+            (k, int(v)) for l in open(self.state_tying_file.get_path()) for k, v in [l.strip().split()[0:2]]
+        )
+
+        alignment_cache = FileArchive(self.alignment_caches[task_id - 1].get_path())
```
vieting, line 324: This assumes that the keys are 0,...,N. In our pipeline, they often start from 1, so it would be nice if that'd be possible here. In any case, it should be detected at graph creation time and not only fail when the job runs.

michelwi: self.alignment is a List[Path], those always start at 0.

vieting: Oh, a list, right. I directly passed the `hidden_paths` of a `FlagDependentFlowAttribute`. But then I'll change it on my end.

vieting:
It might be helpful if we could pass a segment list, e.g. when creating only a subset of a dataset as cv for training.

JackTemaki:
> It might be helpful if we could pass a segment list, e.g. when creating only a subset of a dataset as cv for training.

But is this not something you do afterwards? If it is for training you can load the same HDF twice with different segment lists.

Marvin84:
> > It might be helpful if we could pass a segment list, e.g. when creating only a subset of a dataset as cv for training.
> 
> But is this not something you do afterwards? If it is for training you can load the same HDF twice with different segment lists.

Is this supported on RETURNN side for training with HDF files? Given that we are not using RASR anymore for data loading

JackTemaki:
> > If it is for training you can load the same HDF twice with different segment lists.
> 
> I am wondering if e.g. the whole of LBS was in one big HDF file and you single out 100 sequences for dev, will it significantly slow down because it has to search the specific sequences first?

I think I did this at some point and there was no slowdown, not sure thought.

michelwi:
> > > It might be helpful if we could pass a segment list, e.g. when creating only a subset of a dataset as cv for training.
> > 
> > 
> > But is this not something you do afterwards? If it is for training you can load the same HDF twice with different segment lists.
> 
> Is this supported on RETURNN side for training with HDF files? Given that we are not using RASR anymore for data loading

it is via the parameter `seq_list_filter_file` inherited from the basic [`Dataset` class](https://github.com/rwth-i6/returnn/blob/9db3f9fd7ce26a0cd80138d604136c6b76c895d8/returnn/datasets/basic.py#L122)

JackTemaki (approved):
I think this is good for now, if we need to extend some features we can do this later on!

PR #388, opened 2023-03-31
Bug in BlissToPcmHDFJob

christophmluscher:
`self` missing in `isinstance`.


michelwi (approved):
We refactored that last minute and it seems to have been overlooked. Thanks for fixing!

JackTemaki (approved):

PR #389, opened 2023-04-03
fix empty set litteral

michelwi:
It should be empty set rather than empty dict.


Atticus1806 (approved):

Atticus1806:
Ah makes sense, overlooked that PR I guess.

christophmluscher (approved):

PR #390, opened 2023-04-04
Filter job keep synt tok - Bug fix.

sarahberanek:
Apparently, nobody used the option after implementing :nerd_face:  

if there is no synt tok in a lemma, then - when setting the `check_synt_tok` option - it will always be true, because
`all(empty list) = True`. This fixes it.


Atticus1806:
If I understood correctly the AppTek test fails because the PR is from a private branch and not from a repo branch right?

sarahberanek:
> If I understood correctly the AppTek test fails because the PR is from a private branch and not from a repo branch right?

Is this new? This didn't happen last time. @michelwi will know 😎

michelwi (approved):
> If I understood correctly the AppTek test fails because the PR is from a private branch and not from a repo branch right?

That is correct. Doing PRs from a branch rather than a fork is the preferred method.

michelwi:
> I didnt have permission back in the day, thats why I used this way and now sticked with it. Do you want me to re-open the PR from a branch from i6-core?

I don't think it is necessary to redo this PR.
But please check if you are allowed to create branches now and then use this for future PRs, thanks.

JackTemaki:
> > I didnt have permission back in the day, thats why I used this way and now sticked with it. Do you want me to re-open the PR from a branch from i6-core?
> 
> I don't think it is necessary to redo this PR. But please check if you are allowed to create branches now and then use this for future PRs, thanks.

@sarahberanek I added access for you!

JackTemaki (approved):

sarahberanek:
> > I didnt have permission back in the day, thats why I used this way and now sticked with it. Do you want me to re-open the PR from a branch from i6-core?
> 
> I don't think it is necessary to redo this PR. But please check if you are allowed to create branches now and then use this for future PRs, thanks.

So can we even merge now with this check failed?
Thanks, Nick for the access!

PR #391, opened 2023-04-06
Adjustment to the RasrAlignmentDumpHDFJob 

Marvin84:
If an alignment is empty now we skip that segment and output the name of that segment in an output file.

### DIFF
--- a/returnn/hdf.py
+++ b/returnn/hdf.py
@@ -1,6 +1,7 @@
 __all__ = ["ReturnnDumpHDFJob", "ReturnnRasrDumpHDFJob", "BlissToPcmHDFJob", "RasrAlignmentDumpHDFJob"]
 
 from dataclasses import dataclass
+import glob
 import numpy as np
 import os
 import shutil
@@ -323,7 +324,8 @@ class RasrAlignmentDumpHDFJob(Job):
 
     def merge(self):
         excluded_segments = []
-        for p in self.excluded_segment_sublists:
+        excluded_files = glob.glob("excluded_segments.*")
+        for p in excluded_files:
             if os.path.isfile(p):
                 with open(p, "r") as f:
                     segments = f.read().splitlines()
@@ -371,4 +373,4 @@ class RasrAlignmentDumpHDFJob(Job):
         out_hdf.close()
 
         if len(excluded_segments):
-            write_paths_to_file(self.excluded_segment_sublists[task_id - 1], excluded_segments)
+            write_paths_to_file(f"excluded_segments.{task_id}", excluded_segments)
--- a/returnn/hdf.py
+++ b/returnn/hdf.py
@@ -309,7 +309,7 @@ class RasrAlignmentDumpHDFJob(Job):
         self.alignment_caches = alignment_caches
         self.allophone_file = allophone_file
         self.state_tying_file = state_tying_file
-        self.out_hdf_files = [self.output_path(f"data.hdf.{d}", cached=False) for d in range(len(alignment_caches))]
+        self.out_hdf_files = [self.output_path(f"data.hdf.{d}") for d in range(len(alignment_caches))]
         self.out_excluded_segments = self.output_path(f"excluded.segments", cached=False)
         self.returnn_root = returnn_root
         self.data_type = data_type
--- a/returnn/hdf.py
+++ b/returnn/hdf.py
@@ -310,7 +310,7 @@ class RasrAlignmentDumpHDFJob(Job):
         self.allophone_file = allophone_file
         self.state_tying_file = state_tying_file
         self.out_hdf_files = [self.output_path(f"data.hdf.{d}") for d in range(len(alignment_caches))]
-        self.out_excluded_segments = self.output_path(f"excluded.segments", cached=False)
+        self.out_excluded_segments = self.output_path(f"excluded.segments")
         self.returnn_root = returnn_root
         self.data_type = data_type
         self.rqmt = {"cpu": 1, "mem": 8, "time": 0.5}
--- a/returnn/hdf.py
+++ b/returnn/hdf.py
@@ -309,12 +309,20 @@ class RasrAlignmentDumpHDFJob(Job):
         self.allophone_file = allophone_file
         self.state_tying_file = state_tying_file
         self.out_hdf_files = [self.output_path(f"data.hdf.{d}", cached=False) for d in range(len(alignment_caches))]
+        self.out_excluded_segments = [
+            self.output_path(f"excluded_segments.cache.{d}", cached=False) for d in range(len(alignment_caches))
+        ]
         self.returnn_root = returnn_root
         self.data_type = data_type
         self.rqmt = {"cpu": 1, "mem": 8, "time": 0.5}
 
     def tasks(self):
         yield Task("run", rqmt=self.rqmt, args=range(1, (len(self.alignment_caches) + 1)))
+
+    def _write_text(self, path, segment_list):
+        with open(path, "w") as f:
+            for item in segment_list:
+                f.write("%s\n" % item)
 
     def run(self, task_id):
         state_tying = dict(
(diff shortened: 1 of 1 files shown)


returnn/hdf.py
```diff
@@ -349,3 +362,6 @@ def run(self, task_id):
             )
 
         out_hdf.close()
+
+        if len(excluded_segments):
+            self._write_text(self.out_excluded_segments[task_id - 1], excluded_segments)
```
michelwi, line 367: If you define a file as output, I would say it should always exist. If no segments are excluded, then maybe create an empty file?

Other suggestion: collect the segments of all subtasks and then only create one file with missing segments. (you would need another task for this, I guess)

Marvin84: > collect the segments of all subtasks and then only create one file with missing segments.

I do not want to create too many accesses to the same file.

vieting: I would also prefer not having files in `self.out_excluded_segments` that are not created. If subsequent jobs would use this, they might easily fail. Creating one joint file with all segments seems reasonable to me, it doesn't add much overhead and subsequent jobs would most likely use a joint segment list anyway.

vieting: Just saying they might fail if they assume that all files in `self.out_excluded_segments` exist.

I would vote for a single file including all missing segments, although no strong feelings are involved on my side either :smile:

Atticus1806: > you should just check whether the list is empty or not

You would actually have to check whether to file exists or not right? I am not sure, but is it possible that there might be issues with sisyphus warning regarding missing files? e.g. only 3 and 10 exists, but you pass `out_excluded_segments` to a job (which actually correctly handles the list within the job), still I think there might be a warning. 
I would vote for a single file, which is empty if no segments are excluded.

Marvin84: Ok, I will create one single file that is created also when there is no empty segment.

returnn/hdf.py
```diff
@@ -320,7 +318,8 @@ def tasks(self):
         yield Task("run", rqmt=self.rqmt, args=range(1, (len(self.alignment_caches) + 1)))
 
     def _write_text(self, path, segment_list):
-        with open(path, "w") as f:
+        write_flag = "a" if os.path.isfile(path) else "w"
```
michelwi, line 321: I am pretty certain there will be issues with race conditions when parallel jobs append to the same file at the same time.

michelwi: Sorry, when you criticize you should always give constructive critique and show ways for improvements:

One way around the race conditions would be to still write into single files for each subtask (but those are not outputs and can be omitted if no files are included), and then add one additional task that collects all individual files into the output file which always exists, even if empty.

michelwi: >  I do not know how to solve this if not by creating another job that merges.

Task instead of Job, but yes, this is the only way I see as well.

Marvin84: > Sorry, when you criticize you should always give constructive critique and show ways for improvements

Thanks for the moral lecture. But mine was not a critique, but a lack of knowledge of how I should do that.

> I think that is also how its done in other Jobs, they write subfiles in the work folder and then have a merge task

@Atticus1806 Thanks for the suggestion, will do that.

returnn/hdf.py
```diff
@@ -309,13 +309,20 @@ def __init__(
...
         self.out_hdf_files = [self.output_path(f"data.hdf.{d}", cached=False) for d in range(len(alignment_caches))]
+        self.out_excluded_segments = self.output_path(f"excluded.segments", cached=False)
         self.returnn_root = returnn_root
         self.data_type = data_type
         self.rqmt = {"cpu": 1, "mem": 8, "time": 0.5}
 
     def tasks(self):
         yield Task("run", rqmt=self.rqmt, args=range(1, (len(self.alignment_caches) + 1)))
 
+    def _write_text(self, path, segment_list):
```
christophmluscher, line 320: `staticmethod`?

`i6_core.util.write_paths_to_file` has a very similar logic. maybe you can reuse that

Marvin84: > `i6_core.util.write_paths_to_file` has a very similar logic. maybe you can reuse that

I cannot use this function, because the writing flag is hardcoded inthere to be `w`, which means that it will overwrite the file, meaning you will find the output of the last sub_task that wrote to the file.

Marvin84: However, now with the `merge` task that should not be a problem anymore.

returnn/hdf.py
```diff
@@ -349,3 +361,5 @@ def run(self, task_id):
             )
 
         out_hdf.close()
+
+        self.excluded_segment_list.extend(excluded_segments)
```
michelwi, line 365: Mh.. do you have a way to test this? I fear that since both tasks run on different machines in different python processes, the `self.excluded_segment_list` will be reset for the "merge" subtask.

JackTemaki: Yes this will unfortunately not work, the job.save is not re-written, it has to be written to a file in the work dir, so `write_paths_to_file(segments, f"excluded_segments_{task_id}")` or so

Marvin84: I ran the job and was successful. But now I am not as sure as before, since your argument might also be right.

Marvin84: > If it was somehow successful

@JackTemaki @michelwi I was wrong, that was the old `hash`, and the new code did not work out.

returnn/hdf.py
```diff
@@ -309,12 +309,27 @@ def __init__(
         self.allophone_file = allophone_file
         self.state_tying_file = state_tying_file
         self.out_hdf_files = [self.output_path(f"data.hdf.{d}", cached=False) for d in range(len(alignment_caches))]
+        self.excluded_segment_sublists = [
```
Atticus1806, line 312: I understand that we need to write this in files before merge due to what @michelwi said, but do we want these files to be outputs? I would rather suggest writing them into the work folder.

vieting (approved):

returnn/hdf.py
```diff
@@ -309,12 +310,25 @@ def __init__(
         self.allophone_file = allophone_file
         self.state_tying_file = state_tying_file
         self.out_hdf_files = [self.output_path(f"data.hdf.{d}", cached=False) for d in range(len(alignment_caches))]
+        self.out_excluded_segments = self.output_path(f"excluded.segments", cached=False)
```
Atticus1806, line 313: PR looks good. Wondering `cached` needs to be set, since `False` the default I think and we can assume this does not change? 
But this does not hinder approval.

Atticus1806 (approved):

returnn/hdf.py
```diff
@@ -309,12 +310,25 @@ def __init__(
         self.allophone_file = allophone_file
         self.state_tying_file = state_tying_file
         self.out_hdf_files = [self.output_path(f"data.hdf.{d}", cached=False) for d in range(len(alignment_caches))]
```
JackTemaki, line 312:
```suggestion
        self.out_hdf_files = [self.output_path(f"data.hdf.{d}") for d in range(len(alignment_caches))]
```

Atticus1806: I think you missed the line below I was referring to :D

returnn/hdf.py
```diff
@@ -308,13 +309,26 @@ def __init__(
         self.alignment_caches = alignment_caches
         self.allophone_file = allophone_file
         self.state_tying_file = state_tying_file
-        self.out_hdf_files = [self.output_path(f"data.hdf.{d}", cached=False) for d in range(len(alignment_caches))]
+        self.out_hdf_files = [self.output_path(f"data.hdf.{d}") for d in range(len(alignment_caches))]
+        self.out_excluded_segments = self.output_path(f"excluded.segments", cached=False)
```
Atticus1806, line 313:
```suggestion
        self.out_excluded_segments = self.output_path(f"excluded.segments")
```

PR #393, opened 2023-04-12
BlissToOggZipJob: delete at end of merge task

vieting:
I ran into an issue where the merge task crashed (`rsync` which is used in the task was not installed on the node, but does not really matter here). Unfortunately, the single zip files in the work folder are deleted right at the beginning of the merge task so I had to recreate all of them instead of just rerunning the merge task. I therefore propose to delete them only at the end of the merge task.


Atticus1806:
So this is not an error in the job per se, but quality of life in case merging crashes?

Atticus1806 (approved):

vieting:
> So this is not an error in the job per se, but quality of life in case merging crashes?

Exactly!

JackTemaki (approved):

PR #394, opened 2023-04-14
Generic NextGenHDF writer and a jobf for rasr caches dump

Marvin84:
The `Returnn's` `NextGenHDFDataset` allows for the storage of multiple datasets in one HDF file. Given also some ongoing dicussions in [here](https://github.com/rwth-i6/returnn/issues/331), I am proposing a generic `NextGenHDFWriter`, that writes several datasets (features or labels).
Moreover, there is also one `Sisyphus` job example that uses this specific writer, as an example.

Differently to the `SimpleHDFWriter`, here one can write data in a sparse form and therefore there is no additional squeeze layer needed during training, as @vieting expererienced, recently. Here, the keys of the `label_info_dict` can be used directly in the softmax layer as `target='my_key'`.

Current usecase and necessity of having this in `i6_core`:
This is currently needed for the factored hybrid system for the estimation of context-dependent priors. Moreover, the use of several additional targets in form of multi-task learning is also very common and can be handled easily.

### DIFF
--- a/lib/hdf.py
+++ b/lib/hdf.py
@@ -1,7 +1,8 @@
 import h5py
 import numpy as np
-from typing import Dict, Optional
-import sys
+import logging, sys, shutil, tempfile
+
+from typing import Dict, List, Optional
 
 
 def get_input_dict_from_returnn_hdf(hdf_file: h5py.File) -> Dict[str, np.ndarray]:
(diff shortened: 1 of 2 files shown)


michelwi:
Let me briefly add some comments without having read the code:

Firstly concerning
> NextGenHDFDataset allows for the storage of multiple datasets in one HDF file.
>  for the factored hybrid system for the estimation of context-dependent priors. 

Yes I agree that this is a very neat solution for this special case as your FH multitask approach would always go together hand in hand and never be used separately. However for general multi-task learning as mentioned

> additional targets in form of multi-task learning is also very common and can be handled easily.

I would rather suggest to have it in separate datasets which are combined on-demand with one of the meta-datasets in returnn. This way we do not need to redo the dumping when you add an additional task, or load unnecessary ballast when you remove one.


Secondly concerning:
> Differently to the SimpleHDFWriter, here one can write data in a sparse form and therefore there is no additional squeeze layer needed during training

Are you sure it is not possible? I think if one would put 2D data into the HDF file (i.e. [T,B]) then this could be defined as sparse and directly used in returnn. Let me try it:
```
import numpy as np
from returnn.datasets.hdf import SimpleHDFWriter

w = SimpleHDFWriter("./test.hdf", dim=60,  ndim=1)
data = np.random.randint(min=0, max=59, size=(20, 5))
w.insert_batch(data, [20,20,20,20,20], ["1", "2", "3", "4", "5"])
w.close()
```
and 
```
$ python3 tools/dump-dataset.py "{'class': 'HDFDataset', 'files': ['./test.hdf']}"
Dataset:
  input: 60 x 1
  output: {'data': [60, 1]}
  HDF dataset, sequences: 5, frames: 100
Epoch: 1
Dataset keys: ['data']
Dataset target keys: []
Dump to stdout
[...]
```
This seems to be a valid, sparse dataset. (@vieting )


But I agree that a generic NextGenHDF writer would be a useful addition to our codebase. One additional question to your implementation: Do you think it would make sense to have a generic `DumpNextGenHDFJob` from which the `RasrDumpNextGenHDFJob` inherits to set the exact layout of the alignment based Hybrid system training?

JackTemaki (changes requested):

lib/hdf.py
```diff
@@ -35,3 +36,95 @@ def get_returnn_simple_hdf_writer(returnn_root: Optional[str]):
...
+        feature_group.attrs["parser"] = self.feature_parser_name
+
+        return feature_group.create_group("data")
+
+    def add_sequence_name(self, seq_name):
+        self.sequence_names.append(seq_name)
+
+    def add_data_to_group(self, group_name, seq_name, data):
+        if group_name in self.label_info_dict:
+            data = np.array(data).astype(self.label_data_type)
```
JackTemaki, line 114: This should be left to the user to be defined from the outside. Then the `label_data_type` can also be removed.

lib/hdf.py
```diff
@@ -35,3 +36,95 @@ def get_returnn_simple_hdf_writer(returnn_root: Optional[str]):
...
+        self.temp_file = tempfile.NamedTemporaryFile(suffix="_NextGenHDFWriter_outHDF")
+        self.temp_path = self.temp_file.name
+        self.out_hdf = h5py.File(self.temp_path, "w")
+
+        logging.info(f"processing temporary file { self.temp_path}")
+
+        # root
+        self.root_group = self.out_hdf.create_group("streams")
+
+        for label_name, label_dim in self.label_info_dict.items():
```
JackTemaki, line 84: Following the init restructuring from above, I think it would be better to just loop here over all streams and have a separate `create_group` function that you can override as user to add your own custom streams.

JackTemaki: Sill, I do not know how to cleanly deal with the "label_dim" argument in that case...

PR #396, opened 2023-04-14
Extend type hint

christophmluscher:
@michelwi I think this should be correct?! You agree???

### DIFF
--- a/meta/system.py
+++ b/meta/system.py
@@ -71,7 +71,7 @@ class System:
         self.scorer_args = {}
         self.scorer_hyp_arg = {}
 
-        self.jobs = {"base": {}}  # type: Dict[str, Dict[str, Union[Job, AlignSplitAccumulateSequence]]]
+        self.jobs: Dict[str, Dict[str, Union[Job, AlignSplitAccumulateSequence]]] = {"base": {}}
 
     def set_corpus(
         self, name: str, corpus: CorpusObject, concurrent: int, segment_path: Optional[util.MultiOutputPath] = None
--- a/meta/system.py
+++ b/meta/system.py
@@ -24,6 +24,9 @@ from .mm_sequence import AlignSplitAccumulateSequence
 Path = setup_path(__package__)
 
 
+selector_type = Union[str, List, Tuple[str], Tuple[str, str], Tuple[str, str, int]]
+
+
 class CorpusObject(tk.Object):
     """
     A simple container object to track additional information for a bliss corpus
@@ -49,7 +52,7 @@ class System:
 
         # collections which are a nested dict of corpus_key -> name -> some object / some container
         # the container can be list, tuple or dict
-        self.alignments = {}  # type: Dict[str,Dict[str,Union[List[Any], Any]]]
+        self.alignments = {}  # type: Dict[str, Dict[str, Union[List[Any], Any]]]
         # corpus_key -> alignment_name -> element or list of e.g. FlowAttributes
         # with cache_mode containing caches and bundle
         self.ctm_files = {}
@@ -571,8 +574,8 @@ class System:
         self,
         name: str,
         corpus: str,
-        flow: Union[str, List[str], Tuple[str], rasr.FlowNetwork],
-        feature_scorer: Union[str, List[str], Tuple[str], rasr.FeatureScorer],
+        flow: Union[selector_type, rasr.FlowNetwork],
+        feature_scorer: Union[selector_type, rasr.FeatureScorer],
         **kwargs,
     ):
         """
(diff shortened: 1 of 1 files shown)


Atticus1806:
Maybe this can also be moved into the function head. I can't suggest it as change sadly I think, since these lines were not changed by the PR.

michelwi:
> @michelwi I think this should be correct?! You agree???

I agree that the added `rasr.FlowNetwork` is correct here, but I fail to see how the already existing `rasr.FlagDependentFlowAttribute` would have worked without being wrapped in a generic_cache flow or so...

Also there are many more places where this update is applicable in this file.

Atticus1806 (approved):
Thanks a lot for the large typing changes!

meta/system.py
```diff
@@ -2,24 +2,37 @@
...
+Path = setup_path(__package__)
+
+
+class CorpusObject(tk.Object):
+    """
+    A simple container object to track additional information for a bliss corpus
+    """
+
+    def __init__(self):
+        self.corpus_file = None  # type: Optional[tk.Path] # bliss corpus xml
```
JackTemaki, line 31:
```suggestion
        self.corpus_file: Optional[tk.Path] = None  # bliss corpus xml
```

meta/system.py
```diff
@@ -2,24 +2,37 @@
...
+        self.audio_dir = None  # type: Optional[tk.Path] # audio directory if paths are relative (usually not needed)
```
JackTemaki, line 32:
```suggestion
        self.audio_dir: Optional[tk.Path] = None  # audio directory if paths are relative (usually not needed)
```

meta/system.py
```diff
@@ -2,24 +2,37 @@
...
+        self.audio_format = None  # type: Optional[str] # format type of the audio files, see e.g. get_input_node_type()
```
JackTemaki, line 33:
```suggestion
        self.audio_format: Optional[str] = None  # format type of the audio files, see e.g. get_input_node_type()
```

meta/system.py
```diff
@@ -2,24 +2,37 @@
...
+        self.duration = None  # type: Optional[float] # duration of the corpus, is used to determine job time
```
JackTemaki, line 34:
```suggestion
        self.duration: Optional[float] = None  # duration of the corpus, is used to determine job time
```

meta/system.py
```diff
@@ -57,16 +71,18 @@ def __init__(self):
         self.scorer_args = {}
         self.scorer_hyp_arg = {}
 
-        self.jobs = {"base": {}}  # type: Dict[str,Dict[str,Job]]
+        self.jobs = {"base": {}}  # type: Dict[str, Dict[str, Union[Job, AlignSplitAccumulateSequence]]]
```
christophmluscher, line 74:
```suggestion
        self.jobs: Dict[str, Dict[str, Union[Job, AlignSplitAccumulateSequence]]] = {"base": {}}
```

meta/system.py
```diff
@@ -116,12 +132,13 @@ def add_overlay(self, origin, name):
 
         self.jobs[name] = {}
 
-    def copy_from_system(self, origin_system, origin_name, target_name=None):
+    def copy_from_system(self, origin_system: "System", origin_name: str, target_name: Optional[str] = None):
```
JackTemaki, line 135: Can you try `from __future__ import annotations` to solve the System not being defined problem?

meta/system.py
```diff
@@ -535,12 +565,19 @@ def linear_alignment(self, name, corpus, flow, prefix="", **kwargs):
                 },
             )
 
-    def align(self, name, corpus, flow, feature_scorer, **kwargs):
+    def align(
+        self,
+        name: str,
+        corpus: str,
+        flow: Union[str, List[str], Tuple[str], rasr.FlowNetwork],
+        feature_scorer: Union[str, List[str], Tuple[str], rasr.FeatureScorer],
```
michelwi, line 573: actually `Tuple[str]` means a Tuple containing a single string. But additionally allowed would be 
`Tuple[str, str]` and `Tuple[str, str, int]`

`List[str]` is also only so-so correct, as only Lists with the elements equivalent to the tuples above would be possible.. Well technically also a list with (str, str, int, Any, Any, ...) would work, as all elements behind list[2] are ignored.. But I don't know a good way to type this..

However in most setups the Tuple version is used in practice I think, so whenever we decide to use typechecking I think the tuples should be typed correctly.

Maybe we should define a `selector_type` to make the union shorter.. i.e.
```
selector_type = Union[str, List, Tuple[str], Tuple[str, str], Tuple[str, str, int]]

[...]

flow : Union[selector_type, rasr.FlowNetwork],
feature_scorer: : Union[selector_type, rasr.FeatureScorer],

michelwi (approved):

JackTemaki (approved):

PR #398, opened 2023-04-14
Catch broken cache

christophmluscher:
Does not only delete the cache if it is empty but also when the task failed during the dumping of the cache.

### DIFF
--- a/features/extraction.py
+++ b/features/extraction.py
@@ -160,7 +160,7 @@ class FeatureExtractionJob(rasr.RasrCommand, Job):
     def cleanup_before_run(self, cmd, retry, task_id, *args):
         util.backup_if_exists("feature-extraction.log.%d" % task_id)
         for name in self.out_feature_bundle:
-            util.delete_if_zero("%s.cache.%d" % (name, task_id))
+            util.delete_if_exists("%s.cache.%d" % (name, task_id))
 
     @classmethod
     def create_config(cls, crp, feature_flow, extra_config, extra_post_config, **kwargs):


JackTemaki:
But now you always delete, right?

christophmluscher:
> But now you always delete, right?

Yes. Before the run it cleans up. My understanding is that if it runs again, than anything for that task id should be deleted.

JackTemaki:
Um, are you sure this function is ever called from somewhere?

christophmluscher:
In the `run_script` function of the `RasrCommand` class, `run_cmd` is called, which in turn calls `cleanup_before_run`. so I think it should be called

Atticus1806 (approved):

michelwi (approved):
I do not disagree with this change in principle, but let me add a comment to clarify:

RASR does support picking up interrupted dumping of feature caches. If an existing cache is detected, all segments that already are in the cache will be skipped and only the new ones will be added in the end.
So with this change you add additional work as also the already existing segments need to be re-dumped.

This having said, I oftentimes have found that if the dumping failed for some reason, the crash most often introduced one corrupted segment in the cache that RASR does not recover (or care about), which comes back to bite me some time later in the pipeline. For this reason I have been deleting "interrupted" caches anyway to start from a clean slate.

christophmluscher:
I probably only noticed this when the cache was actually broken.

> RASR does support picking up interrupted dumping of feature caches. If an existing cache is detected, all segments that already are in the cache will be skipped and only the new ones will be added in the end.
So with this change you add additional work as also the already existing segments need to be re-dumped.

I thought this was more for combining features caches in several steps than a fail safe?!

PR #400, opened 2023-04-25
Add PtCheckpoint class, add to ReturnnTrainingJob

JackTemaki:
As discussed, the extra checkpoint class for PyTorch experiments.

(Added all participants as reviewer for visibility)

### DIFF
--- a/returnn/training.py
+++ b/returnn/training.py
@@ -72,6 +72,27 @@ class Checkpoint:
         return "'%s'" % self.ckpt_path
 
 
+class PtCheckpoint:
+    """
+    Checkpoint object pointing to a .pt file
+    """
+
+    def __init__(self, path: tk.Path):
+        """
+        :param path: .pt file
+        """
+        self.path = path
+
+    def _sis_hash(self):
+        return self.path._sis_hash()
+
+    def __str__(self):
+        return self.path.get()
+
+    def __repr__(self):
+        return "'%s'" % self.path
+
+
 class ReturnnTrainingJob(Job):
     """
     Train a RETURNN model using the rnn.py entry point.
@@ -159,13 +180,21 @@ class ReturnnTrainingJob(Job):
             for k in stored_epochs
             if k in self.keep_epochs
         }
-        if self.returnn_config.get("use_tensorflow", False):
+        if self.returnn_config.get("use_tensorflow", False) or self.returnn_config.get("backend") == "tensorflow":
             self.out_checkpoints = {
                 k: Checkpoint(index_path)
                 for k in stored_epochs
                 if k in self.keep_epochs
                 for index_path in [self.output_path("models/epoch.%.3d.index" % k)]
             }
+        if self.returnn_config.get("backend", None) == "torch":
+            self.out_checkpoints = {
+                k: PtCheckpoint(index_path)
+                for k in stored_epochs
+                if k in self.keep_epochs
+                for index_path in [self.output_path("models/epoch.%.3d.pt" % k)]
+            }
+
         self.out_plot_se = self.output_path("score_and_error.png")
         self.out_plot_lr = self.output_path("learning_rate.png")
 


returnn/training.py
```diff
@@ -159,13 +180,21 @@ def __init__(
...
                 for k in stored_epochs
                 if k in self.keep_epochs
                 for index_path in [self.output_path("models/epoch.%.3d.index" % k)]
             }
+        if self.returnn_config.get("backend", None) == "torch":
+            self.out_checkpoints = {
+                k: PtCheckpoint(index_path)
+                for k in stored_epochs
+                if k in self.keep_epochs
+                for index_path in [self.output_path("models/epoch.%.3d.pt" % k)]
```
JackTemaki, line 195: We might want to go for 4 digits here...

returnn/training.py
```diff
@@ -159,13 +180,21 @@ def __init__(
             for k in stored_epochs
             if k in self.keep_epochs
         }
-        if self.returnn_config.get("use_tensorflow", False):
+        if self.returnn_config.get("use_tensorflow", False) or self.returnn_config.get("backend") == "tensorflow":
```
vieting, line 183:
```suggestion
        if self.returnn_config.get("use_tensorflow", False) or self.returnn_config.get("backend", None) == "tensorflow":
```
All other places that call `self.returnn_config.get()` define a default value.

vieting (approved):
Just a small comment

christophmluscher (approved):

returnn/training.py
SimBe195: Or infer from `num_epochs` with minimum 3 to not break old paths.

mmz33 (approved):

returnn/training.py
```diff
@@ -159,13 +180,21 @@ def __init__(
...
                 if k in self.keep_epochs
                 for index_path in [self.output_path("models/epoch.%.3d.index" % k)]
             }
+        if self.returnn_config.get("backend", None) == "torch":
+            self.out_checkpoints = {
+                k: PtCheckpoint(index_path)
+                for k in stored_epochs
+                if k in self.keep_epochs
+                for index_path in [self.output_path("models/epoch.%.3d.pt" % k)]
+            }
```
SimBe195, line 196: Since this is almost purely code duplication, couldn't this be unified into something like
```suggestion
        if self.returnn_config.get("use_tensorflow", False) or self.returnn_config.get("backend") == "tensorflow":
            checkpoint_class = Checkpoint
            suffix = "index"
        if self.returnn_config.get("backend", None) == "torch":
            checkpoint_class = PtCheckpoint
            suffix = "pt"

        self.out_checkpoints = {
            k: checkpoint_class(index_path)
            for k in stored_epochs
            if k in self.keep_epochs
            for index_path in [self.output_path(f"models/epoch.{k:03d}.{suffix}")]
        }
```
?

Atticus1806:
Would renaming `Checkpoint` to `TFCheckpoint` or so break anything? This is only used internally right?

michelwi (commented):
> Would renaming Checkpoint to TFCheckpoint or so break anything? This is only used internally right?

Along these lines but different: How about having TFCheckpoint and PTCheckpoint inherit from a Checkpoint class with a common interface?

returnn/training.py
```diff
@@ -159,13 +180,21 @@ def __init__(
             for k in stored_epochs
             if k in self.keep_epochs
         }
```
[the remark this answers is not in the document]
JackTemaki: This is deprecated anyway. I hope at some point it can be removed.

JackTemaki:
> Would renaming `Checkpoint` to `TFCheckpoint` or so break anything? This is only used internally right?
> 
> Edit: Ofc, code that uses `Checkpoint` needs to be changed, but I am thinking about like hashes etc. should not change I think. And then even though its tedious I would vouch for doing the change. I can look into it later.

Okay lets have a look, if it can be done safely we can do it.

JackTemaki:
For the Checkpoint renaming, this can be independent of this PR, right? Then I would leave this task to @Atticus1806 .

I also updated the typing while we are at it.

returnn/training.py
```diff
@@ -150,22 +171,32 @@ def __init__(
...
+            self.out_models = {
+                k: ReturnnModel(
+                    self.out_returnn_config_file,
+                    self.output_path("models/epoch.%.3d%s" % (k, suffix)),
+                    k,
+                )
+                for k in stored_epochs
+                if k in self.keep_epochs
+            }
+        if self.returnn_config.get("backend", None) == "torch":
```
albertz, line 192:
```suggestion
        elif self.returnn_config.get("backend", None) == "torch":
```

returnn/training.py
```diff
@@ -150,22 +171,32 @@ def __init__(
...
+            self.out_checkpoints = {
+                k: PtCheckpoint(pt_path)
+                for k in stored_epochs
+                if k in self.keep_epochs
+                for pt_path in [self.output_path("models/epoch.%.3d.pt" % k)]
+            }
+
```
albertz, line 199:
```suggestion
        else:
            raise ValueError("'backend' not specified in config")
```
Alternatively, if you really want to support a more custom config without defined backend, we add least should set `self.out_checkpoints = None` and `self.out_models` here, such that those attribs are defined in each case.

returnn/training.py
```diff
@@ -150,22 +171,32 @@ def __init__(
...
+                for k in stored_epochs
+                if k in self.keep_epochs
+            }
+        if self.returnn_config.get("backend", None) == "torch":
+            self.out_checkpoints = {
+                k: PtCheckpoint(pt_path)
+                for k in stored_epochs
+                if k in self.keep_epochs
+                for pt_path in [self.output_path("models/epoch.%.3d.pt" % k)]
+            }
```
albertz, line 198:
```suggestion
            }
            self.out_models = None
```
All existing attribs should always be defined in all branches.

albertz (approved):

PR #401, opened 2023-04-28
Add ffmpeg_input_options to BlissFfmpegJob

SimBe195:
Currently, the `BlissFfmpegJob` only accepts ffmpeg options for the output file. This was an issue when I tried to perform a `BlissChangeEncodingJob` but ffmpeg was not able to read the input files without specifying the correct codec.

This PR adds an `ffmpeg_in_options` parameter to the `BlissFfmpegJob` such that these options are applied when reading an input file. Also, `in_codec` and `in_codec_options` where added to the `BlissChangeEncodingJob` which directly utilize the new `ffmpeg_in_options`.

While we're touching these files, I also added type hints everywhere.

### DIFF
--- a/audio/__init__.py
+++ b/audio/__init__.py
@@ -1,0 +1,2 @@
+from .encoding import *
+from .ffmpeg import *
--- a/audio/encoding.py
+++ b/audio/encoding.py
@@ -13,8 +13,8 @@ class BlissChangeEncodingJob(BlissFfmpegJob):
 
     __sis_hash_exclude__ = {
         "recover_duration": None,
-        "in_codec": None,
-        "in_codec_options": None,
+        "input_codec": None,
+        "input_codec_options": None,
     }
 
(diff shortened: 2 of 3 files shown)


vieting (approved):
The changes look good to me. Also thanks for cleaning up!

I would prefer to use "input" instead of "in" in your newly introduced variable names. To me, `ffmpeg_in_options` intuitively sounds more like you have ffmpeg in some options rather than that these are the options for the ffmpeg input. It's not a must though, so I already approve.

SimBe195:
> The changes look good to me. Also thanks for cleaning up!
> 
> I would prefer to use "input" instead of "in" in your newly introduced variable names. To me, `ffmpeg_in_options` intuitively sounds more like you have ffmpeg in some options rather than that these are the options for the ffmpeg input. It's not a must though, so I already approve.

Okay, did the renaming!

PR #404, opened 2023-05-09
Set lm util to None

Atticus1806:
Fixes error that lm util is not in compiled rasr anymore

### DIFF
--- a/rasr/crp.py
+++ b/rasr/crp.py
@@ -44,6 +44,7 @@ class CommonRasrParameters:
             self.flf_tool_exe = None
             self.kws_tool_exe = None
             self.lattice_processor_exe = None
+            self.lm_util_exe = None
             self.nn_trainer_exe = None
             self.speech_recognizer_exe = None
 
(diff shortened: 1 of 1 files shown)


JackTemaki:
I think this can be removed completely, not only be set to None.

christophmluscher (changes requested):
There are Jobs which depend on `lm_util_exe`:
```
CreateLmImageJob
ComputePerplexityJob
```
@curufinwe @michelwi is this relevant for AppTek?

michelwi:
why is it not in compiled rasr any more?

christophmluscher:
> why is it not in compiled rasr any more?

the flag `TOOLS += Lm` got removed. Not 100% sure if the code for the tool also got removed..

michelwi (changes requested):
> @michelwi is this relevant for AppTek?

yes. And also the failed pipeline is a genuine breaking of hashes of one of our setups. Therefore request changes.

Actually, we should abandon this PR and try to fix that LM_util is not compiled any more.

christophmluscher:
`TOOLS += Lm` probably adding this to `Modules.Make` should fix this..

michelwi:
>  the failed pipeline is a genuine breaking of hashes of one of our setups.

I think the problem is that in FilterSegmentsByAlignmentConfidenceJob https://github.com/rwth-i6/i6_core/blob/e3ea366746eab8c7639ff680d40ce49a6870b6c5/corpus/filter.py#L73 The full crp are hashed. If one member is removed this changes the hash.

The version of this PR where it was set to `None` would, I think, still work for us (the hashes were correct at least).
And we use a different compile job for the AppTek RASR.

michelwi (approved):
I am ok with setting to None here. We do set this elsewhere.

Atticus1806:
@christophmluscher is this fine? Then I would merge

christophmluscher (approved):

PR #405, opened 2023-05-10
add selector for legacy or corrected fsa

christophmluscher:


michelwi (approved):

am/config.py
```diff
@@ -69,6 +71,8 @@ def acoustic_model_config(
     config.hmm.early_recombination = early_recombination
 
     config.tdp.scale = tdp_scale
+    if use_corrected_fsa:
```
Marvin84, line 74: Setting the applicator type to `corrected` without updating other parameters is not going to work. The corrected applicator needs additional settings as shown in the `get_align_config_and_crp_for_corrected_applicator`. Moreover, this is only related to alignment task and should not be set to the general acoustic model config that might be used for also decoding.

Marvin84: My suggestion is to cancel this PR. Since IMO, the setting should be done only via the `get_align_config_and_crp_for_corrected_applicator `.

christophmluscher: oh than I misunderstood this.. I thought you also need this for decoding?! O.o

so you would only add the flag to the alignment step.. so adding `use_corrected_fsa` to some training args?

Marvin84: > I thought you also need this for decoding

No, this is only for the alignment FSA. And I confirm your question above, one should have it only for training and alignment.

Marvin84: The decoding FSA uses a completely different subroutine.

christophmluscher: OK. Thanks for the explanation.

christophmluscher:
Misunderstanding on my side. This is unnecessary.

PR #406, opened 2023-05-10
SRI LM Jobs

christophmluscher:
These are Jobs for the SRI LM toolkit. We will add a full pipeline in `i6_experiments`.

For now, this is a draft PR, in order to talk on a bit higher level about the jobs. Missing parameters and so on.

### DIFF
--- a/lm/srilm.py
+++ b/lm/srilm.py
@@ -370,10 +370,6 @@ class ComputeBestMixJob(Job):
 
         relink("cbm.log", self.out_cbm_file.get_path())
 
-    @classmethod
-    def hash(cls, parsed_args):
-        return super().hash(parsed_args)
-
 
 class InterpolateNgramLmJob(Job):
     """
--- a/lm/srilm.py
+++ b/lm/srilm.py
@@ -12,7 +12,7 @@ import os
 import shutil
 
 from enum import Enum
-from typing import Dict, List, Optional, Union
+from typing import Dict, List, Optional
 
 from sisyphus import tk, Job, Task
 
@@ -374,11 +374,9 @@ class ComputeNgramLmPerplexityJob(Job):
 
 
 class ComputeBestMixJob(Job):
-    def __init__(self, ppl_log: List[tk.Path], compute_best_mix_exe: Optional[tk.Path] = None):
+    def __init__(self, ppl_log: List[tk.Path], compute_best_mix_exe: tk.Path):
         self.ppl_log = ppl_log
-        self.compute_best_mix_exe = (
-            compute_best_mix_exe if compute_best_mix_exe is not None else tk.Path("compute-best-mix")
-        )
+        self.compute_best_mix_exe = compute_best_mix_exe
 
         self.out_lambdas = [self.output_var(f"lambdas{i}") for i, p in enumerate(ppl_log)]
         self.out_cbm_file = self.output_path("cbm.log")
--- a/lm/srilm.py
+++ b/lm/srilm.py
@@ -478,7 +478,7 @@ class InterpolateNgramLmJob(Job):
         """
 
         :param ngram_lms: List of language models to interpolate
-        :param lambdas: Weights of different language models, has to be same order as LMs
+        :param lambdas: Weights of different language models, has to be same order as ngram_lms
         :param ngram_order: Maximum n gram order
         :param interpolation_args: Additional arguments for interpolation
         :param ngram_exe: Path to srilm ngram executable
--- a/lm/srilm.py
+++ b/lm/srilm.py
@@ -64,7 +64,7 @@ class CountNgramsJob(Job):
 
     def tasks(self):
         yield Task("create_files", mini_task=True)
-        yield Task("run", resume="run", rqmt=self.rqmt)
+        yield Task("run", rqmt=self.rqmt)
 
     def create_files(self):
         """creates bash script that will be executed in the run Task"""
@@ -155,10 +155,9 @@ class ComputeNgramLmJob(Job):
 
     def tasks(self):
         yield Task("create_files", mini_task=True)
-        yield Task("run", resume="run", rqmt=self.rqmt_run)
+        yield Task("run", rqmt=self.rqmt_run)
         yield Task(
             "compress",
-            resume="compress",
             rqmt={"mem": 2, "time": 2, "cpu": 1, "fs": self.fs_rqmt},
         )
 
@@ -267,7 +266,7 @@ class ComputeNgramLmPerplexityJob(Job):
 
     def tasks(self):
         yield Task("create_files", mini_task=True)
-        yield Task("run", resume="run", rqmt=self.rqmt)
+        yield Task("run", rqmt=self.rqmt)
         yield Task("get_ppl", mini_task=True)
 
     def create_files(self):
@@ -523,7 +522,7 @@ class PruneLMWithHelperLMJob(Job):
 
     def tasks(self):
         yield Task("create_files", mini_task=True)
-        yield Task("run", resume="run", rqmt=self.rqmt_run)
+        yield Task("run", rqmt=self.rqmt_run)
 
     def create_files(self):
         """creates bash script that will be executed in the run Task"""
--- a/lm/srilm.py
+++ b/lm/srilm.py
@@ -21,7 +21,7 @@ from i6_core.util import create_executable, relink, get_ngram_count_exe, get_ngr
 
 class CountNgramsJob(Job):
     """
-    count ngrams with SRILM
+    count ngrams with srlim
     """
 
     def __init__(
@@ -95,7 +95,7 @@ class CountNgramsJob(Job):
 
 class OptimizeKNDiscountsJob(Job):
     """
-    Uses SRILM to optimize Kneser-Ney discounts for a given dataset
+    Uses srlim to optimize Kneser-Ney discounts for a given dataset
     """
 
     def __init__(
@@ -173,7 +173,7 @@ class OptimizeKNDiscountsJob(Job):
 
 class ComputeNgramLmJob(Job):
     """
-    Generate count based LM with SRILM
+    Generate count based LM with srlim
     """
 
     class DataMode(Enum):
@@ -292,6 +292,7 @@ class ComputeNgramLmJob(Job):
 
 
 class ComputeNgramLmPerplexityJob(Job):
+    """Calculate the Perplexity of an Ngram LM via srlim"""
     def __init__(
         self,
         ngram_order: int,
@@ -307,16 +308,18 @@ class ComputeNgramLmPerplexityJob(Job):
         fs_rqmt: str = "10G",
     ):
         """
-        :param ngram_order:
-        :param lm:
-        :param vocab:
-        :param eval_data:
-        :param ppl_args: -debug 2
-        :param ngram_exe:
-        :param mem_rqmt:
-        :param time_rqmt:
-        :param cpu_rqmt:
+        :param ngram_order: Maximum n gram order
+        :param lm: LM to evaluate
+        :param vocab: Vocabulary file
+        :param eval_data: Data to calculate PPL on
+        :param ppl_args: Extra arguments for the execution call e.g. '-debug 2'
+        :param ngram_exe: Path to srilm ngram exe
+        :param mem_rqmt: Memory requirements of Job (not hashed)
+        :param time_rqmt: Time requirements of Job (not hashed)
+        :param cpu_rqmt: Amount of Cpus required for Job (not hashed)
+        :param fs_rqmt: Space on fileserver required for Job, example: "200G" (not hashed)
         """
+
         self.ngram_order = ngram_order
         self.lm = lm
         self.vocab = vocab
@@ -345,6 +348,7 @@ class ComputeNgramLmPerplexityJob(Job):
         yield Task("get_ppl", mini_task=True)
 
     def create_files(self):
+        """creates bash script that will be executed in the run Task"""
         cmd = [
             f"{self.ngram_exe.get_path()} \\\n",
             f"  -order {self.ngram_order} \\\n",
@@ -365,10 +369,12 @@ class ComputeNgramLmPerplexityJob(Job):
         create_executable("run.sh", cmd)
 
     def run(self):
+        """executes the previously created script and relinks the log file from work folder to output folder"""
         self.sh("./run.sh")
         relink("ppl.log", self.out_ppl_log.get_path())
 
     def get_ppl(self):
+        """extracts various outputs from the ppl.log file"""
         with open(self.out_ppl_log.get_cached_path(), "rt") as f:
             lines = f.readlines()[-2:]
             for line in lines:
@@ -385,6 +391,7 @@ class ComputeNgramLmPerplexityJob(Job):
 
     @classmethod
     def hash(cls, kwargs):
+        """delete the queue requirements from the hashing"""
         del kwargs["mem_rqmt"]
         del kwargs["cpu_rqmt"]
         del kwargs["time_rqmt"]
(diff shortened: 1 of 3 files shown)


JackTemaki:
Some documentation would be nice to understand what jobs and parameters should be doing.

michelwi:
I think it looks good to me on a high level.
 
some parameter names could need a reworking
documentation missing

but I like that each Job does exactly one thing and now we can connect this to a nice pipeline 🎉

Atticus1806:
This is also ready now. Please review.

lm/srilm.py
```diff
@@ -0,0 +1,634 @@
...
+        mem_rqmt: int = 16,
+        time_rqmt: float = 12,
+        cpu_rqmt: int = 1,
+        fs_rqmt: str = "10G",
+    ):
+        """
+        :param ngram_order: Maximum n gram order
+        :param lm: LM to evaluate
+        :param vocab: Vocabulary file
+        :param eval_data: Data to calculate PPL on
```
michelwi, line 318:
```suggestion
        :param eval_data: Data to calculate PPL on
        :param vocab: Vocabulary file
        :param set_unknown_flag: TODO
```
same order as in function definition + missing docstring for `set_unknown_flag`

lm/srilm.py
```diff
@@ -0,0 +1,634 @@
...
+        del kwargs["mem_rqmt"]
+        del kwargs["cpu_rqmt"]
+        del kwargs["time_rqmt"]
+        del kwargs["fs_rqmt"]
+        return super().hash(kwargs)
+
+
+class ComputeBestMixJob(Job):
+    """
+    Compute the best mixture weights from given PPL logs
```
michelwi, line 408:
```suggestion
    Compute the best mixture weights for a combination of count LMs based on the given PPL logs
```
maybe we can also add "linear" or "log-linear" if we know what actually happens here xD

lm/srilm.py
```diff
@@ -0,0 +1,634 @@
...
+    def __init__(self, ppl_log: List[tk.Path], compute_best_mix_exe: Optional[tk.Path] = None):
+        """
+
+        :param ppl_log: List of PPL Logs to compute the weights from
+        :param compute_best_mix_exe: Path to srilm compute_best_mix executable (not hashed)
+        """
+        self.ppl_log = ppl_log
+        self.compute_best_mix_exe = get_compute_best_mix_exe(compute_best_mix_exe)
+
+        self.out_lambdas = [self.output_var(f"lambdas{i}") for i, p in enumerate(ppl_log)]
```
michelwi, line 420:
```suggestion
        self.out_weights = [self.output_var(f"weight{i}") for i, p in enumerate(ppl_log)]
```
rename to be more descriptive. lambda is just a letter of the alphabet and can mean anything.
Also the docstring says the class would compute weights.

michelwi (changes requested):
There is nothing in particular that I require to be changed, but based on the number of suggestions I'll press request changes.

lm/srilm.py
```diff
@@ -0,0 +1,634 @@
...
+        time_rqmt: float = 24,
+        cpu_rqmt: int = 1,
+        fs_rqmt: str = "100G",
+    ):
+        """
+        :param ngram_order: Maximum n gram order
+        :param data: Held-out dataset to optimize discounts on
+        :param vocab: Vocabulary file
+        :param num_discounts: Number of discounts to optimize
+        :param count_file: File to read counts from
```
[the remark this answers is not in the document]
Atticus1806: > what does the number of discounts mean ?

This seems to be some custom flag in the version of srilm that i could not find documentation online for and this wording is what is printed via -help.

michelwi (changes requested):
Remove all Kazuki specific options. This should work with the official srilm version

Atticus1806:
Maybe there was a slight misunderstanding. This works with the official version, just not with the extra flags we called it with.

lm/srilm.py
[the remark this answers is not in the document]
michelwi: > Maybe there was a slight misunderstanding. This works with the official version, just not with the extra flags we called it with in i6_experiments.

Yes, I am confused now. from your previous answer 
>  This seems to be some custom flag in the version of srilm that i could not find documentation online for

I assumed that `-num-discounts` is one of the custom flags that do not exist in the official version. 
Which would imply that the `OptimizeKNDiscountsJob` cannot be used with the official version of srilm as the `run` function will always add the officially unsupported flag to the call.

My which would be that all visible options to the classes are from the official srilm version and any user that provides a modified version may add additional command line args via the `extra_args`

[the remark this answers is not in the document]
Atticus1806: Ah well my bad. So yes that job is not working without the internal version. In our setup this is clearly marked and "optional". I was referring to flags in the other jobs. 

I will think about how to best handle this.

lm/srilm.py
```diff
@@ -0,0 +1,634 @@
...
+        count_exe: Optional[tk.Path] = None,
+        multi_kn_file: Optional[tk.Path] = None,
+        mem_rqmt: int = 48,
+        time_rqmt: float = 24,
+        cpu_rqmt: int = 1,
+        fs_rqmt: str = "100G",
+    ):
+        """
+        :param ngram_order: Maximum n gram order
+        :param data: Either text file or counts file to read from, set data mode accordingly
```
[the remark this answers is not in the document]
christophmluscher: hmm well both are text files but the structure is different.. but the heuristic would need some hardcoded assumptions which I dont like that much.. that is why I decided against auto detection

christophmluscher:
> Remove all Kazuki specific options. This should work with the official srilm version

IIRC this is only for the `OptimizeKNDiscountsJob` the param `num_discounts`. Would you be OK with making this `Optional` and `None` as default?

michelwi:
> > Remove all Kazuki specific options. This should work with the official srilm version
> 
> IIRC this is only for the params `num_discounts` and `multi_kn_file`. Would you be OK with making this `Optional` and `None` as default or so? @Atticus1806 you think this is enough?

This is the least. But I would prefer if these options would vanish completely. Then we do a test how much PPl we lose. If we lose a lot, then in your private setup (or public with a comment) we add this via the extra args.

mh.. for the `extra_args=["--num-discounts", "3"]` I can see this, but if the `multi_kn_file` is a Path, then we maybe cannot do this very easily..

Atticus1806:
The results are the same without the jobs, so we decided on fully removing them.

lm/srilm.py
```diff
@@ -0,0 +1,634 @@
...
+    class DataMode(Enum):
+        TEXT = 1
+        COUNT = 2
+
+    def __init__(
+        self,
+        ngram_order: int,
+        data: tk.Path,
+        data_mode: DataMode,
+        vocab: Optional[tk.Path] = None,
```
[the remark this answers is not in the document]
christophmluscher: OK. then we leave it.

util.py
```diff
@@ -372,3 +372,15 @@ def get_g2p_python(g2p_python: tk.Path) -> tk.Path:
 def get_subword_nmt_repo(subword_nmt_repo: tk.Path) -> tk.Path:
     """gets the path to the root folder of subword-nmt repo"""
     return get_executable_path(subword_nmt_repo, "SUBWORD_NMT_PATH")
+
+
+def get_ngram_count_exe(count_exe: tk.Path):
+    return get_executable_path(count_exe, None, shutil.which("ngram-count"))
```
JackTemaki, line 378: Do we really want to support this? I would prefer no default, so that you always have to give this explicitly. We anyway do not know yet if it would be allowed for us to include SRILM directly in Apptainer images.

JackTemaki (changes requested):

lm/srilm.py
```diff
@@ -0,0 +1,548 @@
...
+            f"  -vocab {self.vocab.get_cached_path()} \\\n" if self.vocab is not None else "  -write-vocab vocab \\\n"
+        )
+        if self.data_mode == ComputeNgramLmJob.DataMode.TEXT:
+            data_str = "-text"
+        elif self.data_mode == ComputeNgramLmJob.DataMode.COUNT:
+            data_str = "-read"
+        else:
+            raise NotImplementedError
+
+        cmd = [
```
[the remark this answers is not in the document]
christophmluscher: idea was for a more readable run script

lm/srilm.py
```diff
@@ -0,0 +1,548 @@
...
+        create_executable("run.sh", cmd)
+
+    def run(self):
+        """executes the previously created script and relinks the log file from work folder to output folder"""
+        self.sh("./run.sh")
+        relink("ppl.log", self.out_ppl_log.get_path())
+
+    def get_ppl(self):
+        """extracts various outputs from the ppl.log file"""
+        with open(self.out_ppl_log.get_cached_path(), "rt") as f:
```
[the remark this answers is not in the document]
christophmluscher: @Atticus1806 is this working for you??

I think @JackTemaki is correct.. but if this works I am understanding something incorrectly...

util.py
christophmluscher: so if the `str` for the `shutil.which` is not found `None` is returned. the fould lead to an assertion error. I think this is ok. but I do not have a strong opinion on this.. @Atticus1806 what do you say?

lm/srilm.py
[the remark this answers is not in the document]
christophmluscher: some of the commands can get quite long..

christophmluscher:
@JackTemaki what is a must change? all things mentioned?

util.py
Atticus1806: When  I think about it I don't really think we should have "hidden defaults" for executables. 99% of the time you want to have this set anyways right? And you get the benefit of 1% "nice behavior" with the risk of having unintended behavior in a case where you just forgot. So I would not mind crashing here.
We also did similar things for `returnn_exe` etc. right? Where we disallowed pulling from `gs` which I would say is similar to this.

lm/srilm.py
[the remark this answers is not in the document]
Atticus1806: I am confident that this worked for me. Maybe my caching is setup incorrectly?

lm/srilm.py
```diff
@@ -0,0 +1,548 @@
...
+        extra_ngram_args: Optional[List[str]] = None,
+        count_exe: Optional[tk.Path] = None,
+        mem_rqmt: int = 48,
+        time_rqmt: float = 24,
+        cpu_rqmt: int = 1,
+        fs_rqmt: str = "100G",
+    ):
+        """
+        :param ngram_order: Maximum n gram order
+        :param data: Either text file or counts file to read from, set data mode accordingly
```
JackTemaki, line 119: Can the count file come from the `CountNgramsJob`s `out_counts`? If yes I think this would be nice to add here.

lm/srilm.py
```diff
@@ -0,0 +1,548 @@
...
+            "qsub_args": f"-l h_fsize={fs_rqmt}",
+        }
+        self.fs_rqmt = fs_rqmt
+
+    def tasks(self):
+        yield Task("create_files", mini_task=True)
+        yield Task("run", resume="run", rqmt=self.rqmt_run)
+        yield Task(
+            "compress",
+            resume="compress",
```
JackTemaki, line 158: Are you sure resuming can not cause issues?

lm/srilm.py
```diff
@@ -0,0 +1,548 @@
...
+        mem_rqmt: int = 48,
+        time_rqmt: float = 24,
+        cpu_rqmt: int = 1,
+        fs_rqmt: str = "100G",
+    ):
+        """
+        :param ngram_order: Maximum n gram order
+        :param data: Either text file or counts file to read from, set data mode accordingly
+        :param data_mode: Defines whether input format is text based or count based
+        :param vocab: Vocabulary file
```
JackTemaki, line 121: What is the expected format of a vocab file? Are there specific jobs that output the desired format?

lm/srilm.py
```diff
@@ -0,0 +1,548 @@
...
+        extra_interpolation_args: Optional[Dict] = None,
+        ngram_exe: Optional[tk.Path] = None,
+        cpu_rqmt: int = 1,
+        mem_rqmt: int = 32,
+        time_rqmt: int = 4,
+        fs_rqmt: str = "50G",
+    ):
+        """
+
+        :param ngram_lms: List of language models to interpolate
```
JackTemaki, line 393: What formats are allowed? Arpa and compressed arpa or also others?

JackTemaki (commented):
@christophmluscher I am not giving explicit approval yet because of the not completed discussion of default exe's. The comments here are not a must but maybe still good to tackle.

util.py
```diff
@@ -300,7 +300,7 @@ def instanciate_delayed(o):
 
 def get_executable_path(
     path: Optional[tk.Path],
-    gs_member_name: str,
+    gs_member_name: Optional[str],
```
JackTemaki, line 303: this variable seems unused

lm/srilm.py
```diff
@@ -0,0 +1,553 @@
...
+
+        Example options/parameters for count_args:
+        -unk
+        """
+        self.ngram_order = ngram_order
+        self.data = data
+        self.count_args = extra_count_args if extra_count_args is not None else ["-unk"]
+        self.count_exe = count_exe
+
+        self.out_counts = self.output_path("counts", cached=True)
```
JackTemaki, line 56: i am more a fan of deciding caching on the input side, but as this anyway inconsisten everywhere lets keep it

lm/srilm.py
```diff
@@ -0,0 +1,553 @@
...
+
+        Example options for ngram_args:
+        -kndiscount -interpolate -debug <int> -addsmooth <int>
+        """
+        self.ngram_order = ngram_order
+        self.data = data
+        self.data_mode = data_mode
+        self.vocab = vocab
+        if ngram_order == 1 and extra_ngram_args is None:
+            extra_ngram_args = "-debug 0 -addsmooth 0"
```
JackTemaki, line 140: maybe add why this is needed

JackTemaki (changes requested):
One small thing, but we also might want to replace `self.sh` with subprocess calls. @michelwi did not like `self.sh` because it opens shell capabilities.

util.py
Atticus1806: It is not right? c.f. Line 337

christophmluscher:
> One small thing, but we also might want to replace self.sh with subprocess calls. @michelwi did not like self.sh because it opens shell capabilities.

IMHO than we should fix the self.sh function in sisyphus...

lm/srilm.py
christophmluscher: not sure why I added this..

JackTemaki:
> > One small thing, but we also might want to replace self.sh with subprocess calls. @michelwi did not like self.sh because it opens shell capabilities.
> 
> IMHO than we should fix the self.sh function in sisyphus...

If possible we would have done that, but there are MT recipes that rely on that feature.

util.py
JackTemaki: Oh, I saw something different, forget about this

lm/srilm.py
```diff
@@ -0,0 +1,553 @@
...
+        lbds = lines[-2].split("(")[1].replace(")", "")
+        lbds = lbds.split()
+
+        for i, v in enumerate(lbds):
+            self.out_weights[i].set(float(v))
+
+        relink("cbm.log", self.out_cbm_file.get_path())
+
+    @classmethod
+    def hash(cls, parsed_args):
```
JackTemaki, line 373: This does nothing?

JackTemaki (changes requested):

Atticus1806:
Everything should be fine now.

Nintendo-Wei:
Looks pretty good to me now, I'll approve.

Nintendo-Wei (approved):

christophmluscher:
@JackTemaki I think all is addressed. Can you check/approve?

JackTemaki (approved):
I still do not like the way we do caching, but ok.

christophmluscher:
what do you mean?

christophmluscher:
output caching?? or the cmd thingy?

JackTemaki:
The output caching, but I think this is more general than this PR.

christophmluscher:
OK. Then I'll merge this, we can still change this if necessary in a later PR

PR #407, opened 2023-05-10
add corrected applicator to linear alignment

christophmluscher:
@Marvin84 can I apply this like this to the linear alignment step?

### DIFF
--- a/am/config.py
+++ b/am/config.py
@@ -1,20 +1,14 @@
 __all__ = ["acoustic_model_config", "get_align_config_and_crp_for_corrected_applicator"]
 
 from dataclasses import dataclass
-from typing import List, Tuple, Union, Optional
+from typing import List, Literal, Tuple, Union, Optional
 
-try:
-    from typing import Literal
-
-    TdpType = Union[float, Literal["infinity"]]
-    TyingType = Literal["global", "global-and-nonword"]
-except ImportError:
-    TdpType = Union[float, str]  # str must be: "infinity"
-    TyingType = str  # str must be : "global", "global-and-nonword"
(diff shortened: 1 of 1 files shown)


Marvin84:
> @Marvin84 can I apply this like this to the linear alignment step?

Looks good.
Once you ran this, please send me your `alignment.config` of the resulting job so that I can double-check.

am/config.py
```diff
@@ -113,7 +113,8 @@ def get_align_config_and_crp_for_corrected_applicator(
     """
     Set the correct type of applicator, default is "legacy". Moreover, set exit penalities to zero
     For a given word sequence the exit penalty is constant with respect to the max/sum
-    :param config:
+    :param crp:
+    :param exit_penalty:
```
Atticus1806, line 117: Could you add some docs pls? :)

Atticus1806:
I have to be honest. I have no clue :sweat_smile:

am/config.py
Marvin84: Docs of what type you mean?

Marvin84:
> I have to be honest. I have no clue 😅

There is a `rasr` PR for this : https://github.com/rwth-i6/rasr/pull/14

Atticus1806:
> There is a rasr PR for this : https://github.com/rwth-i6/rasr/pull/14

Thanks, but I was referring to the python test failing or rather why this only comes up now ^^

Atticus1806 (changes requested):
So if we bump to 3.8 I would revert the "fix" to make the test run.

Atticus1806 (approved):

PR #411, opened 2023-05-17
Add multiple links.

sarahberanek:
Run into KeyError , because of multiple links from a DeltaDelta Node in an old GMM recognition setup.

### DIFF
--- a/rasr/flow.py
+++ b/rasr/flow.py
@@ -237,10 +237,12 @@ class FlowNetwork:
 
         # now add all the links
         for l in internal_links:
-            name_in = l[0].split(":")[0]
-            name_out = l[1].split(":")[0]
+            split_in = l[0].split(":")
+            split_out = l[1].split(":")
+            name_in = split_in[0]
+            name_out = split_out[0]
             if name_in in processed_nodes:
-                net.link(mapping[l[0]], mapping[l[1]])
+                net.link(":".join([mapping[name_in]] + split_in[1:]), ":".join([mapping[name_out]] + split_out[1:]))
             elif name_out in processed_nodes:
                 broken_links.append((l[0], mapping[l[1]]))
 


michelwi:
could you give some example of the links that break the old code?
(and also some of working ones if you have them)

sarahberanek:
> could you give some example of the links that break the old code? (and also some of working ones if you have them)

network:
```
 <node filter="signal-delay" margin-condition="present-not-empty" max-size="5" name="delay" right="2"/>                                                                                                     <link from="mfcc-normalization" to="delay"/>
  <node filter="signal-regression" name="delta" order="1" timestamp-port="0"/>
  <link from="delay:-2" to="delta:-2"/>                                                              
  <link from="delay:-1" to="delta:-1"/>                                                              
  <link from="delay:0" to="delta:0"/>
  <link from="delay:1" to="delta:1"/>
  <link from="delay:2" to="delta:2"/> 
```                                                              


Error message:     
```                                                                                                                                                                                                    
        mapping = {'delay': 'delay', 'deltadelta': 'deltadelta', 'deltadelta-split': 'deltadelta-split', 'concat': 'concat', 'linear-transform': 'linear-transform', 'delta': 'delta'}                     
        l = ('delay:-2', 'delta:-2')                                                                                                                                                                       
        l[0] = 'delay:-2'                                                                                                                                                                                  
        l[1] = 'delta:-2'                                                                                                                                                                                  
    244 elif name_out in processed_nodes:                                                                                                                                                                  
    245     broken_links.append((l[0], mapping[l[1]]))                                                                                                                                                     
                                                                                                                                                                                                           
KeyError: 'delay:-2'   
```

michelwi (approved):
thanks, that makes it easier to spot what is going wrong. 
I think the proposed solution is good.

christophmluscher (approved):

sarahberanek:
Can we merge this now? How is the procedure in i6 core? Could I merge myself after 2 approvals?

PR #412, opened 2023-05-22
Gracefully handle transient FileExistsError

NeoLegends:
Since Ubuntu 22 my sisyphus manager has been crashing occasionally w/ the following obscure error while trying to read the learning rates file to show training progress:
Bildschirmfoto 2023-05-22 um 11 57 40

The error does not make a lot of sense as it is trying to open the file in read mode (so why is it complaining the file exists?!), but it is transient and usually does not happen again after a short while when the sisyphus manager is restarted. We can therefore just ignore it and wait until the job info is computed the next time.

Feel free to merge if you're the one doing the second approval.

### DIFF
--- a/returnn/training.py
+++ b/returnn/training.py
@@ -231,6 +231,8 @@ class ReturnnTrainingJob(Job):
             try:
                 with open(file_path, "rt") as file:
                     return eval(file.read().strip())
+            except FileExistsError:
+                return None
             except FileNotFoundError:
                 return None
 


JackTemaki:
I mean this is in "info" so i6_core is correct I guess. The bigger question is rather: how can a "read open" cause a "file exists" error?

NeoLegends:
>The bigger question is rather: how can a "read open" cause a "file exists" error?

I found https://stackoverflow.com/questions/24761181/why-fopen-s-returns-eexist-errno-17-when-opening-file-for-reading, which suggests that in some cases, the error number is not indicative. Maybe we're hitting this case (`fopen_s`) here.

NeoLegends:
Yes. If you run `file` it exists.

christophmluscher (approved):

Atticus1806 (approved):

PR #413, opened 2023-05-24
Output WER to relpath instead of to abspath

Icemole:
When `prefix=""`, the output was being written to `/wers/...`, which is undesirable since `/wers` is outside of the recipe. With this change, we ensure that the output is always correctly written to the relative path `output/wers/...`.

Note that for `prefix` to work correctly it needs to be a directory, i.e. it needs to finish with `/`, but I think that's a precondition that's already set two lines above, when setting more outputs to `{prefix}recog_...`

### DIFF
--- a/meta/system.py
+++ b/meta/system.py
@@ -721,7 +721,7 @@ class System:
         self.jobs[corpus]["scorer_%s" % name] = scorer
         tk.register_output("%srecog_%s.reports" % (prefix, name), scorer.out_report_dir)
         if scorer.out_wer:
-            tk.register_output("%s/wers/recog_%s.wer" % (prefix, name), scorer.out_wer)
+            tk.register_output("%swers/recog_%s.wer" % (prefix, name), scorer.out_wer)
 
     def optimize_am_lm(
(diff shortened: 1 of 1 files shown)


Icemole:
With this change, the code now outputs to the correct directory (with `prefix=""`):
```
[2023-05-24 02:01:35,250] INFO: Finished output: output/wers/recog_gmm/
```

Atticus1806:
>i.e. it needs to finish with `/`

Is that true? I mean worst case you get smth. like `output/baselinewer` right, which would be a user error?

Atticus1806 (approved):

Icemole:
> [...] would be a user error?

Yes, that was what I was meaning. Indeed, the prefix doesn't need to finish with `/`, but it's more desirable if it finishes that way, I think. When I set `prefix="output"` I get the following output:

```
[2023-05-24 03:33:13,075] INFO: Finished output: output/outputwers/recog_gmm/...
```

Which is undesirable but to be expected, as you said.

christophmluscher (approved):

JackTemaki (approved):

PR #414, opened 2023-05-26
Add exists function to Checkpoints

JackTemaki:
Just a small change regarding checkpoint file existence check that was needed to get my pipeline running.

### DIFF
--- a/returnn/forward.py
+++ b/returnn/forward.py
@@ -111,9 +111,7 @@ class ReturnnForwardJob(Job):
 
         # check here if model actually exists
         if self.model_checkpoint is not None:
-            assert os.path.exists(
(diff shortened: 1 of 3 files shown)


Atticus1806 (approved):

michelwi:
I assume this pipeline is using the `ReturnnForwardJob` with a pytorch model?

Would you then also change the typing info of the init to 
```
model_checkpoint: Optional[Union[Checkpoint,PtCheckpoint]],
```

Also maybe there are other places that could already make use of the new exists function (e.g. in training or search)?

michelwi (approved):
Thanks.

christophmluscher (approved):

PR #416, opened 2023-06-02
Create dummy mixtures with non-zero variance

NeoLegends:
This makes the mixtures still dummy but more sensible, or rather more capable of faking to be a proper mixture. Particularly this change allows using these mixtures to create a dummy `DiagonalMaximumScorer` for jobs that want a feature scorer but don't actually need to score features.

Do you think this warrants a job ctor parameter?

### DIFF
--- a/mm/mixtures.py
+++ b/mm/mixtures.py
@@ -503,9 +503,10 @@ class CreateDummyMixturesJob(Job):
         with open(tk.uncached_path(self.out_mixtures), "wb") as f:
             f.write(b"MIXSET\0\0")
             f.write(struct.pack("II", 2, num_features))
-            args = [1, num_features] + [0.0] * num_features + [1.0]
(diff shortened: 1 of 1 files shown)


Marvin84 (approved):

michelwi:
Your proposal simply changes the numerical value of the variance that is written to file from `0.0` to `1.0`, just so we can calculate an inverse, right?


>Do you think this warrants a job ctor parameter?

I would even call variance 0.0 a bug and this is a fix. Given that the previous variance values could not have been used for anything I would be ok with just changing it without a parameter.

NeoLegends:
> Your proposal simply changes the numerical value of the variance that is written to file from 0.0 to 1.0, just so we can calculate an inverse, right?

Exactly.

christophmluscher (approved):

PR #419, opened 2023-06-19
Add alias to am_lm_scale and lattice2ctm job

Atticus1806:
Add alias to LatticeToCtmJob and OptimizeAMandLMScaleJob

### DIFF
--- a/meta/system.py
+++ b/meta/system.py
@@ -715,6 +715,7 @@ class System:
             parallelize=parallelize_conversion,
             **lattice_to_ctm_kwargs,
         )
+        lat2ctm.add_alias("%s%s_conv_to_ctm_" % (prefix, name))
         self.ctm_files[corpus]["recog_%s" % name] = lat2ctm.out_ctm_file
 
         kwargs = copy.deepcopy(self.scorer_args[corpus])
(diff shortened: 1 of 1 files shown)


JackTemaki (changes requested):

meta/system.py
```diff
@@ -715,6 +715,7 @@ def recog(
             parallelize=parallelize_conversion,
             **lattice_to_ctm_kwargs,
         )
+        lat2ctm.add_alias("%s%s_conv_to_ctm" % (prefix, name))
```
michelwi, line 718: We have already 
```
rec.add_alias("%srecog_%s" % (prefix, name))
```
and 
```
self.jobs[corpus]["lat2ctm_%s" % name] = ...
```
so I guess the most consistent formatting would be
```suggestion
        lat2ctm.add_alias("%slat2ctm_%s" % (prefix, name))
```

meta/system.py
```diff
@@ -748,6 +749,7 @@ def optimize_am_lm(
             scorer_hyp_param_name=self.scorer_hyp_arg[corpus],
             **kwargs,
         )
+        j.add_alias("%s%s_optimize_am_and_lm_scale" % (prefix, name))
```
michelwi, line 752: same as above
```suggestion
        j.add_alias("%soptimize_%s" % (prefix, name))
```

michelwi (commented):
of cause the aliases can be chosen arbitrarily, but here some suggestions for more consistency.

michelwi (approved):

christophmluscher (approved):

PR #420, opened 2023-06-19
Add alias to LMImageAndGlobalCacheJob

Atticus1806:
This PR adds an alias to the LMImageAndGlobalCacheJob within the AdvancedTreeSearchJob. I was not really able to find a way to add this without adding a job input and excluding it from hash. Feel free to suggest better ways if possible.


recognition/advanced_tree_search.py
```diff
@@ -148,6 +148,8 @@ def hash(cls, kwargs):
 
 
 class AdvancedTreeSearchJob(rasr.RasrCommand, Job):
+    __sis_hash_exclude__ = {"lmgc_alias": None}
+
```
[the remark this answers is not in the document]
Atticus1806: Oh I did not see that.

christophmluscher (commented):
since you are touching this file it would be nice to update typing and so on :)

recognition/advanced_tree_search.py
```diff
@@ -150,25 +151,48 @@ def hash(cls, kwargs):
...
+        lm_lookahead: bool = True,
+        lookahead_options: Optional[Dict[str, Any]] = None,
+        create_lattice: bool = True,
+        eval_single_best: bool = True,
+        eval_best_in_lattice: bool = True,
+        use_gpu: bool = False,
+        rtf: int = 10,
+        mem: int = 4,
+        cpu: int = 1,
+        lmgc_mem: int = 12,
```
christophmluscher, line 167:
```suggestion
        rtf: float = 10.,
        mem: float = 4.,
        cpu: int = 1,
        lmgc_mem: float = 12.,
```

?

recognition/advanced_tree_search.py
```diff
@@ -246,25 +270,28 @@ def update_search_parameters(cls, search_parameters):
...
+        feature_scorer: rasr.FeatureScorer,
+        search_parameters: Union[None, Dict[str, Any]],
+        lm_lookahead: bool,
+        lookahead_options: Union[None, Dict[str, Any]],
+        create_lattice: bool,
+        eval_single_best: bool,
+        eval_best_in_lattice: bool,
+        mem: int,
+        cpu: int,
+        lmgc_mem: int,
```
christophmluscher, line 284:
```suggestion
        mem: float,
        cpu: int,
        lmgc_mem: float,
```

christophmluscher (approved):
thx! :)

IIRC some params can be `float` instead of `int`. otherwise fine.

christophmluscher (changes requested):
I ll add a comment in a sec

christophmluscher (approved):
just as a suggestion

PR #422, opened 2023-06-20
Delete empty recordings

hannah220:
Add new option to delete empty recordings in FilterCorpusBySegmentsJob.

### DIFF
--- a/corpus/filter.py
+++ b/corpus/filter.py
@@ -148,12 +148,13 @@ class FilterSegmentsByAlignmentConfidenceJob(Job):
 
 
 class FilterCorpusBySegmentsJob(Job):
-    def __init__(self, bliss_corpus, segment_file, compressed=False, invert_match=False):
+    def __init__(self, bliss_corpus, segment_file, compressed=False, invert_match=False, delete_empty_recordings=False):
         """
         :param Path bliss_corpus:
         :param list[Path]|Path segment_file: a single segment file or a list of segment files
         :param bool compressed:
         :param bool invert_match:
+        :param bool delete_empty_recordings: if true, empty recordings will be removed
         """
         self.bliss_corpus = bliss_corpus
         self.segment_file_list = [segment_file] if isinstance(segment_file, tk.Path) else segment_file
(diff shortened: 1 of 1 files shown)


corpus/filter.py
```diff
@@ -176,12 +180,21 @@ def run(self):
...
                 rec.segments = [x for x in rec.segments if x.fullname() not in segments and x.name not in segments]
             else:
                 rec.segments = [x for x in rec.segments if x.fullname() in segments or x.name in segments]
 
+            if not rec.segments:
+                to_delete.append(idx)
+
+        if self.delete_empty_recordings:
+            for idx in reversed(to_delete):
+                del c.subcorpora[0].recordings[idx]
```
michelwi, line 196: this does only work for the specific corpus structure that we use, but not in general.

I would suggest you use the function [Corpus.remove_recording](https://github.com/rwth-i6/i6_core/blob/db0f0a3b17419b221fdecdf5f8a996ed6b8451e6/lib/corpus.py#L188)

e.g. 
```
if not rec.segments:
  to_delete.append(rec)

[..]
for rec in to_delete:
  c.remove_recording(rec)
```
and then remove the `idx` and `enumerate`
(not tested)

michelwi (changes requested):

corpus/filter.py
hannah220: Thanks for the comment!
I used `remove_recording` like your snippet (also changed to pass a Recording obj and not idx). 
And run a test on a `/nas/data/speech/FR_FR/16kHz/commonvoice/corpus/other.v1/commonvoice-other.v1.xml.gz` corpus, which seems to work.

michelwi (approved):
Thanks for actually testing this. 👍 

Discussion question: do we want to add a warning ala `f"Discarding segment {rec.name} because it is empty"` to the log?

hannah220:
We could, but there can be a lot of warnings generated. 
For example, for the tested corpus there were 27403 deleted recordings.

christophmluscher:
I think this should be tracked in some form.. maybe we can create an additional log file `removed_recordings.log` or similar?

hannah220:
An additional log file is a good idea:+1:

corpus/filter.py
```diff
@@ -148,18 +148,23 @@ def run(self):
...
         :param bool invert_match:
+        :param bool delete_empty_recordings: if true, empty recordings will be removed
         """
         self.bliss_corpus = bliss_corpus
         self.segment_file_list = [segment_file] if isinstance(segment_file, tk.Path) else segment_file
         self.invert_match = invert_match
+        self.delete_empty_recordings = delete_empty_recordings
 
         self.out_corpus = self.output_path("corpus.xml" + (".gz" if compressed else ""))
+        self.out_removed_recordings = self.output_path("removed_recordings.log")
```
michelwi, line 167: This output is always created, but the actual file is only created if `delete_empty_recordings` is set to true.

This may lead to errors if some Job depends on the output, the `FilterCorpusBySegmentsJob` is finished - so sisyphus assumes the file is there - but the actual file on disk is not present.

So I propose either to only create the output if the option is set
```suggestion
        if delete_empty_recordings:
            self.out_removed_recordings = self.output_path("removed_recordings.log")
```

or to create the file in any case. (It will be empty then)

michelwi (changes requested):

JackTemaki (approved):

PR #424, opened 2023-06-22
explicitly set *_NUM_THREADS in Returnn run

JackTemaki:
now that we are using e.g. Apptainer and no external run shell scripts (launcher), I think it is better when the Job itself already sets the variables correctly to the amount of reserved CPU threads. In case you are using a launcher script, you can still overwrite these values as before.

### DIFF
--- a/returnn/extract_prior.py
+++ b/returnn/extract_prior.py
@@ -163,7 +163,10 @@ class ReturnnComputePriorJobV2(Job):
 
     def run(self):
         cmd = self._get_run_cmd()
-        sp.check_call(cmd)
+        env = os.environ.copy()
+        env["OMP_NUM_THREADS"] = str(self.rqmt["cpu"])
+        env["MKL_NUM_THREADS"] = str(self.rqmt["cpu"])
+        sp.check_call(cmd, env=env)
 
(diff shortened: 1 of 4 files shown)


michelwi:
Should this not rather be handled by sisyphus environment management logic? 

Mh.. I just checked, and sisyphus does not expose the `EnvironmentModifier` that lives in `some_job._sis_environment`. In other recipes I have seen that the `gs.DEFAULT_ENVIRONMENT_SET` are modified just before creation of the job, but this is also not very nice.

More nice would be, if the `Job` class got some means to pass some `extra_env: dict` from the constructor to its environment modifier.. But I can also accept your solution.


Do we need something similar for RASR jobs? I remember we had some issues with `OMP_NUM_THREADS` there as well..

JackTemaki:
The AdvancedTreeSearchJob already has that:

```python
extra_code = "export OMP_NUM_THREADS={0}\nexport TF_DEVICE='{1}'".format(
            math.ceil(self.cpu / 2), "gpu" if self.use_gpu else "cpu"
        )
 ```
 
 We have no clear way of managing the environment dynamically. So via Sisyphus we set env vars that are important for the whole environment, but here this is very job specific. I also not like it too much to do it like this, but it is better than nothing and does not hurt.

michelwi (approved):

PR #425, opened 2023-06-27
returnn init: add search

vieting:
In a number of setups, we do something like
```
import i6_core.returnn as returnn

config = returnn.ReturnnConfig(...)
train_job = returnn.ReturnnTrainingJob(...)
```
I'd like to do the same for the search jobs, so added search to the init file.


JackTemaki (approved):

vieting:
Thanks for the quick approval!

PR #426, opened 2023-06-27
Add option for external lmgc scorer

Atticus1806:
This adds the option to add a dummy scorer as external input to the LMGC Job. Similar to the alias update in #419 this should not change hashes. 

Since the scorer is not used but still required it can be fed with smth. like: 
```
acoustic_mixture_path = CreateDummyMixturesJob(
    num_mixtures=returnn_config.config['extern_data']['classes']['dim'],
    num_features=returnn_config.config['extern_data']['data']['dim']).out_mixtures
lmgc_scorer = rasr.GMMFeatureScorer(acoustic_mixture_path)
```

### DIFF
--- a/recognition/advanced_tree_search.py
+++ b/recognition/advanced_tree_search.py
@@ -146,7 +146,7 @@ class AdvancedTreeSearchLmImageAndGlobalCacheJob(rasr.RasrCommand, Job):
     def hash(cls, kwargs):
         config, post_config, num_images = cls.create_config(**kwargs)
         return super().hash(
-            {"config": config, "exe": kwargs["crp"].flf_tool_exe, "post_config": post_config},
+            {"config": config, "exe": kwargs["crp"].flf_tool_exe},
         )
 
 
--- a/recognition/advanced_tree_search.py
+++ b/recognition/advanced_tree_search.py
@@ -145,7 +145,9 @@ class AdvancedTreeSearchLmImageAndGlobalCacheJob(rasr.RasrCommand, Job):
     @classmethod
     def hash(cls, kwargs):
         config, post_config, num_images = cls.create_config(**kwargs)
-        return super().hash({"config": config, "exe": kwargs["crp"].flf_tool_exe})
+        return super().hash(
+            {"config": config, "exe": kwargs["crp"].flf_tool_exe, "post_config": post_config},
+        )
 
 
 class AdvancedTreeSearchJob(rasr.RasrCommand, Job):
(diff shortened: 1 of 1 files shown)


Atticus1806:
>should not change hashes

:(

@michelwi you know what went wrong?

recognition/advanced_tree_search.py
```diff
@@ -145,7 +145,9 @@ def create_config(cls, crp, feature_scorer, extra_config, extra_post_config, **k
     @classmethod
     def hash(cls, kwargs):
         config, post_config, num_images = cls.create_config(**kwargs)
-        return super().hash({"config": config, "exe": kwargs["crp"].flf_tool_exe})
+        return super().hash(
+            {"config": config, "exe": kwargs["crp"].flf_tool_exe, "post_config": post_config},
```
michelwi, line 149: You add the post_config to the things that should be hashed?? This changes the hashes of all the setups that use this Job.

Pls. revert this line.

Atticus1806: Whoops this was a leftover from testing.

michelwi (approved):

recognition/advanced_tree_search.py
```diff
@@ -283,13 +285,16 @@ def create_config(
         cpu: int,
         lmgc_mem: float,
         lmgc_alias: Union[None, str],
+        lmgc_scorer: Union[None, rasr.FeatureScorer],
```
christophmluscher, line 288: What is the difference to `Optional[rasr.FeatureScorer]`? Just curious.

christophmluscher (approved):

recognition/advanced_tree_search.py
Atticus1806: My thinking was: Optional can mean "You can give it, but you do not have to, and then its None"
Here it is always given (due to the call and no default)

Atticus1806: `Optional[...] is a shorthand notation for Union[..., None]` but it seems this was wrong thinking from my side, so I can change it if you want.

christophmluscher: I agree with Willi, IIRC in the surrounding code this is `Union` but in `i6_*` repos it is mainly `Optional`... I would at least keep this per file consistent.. :)

Atticus1806: I hope I didn't overlook anything.

JackTemaki (approved):

PR #429, opened 2023-07-17
Add OnnxExportJob

JackTemaki:
A wrapper around the Returnn-Tool. Be aware that I never tested a "normal" pipeline with this, I just used some dummy `torch_export_to_onnx.py` script to verify this jobs works.

### DIFF
--- a/returnn/compile.py
+++ b/returnn/compile.py
@@ -216,6 +216,7 @@ class OnnxExportJob(Job):
 
     def __init__(
         self,
+        *,
         returnn_config: ReturnnConfig,
         checkpoint: Union[Checkpoint, PtCheckpoint],
         device: str = "cpu",
--- a/returnn/compile.py
+++ b/returnn/compile.py
@@ -209,8 +209,11 @@ class CompileNativeOpJob(Job):
 
 class OnnxExportJob(Job):
     """
-    
+        Export an ONNX model using the appropriate RETURNN tool script.
+
+        Currently only supports PyTorch via tools/torch_export_to_onnx.py
     """
+
     def __init__(
         self,
         returnn_config: ReturnnConfig,
--- a/returnn/compile.py
+++ b/returnn/compile.py
@@ -1,4 +1,4 @@
-__all__ = ["CompileTFGraphJob", "CompileNativeOpJob"]
+__all__ = ["CompileTFGraphJob", "CompileNativeOpJob", "OnnxExportJob"]
 
 from sisyphus import *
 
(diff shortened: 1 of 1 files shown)


michelwi (commented):
No strong objections. Only if it is not planned to include TF model export soon, then we might as well have a Pytorch specific job here.

returnn/compile.py
```diff
@@ -203,3 +205,62 @@ def run(self):
...
+
+class OnnxExportJob(Job):
+    """
+        Export an ONNX model using the appropriate RETURNN tool script.
+
+        Currently only supports PyTorch via tools/torch_export_to_onnx.py
+    """
+
+    def __init__(
+        self,
```
albertz, line 218:
```suggestion
        self,
        *,
```

Atticus1806 (approved):

SimBe195 (approved):

albertz:
> However, the RETURNN tool script calls this function without arguments: https://github.com/rwth-i6/returnn/blob/master/tools/torch_export_to_onnx.py#L174C5-L174C28
> 
> @albertz Is there a convention that `epoch` and `step` should always be given? Then the tool script should be updated.

Oh right, this is an error. I will fix this.

vieting:
Following this: Do we need to set `epoch` here in the job? There might be cases where the checkpoint does not match the model that is created because the export script then uses `epoch=1` by default.

returnn/compile.py
```diff
@@ -203,3 +205,63 @@ def run(self):
...
+        *,
+        returnn_config: ReturnnConfig,
+        checkpoint: Union[Checkpoint, PtCheckpoint],
```
albertz, line 221: Now that I fixed the issue raised by @vieting about `get_model` with `epoch` and `step`, I realize that those should also be arguments here, because the generated model can be dependent on it.

We might want to use those defaults. In many cases, I assume that `get_model` does actually not use it, so such defaults are fine then?

Or do we want to automatically infer it from the checkpoint somehow? But this was based on heuristics, and not sure if this is wanted.

```suggestion
        checkpoint: Union[Checkpoint, PtCheckpoint],
        epoch: int = 1,
        step: int = 0,
```

albertz:
Btw, Black fails.

returnn/compile.py
JackTemaki: >  Or do we want to automatically infer it from the checkpoint somehow? 

Why infer? we store this very explicitly in the checkpoints already, see:
https://github.com/rwth-i6/returnn/blob/master/returnn/torch/engine.py#L613

JackTemaki: So I am against having those as extra parameters for the job.

albertz: Ah yes I forgot. Then I will update the script again. Then it does not need those arguments.

vieting:
Please add the job to `__all__`.

returnn/compile.py
albertz: Done.

albertz (changes requested):
Same as @michelwi.

returnn/compile.py
```diff
@@ -9,10 +9,12 @@
 import os
 import shutil
 import subprocess as sp
+from typing import Optional, Union
```
albertz, line 12:
```suggestion
from typing import Optional
```

PR #431, opened 2023-07-19
Make tf flow functions usable for onnx

vieting:
To set up a recognition using RASR's `tensorflow-forward` flow node, we have helper functions in `returnn.flow.py`. I'd like to generalize them to also support the `onnx-forward` node.

This involves generalizing the existing function to create the precomputed feature flow and adding separate functions for tf and onnx with the specific needs for their configs. The function to integrate it into the base flow is basically just renamed and a version keeping the tf behavior is added.

### DIFF
--- a/returnn/flow.py
+++ b/returnn/flow.py
@@ -9,6 +9,7 @@ from i6_core.returnn.training import Checkpoint
 def make_precomputed_hybrid_feature_flow(
     backend: str,
     rasr_config: rasr.RasrConfig,
+    rasr_post_config: Optional[rasr.RasrConfig] = None,
     fwd_input_name: str = "fwd-input",
 ) -> rasr.FlowNetwork:
     """
@@ -23,6 +24,7 @@ def make_precomputed_hybrid_feature_flow(
 
     :param backend: "tf" or "onnx"
     :param rasr_config: rasr config for the forward node
+    :param rasr_post_config: rasr post config (not hashed) for the forward node
     :param fwd_input_name: naming for the tf network input, usually no need to be changed
     :return: tensorflow-/onnx-forward node flow with output link and related config
     """
@@ -40,6 +42,9 @@ def make_precomputed_hybrid_feature_flow(
 
     flow.config = rasr.RasrConfig()
     flow.config[fwd_node] = rasr_config
+    if rasr_post_config is not None:
+        flow.post_config = rasr.RasrConfig()
+        flow.post_config[fwd_node] = rasr_post_config
 
     return flow
 
@@ -138,16 +143,18 @@ def make_precomputed_hybrid_onnx_feature_flow(
     """
 
     rasr_config = rasr.RasrConfig()
+    rasr_post_config = rasr.RasrConfig()
     for k, v in io_map.items():
         rasr_config.io_map[k] = v
 
     rasr_config.session.file = onnx_model
-    rasr_config.session.inter_op_num_threads = cpu
-    rasr_config.session.intra_op_num_threads = cpu
+    rasr_post_config.session.inter_op_num_threads = cpu
+    rasr_post_config.session.intra_op_num_threads = cpu
 
     return make_precomputed_hybrid_feature_flow(
         backend="onnx",
         rasr_config=rasr_config,
+        rasr_post_config=rasr_post_config,
         fwd_input_name=onnx_fwd_input_name,
     )
 
--- a/returnn/flow.py
+++ b/returnn/flow.py
@@ -35,6 +35,7 @@ def make_precomputed_hybrid_feature_flow(
     flow.add_output("features")
     flow.add_param("id")
 
+    assert backend in ("tf", "onnx"), f"backend not supported: {backend}"
     node_filter = {"tf": "tensorflow-forward", "onnx": "onnx-forward"}[backend]
     fwd_node = flow.add_node(node_filter, f"{backend}-fwd", {"id": "$(id)"})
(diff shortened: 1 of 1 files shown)


vieting:
This will be used e.g. in https://github.com/rwth-i6/i6_experiments/pull/155.

returnn/flow.py
```diff
@@ -52,66 +83,112 @@ def make_precomputed_hybrid_tf_feature_flow(
...
+    :param cpu: number of CPUs to use
+    :return: onnx-forward node flow with output link and related config
+    """
+
+    rasr_config = rasr.RasrConfig()
+    for k, v in io_map.items():
+        rasr_config.io_map[k] = v
+
+    rasr_config.session.file = onnx_model
+    rasr_config.session.inter_op_num_threads = cpu
```
JackTemaki, line 145: This should better be in a non-hashed "post-config", as it has no influence on the result but is purely about speed optimization.

vieting: Do we already have post configs for feature flows somewhere?

JackTemaki: I do not think so, but I checked the FlowNetwork class and it seems to support post configs.

vieting: Sounds like we need a separate PR for that first?

vieting:
The number of threads is now moved to the post config and the hashing issue for flow post configs (#436) is resolved.

JackTemaki (approved):
Recognition with this worked fine, also from the code I see no issues. Potential TF changes should be covered by the AppTek-tests (@michelwi ?), so I guess this is all good.

michelwi:
> Potential TF changes should be covered by the AppTek-tests.

No, we do not use this function.

vieting:
I tested the changes for one of my tf precomputed hybrid setups.

returnn/flow.py
```diff
@@ -52,66 +88,114 @@ def make_precomputed_hybrid_tf_feature_flow(
...
+        fwd_input_name=onnx_fwd_input_name,
+    )
+
+
+def add_fwd_flow_to_base_flow(
     base_flow: rasr.FlowNetwork,
-    tf_flow: rasr.FlowNetwork,
-    tf_fwd_input_name: str = "tf-fwd-input",
+    fwd_flow: rasr.FlowNetwork,
+    fwd_input_name: str = "fwd-input",
```
michelwi, line 165: The default name here is `"fwd-input"`. 
The default names for the `make_precomputed_hybrid_tf_feature_flow` and `make_precomputed_hybrid_onnx_feature_flow` are `"tf-fwd-input"` and `"onnx-fwd-input"` respectively. 

These names always overwrite the default of `make_precomputed_hybrid_feature_flow` (i.e. `"fwd-input"`) and I don't believe that one can use `make_precomputed_hybrid_feature_flow` standalone, because the config will be missing.

I conclude, that the default naming parameters will never match. Or am I missing something?

As the name will anyway never be seen and is arbitrary, I would propose to make your lives easier and just call it `"fwd-input"` in all the cases (except for the legacy `add_tf_flow_to_base_flow`)

vieting: Yes, for onnx we can just name it `"fwd-input". For `add_tf_flow_to_base_flow` we should keep the old name in order not to break the hashes. That's what you propose, right?

returnn/flow.py
```diff
@@ -1,11 +1,54 @@
...
+    Create the feature flow for a simple TF/ONNX network that predicts frame-wise outputs, to be used
+    in combination with the `nn-precomputed-hybrid` feature-scorer setting in RASR.
+
+    The resulting flow is a trivial (for ONNX, the "tf" is replaced by "onnx"):
+
+        <link from="<tf_fwd_input_name>" to="tf-fwd:input"/>
+        <node name="tf-fwd" id="$(id)" filter="tensorflow-forward"/>
+        <link from="tf-fwd:log-posteriors" to="network:features"/>
+
+    :param backend: "tf" or "onnx"
```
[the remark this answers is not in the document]
vieting: I added an assertion so that the error message is even clearer.

returnn/flow.py
michelwi: mh.. would there be a way to auto-detect it? e.g. 
```
input_name = list(fwd_flow.inputs)[0]
```
(I think the flow has just a single input)

vieting: I'm not sure if this is always guaranteed. In the way it is now, the user also never needs to set it. Do you still prefer auto-detecting it?

michelwi: >  In the way it is now, the user also never needs to set it.

if they use the new  `add_fwd_flow_to_base_flow` with TF, then they still have to set it, right?

> I'm not sure if this is always guaranteed. Do you still prefer auto-detecting it?

I do not have a strong opinion and would let that decision to the people that will use this functions.

vieting: > > In the way it is now, the user also never needs to set it.
>
> if they use the new add_fwd_flow_to_base_flow with TF, then they still have to set it, right?

Depends. First, this function is used to avoid code duplications between tf and onnx. Using it independently of either of them is currently no use case. Second, if (as discussed in the other comment) we make `"fwd-input"` the default for everything that isn't tf, the user can always stick to the default here.

christophmluscher (approved):

returnn/flow.py
vieting: @michelwi do you want changes here? Otherwise we could merge.

vieting: Thanks!

PR #432, opened 2023-07-25
add GetBestPtCheckpointJob

vieting:
We have the `GetBestTFCheckpointJob` and I'd like to add the equivalent for torch checkpoints. I added a very similar `GetBestPtCheckpointJob` since this does not interfere with existing setups. We might also think about how to avoid some of the code duplications, but this seems to be the simplest solution.

### DIFF
--- a/returnn/training.py
+++ b/returnn/training.py
@@ -3,6 +3,7 @@ __all__ = [
     "Checkpoint",
     "GetBestEpochJob",
     "GetBestTFCheckpointJob",
+    "GetBestPtCheckpointJob",
     "PtCheckpoint",
     "ReturnnModel",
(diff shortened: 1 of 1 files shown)


JackTemaki (changes requested):
I think for PyTorch we can simplify this a little bit more

JackTemaki (changes requested):
I think for PyTorch we can simplify this a little bit more

vieting:
> I think for PyTorch we can simplify this a little bit more

Right. In this case, the code duplication is also reduced, so we can just add the torch specific job imho.

JackTemaki (approved):

michelwi (approved):

PR #433, opened 2023-07-26
Add rounding scheme rasr_compatible

michelwi:
I found that RASR and the `BlissToPcmHDFJob` use different rounding methods when the start/end time in the corpus does not coincide _exactly_ with a sample. This might in a few cases lead to a different number of frames for features and alignments and break the training.

This implements a rounding scheme that I believe is compatible with RASR.

### DIFF
--- a/returnn/hdf.py
+++ b/returnn/hdf.py
@@ -237,7 +237,7 @@ class BlissToPcmHDFJob(Job):
         :param returnn_root: RETURNN repository
         :param rounding: defines how timestamps should be rounded if they do not exactly fall onto a sample:
             start_and_duration will round down the start time and the duration of the segment
-            rasr_compatible will round up start and round down the end
+            rasr_compatible will round up the start time and round down the end time
         """
         self.set_vis_name("Dump audio to HDF")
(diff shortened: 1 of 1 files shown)


returnn/hdf.py
```diff
@@ -228,6 +235,9 @@ def __init__(
             BaseStrategy(): no handling, assume only one channel
             PickNth(n): Takes audio from n-th channel
         :param returnn_root: RETURNN repository
+        :param rounding: defines how timestamps should be rounded if they do not exactly fall onto a sample:
+            start_and_duration will round down the start time and the duration of the segment
+            rasr_compatible will round up start and round down the end
```
vieting, line 240:
```suggestion
            rasr_compatible will round up the start time and round down the end time
```

vieting (approved):

Icemole (approved):
Hello! @michelwi and I tested for mismatches between the features generated by this PR and the alignments generated by RASR, and I can confirm that everything seems to be fine now! :slightly_smiling_face:

christophmluscher (approved):

PR #434, opened 2023-07-26
Add support for encoding and sparse data in RasrAlignmentDumpHDFJob

michelwi:
Three independent changes:
* Support for sparse data: the Job currently writes out alignments in (B, T, 1) but some datasets expect (B, T). This PR adds the option `sparse` that stores the data in sparse format if enabled.
* Support encoding: In case the segment names contain non-ascii characters, we want to specify a different encoding (e.g. "utf-8"). This is already supported in RASR, but the `lib.rasr_cache.FileArchive` needed a bit of enabling.
* added a `filter_list_keep` so that we can directly remove unwanted segments from the hdf.

### DIFF
--- a/lib/rasr_cache.py
+++ b/lib/rasr_cache.py
@@ -13,7 +13,6 @@ import logging
 import mmap
 import numpy
 import os
-import sys
 import typing
 import zlib
 from struct import pack, unpack
@@ -51,7 +50,8 @@ class FileArchive:
     start_recovery_tag = 0xAA55AA55
     end_recovery_tag = 0x55AA55AA
 
-    def __init__(self, filename, must_exists=False):
+    def __init__(self, filename, must_exists=False, encoding="ascii"):
+        self.encoding = encoding
 
         self.ft = {}  # type: typing.Dict[str,FileInfo]
         if os.path.exists(filename):
@@ -182,12 +182,12 @@ class FileArchive:
         return res
 
     # write routines
-    def write_str(self, s):
+    def write_str(self, s, enc="ascii"):
         """
         :param str s:
         :rtype: int
         """
-        return self.f.write(pack("%ds" % len(s), s.encode("ascii")))
+        return self.f.write(pack("%ds" % len(s.encode(enc)), s.encode(enc)))
 
     def write_char(self, i):
         """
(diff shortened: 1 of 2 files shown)


michelwi:
@albertz returnn contains the same [`FileArchive`](https://github.com/rwth-i6/returnn/blob/da58e5fd495e67215d216c8b92d63f210e97b0bb/returnn/sprint/cache.py#L44) class in `sprint/cache.py`. Would you be interested to keep this in sync?

michelwi:
> But why do we actually have the class here at all when we import other things from RETURNN anyway, e.g. via `get_returnn_simple_hdf_writer`? Why not just import this class here as well from RETURNN?

I do not know why the decision was made to copy the code in one instance and to import it in another instance.

> Yes it would be good to have them in sync.

feel free to review the changes here, I will open a PR with the final version after this is finished.

JackTemaki:
> > But why do we actually have the class here at all when we import other things from RETURNN anyway, e.g. via `get_returnn_simple_hdf_writer`? Why not just import this class here as well from RETURNN?
> 
> I do not know why the decision was made to copy the code in one instance and to import it in another instance.

The sprint cache code is extremely old, and existed in the repo before we even started `i6_core`, and is not directly related to RETURNN (not sure when or how the duplication started, this was way before my time in the ASR team, so there never was an active "decision" on that). The SimpleHDFWriter was introduced by Returnn, and there I did not want to make a copy to avoid discussions, so I used this import function.

JackTemaki (approved):
I could not find any obvious issues, looks good to me.

lib/rasr_cache.py
```diff
@@ -542,8 +542,8 @@ def addAttributes(self, filename, dim, duration):
         ) % (dim, duration)
         self.write_U32(self.start_recovery_tag)
         filename = "%s.attribs" % filename
-        self.write_u32(len(filename))
-        self.write_str(filename)
+        self.write_u32(len(filename.encode(self.encoding)))
```
christophmluscher, line 545: Two lines above, `write_U32` here `write_u32` is there a difference? Or can this be more consistent?

christophmluscher: OK found the functions.. `i` vs `I`???

christophmluscher: `int` or `unsigned int`

christophmluscher (approved):

michelwi:
We are doing a few more tests on our side and I would merge once these finish successfully.

Nintendo-Wei (approved):
It worked with my PT_BR system.

Issue #436, opened 2023-07-27
Wrong hashing in Flow


JackTemaki:
During https://github.com/rwth-i6/i6_core/pull/431 we found that `FlowNetworks` have incorrect hashing of the post_config entry:

```python
    def __sis_state__(self):
        def get_val(a):
            return get_val(a.get(self)) if isinstance(a, FlowAttribute) else a

        nodes = {name: {k: get_val(v) for k, v in attr.items()} for name, attr in self.nodes.items()}
        state = {
            "name": self.name,
            "nodes": nodes,
            "links": self.links,
            "inputs": self.inputs,
            "outputs": self.outputs,
            "params": self.params,
            "hidden_inputs": self.hidden_inputs,
            "config": self.config,
            "post_config": self.post_config,
        }
        return state
```
(The `post_config` should not be added to the state)

For me it is yet unclear if this can be fixed without breaking things, by e.g. setting `post_config` to `None`. As no one ever detected this bug, we might assume no one ever tried to use the `post_config`.

michelwi:
> As no one ever detected this bug, we might assume no one ever tried to use the post_config.

I would not say that. hash Bugs have very long been present in our setups and nobody noticed until we tried to move one setup and then found the hashes to be different..

> unclear if this can be fixed without breaking things, by e.g. setting post_config to None. 

I do not know any flow that uses a post config, but I do not know all flows by heart.
You can make a PR with this change and then we will notice if any hashes are broken on AppTeks site.

vieting:
> You can make a PR with this change and then we will notice if any hashes are broken on AppTeks site.

The AppTek hash test still passes, see #437.

vieting:
Closed with #437.

PR #437, opened 2023-07-27
FlowNetwork post_config hashing

vieting:
As mentioned in #436, the `post_config` is added to the state which should not be the case.

As suggested by @JackTemaki, I set it to `None` here and we can see if this breaks any hashes in the AppTek pipeline test.


vieting:
The tests pass, so following our offline discussion, I think we can merge this, right?

michelwi (approved):
All tests passed. Looks good to me.

JackTemaki (approved):

PR #439, opened 2023-08-08
Call black binary instead of using module

albertz:
Multiple reasons:

- This is more the intended usage of black.
- This also avoids some warnings we always get when we import black.
- When black formatting fails (e.g. syntax error in your Python epilog or so), it was not writing the file at all, so it was very annoying to debug and see where it actually failed. Now with this, you can simply look at the file.

### DIFF
--- a/returnn/config.py
+++ b/returnn/config.py
@@ -6,6 +6,7 @@ import json
 import os
 import pickle
 import pprint
+import shutil
 import string
 import subprocess
 import sys
@@ -139,6 +140,11 @@ class ReturnnConfig:
         if sys.version_info[:2] >= (3, 8):
             self.pprint_kwargs.setdefault("sort_dicts", sort_config)
         self.black_formatting = black_formatting
+        # Get the path to the black executable here in the Sisyphus environment
+        # because inside the worker environment, we usually don't have the original PATH anymore.
+        # This is consistent to the old behavior where we did `import black` instead,
+        # i.e. by default this used the same black as in the Sisyphus environment.
+        self._black_path = shutil.which("black") if black_formatting else None
 
     def get(self, key, default=None):
         if key in self.post_config:
@@ -188,7 +194,7 @@ class ReturnnConfig:
         with open(file_path, "wt", encoding="utf-8") as f:
             f.write(content)
         if self.black_formatting:
-            subprocess.check_call(["black", file_path])
+            subprocess.check_call([self._black_path, file_path])
 
     def _write_network_stages(self, config_path):
         """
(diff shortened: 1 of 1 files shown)


returnn/config.py
```diff
@@ -139,6 +140,11 @@ def __init__(
         if sys.version_info[:2] >= (3, 8):
             self.pprint_kwargs.setdefault("sort_dicts", sort_config)
         self.black_formatting = black_formatting
+        # Get the path to the black executable here in the Sisyphus environment
+        # because inside the worker environment, we usually don't have the original PATH anymore.
+        # This is consistent to the old behavior where we did `import black` instead,
+        # i.e. by default this used the same black as in the Sisyphus environment.
+        self._black_path = shutil.which("black") if black_formatting else None
```
michelwi, line 147: I am wondering if `shutil.which("black")` in the manager (here) can be different than in a worker process on a different node (i.e. below in `_write...`)

Would it then be better to delay the call to `shutil.which("black")` until we are in the worker?

albertz: But that's actually the real intention here, to do it in the manager, not in the worker. In the worker, I would not need the `which` at all, you can simply do `check_call(["black", ...])`.

The reason is as I write in the comment: In the worker, it can happen that `black` is not in the `PATH` anymore. Also, the behavior right now is that the worker runs via `gs.SIS_COMMAND`, which is by default the same Python as the manager, so by default currently you get the Black from the manager environment.

michelwi: But in the old approach we always use `import black` and then use the black that we get there. So in the worker, always a black module must be in the pythonpath and always the black that was present on the worker would be used.

I can see that `shutil.which("black")` would fail if the `PATH` is altered. But the way it is implemented now it fails if the environment changes from manager to worker (i.e. if `self._black_path` of the manager is not available on the worker node)

albertz: > But in the old approach we always use `import black` and then use the black that we get there. So in the worker, always a black module must be in the pythonpath and always the black that was present on the worker would be used.

The worker uses `gs.SIS_COMMAND`. The default is `SIS_COMMAND = [sys.executable, sys.argv[0]]`. `sys.executable` is the Python executable from the Sisyphus manager. So it means, the Python in the worker is specifically explicitly set to be the Python from the manager. Any scripts executed by the worker might use a different Python, but the Python code of the tasks itself (e.g. what you have in `run()` or `create_files()`) is run with the same Python environment as the manager. So the `import black` is the same as in the manager.

> But the way it is implemented now it fails if the environment changes from manager to worker (i.e. if `self._black_path` of the manager is not available on the worker node)

When is this ever not the case? We expect the same filesystem, or not? There is lots of other code (checks via `os.path.exists` etc) which has this expectation that we have the same filesystem.

albertz:
@michelwi @JackTemaki @vieting @curufinwe @mmz33 What is the state here?

albertz:
@michelwi @JackTemaki @vieting @curufinwe @mmz33 ping.

albertz:
@michelwi @JackTemaki @vieting @curufinwe @mmz33 ping.

michelwi (approved):

christophmluscher (approved):

PR #442, opened 2023-08-11
ScliteJob, precision_ndigit option

albertz:
Allows to configure the number of digits after decimal point for the precision of the percentages in the output variables.

In sclite, the precision was always one digit after the decimal point  (https://github.com/usnistgov/SCTK/blob/f48376a203ab17f/src/sclite/sc_dtl.c#L343), thus we recalculate the percentages here.

### DIFF
--- a/recognition/scoring.py
+++ b/recognition/scoring.py
@@ -194,6 +194,7 @@ class ScliteJob(Job):
                     if key == "Aligned words":
                         break  # that should be the last key, can stop now
 
+                assert "Ref. words" in outputs_absolute, "Expected absolute numbers for Ref. words"
                 num_ref_words = outputs_absolute["Ref. words"]
                 assert "Percent Total Error" in outputs_absolute, "Expected absolute numbers for Percent Total Error"
                 outputs_absolute["Percent Word Accuracy"] = num_ref_words - outputs_absolute["Percent Total Error"]
--- a/recognition/scoring.py
+++ b/recognition/scoring.py
@@ -191,6 +191,8 @@ class ScliteJob(Job):
                         assert not output_variables[key][1], f"Expected absolute value for {key}"
                         continue
                     outputs_absolute[key] = int(absolute_s)
+                    if key == "Aligned words":
+                        break  # that should be the last key, can stop now
 
                 num_ref_words = outputs_absolute["Ref. words"]
                 assert "Percent Total Error" in outputs_absolute, "Expected absolute numbers for Percent Total Error"
--- a/recognition/scoring.py
+++ b/recognition/scoring.py
@@ -12,7 +12,7 @@ import subprocess as sp
 import tempfile
 import collections
 import re
-from typing import List, Optional
+from typing import List, Optional, Dict, Tuple
 
 from sisyphus import *
 from i6_core.lib.corpus import *
@@ -51,7 +51,7 @@ class ScliteJob(Job):
         - out_*: the job also outputs many variables, please look in the init code for a list
     """
 
-    __sis_hash_exclude__ = {"sctk_binary_path": None}
+    __sis_hash_exclude__ = {"sctk_binary_path": None, "precision_ndigit": 1}
 
     def __init__(
         self,
@@ -61,6 +61,7 @@ class ScliteJob(Job):
         sort_files: bool = False,
         additional_args: Optional[List[str]] = None,
         sctk_binary_path: Optional[tk.Path] = None,
+        precision_ndigit: int = 1,
     ):
         """
         :param ref: reference stm text file
@@ -69,6 +70,11 @@ class ScliteJob(Job):
         :param sort_files: sort ctm and stm before scoring
         :param additional_args: additional command line arguments passed to the Sclite binary call
         :param sctk_binary_path: set an explicit binary path.
+        :param precision_ndigit: number of digits after decimal point for the precision
+            of the percentages in the output variables.
+            In sclite, the precision was always one digit after the decimal point
+            (https://github.com/usnistgov/SCTK/blob/f48376a203ab17f/src/sclite/sc_dtl.c#L343),
+            thus we recalculate the percentages here.
         """
         self.set_vis_name("Sclite - %s" % ("CER" if cer else "WER"))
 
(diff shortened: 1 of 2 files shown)


Atticus1806:
In general I think this looks good. I am wondering if for this type of job we would want a small testcase. While I think e.g. the regex looks correct, I feel like its easier to verify with a test. And also in general for future possible changes to this job I feel the job is crucial enough to be tested.

albertz:
@christophmluscher @michelwi @JackTemaki @vieting @curufinwe @mmz33 What is the state here?

@Atticus1806 So are you saying such test should be part of this PR, or this is a generic comment which we should do independently of this? What do the other think? Maybe you can work on the test?

Atticus1806:
Whoops. I started writing some simple example, but then forgot to continue. I can make a PR today, then we can merge that quickly and then test it on this one if you agree.

recognition/scoring.py
```diff
@@ -149,31 +156,61 @@ def run(self, output_to_report_dir=True):
...
+                    "Percent Deletions": (self.out_percent_deletions, self.out_num_deletions),
+                    "Percent Insertions": (self.out_percent_insertions, self.out_num_insertions),
+                    "Percent Word Accuracy": (self.out_percent_word_accuracy, None),
+                    "Ref. words": (None, self.out_ref_words),
+                    "Hyp. words": (None, self.out_hyp_words),
+                    "Aligned words": (None, self.out_aligned_words),
+                }
+
+                outputs_absolute: Dict[str, int] = {}
                 for line in f:
```
vieting, line 182: Right now, we always go through all lines, right? We might introduce an early stop (e.g. if the line starts with `"CONFUSION PAIRS"`) because that can be thousands of lines while we are only interested in the first 30 or so.

recognition/scoring.py
```diff
@@ -149,31 +156,61 @@ def run(self, output_to_report_dir=True):
...
+                    m = re.match(pattern, line)
+                    assert m, f"Could not parse line: {line!r}, does not match to pattern r'{pattern}'"
+                    absolute_s = m.group(4)
+                    if not absolute_s:
+                        assert not output_variables[key][1], f"Expected absolute value for {key}"
+                        continue
+                    outputs_absolute[key] = int(absolute_s)
+
+                num_ref_words = outputs_absolute["Ref. words"]
+                assert "Percent Total Error" in outputs_absolute, "Expected absolute numbers for Percent Total Error"
```
vieting, line 196: This assertion is done for `"Percent Total Error"`, but not `"Ref. words"` above. Is there a reason to treat them differently?

vieting (commented):
Generally, the PR looks good to me. Having the test by @Atticus1806 would definitely be nice.

recognition/scoring.py
albertz: Yes. I did not change this logic, but I can change that. `Aligned words` should be the last entry.

recognition/scoring.py
albertz: Not really, I can add that.

albertz:
So #443 is merged now. I rebased this here. And I extended the test to test with custom precision.

@christophmluscher @michelwi @JackTemaki @vieting @curufinwe @mmz33 @Atticus1806 all good now?

tests/job_tests/recognition/test_scoring.py
```diff
@@ -88,3 +88,26 @@ def test_sclite_job():
...
             sclite_job.out_aligned_words.get()
         )
+
+        # Now test custom precision.
+
+        sclite_job = ScliteJob(ref=ref, hyp=hyp, sctk_binary_path=sctk_binary, precision_ndigit=2)
+        sclite_job._sis_setup_directory()
+        sclite_job.run()
+
+        assert sclite_job.out_wer.get() == 58.82, "Wrong WER, %s instead of 58.8" % str(sclite_job.out_wer.get())
```
vieting, line 98:
```suggestion
        assert sclite_job.out_wer.get() == 58.82, "Wrong WER, %s instead of 58.82" % str(sclite_job.out_wer.get())

```

tests/job_tests/recognition/test_scoring.py
```diff
@@ -88,3 +88,26 @@ def test_sclite_job():
...
+        assert sclite_job.out_percent_correct.get() == 47.06, "Wrong percent correct, %s instead of 47.1" % str(
+            sclite_job.out_percent_correct.get()
+        )
+        assert (
+            sclite_job.out_percent_substitution.get() == 41.18
+        ), "Wrong percent substitution, %s instead of 41.2" % str(sclite_job.out_percent_substitution.get())
```
vieting, line 104:
```suggestion
        ), "Wrong percent substitution, %s instead of 41.18" % str(sclite_job.out_percent_substitution.get())

```

tests/job_tests/recognition/test_scoring.py
```diff
@@ -88,3 +88,26 @@ def test_sclite_job():
...
+        assert sclite_job.out_percent_deletions.get() == 11.76, "Wrong percent deletions, %s instead of 11.8" % str(
```
vieting, line 105:
```suggestion
        assert sclite_job.out_percent_deletions.get() == 11.76, "Wrong percent deletions, %s instead of 11.76" % str(

```

tests/job_tests/recognition/test_scoring.py
```diff
@@ -88,3 +88,26 @@ def test_sclite_job():
...
+            sclite_job.out_percent_deletions.get()
+        )
+        assert sclite_job.out_percent_insertions.get() == 5.88, "Wrong percent insertions, %s instead of 4.5" % str(
```
vieting, line 108:
```suggestion
        assert sclite_job.out_percent_insertions.get() == 5.88, "Wrong percent insertions, %s instead of 5.88" % str(
```
This one is really off?

tests/job_tests/recognition/test_scoring.py
```diff
@@ -88,3 +88,26 @@ def test_sclite_job():
...
+            sclite_job.out_percent_insertions.get()
+        )
+        assert (
+            sclite_job.out_percent_word_accuracy.get() == 41.18
+        ), "Wrong percent word accuracy, %s instead of 41.2" % str(sclite_job.out_percent_word_accuracy.get())
```
vieting, line 113:
```suggestion
        ), "Wrong percent word accuracy, %s instead of 41.18" % str(sclite_job.out_percent_word_accuracy.get())

```

vieting (approved):
I'd just reflect the precision in the test error messages as well.

PR #443, opened 2023-08-16
Add test for sclite

Atticus1806:
This adds a test for the `ScliteJob` as a prequisite for #442.


Atticus1806:
Currently there is one problem with the test: 

For some reason the last line of the references is seen as a sentence (c.f. test log):

```
    Performing alignments for file ';;'.                                                                                                                                                                                   1 of 1 Segments For Channel label.
```
Does someone have an idea how this happened? I copied from a real job where I found this last line too.

tests/job_tests/recognition/files/ref.stm
```diff
@@ -0,0 +1,5 @@
+File 1 Test 00.00 04.00 <d0> this one is correct
+File 1 Test 06.00 12.00 <d0> this one is close to correct
+File 1 Test 13.00 19.00 <d0> this one is sadly completely wrong
```
vieting, line 3: Nitpick: This here is the ref, so it's always correct. Maybe it's easier to grasp the test if this is reflected by the ref transcription.

tests/job_tests/recognition/test_scoring.py
```diff
@@ -0,0 +1,109 @@
...
+from sisyphus import tk, setup_path
+
+from i6_core.tools.compile import MakeJob
+from i6_core.tools.git import CloneGitRepositoryJob
+from i6_core.recognition import ScliteJob
+
+Path = setup_path(__package__)
+
+
+def compile_sctk(
```
vieting, line 14: This is taken from `i6_experiments/common/tools`, right? We might move it to i6_core, but then again, the effort is maybe unnecessary just for this test.

Atticus1806: Yes (though slightly modified for the test).
I think for now it should stay there.

tests/job_tests/recognition/files/ref.stm
Atticus1806: Updated. Do you know why there might be the error with the last line not being recognized as comment?

tests/job_tests/recognition/files/hyp.ctm
```diff
@@ -0,0 +1,23 @@
...
+;; TestCorpus/File/3 (13.00-19.00)
+File 1 13.00 01.00 no 1.0000
+File 1 14.00 01.00 clue 1.0000
+File 1 15.00 01.00 what 1.0000
+File 1 16.00 01.00 was 1.0000
+File 1 17.00 01.00 recognized 1.0000
+File 1 18.00 01.00 here 1.0000
+;; TestCorpus/File/4 (20.00-22.00)
+File 1 20.00 01.00 insertion 1.0000
+File 1 21.00 01.00 case 1.0000
```
albertz, line 23: There is a missing endline.

tests/job_tests/recognition/files/ref.stm
```diff
@@ -0,0 +1,5 @@
+File 1 Test 00.00 04.00 <d0> this one is correct
+File 1 Test 06.00 12.00 <d0> this one is a bit longer
+File 1 Test 13.00 19.00 <d0> this one is also quite long
+File 1 Test 20.00 22.00 <d0> case
+;; LABEL "d0" "default0" "all other segments of category 0"
```
albertz, line 5: Missing endline as well.

tests/job_tests/recognition/test_scoring.py
```diff
@@ -0,0 +1,91 @@
...
+
+
+def test_sclite_job():
+    with tempfile.TemporaryDirectory() as tmpdir:
+        from sisyphus import gs
+
+        gs.WORK_DIR = tmpdir
+        sctk_binary = compile_sctk(branch="v2.4.12")
+        hyp = Path("files/hyp.ctm")
+        ref = Path("files/ref.stm")
```
albertz, line 46: Do those relative paths work? Does `pytest` always changes the current dir to the module?

I would make that more explicit. You can do sth like:
```python
_my_dir = os.path.dirname(os.path.absname(__file__))
```
And then here:
```suggestion
        hyp = Path(f"{_my_dir}/files/hyp.ctm")
        ref = Path(f"{_my_dir}/files/ref.stm")
```

Atticus1806: Right now the test works with either the current or your variant. I dont really care which one we use.

JackTemaki (approved):

tests/job_tests/recognition/test_scoring.py
```diff
@@ -0,0 +1,91 @@
...
+    sctk_job.run()
+
+    sctk_make = MakeJob(
+        folder=sctk_job.out_repository,
+        make_sequence=["config", "all", "check", "install", "doc"],
+        link_outputs={"bin": "bin/"},
+    )
+    sctk_make.run()
+    # This is needed for the compilation to work in the i6 environment, otherwise still untested
+    sctk_make._sis_environment.set("CPPFLAGS", "-std=c++11")
```
michelwi, line 35: does it make sense to have this _after_ the call to run?

tests/job_tests/recognition/test_scoring.py
```diff
@@ -0,0 +1,91 @@
...
+        gs.WORK_DIR = tmpdir
+        sctk_binary = compile_sctk(branch="v2.4.12")
+        hyp = Path("files/hyp.ctm")
+        ref = Path("files/ref.stm")
+
+        sclite_job = ScliteJob(ref=ref, hyp=hyp, sctk_binary_path=sctk_binary)
+        sclite_job._sis_setup_directory()
+        sclite_job.run()
+
+        assert sclite_job.out_wer.get() == 58.8, "Wrong WER, %s instead of 58.8" % str(sclite_job.out_wer.get())
```
michelwi, line 52: would you also add a test for a non-default precision?

albertz: This test is supposed to be applied before #442.

tests/job_tests/recognition/test_scoring.py
albertz: Why does this work without the explicit directory? Does pytest always changes the current dir to the module?

tests/job_tests/recognition/test_scoring.py
```diff
@@ -9,7 +9,7 @@
 from i6_core.recognition import ScliteJob
 
 Path = setup_path(__package__)
-
+_my_dir = os.path.dirname(os.path.abspath(__file__))
```
albertz, line 12: Missing newline here.
(Shouldn't black fix this?)

tests/job_tests/recognition/test_scoring.py
Atticus1806: Shouldn't `__package__` always be the location of the file? And then via `setup_path` in sisyphus it should resolve to the right path

tests/job_tests/recognition/test_scoring.py
Atticus1806: Whoops, didnt mean to commit all of this. But still, I guess both versions work, will fix black.

tests/job_tests/recognition/test_scoring.py
Atticus1806: Actually even though both variants work, I think the one with just setup_path seems cleaner to me. But no strong opinion on this.

albertz: Oh, you do `Path = setup_path(__package__)`. I missed that. I thought you just have `from sisyphus.job_path import Path`.

I actually thought that `setup_path` is not recommended anymore?

Atticus1806: I don‘t know to be honest, copied from other tests. Maybe its just deprecated in favor of gs.setup_path?
@JackTemaki @michelwi ?

michelwi: I don't know about deprecating this. If yes, I'd be interested to hear why.

albertz: Maybe I remember wrongly.

In any case, `Path = setup_path(__package__)` is bad, it should be `rel_path = setup_path(__package__)` or so. Names with uppercase starting letter are reserved for modules. Also, you are reusing the same name of an existing class (the `Path` class), which lead to my confusion the first place.

Atticus1806: @michelwi https://github.com/rwth-i6/sisyphus/blob/454943ac0de00f18f4333b73f6a4002c5b742a9f/sisyphus/__init__.py#L7

Atticus1806: seems like just the import without tk

tests/job_tests/recognition/test_scoring.py
```diff
@@ -0,0 +1,91 @@
...
+import tempfile
+from typing import Optional
+
+from sisyphus import tk, setup_path
+
+from i6_core.tools.compile import MakeJob
+from i6_core.tools.git import CloneGitRepositoryJob
+from i6_core.recognition import ScliteJob
+
+Path = setup_path(__package__)
```
albertz, line 11: If you want to keep using this (see separate thread), then this should be:
```suggestion
rel_path = setup_path(__package__)
```

tests/job_tests/recognition/test_scoring.py
Atticus1806: Because its closer to the other tests. Why this function is in sisyphus I dont know, but since we have it I would prefer to stay consistent with the other tests.

albertz: Well, maybe here is not the right place to reconsider this now, but in general, we don't have to repeat bad code over and over just to stay consistent with bad code.

albertz (approved):

tests/job_tests/recognition/test_scoring.py
michelwi: > setup_path and gs will be removed in the future since they are both accessible via tk

Not if I have a say about this. This will break so many of our old setups 🙈

PR #445, opened 2023-08-18
Enable multi-gpu training when "torch" is chosen as the RETURNN backend

Judyxujj:
Enable to launch the torch DDP training using torchrun

### DIFF
--- a/returnn/training.py
+++ b/returnn/training.py
@@ -248,7 +248,7 @@ class ReturnnTrainingJob(Job):
         ]
 
         if self.horovod_num_processes:
-            if self.launch_cmd == "torchrun":
+            if self.distributed_launch_cmd == "torchrun":
                 # use torchrun to lauch DDP training when the backend is torch
                 run_cmd = [
                     "torchrun",
--- a/returnn/training.py
+++ b/returnn/training.py
@@ -240,25 +240,31 @@ class ReturnnTrainingJob(Job):
         ]
 
         if self.horovod_num_processes:
-            # Normally, if the engine (e.g. SGE or Slurm) is configured correctly,
-            # it automatically provides the information on multiple nodes to mpirun,
-            # so it is not needed to explicitly pass on any hostnames here.
-            run_cmd = [
-                "mpirun",
-                "-np",
-                str(self.horovod_num_processes),
-                "-bind-to",
-                "none",
-                "-map-by",
-                "slot",
-                "-mca",
-                "pml",
-                "ob1",
-                "-mca",
-                "btl",
-                "^openib",
-                "--report-bindings",
-            ] + run_cmd
+            if self.returnn_config.get("backend", None) == "torch":
+                # use torchrun to lauch DDP training when the backend is torch
+                run_cmd = ["torchrun", "--nnodes=1", "--nproc-per-node=%s" % str(self.horovod_num_processes)] + run_cmd[
+                    1:
+                ]
+            else:
+                # Normally, if the engine (e.g. SGE or Slurm) is configured correctly,
+                # it automatically provides the information on multiple nodes to mpirun,
+                # so it is not needed to explicitly pass on any hostnames here.
+                run_cmd = [
+                    "mpirun",
+                    "-np",
+                    str(self.horovod_num_processes),
+                    "-bind-to",
+                    "none",
+                    "-map-by",
+                    "slot",
+                    "-mca",
+                    "pml",
+                    "ob1",
+                    "-mca",
+                    "btl",
+                    "^openib",
+                    "--report-bindings",
+                ] + run_cmd
 
         return run_cmd
 
--- a/returnn/training.py
+++ b/returnn/training.py
@@ -242,7 +242,12 @@ class ReturnnTrainingJob(Job):
         if self.horovod_num_processes:
             if self.returnn_config.get("backend", None) == "torch":
                 # use torchrun to lauch DDP training when the backend is torch
-                run_cmd = ["torchrun", "--nnodes=1", f"--nproc-per-node={self.horovod_num_processes}"] + run_cmd[1:]
+                nnodes = self.multi_node_slots if self.multi_node_slots else 1
+                run_cmd = [
+                    "torchrun",
+                    f"--nnodes={nnodes}",
+                    f"--nproc-per-node={self.horovod_num_processes}",
+                ] + run_cmd[1:]
             else:
                 # Normally, if the engine (e.g. SGE or Slurm) is configured correctly,
                 # it automatically provides the information on multiple nodes to mpirun,
--- a/returnn/training.py
+++ b/returnn/training.py
@@ -118,7 +118,7 @@ class ReturnnTrainingJob(Job):
         function of Returnn not all checkpoints are actually available.
     """
 
-    __sis_hash_exclude__ = {"launch_cmd": "mpirun"}
+    __sis_hash_exclude__ = {"distributed_launch_cmd": "mpirun"}
 
     def __init__(
         self,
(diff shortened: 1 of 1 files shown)


albertz:
I'm curious: What happens if you just use `mpirun`? I would have expected that this works as well, and `torchrun` internally just does the same. So, this is not the case?

albertz:
I'm not sure if it is a good idea to just always do this for the PyTorch backend. Maybe we should just introduce a separate option for this, to make it explicit? There might be valid cases to use `mpirun` with PyTorch?

returnn/training.py
```diff
@@ -240,25 +240,31 @@ def _get_run_cmd(self):
...
-                "-mca",
-                "btl",
-                "^openib",
-                "--report-bindings",
-            ] + run_cmd
+            if self.returnn_config.get("backend", None) == "torch":
+                # use torchrun to lauch DDP training when the backend is torch
+                run_cmd = ["torchrun", "--nnodes=1", "--nproc-per-node=%s" % str(self.horovod_num_processes)] + run_cmd[
+                    1:
+                ]
```
michelwi, line 247:
```suggestion
                run_cmd = ["torchrun", "--nnodes=1", f"--nproc-per-node={self.horovod_num_processes}"] + run_cmd[1:]
```
switch to f-string just to prevent ugly line-wrapping

albertz: > Would it be correct to simply set `"--nnodes={self.multi_node_slots}"` or do ne need to configure other things?

I was wondering the same. Even if this was not tested, it seems logical to just put this here.

returnn/training.py
```diff
@@ -240,25 +240,34 @@ def _get_run_cmd(self):
...
-                "btl",
-                "^openib",
-                "--report-bindings",
-            ] + run_cmd
+            if self.returnn_config.get("backend", None) == "torch":
+                # use torchrun to lauch DDP training when the backend is torch
+                nnodes = self.multi_node_slots if self.multi_node_slots else 1
+                run_cmd = [
+                    "torchrun",
+                    f"--nnodes={nnodes}",
```
albertz, line 248:
```suggestion
                run_cmd = [
                    "torchrun",
                    f"--nnodes={self.multi_node_slots or 1}",
```

returnn/training.py
```diff
@@ -240,7 +247,7 @@ def _get_run_cmd(self):
         ]
 
         if self.horovod_num_processes:
-            if self.returnn_config.get("backend", None) == "torch":
+            if self.launch_cmd == "torchrun":
```
albertz, line 251: Then use `elseif self.launch_cmd == "mpirun":` below, and then have a final:
```python
else:
    raise ValueError(f"invalid launch_cmd {self.launch_cmd!r}")
```

curufinwe (approved):

returnn/training.py
```diff
@@ -130,6 +132,7 @@ def __init__(
         time_rqmt: float = 4,
         mem_rqmt: float = 4,
         cpu_rqmt: int = 2,
+        launch_cmd: str = "mpirun",
```
JackTemaki, line 135: As this is only related to horovod, would `horovod_launch_command` not be the better name?

Or are there other purposes for this? Then the default should be `None`.

JackTemaki (dismissed):

returnn/training.py
albertz: `horovod_num_processes` is misleading, we just kept the name to not break setups. It means any kind of multi GPU training, also without Horovod, also PyTorch. So having `horovod_` prefix here does not make sense.
(Maybe we can rename `horovod_num_processes` in some non-breaking way, but I'm not sure, and this should be a separate PR.)

But you are right, a better name might be `distributed_launch_cmd` or sth like that.

The default cannot be `None` here because that would also break setups. E.g. for any setup which has `horovod_num_processes` not None, it should keep the current behavior, i.e. use `mpirun`. I think this default is fine, when the docs explain that it is only used if `horovod_num_processes` is not None.

returnn/training.py
```diff
@@ -149,6 +152,9 @@ def __init__(
         :param time_rqmt:
         :param mem_rqmt:
         :param cpu_rqmt:
+        :param launch_cmd: the command used to launch training jobs
+            Possible values: "mpirun": use mpirun, c.f. https://www.open-mpi.org/doc/v4.0/man1/mpirun.1.php
+                             "torchrun": use torchrun, c.f. https://pytorch.org/docs/stable/elastic/run.html
```
albertz, line 157: It should mention here that this is only used if `horovod_num_processes` is not None.

Judyxujj: @albet @JackTemaki  I added a sentence stating that 
`:param launch_cmd: the command used to launch training jobs, only used if horovod_num_processes is not None `

returnn/training.py
curufinwe, line 135:
```suggestion
        distributed_launch_cmd: str = "mpirun",
```

returnn/training.py
```diff
@@ -149,6 +152,9 @@ def __init__(
         :param time_rqmt:
         :param mem_rqmt:
         :param cpu_rqmt:
+        :param launch_cmd: the command used to launch training jobs, only used if horovod_num_processes is not None
```
curufinwe, line 155:
```suggestion
        :param distributed_launch_cmd: the command used to launch training jobs, only used if horovod_num_processes is not None
```

returnn/training.py
```diff
@@ -118,6 +118,8 @@ class ReturnnTrainingJob(Job):
         function of Returnn not all checkpoints are actually available.
     """
 
+    __sis_hash_exclude__ = {"launch_cmd": "mpirun"}
```
curufinwe, line 121:
```suggestion
    __sis_hash_exclude__ = {"distributed_launch_cmd": "mpirun"}
```

returnn/training.py
```diff
@@ -158,12 +164,14 @@ def __init__(
         :param returnn_root: file path to the RETURNN repository root folder
         """
         assert isinstance(returnn_config, ReturnnConfig)
+        assert launch_cmd in ["mpirun", "torchrun"]
```
curufinwe, line 167:
```suggestion
        assert distributed_launch_cmd in ["mpirun", "torchrun"]
```

returnn/training.py
```diff
@@ -158,12 +164,14 @@ def __init__(
...
         self.check_blacklisted_parameters(returnn_config)
         kwargs = locals()
         del kwargs["self"]
 
         self.returnn_python_exe = util.get_returnn_python_exe(returnn_python_exe)
         self.returnn_root = util.get_returnn_root(returnn_root)
+        self.launch_cmd = launch_cmd
```
curufinwe, line 174:
```suggestion
        self.distributed_launch_cmd = distributed_launch_cmd
```

returnn/training.py
```diff
@@ -240,25 +248,35 @@ def _get_run_cmd(self):
...
-                "slot",
-                "-mca",
-                "pml",
-                "ob1",
-                "-mca",
-                "btl",
-                "^openib",
-                "--report-bindings",
-            ] + run_cmd
+            if self.launch_cmd == "torchrun":
```
curufinwe, line 251:
```suggestion
            if self.distributed_launch_cmd == "torchrun":
```

returnn/training.py
```diff
@@ -240,25 +248,35 @@ def _get_run_cmd(self):
...
+                # use torchrun to lauch DDP training when the backend is torch
+                run_cmd = [
+                    "torchrun",
+                    f"--nnodes={self.multi_node_slots or 1}",
+                    f"--nproc-per-node={self.horovod_num_processes}",
+                ] + run_cmd[1:]
+            elif self.launch_cmd == "mpirun":
```
curufinwe, line 258:
```suggestion
            elif self.distributed_launch_cmd == "mpirun":
```

albertz (approved):

christophmluscher (approved):

christophmluscher:
@JackTemaki friendly ping :)

PR #446, opened 2023-08-21
feat: Add option to generate LM image and GC via two separate jobs

### LINKED ISSUE #430: Hash issues in AdvancedTreeSearchLmImageAndGlobalCacheJob
The way the `AdvancedTreeSearchJob` constructs the `AdvancedTreeSearchLmImageAndGlobalCacheJob` makes it an easy subject to hash issues.

I'm running decodings on LibriSpeech, and found the following job instances in my work folder. All compute the same 1.8G LM image but have different hashes. In my case this is likely due to differing TDP scales, which go into the CRP for the LM image job and change the hash.

Another candidate for (a different kind of) hash issues is the feature scorer, which also goes into the CRP but is explicitly left hashed.

```
❯ shasum $(ls work/i6_core/recognition/advanced_tree_search/AdvancedTreeSearchLmImageAndGlobalCacheJob.*/output/*.image | xargs -n 1 cf)
4cc098b119a7f2e5d692aebacfd92ef1a132bf76  /var/tmp/mgunz/work/asr3/raissi/shared_workspaces/gunz/2023-05--thesis-baselines-tf2/i6_core/recognition/advanced_tree_search/AdvancedTreeSearchLmImageAndGlobalCacheJob.9ZuNbY6vfq4B/output/lm-1.image
4cc098b119a7f2e5d692aebacfd92ef1a132bf76  /var/tmp/mgunz/work/asr3/raissi/shared_workspaces/gunz/2023-05--thesis-baselines-tf2/i6_core/recognition/advanced_tree_search/AdvancedTreeSearchLmImageAndGlobalCacheJob.akOZ3zJXuFWb/output/lm-1.image
4cc098b119a7f2e5d692aebacfd92ef1a132bf76  /var/tmp/mgunz/work/asr3/raissi/shared_workspaces/gunz/2023-05--thesis-baselines-tf2/i6_core/recognition/advanced_tree_search/AdvancedTreeSearchLmImageAndGlobalCacheJob.gqX8MWYitMdb/output/lm-1.image
4cc098b119a7f2e5d692aebacfd92ef1a132bf76  /var/tmp/mgunz/work/asr3/raissi/shared_workspaces/gunz/2023-05--thesis-baselines-tf2/i6_core/recognition/advanced_tree_search/AdvancedTreeSearchLmImageAndGlobalCacheJob.j2O4KrVDm2OD/output/lm-1.image
4cc098b119a7f2e5d692aebacfd92ef1a132bf76  /var/tmp/mgunz/work/asr3/raissi/shared_workspaces/gunz/2023-05--thesis-baselines-tf2/i6_core/recognition/advanced_tree_search/AdvancedTreeSearchLmImageAndGlobalCacheJob.JH9tbiSyYaU5/output/lm-1.image
4cc098b119a7f2e5d692aebacfd92ef1a132bf76  /var/tmp/mgunz/work/asr3/raissi/shared_workspaces/gunz/2023-05--thesis-baselines-tf2/i6_core/recognition/advanced_tree_search/AdvancedTreeSearchLmImageAndGlobalCacheJob.JMMQd6Qt7L4Y/output/lm-1.image
4cc098b119a7f2e5d692aebacfd92ef1a132bf76  /var/tmp/mgunz/work/asr3/raissi/shared_workspaces/gunz/2023-05--thesis-baselines-tf2/i6_core/recognition/advanced_tree_search/AdvancedTreeSearchLmImageAndGlobalCacheJob.KuQpQX2Bh0x3/output/lm-1.image
4cc098b119a7f2e5d692aebacfd92ef1a132bf76  /var/tmp/mgunz/work/asr3/raissi/shared_workspaces/gunz/2023-05--thesis-baselines-tf2/i6_core/recognition/advanced_tree_search/AdvancedTreeSearchLmImageAndGlobalCacheJob.KZg6mlchtLBB/output/lm-1.image
4cc098b119a7f2e5d692aebacfd92ef1a132bf76  /var/tmp/mgunz/work/asr3/raissi/shared_workspaces/gunz/2023-05--thesis-baselines-tf2/i6_core/recognition/advanced_tree_search/AdvancedTreeSearchLmImageAndGlobalCacheJob.NCzSYGARj48g/output/lm-1.image
4cc098b119a7f2e5d692aebacfd92ef1a132bf76  /var/tmp/mgunz/work/asr3/raissi/shared_workspaces/gunz/2023-05--thesis-baselines-tf2/i6_core/recognition/advanced_tree_search/AdvancedTreeSearchLmImageAndGlobalCacheJob.NLuAGBkMBB3P/output/lm-1.image
4cc098b119a7f2e5d692aebacfd92ef1a132bf76  /var/tmp/mgunz/work/asr3/raissi/shared_workspaces/gunz/2023-05--thesis-baselines-tf2/i6_core/recognition/advanced_tree_search/AdvancedTreeSearchLmImageAndGlobalCacheJob.nLvpXSLvNPai/output/lm-1.image
4cc098b119a7f2e5d692aebacfd92ef1a132bf76  /var/tmp/mgunz/work/asr3/raissi/shared_workspaces/gunz/2023-05--thesis-baselines-tf2/i6_core/recognition/advanced_tree_search/AdvancedTreeSearchLmImageAndGlobalCacheJob.onldbKiFZ9Yu/output/lm-1.image
4cc098b119a7f2e5d692aebacfd92ef1a132bf76  /var/tmp/mgunz/work/asr3/raissi/shared_workspaces/gunz/2023-05--thesis-baselines-tf2/i6_core/recognition/advanced_tree_search/AdvancedTreeSearchLmImageAndGlobalCacheJob.osWzDVEIASI1/output/lm-1.image
4cc098b119a7f2e5d692aebacfd92ef1a132bf76  /var/tmp/mgunz/work/asr3/raissi/shared_workspaces/gunz/2023-05--thesis-baselines-tf2/i6_core/recognition/advanced_tree_search/AdvancedTreeSearchLmImageAndGlobalCacheJob.sz9KWZd5IkfX/output/lm-1.image
4cc098b119a7f2e5d692aebacfd92ef1a132bf76  /var/tmp/mgunz/work/asr3/raissi/shared_workspaces/gunz/2023-05--thesis-baselines-tf2/i6_core/recognition/advanced_tree_search/AdvancedTreeSearchLmImageAndGlobalCacheJob.uRxs6uEiLpHJ/output/lm-1.image
```

The global caches that are computed are indeed different, and so splitting the job into two jobs (one for LM image, one for global cache) might be a solution:

```
❯ shasum $(ls work/i6_core/recognition/advanced_tree_search/AdvancedTreeSearchLmImageAndGlobalCacheJob.*/output/*.cache | xargs -n 1 cf)
032e23952d190f2037a5171e4a379e8f8943a51f  /var/tmp/mgunz/work/asr3/raissi/shared_workspaces/gunz/2023-05--thesis-baselines-tf2/i6_core/recognition/advanced_tree_search/AdvancedTreeSearchLmImageAndGlobalCacheJob.9ZuNbY6vfq4B/output/global.cache
5f8cb12a3cc495bab18e07883001696f42a70c08  /var/tmp/mgunz/work/asr3/raissi/shared_workspaces/gunz/2023-05--thesis-baselines-tf2/i6_core/recognition/advanced_tree_search/AdvancedTreeSearchLmImageAndGlobalCacheJob.akOZ3zJXuFWb/output/global.cache
da28cab82e48d176718a6dceb138f77224e21545  /var/tmp/mgunz/work/asr3/raissi/shared_workspaces/gunz/2023-05--thesis-baselines-tf2/i6_core/recognition/advanced_tree_search/AdvancedTreeSearchLmImageAndGlobalCacheJob.gqX8MWYitMdb/output/global.cache
3b5cd599e3f00375335681af962578dd8b7af1a8  /var/tmp/mgunz/work/asr3/raissi/shared_workspaces/gunz/2023-05--thesis-baselines-tf2/i6_core/recognition/advanced_tree_search/AdvancedTreeSearchLmImageAndGlobalCacheJob.j2O4KrVDm2OD/output/global.cache
66ce732c9656b7a32ff9d05760eb35d63720b424  /var/tmp/mgunz/work/asr3/raissi/shared_workspaces/gunz/2023-05--thesis-baselines-tf2/i6_core/recognition/advanced_tree_search/AdvancedTreeSearchLmImageAndGlobalCacheJob.JH9tbiSyYaU5/output/global.cache
2cbe533d5245886945cdd525657ca07e9d3b0831  /var/tmp/mgunz/work/asr3/raissi/shared_workspaces/gunz/2023-05--thesis-baselines-tf2/i6_core/recognition/advanced_tree_search/AdvancedTreeSearchLmImageAndGlobalCacheJob.JMMQd6Qt7L4Y/output/global.cache
16b7cf5a95e06b658a7f40763a5cb8aa4ec2b808  /var/tmp/mgunz/work/asr3/raissi/shared_workspaces/gunz/2023-05--thesis-baselines-tf2/i6_core/recognition/advanced_tree_search/AdvancedTreeSearchLmImageAndGlobalCacheJob.KuQpQX2Bh0x3/output/global.cache
27fb128dd0a76e1bc3be1fc423452ca4e230e91d  /var/tmp/mgunz/work/asr3/raissi/shared_workspaces/gunz/2023-05--thesis-baselines-tf2/i6_core/recognition/advanced_tree_search/AdvancedTreeSearchLmImageAndGlobalCacheJob.KZg6mlchtLBB/output/global.cache
dd2773af1471fb8a6527bfd3fba03e4613ce6869  /var/tmp/mgunz/work/asr3/raissi/shared_workspaces/gunz/2023-05--thesis-baselines-tf2/i6_core/recognition/advanced_tree_search/AdvancedTreeSearchLmImageAndGlobalCacheJob.NCzSYGARj48g/output/global.cache
2128afe006bfb39a6a8bc89c445082dcba8ca0ba  /var/tmp/mgunz/work/asr3/raissi/shared_workspaces/gunz/2023-05--thesis-baselines-tf2/i6_core/recognition/advanced_tree_search/AdvancedTreeSearchLmImageAndGlobalCacheJob.NLuAGBkMBB3P/output/global.cache
2964b6d362cd896e2b144a3d9ba614404457c08a  /var/tmp/mgunz/work/asr3/raissi/shared_workspaces/gunz/2023-05--thesis-baselines-tf2/i6_core/recognition/advanced_tree_search/AdvancedTreeSearchLmImageAndGlobalCacheJob.nLvpXSLvNPai/output/global.cache
34d97b5f4fcdb962cde87d5dcd71a059a2a23829  /var/tmp/mgunz/work/asr3/raissi/shared_workspaces/gunz/2023-05--thesis-baselines-tf2/i6_core/recognition/advanced_tree_search/AdvancedTreeSearchLmImageAndGlobalCacheJob.onldbKiFZ9Yu/output/global.cache
94edb564a1b966c6ae27747625bbff7fb6ad13d6  /var/tmp/mgunz/work/asr3/raissi/shared_workspaces/gunz/2023-05--thesis-baselines-tf2/i6_core/recognition/advanced_tree_search/AdvancedTreeSearchLmImageAndGlobalCacheJob.osWzDVEIASI1/output/global.cache
2e34c5a2d41de5f7c3ba7021c4cbfdfa4749032d  /var/tmp/mgunz/work/asr3/raissi/shared_workspaces/gunz/2023-05--thesis-baselines-tf2/i6_core/recognition/advanced_tree_search/AdvancedTreeSearchLmImageAndGlobalCacheJob.sz9KWZd5IkfX/output/global.cache
36fb52477284c7377bffbabe55d8e56b3912c60d  /var/tmp/mgunz/work/asr3/raissi/shared_workspaces/gunz/2023-05--thesis-baselines-tf2/i6_core/recognition/advanced_tree_search/AdvancedTreeSearchLmImageAndGlobalCacheJob.uRxs6uEiLpHJ/output/global.cache
```

NeoLegends:
Closes #430

Now testing...

### DIFF
--- a/recognition/advanced_tree_search.py
+++ b/recognition/advanced_tree_search.py
@@ -10,6 +10,7 @@ from sisyphus import *
 
 Path = setup_path(__package__)
 
+import copy
 import math
 import os
 import shutil
@@ -167,6 +168,7 @@ class AdvancedTreeSearchJob(rasr.RasrCommand, Job):
         lmgc_mem: float = 12.0,
         lmgc_alias: Optional[str] = None,
         lmgc_scorer: Optional[rasr.FeatureScorer] = None,
+        separate_lmi_gc_generation: bool = False,
         model_combination_config: Optional[rasr.RasrConfig] = None,
         model_combination_post_config: Optional[rasr.RasrConfig] = None,
         extra_config: Optional[rasr.RasrConfig] = None,
@@ -286,18 +288,40 @@ class AdvancedTreeSearchJob(rasr.RasrCommand, Job):
         lmgc_mem: float,
         lmgc_alias: Optional[str],
         lmgc_scorer: Optional[rasr.FeatureScorer],
+        separate_lmi_gc_generation: bool,
         model_combination_config: Optional[rasr.RasrConfig],
         model_combination_post_config: Optional[rasr.RasrConfig],
         extra_config: Optional[rasr.RasrConfig],
         extra_post_config: Optional[rasr.RasrConfig],
         **kwargs,
     ):
-        lm_gc = AdvancedTreeSearchLmImageAndGlobalCacheJob(
-            crp, lmgc_scorer if lmgc_scorer is not None else feature_scorer, extra_config, extra_post_config
-        )
-        if lmgc_alias is not None:
-            lm_gc.add_alias(lmgc_alias)
-        lm_gc.rqmt["mem"] = lmgc_mem
+        def specialize_lm_config(crp, lm_config):
+            crp = copy.deepcopy(crp)
+            crp.language_model = lm_config
+            return crp
+
+        if separate_lmi_gc_generation:
+            gc = BuildGlobalCacheJob(crp, extra_config, extra_post_config).out_global_cache
+
+            arpa_lms = AdvancedTreeSearchLmImageAndGlobalCacheJob.find_arpa_lms(
+                crp.language_model, post_config.lm if post_config is not None else None
+            )
+            lm_images = {
+                (i + 1): lm.CreateLmImageJob(
+                    specialize_lm_config(crp, lm), extra_config=extra_config, extra_post_config=extra_post_config
+                ).out_lm
+                for i, lm in enumerate(arpa_lms)
+            }
+        else:
+            lm_gc = AdvancedTreeSearchLmImageAndGlobalCacheJob(
+                crp, lmgc_scorer if lmgc_scorer is not None else feature_scorer, extra_config, extra_post_config
+            )
+            if lmgc_alias is not None:
+                lm_gc.add_alias(lmgc_alias)
+            lm_gc.rqmt["mem"] = lmgc_mem
+
+            gc = lm_gc.out_global_cache
+            lm_images = lm_gc.out_lm_images
 
         search_parameters = cls.update_search_parameters(search_parameters)
 
@@ -397,14 +421,14 @@ class AdvancedTreeSearchJob(rasr.RasrCommand, Job):
             ]
 
         post_config.flf_lattice_tool.global_cache.read_only = True
-        post_config.flf_lattice_tool.global_cache.file = lm_gc.out_global_cache
+        post_config.flf_lattice_tool.global_cache.file = gc
 
         arpa_lms = AdvancedTreeSearchLmImageAndGlobalCacheJob.find_arpa_lms(
             config.flf_lattice_tool.network.recognizer.lm,
             post_config.flf_lattice_tool.network.recognizer.lm,
         )
         for i, lm_config in enumerate(arpa_lms):
-            lm_config[1].image = lm_gc.out_lm_images[i + 1]
+            lm_config[1].image = lm_images[i + 1]
 
         # Remaining Flf-network
 
@@ -438,11 +462,11 @@ class AdvancedTreeSearchJob(rasr.RasrCommand, Job):
         config._update(extra_config)
         post_config._update(extra_post_config)
 
-        return config, post_config, lm_gc
+        return config, post_config
 
     @classmethod
     def hash(cls, kwargs):
-        config, post_config, lm_gc = cls.create_config(**kwargs)
+        config, post_config = cls.create_config(**kwargs)
         return super().hash(
             {
                 "config": config,


michelwi (changes requested):

NeoLegends:
I just discovered the LM image and GC are set on the post config, not in the normal config. This means splitting the LMGC-Job into two does not change the hash of any existing jobs.

Therefore, I think this flag can be enabled by default, or rather, we can in-line the flag and make it the new default! WDYT?

lm/lm_image.py
```diff
@@ -21,7 +21,7 @@ def __init__(
         extra_config=None,
         extra_post_config=None,
         encoding="utf-8",
-        mem=2,
+        mem=12,
```
michelwi, line 24: are you sure we need this much by default (especially if it is configurable)? I think the memory requirement would be [Size of the LM] + [some MB for the Job].

So while I would agree that 2G are too low in most cases, 6 or 8 would be a more reasonable default value. (What is the current mem/core ratio in the cluster?)

rasr/crp.py
```diff
@@ -79,7 +79,7 @@ def set_executables(self, rasr_binary_path, rasr_arch="linux-x86_64-standard"):
         self.flf_tool_exe = rasr_binary_path.join_right(f"flf-tool.{rasr_arch}")
         self.kws_tool_exe = None  # does not exist
         self.lattice_processor_exe = rasr_binary_path.join_right(f"lattice-processor.{rasr_arch}")
-        self.lm_util_exe = None  # does not exist
+        self.lm_util_exe = rasr_binary_path.join_right(f"lm-util.{rasr_arch}")
```
michelwi, line 82: is lm-util being compiled now by default?

recognition/advanced_tree_search.py
```diff
@@ -397,14 +411,14 @@ def create_config(
...
 
-        arpa_lms = AdvancedTreeSearchLmImageAndGlobalCacheJob.find_arpa_lms(
+        arpa_lms = lm.find_arpa_lms(
             config.flf_lattice_tool.network.recognizer.lm,
             post_config.flf_lattice_tool.network.recognizer.lm,
         )
-        for i, lm_config in enumerate(arpa_lms):
-            lm_config[1].image = lm_gc.out_lm_images[i + 1]
+        for i, (_lm_config, lm_post_config) in enumerate(arpa_lms):
+            lm_post_config.image = lm_images[i + 1]
```
michelwi, line 421: Is there a specific reason why you start counting from 1? here you have to add +1 to i and also above you need to add. 0-based counting would simplify both cases.

recognition/advanced_tree_search.py
```diff
@@ -817,7 +831,7 @@ class BuildGlobalCacheJob(rasr.RasrCommand, Job):
     Standalone job to create the global-cache for advanced-tree-search
     """
 
-    def __init__(self, crp, extra_config=None, extra_post_config=None):
+    def __init__(self, crp, extra_config=None, extra_post_config=None, mem=12):
```
michelwi, line 834: are you sure this much memory is needed? here some small amount should actually be suficcient as we do not need to load the LMs into memory.

Or could it be that in the crp that are given to this job still some LM is present, which is loaded into memory at runtime for this job. In that case you could test setting the `language_model_config.type = "zerogram"` to completely ignore it. (Not 100% sure this works in any case, please test)

lm/lm_image.py
NeoLegends: I had it crash w/ 8GB on LS960 so I went with 12GB.

rasr/crp.py
NeoLegends: Yes, see https://github.com/rwth-i6/rasr/commit/d58a228e80976f4b25c1700648622f658830e2dc

lm/lm_image.py
michelwi: mh.. okay.. I'll leave this to the other users to decide which default they prefer.

recognition/advanced_tree_search.py
michelwi: > It seems to me like it should be the job's duty to use a dummy LM then, in case RASR wants one but doesn't actually use it.

yes, I think as long as the Lm used has the same vocab then it should be fine to replace the LM in this Job with a dummy.

lm/lm_image.py
NeoLegends: It's still [Size of LM] + [some MB for the Job], just that [some MB for the job] is very large ;)

NeoLegends:
I have successfully run recognitions w/ this setup. This is tested now.

lm/lm_image.py
JackTemaki: I would prefer if 8 works, but if even for a standard task like ls960 it is failing we should have 12.

recognition/advanced_tree_search.py
```diff
@@ -167,6 +149,7 @@ def __init__(
         lmgc_mem: float = 12.0,
         lmgc_alias: Optional[str] = None,
         lmgc_scorer: Optional[rasr.FeatureScorer] = None,
+        separate_lmi_gc_generation: bool = False,
```
JackTemaki, line 152: I had difficulties parsing `lmi` in my head right away, maybe fully writing `image` is nicer (but not required).

recognition/advanced_tree_search.py
```diff
@@ -190,6 +173,7 @@ def __init__(
         :param lmgc_mem: Memory requirement for the AdvancedTreeSearchLmImageAndGlobalCacheJob
         :param lmgc_alias: Alias for the AdvancedTreeSearchLmImageAndGlobalCacheJob
         :param lmgc_scorer: Dummy scorer for the AdvancedTreeSearchLmImageAndGlobalCacheJob which is required but unused
+        :param separate_lm_image_gc_generation: Whether to generate the LM image and the global cache via two separate jobs for a more stable hash. Whether or not this flag is set is not part of the hash, so using separate jobs is the default.
```
JackTemaki, line 176: Shouldn't this be a `so NOT using separate jobs is the default?`

NeoLegends:
AppTek hash broke after changing only a comment, did the AppTek pipeline change in the meantime?

Marvin84:
> AppTek hash broke after changing only a comment, did the AppTek pipeline change in the meantime?

@curufinwe @michelwi @christophmluscher any idea?

lm/util.py
```diff
@@ -0,0 +1,24 @@
...
+    res = c._get("image") is not None
+    res = res or (pc is not None and pc._get("image") is not None)
+    return res
+
+
+def find_arpa_lms(lm_config: rasr.RasrConfig, lm_post_config=None) -> List[Tuple[rasr.RasrConfig, rasr.RasrConfig]]:
+    result = []
+
+    if lm_config.type == "ARPA":
+        if not _has_image(lm_config, lm_post_config):
```
Atticus1806, line 16: Is the not correct here? So if there is no image it adds it to result?

NeoLegends: I think the idea is not to overwrite existing LM images.

Marvin84:
@curufinwe @michelwi can we merge this?

michelwi:
> @curufinwe @michelwi can we merge this?

Since the apptek test passes, I see no objections from our site. 

I'll dismiss my old review, but I currently don't have much time to re-review it.. sorry.

lm/util.py
```diff
@@ -0,0 +1,24 @@
+from typing import List, Tuple
+
+import i6_core.rasr as rasr
+
+
+def _has_image(c: rasr.RasrConfig, pc: rasr.RasrConfig):
```
Atticus1806, line 6: I think I would either like to spell config and post_config out or have a small docstring here.

lm/util.py
```diff
@@ -0,0 +1,24 @@
...
+    res = c._get("image") is not None
+    res = res or (pc is not None and pc._get("image") is not None)
+    return res
+
+
+def find_arpa_lms(lm_config: rasr.RasrConfig, lm_post_config=None) -> List[Tuple[rasr.RasrConfig, rasr.RasrConfig]]:
```
Atticus1806, line 12: Is the post config also `rasr.RasrConfig`? Or some other type?

recognition/advanced_tree_search.py
```diff
@@ -206,6 +190,8 @@ def __init__(
             self.config,
             self.post_config,
             self.lm_gc_job,
+            self.gc_job,
+            self.lm_image_jobs,
         ) = AdvancedTreeSearchJob.create_config(**kwargs)
```
Atticus1806, line 195: Is `.create_config` only called within this job? Otherwise we need to be careful with changing the returned variables. But I think you caught the cases here and otherwise the change is easy.

recognition/advanced_tree_search.py
```diff
@@ -286,18 +272,46 @@ def create_config(
...
         extra_post_config: Optional[rasr.RasrConfig],
         **kwargs,
     ):
-        lm_gc = AdvancedTreeSearchLmImageAndGlobalCacheJob(
-            crp, lmgc_scorer if lmgc_scorer is not None else feature_scorer, extra_config, extra_post_config
-        )
-        if lmgc_alias is not None:
-            lm_gc.add_alias(lmgc_alias)
-        lm_gc.rqmt["mem"] = lmgc_mem
+        def add_lm_config_to_crp(crp, lm_config):
```
Atticus1806, line 282:
```suggestion
        def add_lm_config_to_crp(crp: rasr.CommonRasrParameters, lm_config: rasr.RasrConfig):
```

recognition/advanced_tree_search.py
```diff
@@ -286,18 +272,46 @@ def create_config(
...
+            crp = copy.deepcopy(crp)
+            crp.language_model_config = lm_config
+            return crp
+
+        if separate_lm_image_gc_generation:
+            gc_job = BuildGlobalCacheJob(crp, extra_config, extra_post_config)
```
Atticus1806, line 288: I am wondering if we should already think about adding aliases here or if this overcomplicates the matter for now.

recognition/advanced_tree_search.py
```diff
@@ -397,14 +411,14 @@ def create_config(
...
+        post_config.flf_lattice_tool.global_cache.file = gc
 
-        arpa_lms = AdvancedTreeSearchLmImageAndGlobalCacheJob.find_arpa_lms(
+        arpa_lms = lm.find_arpa_lms(
             config.flf_lattice_tool.network.recognizer.lm,
             post_config.flf_lattice_tool.network.recognizer.lm,
         )
-        for i, lm_config in enumerate(arpa_lms):
-            lm_config[1].image = lm_gc.out_lm_images[i + 1]
+        for i, (_lm_config, lm_post_config) in enumerate(arpa_lms):
```
Atticus1806, line 420: Same as above

recognition/advanced_tree_search.py
```diff
@@ -286,18 +272,46 @@ def create_config(
...
+
+        if separate_lm_image_gc_generation:
+            gc_job = BuildGlobalCacheJob(crp, extra_config, extra_post_config)
+
+            arpa_lms = lm.find_arpa_lms(crp.language_model_config, None)
+            lm_image_jobs = {
+                (i + 1): lm.CreateLmImageJob(
+                    add_lm_config_to_crp(crp, lm_config), extra_config=extra_config, extra_post_config=extra_post_config
+                )
+                for i, (lm_config, _lm_post_config) in enumerate(arpa_lms)
```
Atticus1806, line 295: Is there a reason for the `_`. Maybe I am overlooking something in the web view. If its not used you could just fully replace it with `_`

recognition/advanced_tree_search.py
Atticus1806: What about this. @michelwi any more comments? Otherwise I think we can skip this for now.

Atticus1806:
> can we merge this?

With two approves yes :). I just got a few comments then I can approve, but please ask someone else to also review.

lm/util.py
Marvin84: I agree with @Atticus1806

recognition/advanced_tree_search.py
Marvin84: I do not think we need alias for the global cache

recognition/advanced_tree_search.py
michelwi: I think the `post_config here is unused, because 
* both `config` and `post_confic` are extracted from the `crp`
* only the `config` is modified
* the `crp` with the old `post_config` is still being used

+1 to can be just `_` to make clear it is unused
```suggestion
                for i, (lm_config, _) in enumerate(arpa_lms)
```

michelwi: I think the `post_config here is unused, because 
* both `config` and `post_confic` are extracted from the `crp`
* only the `config` is modified
* the `crp` with the old `post_config` is still being used

+1 to can be just `_` to make clear it is unused
```suggestion
                for i, (lm_config, _) in enumerate(arpa_lms)
```

recognition/advanced_tree_search.py
michelwi: yes, we can keep it this way.

michelwi (commented):
I think many open comments should be addressed before merging.

recognition/advanced_tree_search.py
```diff
@@ -286,18 +272,46 @@ def create_config(
...
-        lm_gc.rqmt["mem"] = lmgc_mem
+        def add_lm_config_to_crp(crp, lm_config):
+            crp = copy.deepcopy(crp)
+            crp.language_model_config = lm_config
+            return crp
+
+        if separate_lm_image_gc_generation:
+            gc_job = BuildGlobalCacheJob(crp, extra_config, extra_post_config)
+
+            arpa_lms = lm.find_arpa_lms(crp.language_model_config, None)
```
michelwi, line 290: The function `lm.find_arpa_lms` only returns the LMs that do not already have an lm.image - because for those that already have an image we do not need to create a new one.

But the lm.image usually is defined in the post_config. When we here not pass the post config, then all (arpa) LMs are found and returned. Therefore we do extra work here.
And even worse: below in line 404/418 we call the function again, but with the post config. So if the original crp contain a mix of arpa LMs out of which some have already images and others do not, then the items in `arpa_lms` differ between the calls. And since we use the index to match the image to the LM, the mapping will be off and the wrong image will be assigned.

```suggestion
            arpa_lms = lm.find_arpa_lms(crp.language_model_config, crp.language_model_post_config)
```

recognition/advanced_tree_search.py
```diff
@@ -397,14 +411,14 @@ def create_config(
             ]
 
         post_config.flf_lattice_tool.global_cache.read_only = True
-        post_config.flf_lattice_tool.global_cache.file = lm_gc.out_global_cache
+        post_config.flf_lattice_tool.global_cache.file = gc
 
-        arpa_lms = AdvancedTreeSearchLmImageAndGlobalCacheJob.find_arpa_lms(
+        arpa_lms = lm.find_arpa_lms(
```
michelwi, line 416: We already have called lm.find_arpa_lms above. Maybe move the above call out of the if condition and then reuse the `arpa_lms` from above here. This would also avoid mismatching items in the list as I outlined above.

recognition/advanced_tree_search.py
michelwi: the `AdvancedTreeSearchWithRescoringJob` below in this file inherits from `AdvancedTreeSearchJob` and uses `super().create_config`. that should be fixed.

michelwi (changes requested):

michelwi (approved):
I think all open comments are addressed and the pipeline is green.

@Atticus1806 , @JackTemaki, @NeoLegends would you be able to take another look? Thx.

recognition/advanced_tree_search.py
```diff
@@ -117,7 +100,7 @@ def create_config(cls, crp, feature_scorer, extra_config, extra_post_config, **k
         config.flf_lattice_tool.network.recognizer.feature_extraction.file = "dummy.flow"
         config.flf_lattice_tool.network.recognizer.lm.scale = 1.0
 
-        arpa_lms = cls.find_arpa_lms(
+        arpa_lms = lm.find_arpa_lms(
```
Icemole:
```suggestion
        arpa_lms = lm.util.find_arpa_lms(
```

Following the suggestion above.

Icemole (commented):
I only have some minor comments, otherwise the functionality looks good to me. Perhaps @michelwi can take a look and potentially discuss them?

With respect to AppTek hashes: if it's generally agreed that `JOINED` is a better alternative, we could change the few (?) jobs ourselves. I'm more worried about users outside of AppTek (such as the i6 research team) suddenly seeing that their hashes change after a `git pull`...

michelwi:
> With respect to [...] hashes

`JOINED` is the old way, which is generally agreed upon the worse way to do it. We only keep it as default precisely not to mess with everyones setups.

PR #447, opened 2023-08-29
Modifying `ApplyBPEModelToLexiconJob` and `ApplyBPEToTextJob`

AtanasGruev:
The jobs deal with BPE-adaptation of Bliss lexica and text files specifying corpus vocabularies. These have now been modified to solve some detected problems, e.g. how "bpe.vocab" is used to provide the vocabulary (needs to be adapted due to dictionary structure) and how the [SILENCE] special lemma allows for an empty orthography (potentially leading to lexicon mismatches).


JackTemaki (changes requested):
While the ApplyBPEModelToLexiconJob is probably not usable in its current state, you can not completely re-write the jobs without careful consideration. Also, there seems to be a lot of explicit things to be put in there. I think we should have an offline discussion about this, as the PR needs severe modifications in order to be mergeable.

JackTemaki:
Lets close this for now, and find new and proper solutions, also related to #321

PR #448, opened 2023-08-30
Fix number of CPUs for AdvancedTreeSearch

NeoLegends:
I am measuring RTFs, for this I need precise specification of the number of cores a job is supposed to run on.

Specifying `cpu=2` in the constructor of `AdvancedTreeSearchJob` in practice meant the jobs only consumes 100% (un-normalized, so only one core) of CPU, because `OMP_NUM_THREADS` is set to `cpu / 2`, i.e. `1`.

Removing the division by two leads the job to consume the desired 200%, i.e. 2 cores. This division by two is very strange to me and I'm struggling to think of the original motivation here. I'm not even sure it's a bug -- other than what I could observe from my use case, where the behavior indeed seemed buggy, there might be something I'm missing.

@JackTemaki You wrote that piece of code in the first place, so you're likely the most qualified to answer here. :) I'd love to know the original motivation behind the `/ 2`.

### DIFF
--- a/recognition/advanced_tree_search.py
+++ b/recognition/advanced_tree_search.py
@@ -239,7 +239,7 @@ class AdvancedTreeSearchJob(rasr.RasrCommand, Job):
         self.feature_flow.write_to_file("feature.flow")
         util.write_paths_to_file(self.out_lattice_bundle, self.out_single_lattice_caches.values())
         extra_code = "export OMP_NUM_THREADS={0}\nexport TF_DEVICE='{1}'".format(
-            math.ceil(self.cpu / 2), "gpu" if self.use_gpu else "cpu"
+            math.ceil(self.cpu), "gpu" if self.use_gpu else "cpu"
         )
         self.write_run_script(self.exe, "recognition.config", extra_code=extra_code)
 


JackTemaki (approved):
After an offline discussion, it seems that this number was optimized for different other settings (and BLSTM instead of Conformer), and indeed OMP_NUM_THREADS should better default to the number of CPUs.

Atticus1806 (approved):

JackTemaki:
So, its not a bug but a well intended but wrong feature ;) We should keep a close look on the threading behavior for both TF and ONNX.

JackTemaki:
@NeoLegends more than one year later I am not so sure anymore if this is correct. The percentage in htop is not a good measurement if you have Hyperthreading enabled, and over-usage of threads is usually more risky than under-usage. See plot and warning here:
https://pytorch.org/docs/stable/notes/cpu_threading_torchscript_inference.html

I came across this while debugging a load issue caused by a number of PyTorch forward jobs from @Atticus1806 , which had a 100% perfect occupation of the node, but caused a load error.

PR #449, opened 2023-09-06
add FilterSegmentsByRegexJob

vieting:
I needed a job to filter the segments based on whether they contain a certain string or not and therefore added this job. Do you think it's generally helpful? Then we could add it here as well.

### DIFF
--- a/corpus/filter.py
+++ b/corpus/filter.py
@@ -1,13 +1,15 @@
 __all__ = [
+    "FilterSegmentsByListJob",
+    "FilterSegmentsByRegexJob",
     "FilterSegmentsByAlignmentConfidenceJob",
     "FilterCorpusBySegmentsJob",
-    "FilterSegmentsByListJob",
     "FilterCorpusRemoveUnknownWordSegmentsJob",
     "FilterCorpusBySegmentDurationJob",
 ]
 
 import gzip
 import logging
+import re
 import xml.etree.cElementTree as ET
 
 import numpy as np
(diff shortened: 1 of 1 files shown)


corpus/filter.py
```diff
@@ -69,6 +71,45 @@ def run(self):
                 )
 
 
+class FilterSegmentsByRegexJob(Job):
+    def __init__(self, segment_files, filter_regex, invert_match=False):
```
Atticus1806, line 75: Can you move the typing to the function head?

vieting: Sure, I can do that. I basically adapted the `FilterSegmentsByListJob` which also has the typing like this (like most other jobs in this file). Should we also refactor them or just do it for the job that is added here?

corpus/filter.py
```diff
@@ -69,6 +71,45 @@ def run(self):
...
+        self.out_segment_path = MultiOutputPath(self, "segments.$(TASK)", self.out_single_segment_files)
+
+    def tasks(self):
+        yield Task("run", resume="run", mini_task=True)
+
+    def run(self):
+        pattern = re.compile(self.filter_regex)
+        for idx, segment_file in self.segment_files.items():
+            segment_list = [line.rstrip() for line in open(segment_file.get_path(), "r")]
+            non_empty = False
```
michelwi, line 99:
```suggestion
            output_empty = True
```
I would try to avoid double negative

corpus/filter.py
```diff
@@ -69,6 +71,45 @@ def run(self):
...
+            with open(self.out_single_segment_files[idx].get_path(), "wt") as segment_file_filtered:
+                for segment in segment_list:
+                    if (self.invert_match and pattern.match(segment)) or (
+                        not self.invert_match and not pattern.match(segment)
+                    ):
+                        segment_file_filtered.write(segment + "\n")
+                        non_empty = True
```
michelwi, line 106:
```suggestion
                        output_empty = False
```

corpus/filter.py
```diff
@@ -69,6 +71,45 @@ def run(self):
...
+            if not non_empty:
```
michelwi, line 107:
```suggestion
            if output_empty:
```
definitively more readable like this.

michelwi (commented):
I think this will be a useful addition.

corpus/filter.py
Atticus1806: If you want I dont think anyone would oppose more type changes :)

michelwi (approved):

vieting:
Thanks for the quick reviews @Atticus1806 and @michelwi!

PR #450, opened 2023-09-06
fix typing in corpus/filter.py

vieting:
Since the typing in `corpus/filter.py` came up in #449, I just updated it.

### DIFF
--- a/corpus/filter.py
+++ b/corpus/filter.py
@@ -12,12 +12,11 @@ import logging
 import numpy as np
 import re
 import xml.etree.cElementTree as ET
-from typing import Dict
+from typing import Dict, List, Optional, Union
 
-from i6_core.util import MultiOutputPath
-
+from i6_core import rasr
 from i6_core.lib import corpus
-from i6_core.util import chunks, uopen
+from i6_core.util import chunks, uopen, MultiOutputPath
 
 from sisyphus import *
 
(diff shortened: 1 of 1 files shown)


JackTemaki (changes requested):
Just one small question

Atticus1806 (approved):

corpus/filter.py
```diff
@@ -111,14 +109,20 @@ def run(self):
 
 
 class FilterSegmentsByAlignmentConfidenceJob(Job):
-    def __init__(self, alignment_logs, percentile, crp=None, plot=True, absolute_threshold=None):
+    def __init__(
+        self,
+        alignment_logs: Dict[int, Path],
+        percentile: float,
+        absolute_threshold: Optional[float] = None,
```
JackTemaki, line 116: I wondered why the apptek-tests failed, but you changed the order here which could be a potential cause. This is not a good idea unless there was a `*` before, enforcing only named parameters.

JackTemaki (changes requested):

corpus/filter.py
michelwi: please disregard the apptek test for the next few minutes, this is a problem on our side that should be fixed soon.

vieting: Oh right, I changed it to match the order in the docstring but it's certainly a better idea to do it the other way around.

corpus/filter.py
```diff
@@ -12,26 +12,24 @@
...
+from i6_core.util import chunks, uopen, MultiOutputPath
 
 from sisyphus import *
 
 Path = setup_path(__package__)
 
 
 class FilterSegmentsByListJob(Job):
-    def __init__(self, segment_files, filter_list, invert_match=False):
+    def __init__(self, segment_files: Dict[int, Path], filter_list: Union[List, Path], invert_match: bool = False):
```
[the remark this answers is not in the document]
JackTemaki: Nope, this is actually `str`...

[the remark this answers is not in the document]
vieting: Oh right, sorry. Somehow thought it would be multiple segment files then.

corpus/filter.py
michelwi: > please disregard the apptek test for the next few minutes

the problem was resolved and the test passes now. pls continue as you were :)

JackTemaki (approved):

PR #452, opened 2023-09-21
Onnx Feature Scorer

Marvin84:
Starting with the `apptek` version, I renamed some variables in order to decouple this from the old GMM feature scorer. Moreover, 2 internal rasr variables with their default values are added.

The outdated wrong FH label scorer should be deleted.

### DIFF
--- a/rasr/feature_scorer.py
+++ b/rasr/feature_scorer.py
@@ -16,7 +16,7 @@ Path = setup_path(__package__)
 
 import os
 
-from typing import Union, Dict
+from typing import Union, Dict, Optional
 
 from .config import *
 from i6_core.util import get_returnn_root
@@ -138,7 +138,7 @@ class OnnxFeatureScorer(FeatureScorer):
         io_map: Dict[str, str],
         label_log_posterior_scale: float = 1.0,
         label_prior_scale: float = 0.7,
-        label_log_prior_file: tk.Path = None,
+        label_log_prior_file: Optional[tk.Path] = None,
         apply_log_on_output: bool = False,
         negate_output: bool = True,
         intra_op_threads: int = 1,
--- a/rasr/feature_scorer.py
+++ b/rasr/feature_scorer.py
@@ -130,10 +130,10 @@ class PrecomputedHybridFeatureScorer(FeatureScorer):
 class OnnxFeatureScorer(FeatureScorer):
     def __init__(
         self,
+        *,
         mixtures,
         model,
         io_map,
-        *args,
         label_log_posterior_scale=1.0,
         label_prior_scale=0.7,
         label_log_prior_file=None,
--- a/rasr/feature_scorer.py
+++ b/rasr/feature_scorer.py
@@ -7,6 +7,7 @@ __all__ = [
     "ReturnnScorer",
     "InvAlignmentPassThroughFeatureScorer",
     "PrecomputedHybridFeatureScorer",
+    "OnnxFeatureScorer",
 ]
 
 from sisyphus import *
@@ -126,25 +127,50 @@ class PrecomputedHybridFeatureScorer(FeatureScorer):
         self.config.normalize_mixture_weights = False
 
 
-class TFLabelContextFeatureScorer(FeatureScorer):
+class OnnxFeatureScorer(rasr.FeatureScorer):
     def __init__(
         self,
-        fs_tf_config,
-        contextPriorFile,
-        diphonePriorFile,
-        prior_mixtures,
-        prior_scale,
+        mixtures,
+        model,
+        io_map,
+        *args,
+        label_log_posterior_scale=1.0,
+        label_prior_scale=0.7,
+        label_log_prior_file=None,
+        apply_log_on_output=False,
+        negate_output=True,
+        intra_op_threads=1,
+        inter_op_threads=1,
+        **kwargs,
     ):
-        super().__init__()
+        """
+        :param str mixtures: path to a *.mix file e.g. output of either EstimateMixturesJob or CreateDummyMixturesJob
+        :param str model: path of a model e.g. output of ExportPyTorchModelToOnnxJob
+        :param dict io_map: mapping between internal rasr identifiers and the model related input/output
+        :param float label_log_posterior_scale: scales for the log probability of a label e.g. 1.0 is recommended
+        :param float label_prior_scale: scale for the prior log probability of a label reasonable e.g. values in [0.1, 0.7] interval
+        :param str label_log_prior_file: xml file containing log prior probabilities e.g. estimated from the model via povey method
+        :param bool apply_log_on_output: whether to apply the log-function on the output, usefull if the model outputs softmax instead of log-softmax
+        :param bool negate_output: wheter negate output (because the model outputs log softmax and not negative log softmax
+        """
+        super().__init__(*args, **kwargs)
 
-        self.config = RasrConfig()
-        self.config.feature_scorer_type = "tf-label-context-scorer"
-        self.config.file = prior_mixtures
-        self.config.num_label_contexts = 46
-        self.config.prior_scale = prior_scale
-        self.config.context_prior = contextPriorFile
-        self.config.diphone_prior = diphonePriorFile
-        self.config.normalize_mixture_weights = False
-        self.config.loader = fs_tf_config.loader
-        self.config.input_map = fs_tf_config.input_map
-        self.config.output_map = fs_tf_config.output_map
+        self.config.feature_scorer_type = "onnx-feature-scorer"
+        self.config.file = mixtures
+        self.config.scale = label_log_posterior_scale
+        self.config.priori_scale = label_prior_scale
+        if label_log_prior_file is not None:
+            self.config.prior_file = label_log_prior_file
+
+        self.config.session.file = model
+
+        if label_log_prior_file:
+            self.config.apply_log_on_output = apply_log_on_output
+        if not negate_output:
+            self.config.negate_output = negate_output
+
+        self.post_config.session.intra_op_num_threads = intra_op_threads
+        self.post_config.session.inter_op_num_threads = inter_op_threads
+
+        for k, v in io_map.items():
+            self.config.io_map[k] = v
--- a/rasr/feature_scorer.py
+++ b/rasr/feature_scorer.py
@@ -127,7 +127,7 @@ class PrecomputedHybridFeatureScorer(FeatureScorer):
         self.config.normalize_mixture_weights = False
 
 
-class OnnxFeatureScorer(rasr.FeatureScorer):
+class OnnxFeatureScorer(FeatureScorer):
     def __init__(
         self,
         mixtures,
--- a/rasr/feature_scorer.py
+++ b/rasr/feature_scorer.py
@@ -133,14 +133,14 @@ class OnnxFeatureScorer(FeatureScorer):
     def __init__(
         self,
         *,
-        mixtures: Union[str, tk.Path],
-        model: Union[str, tk.Path],
+        mixtures: tk.Path,
+        model: tk.Path,
         io_map: Dict[str, str],
         label_log_posterior_scale: float = 1.0,
         label_prior_scale: float = 0.7,
-        label_log_prior_file: Union[str, tk.Path] = None,
-        apply_log_on_output: Bool = False,
-        negate_output: Bool = True,
+        label_log_prior_file: tk.Path = None,
+        apply_log_on_output: bool = False,
+        negate_output: bool = True,
         intra_op_threads: int = 1,
         inter_op_threads: int = 1,
         **kwargs,
@@ -170,7 +170,7 @@ class OnnxFeatureScorer(FeatureScorer):
 
         self.config.session.file = model
 
-        if label_log_prior_file:
+        if apply_log_on_output:
             self.config.apply_log_on_output = apply_log_on_output
         if not negate_output:
             self.config.negate_output = negate_output
--- a/rasr/feature_scorer.py
+++ b/rasr/feature_scorer.py
@@ -16,7 +16,7 @@ Path = setup_path(__package__)
 
 import os
 
-from typing import Union, Dict, Bool
+from typing import Union, Dict
 
 from .config import *
 from i6_core.util import get_returnn_root
--- a/rasr/feature_scorer.py
+++ b/rasr/feature_scorer.py
@@ -151,13 +151,13 @@ class OnnxFeatureScorer(FeatureScorer):
         :param io_map: mapping between internal rasr identifiers and the model related input/output. Default key values
                             are "features" and "output", and optionally "features-size", e.g.
                             io_map = {"features": "data", "output": "classes"}
-        :param float label_log_posterior_scale: scales for the log probability of a label e.g. 1.0 is recommended
-        :param float label_prior_scale: scale for the prior log probability of a label reasonable e.g. values in [0.1, 0.7] interval
-        :param str|tk.Path label_log_prior_file: xml file containing log prior probabilities e.g. estimated from the model via povey method
-        :param bool apply_log_on_output: whether to apply the log-function on the output, usefull if the model outputs softmax instead of log-softmax
-        :param bool negate_output: whether negate output (because the model outputs log softmax and not negative log softmax
-        :param int intra_op_threads: Onnxruntime session's number of parallel threads within each operator
-        :param int inter_op_threads: Onnxruntime session's number of parallel threads between operators used only for parallel execution mode
+        :param label_log_posterior_scale: scales for the log probability of a label e.g. 1.0 is recommended
+        :param label_prior_scale: scale for the prior log probability of a label reasonable e.g. values in [0.1, 0.7] interval
+        :param label_log_prior_file: xml file containing log prior probabilities e.g. estimated from the model via povey method
+        :param apply_log_on_output: whether to apply the log-function on the output, usefull if the model outputs softmax instead of log-softmax
+        :param negate_output: whether negate output (because the model outputs log softmax and not negative log softmax
+        :param intra_op_threads: Onnxruntime session's number of parallel threads within each operator
+        :param inter_op_threads: Onnxruntime session's number of parallel threads between operators used only for parallel execution mode
         """
         super().__init__(*args, **kwargs)
 
--- a/rasr/feature_scorer.py
+++ b/rasr/feature_scorer.py
@@ -137,7 +137,7 @@ class OnnxFeatureScorer(FeatureScorer):
         model: tk.Path,
         io_map: Dict[str, str],
         label_log_posterior_scale: float = 1.0,
-        label_prior_scale: float = 0.7,
+        label_prior_scale: float,
         label_log_prior_file: Optional[tk.Path] = None,
         apply_log_on_output: bool = False,
         negate_output: bool = True,
@@ -169,11 +169,8 @@ class OnnxFeatureScorer(FeatureScorer):
             self.config.prior_file = label_log_prior_file
 
         self.config.session.file = model
-
-        if apply_log_on_output:
-            self.config.apply_log_on_output = apply_log_on_output
-        if not negate_output:
-            self.config.negate_output = negate_output
+        self.config.apply_log_on_output = apply_log_on_output
+        self.config.negate_output = negate_output
 
         self.post_config.session.intra_op_num_threads = intra_op_threads
         self.post_config.session.inter_op_num_threads = inter_op_threads
--- a/rasr/feature_scorer.py
+++ b/rasr/feature_scorer.py
@@ -144,15 +144,16 @@ class OnnxFeatureScorer(FeatureScorer):
         **kwargs,
     ):
         """
-        :param str mixtures: path to a *.mix file e.g. output of either EstimateMixturesJob or CreateDummyMixturesJob
-        :param str model: path of a model e.g. output of ExportPyTorchModelToOnnxJob
+        :param str|tk.Path mixtures: path to a *.mix file e.g. output of either EstimateMixturesJob or CreateDummyMixturesJob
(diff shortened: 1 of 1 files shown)


Atticus1806:
I thought this job was about adding a new feature scorer not replacing and old?

Also I think rasr is not defined.

christophmluscher:
I agree with @Atticus1806 we should add a Featurescorer not replace

rasr/feature_scorer.py
```diff
@@ -126,25 +127,50 @@ def __init__(self, prior_mixtures, scale=1.0, priori_scale=0.0, prior_file=None)
         self.config.normalize_mixture_weights = False
 
 
-class TFLabelContextFeatureScorer(FeatureScorer):
+class OnnxFeatureScorer(rasr.FeatureScorer):
```
christophmluscher, line 130:
```suggestion
class OnnxFeatureScorer(FeatureScorer):
```

christophmluscher:
I would add onnx feature scorer and later refactor the FH label scorer

Marvin84:
> I thought this PR was about adding a new feature scorer not replacing and old?

> `I would add onnx feature scorer and later refactor the FH label scorer`

We discussed this some time ago and I wondered why this FS is still there. The code that is here would not work in any setup or branch of `rasr`. First of all, this is an outdated FH FS and has ended up in this module only by error. Secondly, the merging of the 24bit now we can have FH in the master of rasr however, there is still no feature scorer in the master. Therefore, this code should be deleted.

Marvin84:
> I agree with @Atticus1806 we should add a Featurescorer not replace
> 
> Or is the tf label scorer removed from RASR code?

It has never been in the master version. This is from 4 years ago. Refactoring it would not bring anything since the new FS is completely different

rasr/feature_scorer.py
```diff
@@ -126,25 +127,50 @@ def __init__(self, prior_mixtures, scale=1.0, priori_scale=0.0, prior_file=None)
...
+        negate_output=True,
+        intra_op_threads=1,
+        inter_op_threads=1,
+        **kwargs,
     ):
-        super().__init__()
+        """
+        :param str mixtures: path to a *.mix file e.g. output of either EstimateMixturesJob or CreateDummyMixturesJob
+        :param str model: path of a model e.g. output of ExportPyTorchModelToOnnxJob
+        :param dict io_map: mapping between internal rasr identifiers and the model related input/output
```
JackTemaki, line 149: You could add that the default expected keys are `features`, `output` and optionally `features-size`.

rasr/feature_scorer.py
```diff
@@ -126,25 +127,50 @@ def __init__(self, prior_mixtures, scale=1.0, priori_scale=0.0, prior_file=None)
...
         self,
-        fs_tf_config,
-        contextPriorFile,
-        diphonePriorFile,
-        prior_mixtures,
-        prior_scale,
+        mixtures,
+        model,
+        io_map,
+        *args,
```
JackTemaki, line 136: *args can be removed, the `FeatureScorer` anyway does not have any extra parameters. In addition, it would maybe be good to place a `*` after self to enforce all parameters to be passed as kwargs.

rasr/feature_scorer.py
```diff
@@ -126,25 +127,50 @@ def __init__(self, prior_mixtures, scale=1.0, priori_scale=0.0, prior_file=None)
...
+        :param str mixtures: path to a *.mix file e.g. output of either EstimateMixturesJob or CreateDummyMixturesJob
+        :param str model: path of a model e.g. output of ExportPyTorchModelToOnnxJob
+        :param dict io_map: mapping between internal rasr identifiers and the model related input/output
+        :param float label_log_posterior_scale: scales for the log probability of a label e.g. 1.0 is recommended
+        :param float label_prior_scale: scale for the prior log probability of a label reasonable e.g. values in [0.1, 0.7] interval
+        :param str label_log_prior_file: xml file containing log prior probabilities e.g. estimated from the model via povey method
+        :param bool apply_log_on_output: whether to apply the log-function on the output, usefull if the model outputs softmax instead of log-softmax
+        :param bool negate_output: wheter negate output (because the model outputs log softmax and not negative log softmax
+        """
+        super().__init__(*args, **kwargs)
```
JackTemaki, line 157: I think the parameters here can be dropped, they are not supplied with other FeatureScorers

rasr/feature_scorer.py
```diff
@@ -126,25 +127,51 @@ def __init__(self, prior_mixtures, scale=1.0, priori_scale=0.0, prior_file=None)
...
+        :param str mixtures: path to a *.mix file e.g. output of either EstimateMixturesJob or CreateDummyMixturesJob
+        :param str model: path of a model e.g. output of ExportPyTorchModelToOnnxJob
+        :param dict io_map: mapping between internal rasr identifiers and the model related input/output. Default key values
+                            are "features" and "output", and optionally "features-size"
+        :param float label_log_posterior_scale: scales for the log probability of a label e.g. 1.0 is recommended
+        :param float label_prior_scale: scale for the prior log probability of a label reasonable e.g. values in [0.1, 0.7] interval
+        :param str label_log_prior_file: xml file containing log prior probabilities e.g. estimated from the model via povey method
+        :param bool apply_log_on_output: whether to apply the log-function on the output, usefull if the model outputs softmax instead of log-softmax
+        :param bool negate_output: wheter negate output (because the model outputs log softmax and not negative log softmax
+        """
```
JackTemaki, line 156: Please also add the entries for the op_threads, you can e.g. just point to:

 https://onnxruntime.ai/docs/performance/tune-performance/threading.html

JackTemaki (changes requested):
some small things

rasr/feature_scorer.py
Marvin84: I think `apply_log_on_output ` is important to be there, and once this is here I would also put the other one for consistency. Here, we are giving the option of using the model directly with the softmax output and not with log softmax, it could be handy

rasr/feature_scorer.py
```diff
@@ -126,25 +127,53 @@ def __init__(self, prior_mixtures, scale=1.0, priori_scale=0.0, prior_file=None)
...
+        :param str mixtures: path to a *.mix file e.g. output of either EstimateMixturesJob or CreateDummyMixturesJob
+        :param str model: path of a model e.g. output of ExportPyTorchModelToOnnxJob
+        :param dict io_map: mapping between internal rasr identifiers and the model related input/output. Default key values
+                            are "features" and "output", and optionally "features-size"
+        :param float label_log_posterior_scale: scales for the log probability of a label e.g. 1.0 is recommended
+        :param float label_prior_scale: scale for the prior log probability of a label reasonable e.g. values in [0.1, 0.7] interval
+        :param str label_log_prior_file: xml file containing log prior probabilities e.g. estimated from the model via povey method
+        :param bool apply_log_on_output: whether to apply the log-function on the output, usefull if the model outputs softmax instead of log-softmax
+        :param bool negate_output: wheter negate output (because the model outputs log softmax and not negative log softmax
```
Atticus1806, line 155:
```suggestion
        :param bool negate_output: whether negate output (because the model outputs log softmax and not negative log softmax)
```

rasr/feature_scorer.py
```diff
@@ -126,25 +127,53 @@ def __init__(self, prior_mixtures, scale=1.0, priori_scale=0.0, prior_file=None)
...
+        intra_op_threads=1,
+        inter_op_threads=1,
+        **kwargs,
     ):
-        super().__init__()
+        """
+        :param str mixtures: path to a *.mix file e.g. output of either EstimateMixturesJob or CreateDummyMixturesJob
+        :param str model: path of a model e.g. output of ExportPyTorchModelToOnnxJob
+        :param dict io_map: mapping between internal rasr identifiers and the model related input/output. Default key values
+                            are "features" and "output", and optionally "features-size"
```
Atticus1806, line 150:
```suggestion
                            are "features" and "output", and optionally "features-size",
                            e.g. io_map = {"features": "data", "output": "classes"}
```

rasr/feature_scorer.py
```diff
@@ -126,25 +127,53 @@ def __init__(self, prior_mixtures, scale=1.0, priori_scale=0.0, prior_file=None)
         self.config.normalize_mixture_weights = False
 
 
-class TFLabelContextFeatureScorer(FeatureScorer):
+class OnnxFeatureScorer(FeatureScorer):
     def __init__(
```
Atticus1806, line 131: Please add the types in the function head so the IDE can use them.

rasr/feature_scorer.py
```diff
@@ -126,25 +127,53 @@ def __init__(self, prior_mixtures, scale=1.0, priori_scale=0.0, prior_file=None)
...
+        label_log_prior_file=None,
+        apply_log_on_output=False,
+        negate_output=True,
+        intra_op_threads=1,
+        inter_op_threads=1,
+        **kwargs,
     ):
-        super().__init__()
+        """
+        :param str mixtures: path to a *.mix file e.g. output of either EstimateMixturesJob or CreateDummyMixturesJob
```
Atticus1806, line 147: Is this really a string? I would expect smth. like `tk.Path` or `Union[tk.Path, str]`

rasr/feature_scorer.py
```diff
@@ -126,25 +127,53 @@ def __init__(self, prior_mixtures, scale=1.0, priori_scale=0.0, prior_file=None)
...
+        :param str model: path of a model e.g. output of ExportPyTorchModelToOnnxJob
```
Atticus1806, line 148: typing question, same as mixtures.

rasr/feature_scorer.py
```diff
@@ -126,25 +127,53 @@ def __init__(self, prior_mixtures, scale=1.0, priori_scale=0.0, prior_file=None)
...
+        :param dict io_map: mapping between internal rasr identifiers and the model related input/output. Default key values
```
Atticus1806, line 149: please also add the dict mapping, e.g. `Dict[str, str]`

rasr/feature_scorer.py
```diff
@@ -126,25 +127,53 @@ def __init__(self, prior_mixtures, scale=1.0, priori_scale=0.0, prior_file=None)
...
+                            are "features" and "output", and optionally "features-size"
+        :param float label_log_posterior_scale: scales for the log probability of a label e.g. 1.0 is recommended
+        :param float label_prior_scale: scale for the prior log probability of a label reasonable e.g. values in [0.1, 0.7] interval
+        :param str label_log_prior_file: xml file containing log prior probabilities e.g. estimated from the model via povey method
```
Atticus1806, line 153: same as mixtures.

rasr/feature_scorer.py
JackTemaki: Oh, I meant you can just but the link, but this is also fine.

rasr/feature_scorer.py
Atticus1806: I am not sure about the spacing here, maybe one or two spaces for the new line missing.

Atticus1806 (commented):
If its deprecated and all tests pass I am not against removing, I don't know the details I just noticed that this was different from what I expected :).

Just make sure with AppTek I guess (do their tests claim completeness? If so I guess its fine if tests pass)

Once that is clear and the typing is changeed I think it can be approved :)

rasr/feature_scorer.py
JackTemaki: I do not really understand you answer, what does this have to do with an empty constructor?

rasr/feature_scorer.py
Marvin84: what type you mean?

Atticus1806: E.g. `label_log_posterior_scale: float = 1.0`

Atticus1806: instead of doing typing in the docstring

Marvin84: I prefer doing docstring so that some explanation is provided

Atticus1806: Leave the docstring itself I am only talking about  the type hint.

Marvin84: I suggest you open a new PR where you do this consistently for all classes in this module, because it does not make sense to do this for one and not for all others

rasr/feature_scorer.py
Marvin84: I think example is clear here.

rasr/feature_scorer.py
```diff
@@ -126,25 +127,54 @@ def __init__(self, prior_mixtures, scale=1.0, priori_scale=0.0, prior_file=None)
...
+        label_log_prior_file=None,
+        apply_log_on_output=False,
+        negate_output=True,
+        intra_op_threads=1,
+        inter_op_threads=1,
+        **kwargs,
     ):
-        super().__init__()
+        """
+        :param str|tk.Path mixtures: path to a *.mix file e.g. output of either EstimateMixturesJob or CreateDummyMixturesJob
```
Atticus1806, line 147: @michelwi might correct me here, but I think using | for `Union` is deprecated.

rasr/feature_scorer.py
Marvin84: You mean for all classes in this module?

Atticus1806: Mainly for the new/updated code. Sometimes people then directly change it for the whole file, but this is up to you.

Marvin84: > It was decided

I do not recall this decision. What is exactly the decision? So I need to now add documentation for all classes and add typings for all of them in `feature_scorer.py`?

rasr/feature_scorer.py
Atticus1806: But this also just might be flavor, maybe someone like @JackTemaki knows?

rasr/feature_scorer.py
Marvin84: > Mainly for the new/updated code.

IMO, this does not make sense here, we have similar classes and they should all follow similar treatment. I can add the typing to this, but I suggest you open another PR for other classes

Atticus1806: No. I am just asking about moving the typing from the docstring to the function header of the updated class as it is done in every newer PR updating code in this repo.

Marvin84: ok, will do

Atticus1806: I dont think its that big of a problem to have both variants in one file, we can update them as we go along as we usually do.

JackTemaki: Benedikt is correct about this. Also, we should not use `str` for files but only `tk.Path`.

Atticus1806 (approved):
Thanks :)

rasr/feature_scorer.py
```diff
@@ -126,25 +129,54 @@ def __init__(self, prior_mixtures, scale=1.0, priori_scale=0.0, prior_file=None)
...
-        prior_mixtures,
-        prior_scale,
+        *,
+        mixtures: Union[str, tk.Path],
+        model: Union[str, tk.Path],
+        io_map: Dict[str, str],
+        label_log_posterior_scale: float = 1.0,
+        label_prior_scale: float = 0.7,
+        label_log_prior_file: Union[str, tk.Path] = None,
+        apply_log_on_output: Bool = False,
```
Atticus1806, line 142: Ahh I was too quick, I think this should just be `bool` without import.

rasr/feature_scorer.py
```diff
@@ -126,25 +129,54 @@ def __init__(self, prior_mixtures, scale=1.0, priori_scale=0.0, prior_file=None)
...
-        diphonePriorFile,
-        prior_mixtures,
-        prior_scale,
+        *,
+        mixtures: Union[str, tk.Path],
+        model: Union[str, tk.Path],
+        io_map: Dict[str, str],
+        label_log_posterior_scale: float = 1.0,
+        label_prior_scale: float = 0.7,
+        label_log_prior_file: Union[str, tk.Path] = None,
```
SimBe195, line 141: Wouldn't we want the parameters to be called the same as in other feature scorers and in the rasr config itself? I.e.  `mixtures` -> `prior_mixtures`, `label_prior_scale` -> `priori_scale` and `label_log_prior_file` -> `prior_file`.

rasr/feature_scorer.py
SimBe195: >  I think using | for Union is deprecated.

@Atticus1806 Actually it's the other way around. Using `|` is the more "modern" way. Type annotations with `|` instead of `typing.Union` were introduced in python 3.10. But since i6_core is currently based on python3.8 `Union` is correct here.

Atticus1806: Ahh good to know. But if 3.8 already "allows" it I would not mind keeping it like this.

rasr/feature_scorer.py
Marvin84: No, I explicitly explained above that I would like to make the variables more expressive and correct, by decoupling it from the old GMM based feature scorers

rasr/feature_scorer.py
```diff
@@ -126,25 +129,54 @@ def __init__(self, prior_mixtures, scale=1.0, priori_scale=0.0, prior_file=None)
...
+        self.config.scale = label_log_posterior_scale
+        self.config.priori_scale = label_prior_scale
+        if label_log_prior_file is not None:
+            self.config.prior_file = label_log_prior_file
+
+        self.config.session.file = model
+
+        if label_log_prior_file:
+            self.config.apply_log_on_output = apply_log_on_output
+        if not negate_output:
```
curufinwe, line 175: Why put this under an if-condition? If you have it you can just always write it into the config.

rasr/feature_scorer.py
```diff
@@ -126,25 +129,54 @@ def __init__(self, prior_mixtures, scale=1.0, priori_scale=0.0, prior_file=None)
...
+        self.config.feature_scorer_type = "onnx-feature-scorer"
+        self.config.file = mixtures
+        self.config.scale = label_log_posterior_scale
+        self.config.priori_scale = label_prior_scale
+        if label_log_prior_file is not None:
+            self.config.prior_file = label_log_prior_file
+
+        self.config.session.file = model
+
+        if label_log_prior_file:
```
curufinwe, line 173: are you sure you want to check for label_log_prior_file?

rasr/feature_scorer.py
```diff
@@ -126,25 +129,54 @@ def __init__(self, prior_mixtures, scale=1.0, priori_scale=0.0, prior_file=None)
...
-        contextPriorFile,
-        diphonePriorFile,
-        prior_mixtures,
-        prior_scale,
+        *,
+        mixtures: Union[str, tk.Path],
+        model: Union[str, tk.Path],
+        io_map: Dict[str, str],
+        label_log_posterior_scale: float = 1.0,
+        label_prior_scale: float = 0.7,
```
curufinwe, line 140: I know this was in the Apptek scorer, but actually providing a default here is maybe counter-productive.

rasr/feature_scorer.py
```diff
@@ -126,25 +129,54 @@ def __init__(self, prior_mixtures, scale=1.0, priori_scale=0.0, prior_file=None)
...
+class OnnxFeatureScorer(FeatureScorer):
     def __init__(
         self,
-        fs_tf_config,
-        contextPriorFile,
-        diphonePriorFile,
-        prior_mixtures,
-        prior_scale,
+        *,
+        mixtures: Union[str, tk.Path],
```
JackTemaki, line 136: Removing Union, leaving only tk.Path for all str/Path cases and Bool->bool, then I think from my side everything is good.

rasr/feature_scorer.py
Marvin84: no, this is an error, good catch. this was `apply_log_on_output`. Evil copy pasting

rasr/feature_scorer.py
SimBe195: Okay, I didn't see the conversation. If other people also agree then this renaming is fine for me.

rasr/feature_scorer.py
Marvin84: I agree. But I would still keep the default `label_log_posterior_scale`, messing around with that can be catastrophic, whereas prior scale might change just a little

curufinwe:
> Just make sure with AppTek I guess (do their tests claim completeness? If so I guess its fine if tests pass)

We never used the deleted scorer, so it should be fine.
The tests we have cover almost everything that is related to the training / running of our production systems so if they are green you should be ok.

SimBe195 (approved):

Marvin84:
> PS: the current failing check is because of this:

Yes, solved

rasr/feature_scorer.py
```diff
@@ -126,25 +129,54 @@ def __init__(self, prior_mixtures, scale=1.0, priori_scale=0.0, prior_file=None)
...
-        diphonePriorFile,
-        prior_mixtures,
-        prior_scale,
+        *,
+        mixtures: tk.Path,
+        model: tk.Path,
+        io_map: Dict[str, str],
+        label_log_posterior_scale: float = 1.0,
+        label_prior_scale: float = 0.7,
+        label_log_prior_file: tk.Path = None,
```
michelwi, line 141:
```suggestion
        label_log_prior_file: Optional[tk.Path] = None,
```

I suggest Optional as it can be None. Import Optional from typing

Marvin84: done

JackTemaki (approved):

rasr/feature_scorer.py
JackTemaki: This is new code added to i6_core, so there is no strict requirement to care about that. We anyway can keep our local version and just switch for new setups.

Marvin84: ok, changed

Marvin84:
@curufinwe I give you all passed tests and 3 approvals, do I get a merge from you? :)

NeoLegends (approved):
Nice

PR #453, opened 2023-09-27
Copy serialization code from i6_experiments

kuacakuaca:
Copied 
- [common/setups/serialization.py](https://github.com/rwth-i6/i6_experiments/blob/main/common/setups/serialization.py)
- [common/setups/returnn_pytorch](https://github.com/rwth-i6/i6_experiments/tree/main/common/setups/returnn_pytorch)


JackTemaki:
> So my understanding is that this is already reviewed in i6_experiments right? So I will approve without having a deeper look at the code.

Parts were reviewed, but not everything, I will do some further cleanup here on this branch.

Atticus1806 (changes requested):
Then ill change to request changes for now

albertz:
I wonder a bit about `class Collection`, which is both in [returnn_pytorch/serialization.py](https://github.com/rwth-i6/i6_core/pull/453/files#diff-a58955ede41f08f690a17ea03f2ebab6de1a82fcb512ccaef14a3c00c95c951c) and [serialization.py](https://github.com/rwth-i6/i6_core/pull/453/files#diff-809aa84d81dc86f3cb5e2cdf10df13c5e1f9166f08f79ffaf7e53e60c544d016), and which is quite similar, only that the one in returnn_pytorch is more extended and has a number of things hardcoded in it. I see multiple problems:

* Having the same class with the same name and with seemingly the same functionality is confusing.
* Having things hardcoded (adding `import os`, `import sys`, `sys.path.insert(...)` etc) is bad. This should be a choice by the user.
* `returnn_pytorch.Collection` is not PyTorch specific but also generic.
* The base `Collection` class can do exactly the same as `returnn_pytorch.Collection` but is simpler. The hardcoded extra content can be added via `NonhashedCode`. The extra optional logic for copying the packages can be done via other custom `SerializerObject`s, e.g. some `MakeLocalPackageCopy` objects, which do the copying.

albertz:
> @kuacakuaca the location you put the code is not optimal, I will do some changes now.

Yes, I also wanted to say that. Not sure where it should be, maybe in an own package `serialization`. But the current location is bad.

kuacakuaca:
> @kuacakuaca the location you put the code is not optimal, I will do some changes now.

feel free to change.

JackTemaki:
> I wonder a bit about `class Collection`, which is both in [returnn_pytorch/serialization.py](https://github.com/rwth-i6/i6_core/pull/453/files#diff-a58955ede41f08f690a17ea03f2ebab6de1a82fcb512ccaef14a3c00c95c951c) and [serialization.py](https://github.com/rwth-i6/i6_core/pull/453/files#diff-809aa84d81dc86f3cb5e2cdf10df13c5e1f9166f08f79ffaf7e53e60c544d016), and which is quite similar, only that the one in returnn_pytorch is more extended and has a number of things hardcoded in it. I see multiple problems:
> 
>     * Having the same class with the same name and with seemingly the same functionality is confusing.
> 
>     * Having things hardcoded (adding `import os`, `import sys`, `sys.path.insert(...)` etc) is bad. This should be a choice by the user.
> 
>     * `returnn_pytorch.Collection` is not PyTorch specific but also generic.
> 
>     * The base `Collection` class can do exactly the same as `returnn_pytorch.Collection` but is simpler. The hardcoded extra content can be added via `NonhashedCode`. The extra optional logic for copying the packages can be done via other custom `SerializerObject`s, e.g. some `MakeLocalPackageCopy` objects, which do the copying.

Yes this can be re-written even without breaking hashes, I would just remove the second Collection completely. It was not intended to be copied here, but rather just a WIP class for (my) desired behavior of the copy mechanism.

@albertz how are you currently setting the sys.path for the recipes in the config? Just by a manual code block (e.g. NonhashedCode)?

JackTemaki:
> > @kuacakuaca the location you put the code is not optimal, I will do some changes now.
> 
> Yes, I also wanted to say that. Not sure where it should be, maybe in an own package `serialization`. But the current location is bad.

Yes, I would create a new high-level package `serialization`, as most/all of the code is also not really RETURNN specific

albertz:
> how are you currently setting the sys.path for the recipes in the config? Just by a manual code block (e.g. NonhashedCode)?

[Here](https://github.com/rwth-i6/i6_experiments/blob/0be8662664e42b4d456758af488fc87a35d3abf4/users/zeyer/experiments/exp2023_04_25_rf/train.py#L86) is an example:
```python
    returnn_train_config = ReturnnConfig(
        returnn_train_config_dict,
        python_epilog=[
            serialization.Collection(
                [
                    serialization.NonhashedCode(get_import_py_code()),
                    serialization.NonhashedCode(
                        nn.ReturnnConfigSerializer.get_base_extern_data_py_code_str_direct(extern_data_raw)
                    ),
                    serialization.Import(model_def, "_model_def", ignore_import_as_for_hash=True),
                    serialization.Import(_returnn_v2_get_model, "get_model"),
                    serialization.Import(train_def, "_train_def", ignore_import_as_for_hash=True),
                    serialization.Import(_returnn_v2_train_step, "train_step"),
                    serialization.ExplicitHash({"version": 1, "extra": extra_hash}),
                    serialization.PythonEnlargeStackWorkaroundNonhashedCode,
                    serialization.PythonCacheManagerFunctionNonhashedCode,
                    serialization.PythonModelineNonhashedCode,
                ]
                + list(epilog)
            )
        ],
        post_config=dict(  # not hashed
            log_batch_size=True,
            cleanup_old_models=True,
            # ...
        ),
        sort_config=False,
    )
```

And that `get_import_py_code` is defined [here](https://github.com/rwth-i6/i6_experiments/blob/0be8662664e42b4d456758af488fc87a35d3abf4/users/zeyer/utils/serialization.py#L11C1-L25C28) as:
```python
def get_import_py_code() -> str:
    content = ["import os\nimport sys\n"]
    root_path = os.path.join(gs.BASE_DIR, gs.RECIPE_PREFIX)
    content.append(f"sys.path.insert(0, {root_path!r})\n")
    content.append(f"sys.path.insert(1, {os.path.dirname(sisyphus.__path__[0])!r})\n")
    return "".join(content)
```

JackTemaki:
> And that `get_import_py_code` is defined [here](https://github.com/rwth-i6/i6_experiments/blob/0be8662664e42b4d456758af488fc87a35d3abf4/users/zeyer/utils/serialization.py#L11C1-L25C28) as:
> 
> ```python
> def get_import_py_code() -> str:
>     content = ["import os\nimport sys\n"]
>     root_path = os.path.join(gs.BASE_DIR, gs.RECIPE_PREFIX)
>     content.append(f"sys.path.insert(0, {root_path!r})\n")
>     content.append(f"sys.path.insert(1, {os.path.dirname(sisyphus.__path__[0])!r})\n")
>     return "".join(content)
> ```

Ok thanks! As this is basically the same as in my hardcoded Collection, and you cannot really build proper setups without this, I would suggest we also add this code here in the PR. as e.g. `SisyphusImportNonhashedCode` or so.

JackTemaki:
> > And that `get_import_py_code` is defined [here](https://github.com/rwth-i6/i6_experiments/blob/0be8662664e42b4d456758af488fc87a35d3abf4/users/zeyer/utils/serialization.py#L11C1-L25C28) as:
> > ```python
> > def get_import_py_code() -> str:
> >     content = ["import os\nimport sys\n"]
> >     root_path = os.path.join(gs.BASE_DIR, gs.RECIPE_PREFIX)
> >     content.append(f"sys.path.insert(0, {root_path!r})\n")
> >     content.append(f"sys.path.insert(1, {os.path.dirname(sisyphus.__path__[0])!r})\n")
> >     return "".join(content)
> > ```
> 
> Ok thanks! As this is basically the same as in my hardcoded Collection, and you cannot really build proper setups without this, I would suggest we also add this code here in the PR. as e.g. `SisyphusImportNonhashedCode` or so.

Or as a separate PR, this would be better I think.

serialization/__init__.py
```diff
@@ -0,0 +1,2 @@
+from base import *
+from pytorch import *
```
albertz, line 2: This code is wrong. Did you test this? I wonder how this could have worked without an exception.
```suggestion
from .base import *
from .pytorch import *
```
However, I'm not sure if we really want to import everything here?
Maybe only the stuff from `base`?
```suggestion
from .base import *
```

serialization/pytorch.py
albertz, line 1: I think naming the file `torch.py` would be more standard?

serialization/pytorch.py
```diff
@@ -0,0 +1,126 @@
...
+import torch
+from sisyphus.delayed_ops import DelayedBase
+
+if TYPE_CHECKING:
+    from i6_models.config import ModelConfiguration
+
+from .base import Call, Import
+
+
+def build_config_constructor_serializers(
```
albertz, line 22: If this is imported directly in the package `__init__`, the function name should have `torch` in the name. If this is not imported directly in the package, it's ok without `torch`.

But then, this is not just about PyTorch, but also specifically about i6_models. So I think the function name should make this more clear.
```suggestion
def build_i6_models_config_constructor_serializers(
```

serialization/__init__.py
JackTemaki: Ah sure this is wrong

albertz (approved):

serialization/base.py
```diff
@@ -0,0 +1,469 @@
...
+
+class CodeFromFunction(SerializerObject):
+    """
+    Insert code from function.
+    """
+
+    def __init__(self, name: str, func: FunctionType, *, hash_full_python_code: bool = False):
+        """
+        :param name: name of the function as exposed in the config
+        :param func:
```
christophmluscher, line 248: missing docstring or on purpose?

serialization/base.py
```diff
@@ -0,0 +1,469 @@
...
+    # noinspection PyShadowingBuiltins
+    def __init__(self, hash: Any):
+        super().__init__()
+        self.hash = hash
+
+    def get(self) -> str:
+        """get"""
+        return ""
+
+    def _sis_hash(self):
```
christophmluscher, line 369:
```suggestion
    def _sis_hash(self) -> bytes:
```

serialization/base.py
```diff
@@ -0,0 +1,469 @@
...
+        super().__init__()
+        self.filename = filename
+        self.hash_full_content = hash_full_content
+
+    def get(self):
+        """get"""
+        with uopen(self.filename, "rt") as f:
+            return f.read()
+
+    def _sis_hash(self):
```
christophmluscher, line 347:
```suggestion
    def _sis_hash(self) -> bytes:
```

serialization/base.py
```diff
@@ -0,0 +1,469 @@
...
+    def __init__(self, filename: tk.Path, hash_full_content: bool = False):
+        """
+        :param filename:
+        :param hash_full_content: False -> hash filename, True -> hash content (but not filename)
+        """
+        super().__init__()
+        self.filename = filename
+        self.hash_full_content = hash_full_content
+
+    def get(self):
```
christophmluscher, line 342:
```suggestion
    def get(self) -> str:
```

serialization/base.py
```diff
@@ -0,0 +1,469 @@
...
+class NonhashedCodeFromFile(_NonhashedSerializerObject):
+    """
+    Insert code from file content which is not hashed (neither the file name nor the content).
+    """
+
+    def __init__(self, filename: tk.Path):
+        super().__init__()
+        self.filename = filename
+
+    def get(self):
```
christophmluscher, line 322:
```suggestion
    def get(self) -> str:
```

serialization/base.py
```diff
@@ -0,0 +1,469 @@
...
+class NonhashedCode(_NonhashedSerializerObject):
+    """
+    Insert code from raw string which is not hashed.
+    """
+
+    def __init__(self, code: Union[str, tk.Path]):
+        super().__init__()
+        self.code = code
+
+    def get(self):
```
christophmluscher, line 308:
```suggestion
    def get(self) -> str:
```

serialization/base.py
```diff
@@ -0,0 +1,469 @@
...
+                    f"    return {self.func.__name__}\n",
+                    f"{self.name} = _{self.name}_{code_hash}()\n",
+                ]
+            )
+
+    def get(self):
+        """get"""
+        return self._code
+
+    def _sis_hash(self):
```
christophmluscher, line 280:
```suggestion
    def _sis_hash(self) -> bytes:
```

serialization/base.py
```diff
@@ -0,0 +1,469 @@
...
+                [
+                    f"def _{self.name}_{code_hash}():\n",
+                    textwrap.indent(self._func_code, "    "),
+                    "\n",
+                    f"    return {self.func.__name__}\n",
+                    f"{self.name} = _{self.name}_{code_hash}()\n",
+                ]
+            )
+
+    def get(self):
```
christophmluscher, line 276:
```suggestion
    def get(self) -> str:
```

serialization/base.py
```diff
@@ -0,0 +1,469 @@
...
+    """
+
+    def __init__(self, import_path: tk.Path):
+        super().__init__()
+        self.import_path = import_path
+
+    def get(self) -> str:
+        return f'sys.path.insert(0, "{self.import_path.get()}")\n'
+
+    def _sis_hash(self):
```
christophmluscher, line 236:
```suggestion
    def _sis_hash(self) -> bytes:
```

serialization/base.py
```diff
@@ -0,0 +1,469 @@
...
+        return string.Template(self.TEMPLATE).substitute(
+            {
+                "KWARGS": str(instanciate_delayed(arguments)),
+                "IMPORT_PATH": self.module,
+                "IMPORT_NAME": self.object_name,
+                "OBJECT_NAME": self.import_as if self.import_as is not None else self.object_name,
+            }
+        )
+
+    def _sis_hash(self):
```
christophmluscher, line 217:
```suggestion
    def _sis_hash(self) -> bytes:
```

serialization/base.py
```diff
@@ -0,0 +1,469 @@
...
+        if self.import_as:
+            return f"from {self.module} import {self.object_name} as {self.import_as}\n"
+        return f"from {self.module} import {self.object_name}\n"
+
+    def _sis_hash(self):
+        if self.import_as and not self.ignore_import_as_for_hash:
+            return sis_hash_helper({"code_object": self.code_object, "import_as": self.import_as})
+        return sis_hash_helper(self.code_object)
+
+    def __hash__(self):
```
christophmluscher, line 148:
```suggestion
    def __hash__(self) -> int:
```

serialization/base.py
```diff
@@ -0,0 +1,469 @@
...
+        if self.import_as and not self.ignore_import_as_for_hash:
+            return hash({"code_object": self.code_object, "import_as": self.import_as})
```
christophmluscher, line 150: maybe I am understanding something wrong.. but this would be the builtin hash func, yes? AFAIK that does not take dicts as an arg.. so is this `short_hash` or am I missing something?

serialization/base.py
```diff
@@ -0,0 +1,469 @@
...
+        return hash(self.code_object)
```
christophmluscher, line 151: here the same

serialization/base.py
christophmluscher:
```suggestion
            return short_hash({"code_object": self.code_object, "import_as": self.import_as})
```

serialization/base.py
christophmluscher:
```suggestion
        return short_hash(self.code_object)
```

serialization/base.py
```diff
@@ -0,0 +1,469 @@
...
+
```
curufinwe, line 152:
```suggestion
```

christophmluscher (approved):

curufinwe (approved):

serialization/base.py
```diff
@@ -0,0 +1,464 @@
...
+        }
+        return sis_hash_helper(h)
+
+
+class Import(SerializerObject):
+    """
+    A class to indicate a module or function that should be imported within the returnn config
+
+    When passed to the ReturnnCommonSerializer it will automatically detect the local package in case of
+    `make_local_package_copy=True`, unless specific package paths are given.
```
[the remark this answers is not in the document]
JackTemaki: Ah yes, this only exists in some of the more complex `Collection` implementations in `i6_experiments/common`. We should either move that functionality or remove the comment here.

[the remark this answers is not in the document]
albertz: I think it was discussed before that we wanted to remove this to simplify it.

PR #454, opened 2023-10-05
Fix the MergeAudioDirsJob for fairseq training

Judyxujj:
With the current version, if file name from two directories are the same, there will be `file already exists error` when doing the symlink `os.symlink(file, dst)` since the dst are the same
```
a.wav -> dir_path_1/a.wav
a.wav -> dir_path_2/a.wav
```
This PR fixes this issue by renaming dst for the second file
```
a.wav -> dir_path_1/a.wav
a_2.wav -> dir_path_2/a.wav
```

### DIFF
--- a/fairseq/manifest.py
+++ b/fairseq/manifest.py
@@ -43,7 +43,19 @@ class MergeAudioDirsJob(Job):
                     continue
                 base_name = os.path.basename(file)
                 dst = os.path.join(self.out_audio_dir_path, base_name)
-                os.symlink(file, dst)
+                if not os.path.exists(dst) and os.path.realpath(dst) != file:
+                    os.symlink(file, dst)
+                elif os.path.exists(dst) and os.path.realpath(dst) != file:
+                    i = 2
+                    new_dst = f"{os.path.splitext(dst)[0]}_{i}.{self.file_extension}"
+                    while os.path.exists(new_dst):
+                        if os.path.realpath(new_dst) == file:
+                            break
+                        else:
+                            i += 1
+                        new_dst = f"{os.path.splitext(dst)[0]}_{i}.{self.file_extension}"
+                    if os.path.realpath(new_dst) != file:
+                        os.symlink(file, new_dst)
 
 
 class CreateManifestJob(Job):
--- a/fairseq/manifest.py
+++ b/fairseq/manifest.py
@@ -57,7 +57,7 @@ class MergeAudioDirsJob(Job):
                     else:
                         i += 1
                     new_dst = f"{os.path.splitext(dst)[0]}_{i}.{self.file_extension}"
-                if os.path.realpath(new_dst) != file:
+                if not os.path.exists(dst) and os.path.realpath(new_dst) != file:
                     os.symlink(file, new_dst)
 
 
--- a/fairseq/manifest.py
+++ b/fairseq/manifest.py
@@ -52,6 +52,7 @@ class MergeAudioDirsJob(Job):
             while not creation_complete:
                 while os.path.exists(dst):
                     if os.path.realpath(dst) == file:
+                        creation_complete = True
                         break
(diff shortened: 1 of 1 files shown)


christophmluscher (approved):

michelwi (changes requested):

fairseq/manifest.py
```diff
@@ -29,21 +30,39 @@ def __init__(
         self.audio_dir_paths = audio_dir_paths
         self.file_extension = file_extension
         self.path_must_contain = path_must_contain
+        self.concurrent = len(audio_dir_paths)
 
         self.out_audio_dir_path = self.output_path("audio", directory=True)
 
+        self.rqmt = {"time": 8, "cpu": 4, "mem": 4}
```
michelwi, line 37: The job does some symlinking. This should be fast. Why do we go from `mini_task=True` to 4 cpu cores for 8 hours?

fairseq/manifest.py
```diff
@@ -29,21 +30,39 @@ def __init__(
...
+                        break
+                    dst = f"{os.path.splitext(dst)[0]}_{i}.{self.file_extension}"
+                    i += 1
+                else:
+                    try:
+                        os.symlink(file, dst)
+                        creation_complete = True
+                    except OSError as err:
+                        if err.errno != errno.EEXIST:
+                            raise err
```
michelwi, line 65: this double while loop looks unnecessary complicated, did it not work with the else as I indicated? We do not need to have an else here, I would prefer the old solution over this 🙈 

Also, in this line
```
dst = f"{os.path.splitext(dst)[0]}_{i}.{self.file_extension}"
```
would you not append an underscore in every iteration?
```
a.wav
a_2.wav
a_2_3.wav
...
```

fairseq/manifest.py
```diff
@@ -29,21 +30,39 @@ def __init__(
...
+                        creation_complete = True
+                        break
+                    dst = f"{os.path.splitext(dst)[0]}_{i}.{self.file_extension}"
+                    i += 1
+                else:
+                    try:
+                        os.symlink(file, dst)
+                        creation_complete = True
+                    except OSError as err:
+                        if err.errno != errno.EEXIST:
```
michelwi, line 64: Why would there be an `OSError` with `errno.EEXIST` (i.e. `FileExistsError`)? You have just completed the while loop and this implies that it did not exist.

Oh, you are running several tasks in parallel and there could be race conditions. if `folder1/a.wav` is in   `self.audio_dir_paths[0]` and `folder2/a.wav` is in `self.audio_dir_paths[1]`. But then the correct solution is not to ignore the error but make a thread save implementation.... which I now realize is the purpose of the double while loop..

Ok. nevermind my previous comment then. (But do we really need parallelism for symlinking a bunch of files?)

Except for the issue with the ever expanding underscores. That should probably be fixed.

fairseq/manifest.py
Judyxujj: For the wav2vec 2 training, after the audio preprocessing, we had around 13 millions of small files, that is why I do not parallelism here

fairseq/manifest.py
Judyxujj: what is the issue for the ever expanding underscores, sorry I couldn't find the relevant comments, could you please refer me to that?

michelwi:
```
while os.path.exists(dst):
    dst = f"{os.path.splitext(dst)[0]}_{i}.{self.file_extension}"
```
this will keep appending underscores to the filename:
```
a.wav
a_2.wav
a_2_3.wav
...
```

curufinwe:
I don't think we will need this anymore, but the question remains: are the proposed changes better than the current state even if not perfect @michelwi ? If yes I would suggest we merge and close.

michelwi (approved):
I think it is a simple fix. But one would need to test it and I don't have a setup for that. 
And I don't care too much, so I don't mind.

PR #455, opened 2023-10-17
enable partial check out of only some files from a repo

moothiringote:


christophmluscher (approved):

tools/git.py
```diff
@@ -31,27 +33,44 @@ def __init__(
...
             args.extend(["-b", self.branch])
         repository_dir = self.out_repository.get_path()
         args += [repository_dir]
         logging.info("running command: %s" % " ".join(args))
         sp.run(args, check=True)
 
-        if self.commit is not None:
+        if self.files_to_checkout is not None:
+            args = ["git", "checkout"]
+            commit = self.commit if self.commit is not None else ""
```
Icemole, line 63: In case that `self.commit` is not provided, I have tried the following:
```
$ git clone --no-checkout https://github.com/rwth-i6/i6_core.git
$ cd i6_core
$ git checkout -- .gitignore
```
The result is:
```
error: pathspec '.gitignore' did not match any file(s) known to git.
```
From experimenting, the correct checkout seems to be done by adding any branch name right after checkout, but it must be a valid one, like:
```
git checkout main -- .gitignore
```
Which properly tracks the `.gitignore` file. Therefore, the straightforward solution would be the following:
```suggestion
            commit = self.commit if self.commit is not None else "HEAD"
```

An alternative would be to clone the entire repository normally by not using `--no-checkout` above in order to have the whole repository available, and the condition to be checked in this conditional branch would be:
```
if self.files_to_checkout is not None and self.commit is not None:
```

tools/git.py
```diff
@@ -31,27 +33,44 @@ def __init__(
         :param commit: Git commit hash
         :param checkout_folder_name: Name of the output path repository folder
         :param clone_submodules: Flag to clone submodules if set to True
+        :param files_to_checkout: List of files to be checked out sparsely. If not set, the entire repo is checked out (default behaviour).
```
Icemole, line 36: Would the user need to use multiple files from different commits at some point? Maybe this parameter can be specified as a dictionary of commits to lists of files. Would that be confusing for the user?

tools/git.py
Icemole: Requesting changes only because of this.

Icemole (changes requested):

tools/git.py
moothiringote: Opos, I didn't test that condition - good that you did. I'll go for the first option as the second one would still check out the whole repo - something that should be avoided if files_to_checkout is set.

tools/git.py
moothiringote: I think that is a bit too much for my use-case because all the commits are specified at repo-level. I will not add this now, but if everyone thinks it is useful, please let me know.

Icemole (approved):

PR #458, opened 2023-10-27
Deduplication helpers for lexicon modification jobs

Icemole:
This PR introduces some improvements to how the lemma duplication is handled in `lexicon.conversion.LexiconUniqueOrthJob` and `lexicon.modification.MergeLexiconJob`.

For the former, I added a flag that makes special lemmas go through the unique lemma creation. In my view this is the most relevant part of the PR.

For the latter, I added a flag to deduplicate any duplicated lemmas that might appear as a result of the merging. This change might be debatable since `MergeLexiconJob` combined with `LexiconUniqueOrthJob` has the same result. I can remove this flag if it's too redundant; let me know about this.

Finally, I also added a `Lemma._equals` method and overwrote `==` and `!=` for the `Lemma` class, which I think might be useful in the long term (I actually used this in `MergeLexiconJob`).

### DIFF
--- a/lexicon/conversion.py
+++ b/lexicon/conversion.py
@@ -99,9 +99,9 @@ class FilterLexiconByWordListJob(Job):
 class LexiconUniqueOrthJob(Job):
     """Merge lemmata with the same orthography."""
 
-    __sis_hash_exclude__ = {"merge_multi_orths_lemmata": False}
+    __sis_hash_exclude__ = {"merge_multi_orths_lemmata": False, "deduplicate_special_lemmata": False}
 
-    def __init__(self, bliss_lexicon, merge_multi_orths_lemmata=False):
+    def __init__(self, bliss_lexicon, merge_multi_orths_lemmata=False, deduplicate_special_lemmata=False):
         """
         :param tk.Path bliss_lexicon: lexicon file to be handeled
         :param bool merge_multi_orths_lemmata: if True, also lemmata containing
@@ -119,11 +119,14 @@ class LexiconUniqueOrthJob(Job):
                     ** having a synt <=> synt is not None
                 this could lead to INFORMATION LOSS if there are several
                 different synt token sequences in the to-be-merged lemmata
+        :param bool deduplicate_special_lemmata: if True, special lemmata will also be considered
+        in the unique process.
         """
         self.set_vis_name("Make Lexicon Orths Unique")
 
         self.bliss_lexicon = bliss_lexicon
         self.merge_multi_orths_lemmata = merge_multi_orths_lemmata
+        self.deduplicate_special_lemmata = deduplicate_special_lemmata
 
         self.out_bliss_lexicon = self.output_path(os.path.basename(tk.uncached_path(bliss_lexicon)), cached=True)
 
@@ -137,7 +140,7 @@ class LexiconUniqueOrthJob(Job):
         orth2lemmata = collections.defaultdict(list)
 
         for lemma in lex.lemmata:
-            if lemma.special:
+            if lemma.special and not self.deduplicate_special_lemmata:
                 continue
             num_orths = len(lemma.orth)
             if num_orths < 1:
--- a/lexicon/modification.py
+++ b/lexicon/modification.py
@@ -130,16 +130,22 @@ class MergeLexiconJob(Job):
     will create a new lexicon that might be incompatible to previously generated alignments.
     """
 
-    def __init__(self, bliss_lexica, sort_phonemes=False, sort_lemmata=False, compressed=True):
+    __sis_hash_exclude__ = {"deduplicate_lemmata": False}
+
+    def __init__(
+        self, bliss_lexica, sort_phonemes=False, sort_lemmata=False, compressed=True, deduplicate_lemmata=False
+    ):
         """
         :param list[Path] bliss_lexica: list of bliss lexicon files (plain or gz)
         :param bool sort_phonemes: sort phoneme inventory alphabetically
         :param bool sort_lemmata: sort lemmata alphabetically based on first orth entry
         :param bool compressed: compress final lexicon
+        :param bool deduplicate_lemmata: whether to deduplicate lemmatas, only applied when sort_lemmata=True
         """
         self.lexica = bliss_lexica
         self.sort_phonemes = sort_phonemes
         self.sort_lemmata = sort_lemmata
+        self.deduplicate_lemmata = deduplicate_lemmata
 
         self.out_bliss_lexicon = self.output_path("lexicon.xml.gz" if compressed else "lexicon.xml")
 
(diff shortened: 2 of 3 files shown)


lexicon/conversion.py
```diff
@@ -119,11 +119,14 @@ def __init__(self, bliss_lexicon, merge_multi_orths_lemmata=False):
                     ** having a synt <=> synt is not None
                 this could lead to INFORMATION LOSS if there are several
                 different synt token sequences in the to-be-merged lemmata
+        :param bool deduplicate_special_lemmata: if True, special lemmata will also be considered
+        in the unique process.
```
albertz, line 123:
```suggestion
            in the unique process.
```

lexicon/modification.py
```diff
@@ -178,7 +184,12 @@ def run(self):
                 for lemma in lex.lemmata:
                     # sort by first orth entry
                     orth_key = lemma.orth[0] if lemma.orth else ""
+                    if self.deduplicate_lemmata:
+                        # don't add the lemma when there's already an equal lemma
+                        if len(lemma_dict[orth_key]) > 0 and lemma == lemma_dict[orth_key][0]:
+                            continue
                     lemma_dict[orth_key].append(lemma)
+            print(lemma_dict)
```
albertz, line 192: Was this just for debugging, or is this intended?
```suggestion
```

lib/lexicon.py
```diff
@@ -104,6 +107,42 @@ def from_element(cls, e):
         synt = None if not synt else synt[0]
         return Lemma(orth, phon, synt, eval, special)
 
+    def _equals(self, other: Lemma, same_order: bool = True) -> bool:
```
albertz, line 110:
```suggestion
    def _equals(self, other: Lemma, *, same_order: bool = True) -> bool:
```

albertz, line 110: Why is the default `same_order=True` here, but the only usage of this function actually uses `same_order=False`?

The default should reflect the most reasonable expected common usage, or if this might be ambiguous, there should not be a default.

lib/lexicon.py
```diff
@@ -104,6 +107,42 @@ def from_element(cls, e):
...
+            return (
+                set(self.orth) == set(other.orth)
+                and set(self.phon) == set(other.phon)
+                and self.special == other.special
+                and equal_synt
+                and set(itertools.chain(*self.eval)) == set(itertools.chain(*other.eval))
+            )
+
+    def __eq__(self, other: Lemma) -> bool:
+        return self._equals(other, same_order=False)
```
albertz, line 141: Is there a reason you want `same_order=False` here?

It probably depends on the use case of `__eq__` here (what actually is your use case?) but it feels kind of arbitrary.

lexicon/modification.py
Icemole: Just debugging :sweat_smile:

lib/lexicon.py
Icemole: This should be discussed, but in my view a lemma is equal to another even if it has different orth order, or different pronunciation order. This is why it's set to `False` here.

lib/lexicon.py
Icemole: You're right, probably the most common usage of equality would be checking without explicit order. At least I think it that way. What would other reviewers expect from the default of this function?

lib/lexicon.py
albertz: I think it depends on the use case. Or if you say, only `same_order=False` makes sense anyway, then why is this an option at all for `_equals`.

What actually is your use case?

Icemole: My use case is the following: in `MergeLexiconJob` I added functionality (present in this PR) that will remove "duplicated" lemmas. According to my definition, duplicated lemmas are those lemmas that have the same orthographies and pronunciations. Note that my definition implies no ordering, so I would be fine with `same_order=False`.

However, I wanted to give the freedom to let the user be able to decide whether the order in their orths and prons matters or not. Maybe some users might want to store the orths and prons of a lexicon in lexicographical order, and thus the order would matter. Maybe some other users already have a specific ordering, and therefore the addition of a lemma with a different order helps them notice that something's wrong with their pipeline.

Summarizing: `same_order=False` is enough for my needs/use case, but I added this flag since I was unsure of the _other_ use cases for this function, since each user might have a particular use case.

If you think it won't be used and only `==`/`!=` will be used as I have been using it (with `same_order=False`), let me know. I won't use this flag, and if keeping it or removing it will spark a debate, I'd rather just remove it now and let another user, who might find it more necessary, implement it.

albertz: So, if you think there is no clear non-ambiguous definition of an equal-relation, thus having such flag makes sense, then I'm not sure if defining `__eq__` makes sense here. Or usually `__eq__` should always be most restrictive, i.e. use `same_order=True`.

Now `__eq__` is very arbitrary, supporting exactly your case, but as you say yourself, there might be other cases where the equality should only be with `same_order=True`.

lexicon/conversion.py
```diff
@@ -119,11 +119,14 @@ def __init__(self, bliss_lexicon, merge_multi_orths_lemmata=False):
                     ** having a synt <=> synt is not None
                 this could lead to INFORMATION LOSS if there are several
                 different synt token sequences in the to-be-merged lemmata
+        :param bool deduplicate_special_lemmata: if True, special lemmata will also be considered
```
albertz, line 122: If you think there are possibly cases where `same_order=True` makes sense, maybe this should not just be a bool but sth like `deduplicate_special_lemmata_type` or `deduplicate_special_lemmata_opts`?

lib/lexicon.py
Icemole: I agree, `__eq__` matches my use case but it doesn't cover the most restrictive use case.

What do others think? Is it too overkill to have `same_order` to test lemma equality? Should we just have the case of "two lemmas are equal whenever they contain the same data, regardless of their order"?

At every comment I'm more inclined to remove `same_order` and just check for what I said above, since I think no realistic use case would consider the ordering of data inside a lemma... At least I can't think of any specific use case for this...

lexicon/modification.py
```diff
@@ -178,6 +184,10 @@ def run(self):
                 for lemma in lex.lemmata:
                     # sort by first orth entry
                     orth_key = lemma.orth[0] if lemma.orth else ""
+                    if self.deduplicate_lemmata:
+                        # don't add the lemma when there's already an equal lemma
+                        if len(lemma_dict[orth_key]) > 0 and lemma == lemma_dict[orth_key][0]:
```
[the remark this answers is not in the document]
Icemole: Yes, this is the use case what I was referring to, I might not have made this clear.

Right now I'm even more inclined to remove the `same_order` flag altogether and just keep `same_order=False` as the default. I'm waiting for what the rest of the reviewers might have to say about this.

In general, thinking about use cases as you said, I couldn't think of any actual use case for `same_order=True` that wouldn't be covered by `same_order=False`, only hypothetical situations in which the user might want the same order, but I don't think that would have any sort of impact on the final lexicon.

lib/lexicon.py
albertz: In general, we should not add code/logic for cases which are only hypothetical and not used currently, but only for things we really are using currently. Thus only the logic `same_order=False`. And then `__eq__` is also non-ambiguous.

But let's wait what other think about this.

JackTemaki (changes requested):
I would like to have the option of `same_order` in there, as there are some setups that just pick the first phon sequence, and then the order matters.

Thus `_equals` should actually be `equals`, so that it is not private but is supposed to be used outside.

PR #460, opened 2023-11-20
Add error_threshold param to ffmpeg conversion jobs

curufinwe:
I have a use-case where some files have invalid data, but correct headers. I cannot change that and the only way to detect if they are broken is trying to play them or trying to convert them. Since I can't do anything about them being broken I'm ok with not converting them. To that end I have added an optional error_threshold parameter to the conversion job to allow a certain number of errors before failing. By default that threshold is 0 (current behavior).

### DIFF
--- a/audio/ffmpeg.py
+++ b/audio/ffmpeg.py
@@ -81,6 +81,7 @@ class BlissFfmpegJob(Job):
         ffmpeg_binary: Optional[Union[str, tk.Path]] = None,
         hash_binary: bool = False,
         ffmpeg_input_options: Optional[List[str]] = None,
+        error_threshold: int = 0,
     ):
         """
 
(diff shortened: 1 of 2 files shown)


audio/ffmpeg.py
```diff
@@ -100,9 +102,14 @@ def __init__(
...
         self.ffmpeg_binary = ffmpeg_binary if ffmpeg_binary else "ffmpeg"
         self.hash_binary = hash_binary
+        self.error_threshold = error_threshold
+        self.num_errors = 0
+        self.failed_files = []
 
         self.out_audio_folder = self.output_path("audio/", directory=True)
         self.out_corpus = self.output_path("corpus.xml.gz")
+        if self.error_threshold > 0:
+            self.out_failed_files = self.output_path("failed_files.txt")
```
albertz, line 112:
```suggestion
        self.out_failed_files = None
        if self.error_threshold > 0:
            self.out_failed_files = self.output_path("failed_files.txt")
```

audio/ffmpeg.py
```diff
@@ -132,6 +139,10 @@ def run(self):
         else:
             c.dump(tk.uncached_path(self.out_corpus))
 
+        if hasattr(self, "out_failed_files"):
```
albertz, line 142:
```suggestion
        if self.out_failed_files is not None:
```

Marvin84 (approved):

albertz (approved):

PR #468, opened 2023-12-06
BlissToOggZipJob: fix rsync for many files

vieting:
When merging the different zip files in the `BlissToOggZipJob`'s `merge` task, the job executes `$ rsync -av /var/tmp/sis_x8jg4syk/ogg.*/* /var/tmp/sis_x8jg4syk/out.ogg` which in my case crashed with `/bin/sh: 1: rsync: Argument list too long`. Removing the second `*` solved the issue for me.

### DIFF
--- a/returnn/oggzip.py
+++ b/returnn/oggzip.py
@@ -163,7 +163,7 @@ class BlissToOggZipJob(Job):
                 "-av",
                 os.path.join(
                     tmp_dir,
-                    os.path.basename(self.zip_subarchives.path_template).replace(".$(TASK).zip", ".*/*"),
+                    os.path.basename(self.zip_subarchives.path_template).replace(".$(TASK).zip", ".*/"),
                 ),
                 output_folder,
             ]


albertz:
Ah, I did not really knew about the behavior with a trailing `/` in the source. But I just tested it and it seems to do exactly what is intended here.

albertz (approved):

vieting:
@michelwi Could you please check the AppTek hash test? It also crashes in https://github.com/rwth-i6/i6_core/pull/469 which is totally unrelated.

michelwi:
The failed AppTek test looks like you are coming from an old version of i6_core, could you rebase to the current master head and try again please?

vieting:
> The failed AppTek test looks like you are coming from an old version of i6_core, could you rebase to the current master head and try again please?

I see, my bad. Tests are running again.

vieting:
> > behavior with a trailing / in the source.
> 
> a similar difference exists with trailing / in the target..
> 
> I think one difference is that the order of file in the zip file might change without the last star but is guaranteed to be sorted when using the star. I don't think this is a problem here, just wanted to mention.

I also think the order does not matter here. If you agree with the change, would you mind approving the PR @michelwi? Then we could merge.

vieting:
Thanks @albertz and @michelwi!

PR #469, opened 2023-12-06
ComputeBestMixJob, InterpolateNgramLmJob, fix check_call

Atticus1806:
For some reason we left 2 Jobs as command strings instead of creating an executable.
This does not work (would only work with `shell=True`).
Since we did not want this back when #406 was done, I propose to change the two jobs too. 
Successfully tested this just now.

### DIFF
--- a/lm/srilm.py
+++ b/lm/srilm.py
@@ -441,7 +441,7 @@ class InterpolateNgramLmJob(Job):
             if i == 0:
                 c = ["-lambda", str(lmbd.get())]
             elif i == 1:
-                c = [""]
+                c = []
             else:
                 c = [f"-mix-lambda{i}", str(lmbd.get())]
             cmd += c
--- a/lm/srilm.py
+++ b/lm/srilm.py
@@ -426,24 +426,24 @@ class InterpolateNgramLmJob(Job):
     def _create_cmd(self) -> List[str]:
         """creates bash script that will be executed in the run Task"""
         cmd = [self.ngram_exe.get_path()]
-        cmd += ["-order", f"{self.ngram_order} -unk"]
+        cmd += ["-order", str(self.ngram_order), "-unk"]
 
         for i, lm in enumerate(self.ngram_lms):
             if i == 0:
-                c = ["-lm", f"{lm.get_path()}"]
+                c = ["-lm", lm.get_path()]
             elif i == 1:
-                c = ["-mix-lm", f"{lm.get_path()}"]
+                c = ["-mix-lm", lm.get_path()]
             else:
-                c = [f"-mix-lm{i}", f"{lm.get_path()}"]
+                c = [f"-mix-lm{i}", lm.get_path()]
             cmd += c
 
         for i, lmbd in enumerate(self.weights):
             if i == 0:
-                c = ["-lambda", f"{lmbd.get()}"]
+                c = ["-lambda", str(lmbd.get())]
             elif i == 1:
-                c = [f""]
+                c = [""]
             else:
-                c = [f"-mix-lambda{i}", f"{lmbd.get()}"]
+                c = [f"-mix-lambda{i}", str(lmbd.get())]
             cmd += c
         cmd += ["-write-lm", "interpolated.lm.gz"]
 
--- a/lm/srilm.py
+++ b/lm/srilm.py
@@ -353,7 +353,7 @@ class ComputeBestMixJob(Job):
     def run(self):
         """Call the srilm script and extracts the different weights from the log, then relinks log to output folder"""
         cmd = self._create_cmd()
-        subprocess.check_call(cmd, stdout=open("cbm.log", "wb"), stderr=subprocess.STDOUT)
+        subprocess.check_call(cmd, stdout=open("cbm.log", "wt"), stderr=subprocess.STDOUT)
 
         lines = open("cbm.log", "rt").readlines()
         lbds = lines[-2].split("(")[1].replace(")", "")
--- a/lm/srilm.py
+++ b/lm/srilm.py
@@ -338,26 +338,24 @@ class ComputeBestMixJob(Job):
         self.out_cbm_file = self.output_path("cbm.log")
 
     def tasks(self):
+        yield Task("create_files", mini_task=True)
         yield Task("run", mini_task=True)
 
-    def _get_cmd(self) -> str:
-        """creates command string for the bash call"""
-        cmd = self.compute_best_mix_exe.get_path()
-
-        cmd += " "
+    def create_files(self):
+        """creates bash script that will be executed in the run Task"""
+        cmd = [self.compute_best_mix_exe.get_path()]
 
         ppl_log = [x.get_path() for x in self.ppl_logs]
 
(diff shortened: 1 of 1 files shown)


Atticus1806:
@michelwi @curufinwe is your pipeline using this? If so why does it fail?

vieting (approved):
Did you find out the reason why `subprocess` does not work anymore?

albertz:
I don't understand. So it never worked before, and we had this untested and broken code, and only this fixes it now? Why did we have untested (unused?) code?

michelwi:
The failed AppTek test looks like you are coming from an old version of i6_core, could you rebase to the current master head and try again please?

Atticus1806:
When the PR was created `self.sh` was used. During the review this was changed to `subprocess.check_call`.
Tbh. I cannot recall if this was then tested or not, or how this slipped through. I just know that when executing the old finished job from back then it does not work either.

albertz:
I think it's bad practice that we use this `create_executable`. None of the strings are properly escaped. This can cause lots of strange bugs, which will only happen occasionally and will be hard to debug, so the worst kind of bugs.

Atticus1806:
>Actually using subprocess.check_call directly is much better and safer, but you have to pass the arguments properly as a list, not just as a joined string.

I tried this, this did not work. The problem seems to be the ` &> cbm.log`, which fails for awk with `FileNotFoundError: [Errno 2] No such file or directory:`. Using `shell=True` with subprocess would work too, but I think we did not want this.

albertz:
> I tried this, this did not work. The problem seems to be the ` &> cbm.log`,

This is what I mean what you should *not* do. You should *not* use `shell=True` (which has the same problems as `create_executable`, what I mentioned before), but instead write it directly in a correct way.

See the documentation of [subprocess](https://docs.python.org/3/library/subprocess.html). If you want to forward stdout/stderr to a file, it would be sth like:
```python
run(..., check=True, stdout=open("cbm.log", "wb"), stderr=STDOUT)
```

Atticus1806:
> This is what I mean what you should not do

That was why I was looking for a different way to do it :D 
Changed it back with your suggestion applied to subprocess. I dont really have a way to test the other jobs right now, so I would leave them as is for now and just apply it to the two broken jobs.

lm/srilm.py
```diff
@@ -427,35 +423,35 @@ def __init__(
...
         yield Task("run", rqmt=self.rqmt)
 
-    def _get_cmd(self) -> str:
-        """creates command string for the bash call"""
-        cmd = self.ngram_exe.get_path()
-        cmd += f" -order {self.ngram_order} -unk"
+    def _create_cmd(self) -> List[str]:
+        """creates bash script that will be executed in the run Task"""
+        cmd = [self.ngram_exe.get_path()]
+        cmd += ["-order", f"{self.ngram_order} -unk"]
```
albertz, line 429:
```suggestion
        cmd += ["-order", str(self.ngram_order), "-unk"]
```

lm/srilm.py
```diff
@@ -427,35 +423,35 @@ def __init__(
...
 
         for i, lm in enumerate(self.ngram_lms):
             if i == 0:
-                c = f" -lm {lm.get_path()}"
+                c = ["-lm", f"{lm.get_path()}"]
```
albertz, line 433:
```suggestion
                c = ["-lm", lm.get_path()]
```

lm/srilm.py
```diff
@@ -427,35 +423,35 @@ def __init__(
...
             elif i == 1:
-                c = f" -mix-lm {lm.get_path()}"
+                c = ["-mix-lm", f"{lm.get_path()}"]
             else:
-                c = f" -mix-lm{i} {lm.get_path()}"
+                c = [f"-mix-lm{i}", f"{lm.get_path()}"]
```
albertz, line 437:
```suggestion
                c = [f"-mix-lm{i}", lm.get_path()]
```

lm/srilm.py
```diff
@@ -427,35 +423,35 @@ def __init__(
...
             cmd += c
 
         for i, lmbd in enumerate(self.weights):
             if i == 0:
-                c = f" -lambda {lmbd.get()}"
+                c = ["-lambda", f"{lmbd.get()}"]
             elif i == 1:
-                c = f""
+                c = [f""]
```
albertz, line 444:
```suggestion
                c = [""]
```

lm/srilm.py
```diff
@@ -427,35 +423,35 @@ def __init__(
...
             else:
-                c = f" -mix-lambda{i} {lmbd.get()}"
+                c = [f"-mix-lambda{i}", f"{lmbd.get()}"]
```
albertz, line 446:
```suggestion
                c = [f"-mix-lambda{i}", str(lmbd.get())]
```

lm/srilm.py
```diff
@@ -427,35 +423,35 @@ def __init__(
...
             cmd += c
-        cmd += " -write-lm interpolated.lm.gz"
+        cmd += ["-write-lm", "interpolated.lm.gz"]
 
         return cmd
 
     def run(self):
-        """delete the executable from the hashing"""
-        cmd = self._get_cmd()
+        """run the previously created bash script"""
```
albertz, line 453:
```suggestion
        """run the command"""
```

lm/srilm.py
```diff
@@ -427,35 +423,35 @@ def __init__(
     def tasks(self):
         yield Task("run", rqmt=self.rqmt)
 
-    def _get_cmd(self) -> str:
-        """creates command string for the bash call"""
-        cmd = self.ngram_exe.get_path()
-        cmd += f" -order {self.ngram_order} -unk"
+    def _create_cmd(self) -> List[str]:
+        """creates bash script that will be executed in the run Task"""
```
albertz, line 427:
```suggestion
        """:return: the command"""
```

albertz (approved):
Looks good now. But I did not actually test this. Did you actually test it?

Atticus1806:
So I did not test your suggestions to change the strings, but I tested the variant before (which had the f-strings). I think that should suffice.

vieting (approved):

lm/srilm.py
```diff
@@ -353,7 +353,7 @@ def _create_cmd(self) -> List[str]:
     def run(self):
         """Call the srilm script and extracts the different weights from the log, then relinks log to output folder"""
         cmd = self._create_cmd()
-        subprocess.check_call(cmd, stdout=open("cbm.log", "wb"), stderr=subprocess.STDOUT)
+        subprocess.check_call(cmd, stdout=open("cbm.log", "wt"), stderr=subprocess.STDOUT)
```
albertz, line 356: I don't understand this change. It should be binary, or not?

Atticus1806: Should it? This is just a textfile i think.

albertz: It does not matter what it is. It matters whether you get a binary or text stream from the `subprocess` module, and I'm pretty sure by default you get a binary stream (over `bytes`, not `str`). So you store it in binary.

PR #471, opened 2024-01-02
AverageTorchCheckpointsJob

albertz:


vieting (approved):
The interface is not identical to the `AverageTFCheckpointsJob` but I think that's sensible because it avoids fixed assumptions about the checkpoint names and is cleaner here.

Atticus1806 (approved):
What @vieting says.

PR #473, opened 2024-01-10
replace relink with shutil.copy

christophmluscher:
`relink` function breaks if the given vocab file is not on the same file system, since it tries to create a hard link. replacing with `shutil.copy` should solve this issue.


JackTemaki (approved):

Nintendo-Wei (approved):

michelwi (approved):
We waste a bit of space then, but for vocab file sized files I would not bother to change the code.

PR #474, opened 2024-01-11
add VocabularyFromTextJob

Nintendo-Wei:


lm/vocabulary.py
```diff
@@ -146,3 +146,47 @@ def run(self):
...
+
+class VocabularyFromTextJob(Job):
+    """
+    Extract vocabulary from given text files based on frequency.
+    """
+
+    def __init__(self, file_paths, size=1000000, counts=False):
+        """
+        :param List[Path] file_paths: paths to the text files
+        :param int size: expected size of the vocabulary
```
christophmluscher, line 159: I find `size` as a name not that a clear. is this the size the user expects or is this the max limit? then maybe something like `max_num_words`?

lm/vocabulary.py
```diff
@@ -146,3 +146,47 @@ def run(self):
...
+                        words = line.strip().split()
+                        counter.update(words)
+            except IOError as e:
+                print(f"Error reading file {file_path}: {e}")
+
+        cutoff = min(self.size, len(counter))
+
+        with open(self.out_vocabulary, "w") as vocabulary:
+            if self.counts:
+                for index, (word, count) in enumerate(counter.most_common(cutoff)):
```
christophmluscher, line 188: just wondering why is the enumeration needed here?

christophmluscher (commented):
overall looks good. only a few basic issues from my side

lm/vocabulary.py
```diff
@@ -146,3 +146,47 @@ def run(self):
...
+                    vocabulary.write(f"{word} {count}\n")
+            else:
+                for word in counter.most_common(cutoff):
+                    vocabulary.write(f"{word[0]}\n")
```
christophmluscher, line 192: here it is `word[0]` a few lines above `word` not sure if this is correct?

lm/vocabulary.py
```diff
@@ -146,3 +146,47 @@ def run(self):
...
+        counter = Counter()
+
+        for file_path in self.file_paths:
+            try:
+                with uopen(file_path, "rt") as input_file:
+                    for line in input_file:
+                        words = line.strip().split()
+                        counter.update(words)
+            except IOError as e:
+                print(f"Error reading file {file_path}: {e}")
```
michelwi, line 182: logging.error rather than print? Or logging.warning?

Are you intentionally discarding the exception here? Is this not something that should cause the job to fail and the user to investigate?

lm/vocabulary.py
```diff
@@ -146,3 +146,47 @@ def run(self):
...
+class VocabularyFromTextJob(Job):
+    """
+    Extract vocabulary from given text files based on frequency.
+    """
+
+    def __init__(self, file_paths, size=1000000, counts=False):
+        """
+        :param List[Path] file_paths: paths to the text files
+        :param int size: expected size of the vocabulary
+        :param bool counts: whether write the counts of the words into the vocabulary file
```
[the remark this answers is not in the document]
michelwi: I find `counts` as a flag to activate behaviour not clear. maybe better `add_counts` or `include_counts`.

Or what it actually does is select the output format so one could do an enum and select supported formats based on this (in case we want to have a different output format in the future it is easier to extend).

lm/vocabulary.py
```diff
@@ -146,3 +146,47 @@ def run(self):
...
             for word in sorted(vocabulary):
                 fout.write(f"{word}\n")
+
+
+class VocabularyFromTextJob(Job):
+    """
+    Extract vocabulary from given text files based on frequency.
+    """
+
+    def __init__(self, file_paths, size=1000000, counts=False):
```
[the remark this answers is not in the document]
michelwi: Additionally I suggest to make the default number more readable
```suggestion
    def __init__(self, file_paths, size=1_000_000, counts=False):
```

combined suggestion for your convenience:
```suggestion
    def __init__(self, file_paths: List[tk.Path], size: int = 1_000_000, counts: bool = False):
```

lm/vocabulary.py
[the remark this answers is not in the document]
michelwi: Tying into this discussion: Different classes in this File have different input formats (`FromLexicon`, `FromLm`, and now `FromText`) and I understand we need different parsers in each case. But there are also different output formats in each case:
* Lexicon: docstring says `"{word}: {index}\n"` but implementation seems `"{word} {index}\n"`
* Lm: only `"{word}\n"`
* Text: both `"{word}\n"` and `"{word} {count}\n"`

This is very inconsistent. Should we aim to consistencise it? 
What are the necessary vocab file formats across srilm, kenlm, subwordnmt, returnn, pytorch, ... ?

michelwi (changes requested):
Request changes (or good explanation) w.r.t. the `word` vs `word[0]` issue raised by Chris + the supressed exception.

lm/vocabulary.py
Nintendo-Wei: This is correct as word is a tuple here (unlike the word above), but it is indeed ugly, I will change it.

lm/vocabulary.py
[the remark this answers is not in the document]
Nintendo-Wei: For SRILM and KenLM, we need `"{word}\n"`
For RETURNN, we need `"{word} {index}\n"`
The new job will output `"{word}\n"` by default, and it could also output `"{word} {count}\n"` if the user want, e.g. if the user want to check the counts or merge different vocabularies.

michelwi (approved):
If I understand correctly, the `num_words` is just the maximum that we would take. If the text contains fewer counts then the vocab will be smaller.

Given that the default value for `num_words` is one million, will these "almost always" be exhausted? Would it otherwise make sense to define an additional output that would give the actual size?

Orthogonaliy, would it make sense to have additional exclusion criteria? (e.g. must appear at least x times to be included)

Nintendo-Wei:
> If I understand correctly, the num_words is just the maximum that we would take. If the text contains fewer counts then the vocab will be smaller.
Yes.

> Given that the default value for num_words is one million, will these "almost always" be exhausted?
Yes, 1 million is almost never enough for almost all languages if we consider all available text data for that language.

> Would it otherwise make sense to define an additional output that would give the actual size?
That would be possible, but I have never used such a value.

> Orthogonaliy, would it make sense to have additional exclusion criteria? (e.g. must appear at least x times to be included)
That could actually be more interesting. This is also the reason why I added the option to output counts. And indeed, I often determined the vocabulary based on the counts. Adding such a threshold might be useful. But we will have a wrapper function in apptek_asr for this job to parallelize the job etc, we will think about where we want to add this parameter.

lm/vocabulary.py
```diff
@@ -146,3 +152,47 @@ def run(self):
...
+        """
+        :param file_paths: paths to the text files
+        :param num_words: expected size of the vocabulary
+        :param include_counts: whether write the counts of the words into the vocabulary file
+        """
+        self.file_paths = file_paths
+        self.num_words = num_words
+        self.include_counts = include_counts
+
+        self.out_vocabulary = self.output_path("vocabulary.txt")
```
christophmluscher, line 172:
```suggestion
        self.out_vocabulary = self.output_path("vocabulary.txt")
        self.out_counter = self.output_var("counter")
```

lm/vocabulary.py
```diff
@@ -146,3 +152,47 @@ def run(self):
...
+
+        with open(self.out_vocabulary, "w") as vocabulary:
+            if self.include_counts:
+                for (word, count) in counter.most_common(cutoff):
+                    vocabulary.write(f"{word} {count}\n")
+            else:
+                for (word, _) in counter.most_common(cutoff):
+                    vocabulary.write(f"{word}\n")
+
+        return counter
```
christophmluscher, line 198:
```suggestion
        self.out_counter.set(counter)
```

christophmluscher (commented):
Sry, I think was unclear when I said it would be good to return the counter!

This is what I meant, assign to a value so that it could be used in a later step, for what ever reason :)

christophmluscher (approved):

lm/vocabulary.py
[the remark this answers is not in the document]
Icemole: For RETURNN: what's `index`? 0, 1, ...? If so, then we might be able to use another job to obtain such an index (or even a loop enclosed in `enumerate`).

I find `f"{word} {count}\n"` to be the most consistent output, but I'm afraid that changing current functionality breaks some developed jobs.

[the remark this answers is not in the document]
michelwi: > For RETURNN: what's index? 0, 1, ...? 

yes, but we usually map sentence begin and sentence end to the same token. So it will be 0, 0, 1, 2, ... (depending on position of `<s>` and `<\s>`)

Atticus1806 (approved):
In general looks good to me. The only thing I am wondering about is the `include_counts` parameter. 
From my point of view the Job does the same thing with or without that flag, only the output (format) changes. Would it make sense to just output both always? E.g. `self.out_vocab` and `self.out_vocab_w_counts` or so? I guess it depends on how often you need both variants in the same pipeline, maybe almost never? Then rerunning the job when the other output is required is fine I guess. 
So I will approve here, but just want to bring the point into the discussion.

PR #475, opened 2024-01-18
Change BlissFfmpegJob threading

JackTemaki:
While working in a different cluster environment I found out that the job was spawing way more threads than it should, leading to conflicts with the cluster settings and also having an impact on efficiency.

These changes force ffmpeg to use only one thread for the input processing and one thread for the output processing. The pool size is adjusted accordingly. 

I also increased the number of default cores, but this is debatable. The extra thread for the python process might not be strictly necessary for maximum performance (I did not test for that), but the cluster environment kills jobs that uses more threads than requested, and with this settings it worked. 

From commit message
 - force ffmpeg to use 1 thread for input and output processing each
 - set higher number of default cpu cores
 - remote some tk.uncached_path calls

### DIFF
--- a/audio/ffmpeg.py
+++ b/audio/ffmpeg.py
@@ -82,6 +82,7 @@ class BlissFfmpegJob(Job):
         hash_binary: bool = False,
         ffmpeg_input_options: Optional[List[str]] = None,
         error_threshold: int = 0,
+        cpu_rqmt: int = 9,
     ):
         """
 
@@ -94,6 +95,7 @@ class BlissFfmpegJob(Job):
                             in which case the binary needs to be hashed
         :param ffmpeg_input_options: list of ffmpeg parameters thare are applied for reading the input files
         :param error_threshold: Allow upto this many files to fail conversion before failing this job
+        :param cpu_rqmt: number of cpu cores to use
         """
         self.corpus_file = corpus_file
         self.ffmpeg_input_options = ffmpeg_input_options
@@ -113,7 +115,7 @@ class BlissFfmpegJob(Job):
             self.out_failed_files = self.output_path("failed_files.txt")
 
         # e.g. 1 core for python and 4x2 cores for ffmpeg, one for input processing and one for output processing
-        self.rqmt = {"time": 4, "cpu": 9, "mem": 8}
+        self.rqmt = {"time": 4, "cpu": cpu_rqmt, "mem": 8}
 
     def tasks(self):
         yield Task("run", rqmt=self.rqmt)
@@ -221,4 +223,5 @@ class BlissFfmpegJob(Job):
             d.pop("ffmpeg_binary")
         if kwargs["error_threshold"] == 0:
             d.pop("error_threshold")
+        d.pop("cpu_rqmt")
         return super().hash(d)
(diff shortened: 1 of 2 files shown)


audio/ffmpeg.py
```diff
@@ -112,7 +112,8 @@ def __init__(
         if self.error_threshold > 0:
             self.out_failed_files = self.output_path("failed_files.txt")
 
-        self.rqmt = {"time": 4, "cpu": 4, "mem": 8}
+        # e.g. 1 core for python and 4x2 cores for ffmpeg, one for input processing and one for output processing
```
michelwi, line 115: ~Where does the 4x2 come from? Can we configure ffmpeg to use less threads? Will this number now be lower as you set `"-threads", "1"` for ffmpeg?~

Oh I see, we use `multiprocessing.pool` that is configured via the requested number of cores. But still where does the factor of 4 come from? Should we make this a parameter to the job to (more) expose it to the user?

audio/ffmpeg.py
```diff
@@ -124,11 +125,11 @@ def tasks(self):
 
     def run(self):
         c = corpus.Corpus()
-        c.load(tk.uncached_path(self.corpus_file))
+        c.load(self.corpus_file.get_path())
 
         from multiprocessing import pool
 
-        p = pool.Pool(self.rqmt["cpu"])
+        p = pool.Pool(self.rqmt["cpu"] // 2)
```
michelwi, line 132: should we remove 3 here for python, input and output?

audio/ffmpeg.py
```diff
@@ -195,13 +196,9 @@ def _perform_ffmpeg(self, recording: corpus.Recording):
...
             logging.info(f"try converting {target}")
-            command_head = [
-                self.ffmpeg_binary,
-                "-hide_banner",
-                "-y",
-            ]
+            command_head = [self.ffmpeg_binary, "-hide_banner", "-y", "-threads", "1"]
             command_in = ["-i", recording.audio]
-            command_out = [target]
+            command_out = ["-threads", "1", target]
```
michelwi, line 201: is this duplicate `"-threads", "1"` intentional (e.g. ffmpeg accepts this option once for input and once for output)?

audio/ffmpeg.py
JackTemaki: Yes we could expose that parameter, this is just arbitrary.

audio/ffmpeg.py
JackTemaki: Why remove 3? You can remove 1 for Python, but doing the //2 for odds already does that.

audio/ffmpeg.py
JackTemaki: Yes this is intentional, from my observation and some internet research this seems to be the case.

audio/ffmpeg.py
michelwi: from your comment:

> e.g. 1 core for python and 4x2 cores for ffmpeg, one for input processing and one for output processing

oh.. sorry, I am bad at math this morning.. I thought you meant `1 + 4x2 + 1 + 1` here but now I realize that the result is 9 which is `1 + 4x(1+1)` (And I was searching for the extra two threads for "input and output processing").

Ok, then disregard this comment.

audio/ffmpeg.py
michelwi: Ok, I weakly suggest adding this as a parameter (as the new default of 9 cores might be a lot for some cluster environments.. or if you want to more effectively use the `2^x` processors of your node)

Otherwise I'm fine with the changes.

michelwi (approved):

Atticus1806 (approved):

christophmluscher (approved):

PR #477, opened 2024-01-23
CloneGitRepositoryJob: allow applying patches

vieting:
This PR extends the `CloneGitRepositoryJob` to allow applying patches.

### DIFF
--- a/tools/git.py
+++ b/tools/git.py
@@ -15,7 +15,7 @@ class CloneGitRepositoryJob(Job):
     Clone a git repository given optional branch name and commit hash
     """
 
-    __sis_hash_exclude__ = {"clone_submodules": False, "files_to_checkout": None}
+    __sis_hash_exclude__ = {"clone_submodules": False, "files_to_checkout": None, "patches": None}
 
     def __init__(
         self,
@@ -25,6 +25,7 @@ class CloneGitRepositoryJob(Job):
         checkout_folder_name: str = "repository",
         clone_submodules: bool = False,
         files_to_checkout: Optional[List[str]] = None,
+        patches: Optional[List[str]] = None,
     ):
         """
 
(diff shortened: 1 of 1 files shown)


tools/git.py
```diff
@@ -74,6 +78,11 @@ def run(self):
             args = ["git", "checkout", self.commit]
             logging.info("running command: %s" % " ".join(args))
             sp.run(args, cwd=repository_dir, check=True)
+        if self.patches is not None:
```
Marvin84, line 81: It is not clear to me how this would work via git apply, is this tested with an example?

JackTemaki (approved):
Seems good to me. I first thought "why not use files instead of strings", but then you would probably want to store them inside the recipes, and then Sisyphus would complain when loading them as `tk.Path`. So maybe the way here is to load the files in the manager, and then pass the string content to be hashed. At least that sounds to me like a valid way.

tools/git.py
vieting: You mean that the patch is applied properly, right? Yes, I did test this.

vieting:
> Seems good to me. I first thought "why not use files instead of strings", but then you would probably want to store them inside the recipes, and then Sisyphus would complain when loading them as `tk.Path`. So maybe the way here is to load the files in the manager, and then pass the string content to be hashed. At least that sounds to me like a valid way.

Yes, I had a similar train of thought. I would load (open and read content as string) them inside the recipes using relative paths. It also seems a little more flexible and the content is directly hashed which is not the case when using files (at least not by default). If anyone has a better suggestion using files instead of strings I would be open to it though.

tools/git.py
vieting: See documentation of `git apply` [here](https://git-scm.com/docs/git-apply):
>  \- can be used to read from the standard input.

Therefore I pass the patch as a string as input.

michelwi (approved):
> Yes, I had a similar train of thought.

My train of thought for quite a while in the past was going the route of
* `ApplyPatchJob` - but do we really copy the whole repo another time just to apply one patch?
* `tk.register_callback` and the callback would apply the patch. But this is not only fragile w.r.t. manager restarts, it also violates the principle that "one Job and this Job writes to its job folder and only its job folder". 
* We should anyway not do this, if a change is necessary, then it should be pushed to the repo.

But if we have any implementation, I like this one the most.

vieting:
> We should anyway not do this, if a change is necessary, then it should be pushed to the repo.

Yes, I agree that the regular work practice should be to push necessary changes to the repo!

PR #479, opened 2024-01-24
Fix HeadJob Hash

### LINKED ISSUE #478: Hash Breaks in HeadJob
In #467, there are modifications to the `HeadJob`, in particular:
```
-        self.out = self.output_path("out.gz")
+        self.out = self.output_path("out.gz") if self.zip_output else self.output_path("out")
```
In general, this makes sense of course. However, it breaks setups that use this job with `zip_output=False` because (a) if the job did already run, subsequent jobs will now look for the output file `"out"` which does not exist (only `"out.gz"` exists) and (b) subsequent jobs will have different hashes because of the changed input.

In particular, I have a setup which has a `TailJob` (inherits this issue from `HeadJob`) as input to returnn trainings (segment list). Now, all trainings would re-run. Since the change in general makes sense, I wouldn't necessarily revert it though. A fix for my case would be to reassign the out file.
```
tail_job = text.TailJob(train_segments, num_lines=300, zip_output=False)
tail_job.out = tail_job.output_path("out.gz")
```
This is a bit hacky though and sisyphus outputs warnings:
```
WARNING: Added output out.gz more than once to Job<work/i6_core/text/processing/TailJob.RiSM6fe2XipO>
```

What do you guys think?

vieting:
Fix #478 using @michelwi's suggested solution.

### DIFF
--- a/text/processing.py
+++ b/text/processing.py
@@ -186,7 +186,8 @@ class HeadJob(Job):
         self.zip_output = zip_output
 
         self.out = self.output_path("out.gz") if self.zip_output else self.output_path("out")
-        self.out.hash_overwrite = (self, "out.gz")
+        if not self.zip_output:
+            self.out.hash_overwrite = (self, "out.gz")  # keep old hashing behavior
         self.length = self.output_var("length")
 
     def tasks(self):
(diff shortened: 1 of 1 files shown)


michelwi:
we could break this into explicit if cases and only apply the hash overwrite in case of unzipped output.

Otherwise as I am the de facto author of this code I verbally approve.

JackTemaki (approved):
I guess it is okay that way...

Atticus1806 (approved):

PR #481, opened 2024-01-25
Caching behavior in ConcatenateJob

vieting:
The `ConcatenateJob` has currently hard-coded caching behavior.
```
gs.file_caching(str(i)) for i in self.text_files
```
In a case where I wanted to concatenate feature cache bundle files, this is suboptimal because caching the bundle (I think it's determined via the ending `".bundle"`) means caching all files inside that. That (a) takes unnecessarily long because we don't need to cache the content files and (b) - more importantly - is also wrong because the concatenated output file will contain the cached paths starting with `/var/tmp/`.

I used `Path` objects and set `cached=False`. The proposed change results in the intended behavior. Should we keep the compatibility for string inputs?

Also, I refactored the init function a bit.

### DIFF
--- a/text/processing.py
+++ b/text/processing.py
@@ -9,6 +9,7 @@ __all__ = [
 
 import os
 from collections.abc import Iterable
+from typing import List
 
 from sisyphus import Job, Task, Path, global_settings as gs
 from sisyphus.delayed_ops import DelayedBase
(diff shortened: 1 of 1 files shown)


michelwi (approved):

Atticus1806:
In general I would consider dropping string inputs generally. But since this would break potentially a lot, I would prefer to keep it in. But I really like the approach of not having it as a type.

text/processing.py
```diff
@@ -121,11 +122,11 @@ class ConcatenateJob(Job):
 
     __sis_hash_exclude = {"zip_out": True, "out_name": "out"}
 
-    def __init__(self, text_files, zip_out=True, out_name="out"):
+    def __init__(self, text_files: List[Path], zip_out: bool = True, out_name: str = "out"):
```
Atticus1806, line 125: Is List required here? Or would Iterable suffice?

vieting: Not sure if all iterables support `len()`, otherwise everything could be written such that it works for any iterable I guess. I just used list because it was annotated like this before.

Atticus1806 (approved):

vieting:
> In general I would consider dropping string inputs generally. But since this would break potentially a lot, I would prefer to keep it in. But I really like the approach of not having it as a type.

I also thought about this but was afraid that it breaks things, too. So I decided not to tackle that issue here.

PR #482, opened 2024-01-31
add RasrFeatureDumpHDFJob

vieting:
I find the `RasrAlignmentDumpHDFJob` useful for some of my pipelines and added a corresponding `RasrFeatureDumpHDFJob`. It is analog to the existing job for alignment caches. It shares quite some code so it would be conceivable to generalize things here, but there are a number of differences in the processing and the inputs also differ. Therefore I figured it would be easier and more readable to have two separate jobs. We can discuss this though if you disagree.

### DIFF
--- a/returnn/hdf.py
+++ b/returnn/hdf.py
@@ -325,7 +325,7 @@ class BlissToPcmHDFJob(Job):
 
 class RasrFeatureDumpHDFJob(Job):
     """
-    This Job reads Rasr feature caches and dump them in hdf files.
+    This Job reads Rasr feature caches and dumps them in hdf files.
     """
 
     def __init__(
@@ -337,7 +337,7 @@ class RasrFeatureDumpHDFJob(Job):
         filter_list_keep: Optional[tk.Path] = None,
     ):
         """
-        :param feature_caches: e.g. output of an FeatureExtractionJob
+        :param feature_caches: e.g. output of a FeatureExtractionJob
         :param data_type: type that is used to store the data
         :param returnn_root: file path to the RETURNN repository root folder
         :param encoding: encoding of the segment names in the cache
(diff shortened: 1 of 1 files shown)


Marvin84 (approved):

michelwi (approved):
I tend to agree.

Atticus1806:
I am wondering, does this job achieve something that the default `ReturnnDumpHDFJob` cannot? Or am I missing something?

I am usually doing something like (also for the alignments): 
```
    feat_dataset = {
        "class": "SprintCacheDataset",
        "data": {
            "data": {
                "filename": features,
                "data_type": "feat",
                "allophone_labeling": {
                    "silence_phone": allophone_labeling.silence_phone,
                    "allophone_file": allophone_labeling.allophone_file,
                    "state_tying_file": allophone_labeling.state_tying_file,
                },
            }
        },
        "seq_list_filter_file": segment_list,
    }
```

vieting:
You can certainly achieve the same in different ways. I never used `SprintCacheDataset`. Does it work without setting up RASR? I guess for features, you wouldn't even need the allophone settings.

Atticus1806:
I pass this as is into the `ReturnnDumpHDFJob`, which from what I know does not "see" RASR. 

The job itself seems fine to me, I was just wondering if this does anything new. Even if not, if this is more easy to handle we can certainly add this imo.

michelwi:
> Does [SprintCacheDataset] work without setting up RASR?

It uses the same code ("gleichen" nicht "selben") as the `FileArchive`. -> yes.

vieting:
So what do you think? Is the `RasrFeatureDumpHDFJob` still helpful in addition?

vieting:
Following the principle of not adding jobs for something that is already doable by another more generic job, I decided not to merge this PR. In my setup, I verified that using the `ReturnnDumpHDFJob` is also simple and has the same result. We could nevertheless add a job here which inherits from `ReturnnDumpHDFJob` and just prepares the `data` dict based on a bundle path, we could add a comment on how to do it with the existing job or we could just drop this PR altogether. Any opinions?

christophmluscher (approved):

christophmluscher:
Code looks good. But concerns about necessity. I ll close for now.

PR #484, opened 2024-02-27
Corpus: add get_segment_by_name

vieting:
I found it handy to have a method that returns a segment by its name from the corpus.

### DIFF
--- a/lib/corpus.py
+++ b/lib/corpus.py
@@ -166,6 +166,15 @@ class Corpus(NamedEntity, CorpusSection):
         for sc in self.subcorpora:
             yield from sc.segments()
 
+    def get_segment_by_name(self, name: str) -> Segment:
+        """
+        :return: the segment specified by its name
+        """
+        for seg in self.segments():
+            if seg.fullname() == name:
+                return seg
+        assert False, f"Segment '{name}' was not found in corpus"
+
     def all_recordings(self) -> Iterable[Recording]:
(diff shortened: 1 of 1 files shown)


Atticus1806:
Do we maybe also want to extend this check to possibly cover `segment.name` so `name in [segment.fullname(), segment.name]`? I am wondering if one would always call this with knowing the fullname.

albertz (approved):
If this function is used often, it would make sense to a dict full name -> seg. Actually, that's my only criticism here, that from such a function `get_segment_by_name`, I would expect that it is efficient, while this here is not efficient. But maybe this does not matter for your current use case, and then we should not overcomplicate it for now.

albertz:
> Do we maybe also want to extend this check to possibly cover `segment.name` so sth. like `name in [segment.fullname(), segment.name]`?

It should be a flag which switches the behavior to one or the other, or maybe both if you really want that, but not by default. Also, we should only add this once this is really needed by anyone.

vieting:
> Do we maybe also want to extend this check to possibly cover segment.name so sth. like name in [segment.fullname(), segment.name]?

It would be a poor choice to do this by default, because in many of our corpora the segments within one recording are just named `"0001"`, `"0002"`, etc. So this would then match too many segments. I agree to Albert, we can add this if needed.

Atticus1806 (approved):

PR #487, opened 2024-03-22
CreateBPELexiconJob

JackTemaki:
This job is used for the BPE-CTC setup with Flashlight I want to push to common, but also @AtanasGruev has a very similar job for the RASR-BPE setups, so this job should be re-usable. 

Includes a simple test case that can be extended if needed.


lexicon/bpe.py
```diff
@@ -0,0 +1,142 @@
...
+
+from sisyphus import Job, Task, tk, setup_path
+
+from i6_core.lib.lexicon import Lexicon, Lemma
+import i6_core.util as util
+
+
+class CreateBPELexiconJob(Job):
+    """
+    Create a Bliss lexicon from bpe transcriptions that can be used e.g, for lexicon constrained BPE search.
```
michelwi, line 17: > from bpe transcriptions

is that the correct term here? 

Isn't what the Job does rather
```suggestion
    In a Bliss lexicon replace the phonetic representation with a BPE decomposition of the words that can be used e.g, for lexicon constrained BPE search.
```

michelwi (changes requested):

lexicon/bpe.py
```diff
@@ -0,0 +1,142 @@
...
+    """
+
+    def __init__(
+        self,
+        base_lexicon_path: tk.Path,
+        bpe_codes: tk.Path,
+        bpe_vocab: tk.Path,
+        subword_nmt_repo: tk.Path,
+        unk_label: str = "UNK",
+        vocab_blacklist: Optional[Union[List[str], Set[str]]] = None,
```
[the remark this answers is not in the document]
JackTemaki: I thought about this, but then a lot of things are iterable in python that I do not want here, e.g. dicts

lexicon/bpe.py
```diff
@@ -0,0 +1,142 @@
...
+            "bpes",
+        ]
+        sp.run(args, check=True)
+
+        with util.uopen("bpes", "rt") as f:
+            bpe_tokens = [l.strip() for l in f]
+
+        w2b = {w: b for w, b in zip(lm_tokens, bpe_tokens)}
+
+        for w, b in w2b.items():
```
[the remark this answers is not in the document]
michelwi: should I propose some code that I think is good, or would you rather keep it as is until somebody needs it?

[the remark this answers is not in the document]
JackTemaki: Feel free to propose code :)

michelwi (approved):
Now I approve, but I touched half of the code.. 🤷

JackTemaki:
@michelwi thanks a lot for your contribution!

@SimBe195 @Atticus1806 please check if possible!

lexicon/bpe.py
```diff
@@ -0,0 +1,141 @@
...
+        self.base_lexicon_path = base_lexicon_path
+        self.bpe_codes = bpe_codes
+        self.bpe_vocab = bpe_vocab
+        self.subword_nmt_repo = subword_nmt_repo
+        self.unk_label = unk_label
+        if vocab_blacklist is None:
+            self.vocab_blacklist = set()
+        else:
+            # convert list to set for faster "in" check
+            self.vocab_blacklist = set(vocab_blacklist)
```
SimBe195, line 54: Not very important but I don't think this is helpful. For short lists the "in" check is just as fast or even faster than for sets.

lexicon/bpe.py
```diff
@@ -0,0 +1,141 @@
...
+
+    def _fill_vocab_and_lexicon(self):
+        lexicon = Lexicon()
+        vocab = set()
+        with util.uopen(self.bpe_vocab.get_path(), "rt") as f, util.uopen("fake_count_vocab.txt", "wt") as vocab_file:
+            for line in f:
+                line = line.strip()
+                if line == "{" or line == "}":
+                    continue
+                symbol = line.split(":")[0][1:-1]
```
SimBe195, line 82: Maybe add a comment that indicates how `line` looks like to show what this indexing does.

lexicon/bpe.py
```diff
@@ -0,0 +1,141 @@
...
+        lm_tokens, special_lemmas = self._fill_lm_tokens(base_lexicon)
+
+        with util.uopen("words", "wt") as f:
+            for t in lm_tokens:
+                f.write(f"{t}\n")
+
+        vocab, lexicon = self._fill_vocab_and_lexicon()
+
+        # add special lemmas back to lexicon
+        if self.keep_special_lemmas is True:
```
SimBe195, line 105:
```suggestion
        if self.keep_special_lemmas:
```

lexicon/bpe.py
```diff
@@ -0,0 +1,141 @@
...
+    def tasks(self):
+        yield Task("run", resume="run", mini_task=True)
+
+    def _fill_lm_tokens(self, base_lexicon: Lexicon):
+        lm_tokens = set()
+        special_lemmas = []
+        for lemma in base_lexicon.lemmata:
+            if lemma.special is None:
+                for orth in lemma.orth:
+                    lm_tokens.add(orth)
```
SimBe195, line 68:
```suggestion
                lm_tokens.update(lemma.orth)
```

SimBe195 (approved):
Approved with some minor nitpicks.

lexicon/bpe.py
Atticus1806: I would keep the `is True` check, otherwise other objects might evaluate this to true and I don't think we should allow this.

lexicon/bpe.py
JackTemaki: But for short lists the speed is then not relevant, right? I am not sure if someone would ever put very large blacklists here, but I thought set is more consistent. I can change the comment though.

Atticus1806 (approved):

PR #488, opened 2024-03-25
Implements SRILM Discount Job

christophmluscher:


lm/srilm.py
```diff
@@ -94,6 +95,74 @@ def hash(cls, kwargs):
...
+        counts: tk.Path,
+        vocab: tk.Path,
+        count_exe: tk.Path,
+        *,
+        extra_discount_args: Optional[List[str]] = None,
+        cpu_rqmt: int = 1,
+        mem_rqmt: int = 48,
+        time_rqmt: float = 24,
+        fs_rqmt: str = "100G",
+    ):
```
michelwi, line 115: docstring pls if possible

lm/srilm.py
```diff
@@ -94,6 +95,74 @@ def hash(cls, kwargs):
...
+
+    def create_files(self):
+        """creates bash script to compute discounts"""
+        cmd = [
+            f"{self.count_exe.get_path()} \\\n",
+            f"  -order {self.ngram_order} \\\n",
+            f"  -vocab {self.vocab.get_cached_path()} \\\n",
+        ]
+
+        cmd += [
```
michelwi, line 143: why split this?
```suggestion
```

lm/srilm.py
```diff
@@ -94,6 +95,74 @@ def hash(cls, kwargs):
...
+        cpu_rqmt: int = 1,
+        mem_rqmt: int = 48,
+        time_rqmt: float = 24,
+        fs_rqmt: str = "100G",
+    ):
+        self.ngram_order = ngram_order
+        self.counts = counts
+        self.vocab = vocab
+        self.count_exe = count_exe
+        self.discount_args = extra_discount_args
```
michelwi, line 120: cf line 146
```
>>> f"  {' '.join(None)} -memuse\n"
  File "<stdin>", line 1, in <module>
TypeError: can only join an iterable
```
therefore 
```suggestion
        self.discount_args = extra_discount_args or []
```

This might in line 146 add an extra space if there are no extra args, but that would only be cosmetical in a file that nobody reads anyway..

michelwi (changes requested):

lm/srilm.py
```diff
@@ -94,6 +95,82 @@ def hash(cls, kwargs):
...
+        ngram_order: int,
+        counts: tk.Path,
+        vocab: tk.Path,
+        count_exe: tk.Path,
+        *,
+        extra_discount_args: Optional[List[str]] = None,
+        cpu_rqmt: int = 1,
+        mem_rqmt: int = 48,
+        time_rqmt: float = 24,
+        fs_rqmt: str = "100G",
```
Atticus1806, line 114: Why is this different from the other requirements (in not being an int)? Is this some practical benefit because of how it is passed within the job?

lm/srilm.py
```diff
@@ -94,6 +95,82 @@ def hash(cls, kwargs):
...
+    ):
+        """
+        :param ngram_order: order of the ngram counts, typically 3 or 4.
+        :param counts: file with the ngram counts, see  :class:`CountNgramsJob.out_counts`.
```
Atticus1806, line 118:
```suggestion
        :param counts: file with the ngram counts, see :class:`CountNgramsJob.out_counts`.
```

Atticus1806 (approved):
Nothing that prevents merging, rather beauty suggestions.

lm/srilm.py
michelwi: this requirement is not implemented per se in sisyphus, rather we use the fallback option "add any arbitrary string to the SGE call".. (cf. line 139)

Speaking of which.. @christophmluscher should `qsub_args` not rather be `sbatch_args` now for slurm?

JackTemaki: My suggestion would be to add both qsub_args and sbatch_args, the engine anyway only picks the needed entry.

lm/srilm.py
```diff
@@ -111,6 +188,7 @@ def __init__(
         count_exe: tk.Path,
         *,
         vocab: Optional[tk.Path] = None,
+        discounts: Optional[tk.Path] = None,
```
JackTemaki, line 191: This is a hash breaking change, you would need a `__sis_hash_exclude__` for that.

JackTemaki (changes requested):

lm/srilm.py
christophmluscher: >Why is this different from the other requirements (in not being an int)? Is this some practical benefit because of how it is passed within the job?

I am staying consistent with the other jobs in this regard. I dont remember the design decision for `str`

JackTemaki (changes requested):

lm/srilm.py
```diff
@@ -124,6 +217,7 @@ def __init__(
         :param data_mode: Defines whether input format is text based or count based
         :param vocab: Vocabulary file, one word per line
         :param extra_ngram_args: Extra arguments for the execution call e.g. ['-kndiscount']
+        :param use_modified_srilm: use the modified SRILM version.
```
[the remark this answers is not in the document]
christophmluscher: I have no idea. This was modified at some point in the past.
But we used it to create basically all our LMs.

lm/srilm.py
```diff
@@ -94,11 +95,103 @@ def hash(cls, kwargs):
...
+
+        self.count_exe = count_exe
+
+        self.out_discounts = self.output_path("discounts", cached=True)
+
+        self.rqmt_run = {
+            "cpu": cpu_rqmt,
+            "mem": mem_rqmt,
+            "time": time_rqmt,
+            "qsub_args": f"-l h_fsize={fs_rqmt}",
```
curufinwe, line 145: Should this really have SGE specific things hardcoded in it? I think it would be better to add them outside the job if they're required.

lm/srilm.py
```diff
@@ -94,11 +95,103 @@ def hash(cls, kwargs):
...
+            f"  -read {self.counts.get_cached_path()} \\\n",
+            f"  {' '.join(self.discount_args)} -memuse\n",
+        ]
+
+        create_executable("run.sh", cmd)
+
+    def run(self):
+        """executes the previously created bash script and relinks outputs from work folder to output folder"""
+        subprocess.check_call("./run.sh")
+        relink("discounts", self.out_discounts.get_path())
```
curufinwe, line 173: Unrelated to this PR, but could maybe be fixed here aswell: relink checks existance of the destination by checking for os.path.exists(dst) which will return false if the file the link is pointing to does not exist. I think it's probably not intended behavior. One solution would be to check also for `os.path.islink(dst)`. This is just FYI, fixing that is not required for approval.

curufinwe (changes requested):

lm/srilm.py
christophmluscher: I removed the SGE specific rqmts from the job I added. I am not sure about the other jobs since that might affect other setups...

lm/srilm.py
christophmluscher: I added a check in the relink func

curufinwe (approved):

JackTemaki (approved):

PR #489, opened 2024-03-26
Add alias prefix to filterbank features

Marvin84:
Other feature extraction functions of the `system` have explicit `prefix` parameter, which this PR adds also to the filterbanks

### DIFF
--- a/meta/system.py
+++ b/meta/system.py
@@ -323,14 +323,15 @@ class System:
         self.add_derivatives(corpus, "mfcc", num_deriv, num_features)
         self.add_derivatives(corpus, "uncached_mfcc", num_deriv, num_features)
 
-    def fb_features(self, corpus: str, **kwargs):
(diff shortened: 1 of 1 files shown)


vieting (approved):
The PR's title could be more descriptive of what is done here, otherwise ok.

JackTemaki (approved):

Marvin84:
> The PR's title could be more descriptive of what is done here, otherwise ok.

updated

JackTemaki:
@Marvin84 please do not do pull requests from personal fork branches, but always from branches within i6_core

michelwi:
@JackTemaki please do not merge a PR from a fork without confirming with us/me. I will manually trigger a pipeline now to check if this breaks any important recipes.

JackTemaki:
> @JackTemaki please do not merge a PR from a fork without confirming with us/me. I will manually trigger a pipeline now to check if this breaks any important recipes.

I was confident enough about this one... but you are right this is against protocol

PR #490, opened 2024-03-27
Tedlium not distinct corpora names for the 3 partitions

Marvin84:
The corpus name for all train, dev, and test partition is the same. This PR suggests a second version of the job that modifies simply the corpus name.

### DIFF
--- a/datasets/tedlium2.py
+++ b/datasets/tedlium2.py
@@ -219,6 +219,7 @@ class CreateTEDLIUM2BlissCorpusJob(Job):
 class CreateTEDLIUM2BlissCorpusJobV2(CreateTEDLIUM2BlissCorpusJob):
     """
     It has the same logic as CreateTEDLIUM2BlissCorpusJob but the corpus name is distinct for each partition
+    and the common text normalization for eliminating space before apostroph, similarly to Kaldi and ESPNet is applied
     """
 
     def make_corpus(self):
--- a/datasets/tedlium2.py
+++ b/datasets/tedlium2.py
@@ -261,12 +261,15 @@ class CreateTEDLIUM2BlissCorpusJobV2(CreateTEDLIUM2BlissCorpusJob):
                     last_rec_name = rec_name
                     seg_id = 1
 
+                #This solved the TedLium known problem of additional space before apostrof
+                normalized_text = text.replace(" '", "'")
+
                 segment = corpus.Segment()
                 segment.name = str(seg_id)
                 segment.start = float(start)
                 segment.end = float(end)
                 segment.speaker_name = spk_name
-                segment.orth = text
+                segment.orth = normalized_text
 
                 recording.add_segment(segment)
                 seg_id += 1
--- a/datasets/tedlium2.py
+++ b/datasets/tedlium2.py
@@ -318,7 +318,7 @@ class CreateTEDLIUM2BlissCorpusJobV2(CreateTEDLIUM2BlissCorpusJob):
                 segment.start = float(start)
                 segment.end = float(end)
                 segment.speaker_name = spk_name
-                segment.orth = normalized_text
+                segment.orth = text
 
                 recording.add_segment(segment)
                 seg_id += 1
--- a/datasets/tedlium2.py
+++ b/datasets/tedlium2.py
@@ -221,9 +221,6 @@ class CreateTEDLIUM2BlissCorpusJobV2(CreateTEDLIUM2BlissCorpusJob):
     It has the same logic as CreateTEDLIUM2BlissCorpusJob but the corpus name is distinct for each partition
     """
 
-    def __init__(self, corpus_folders):
-        super().__init__(corpus_folders)
-
     def make_corpus(self):
         """
         create bliss corpus from stm file (always include speakers)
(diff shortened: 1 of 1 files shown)


christophmluscher:
I solve it very similar to this PR: `c.name = f"TED-LIUM-release3-{corpus_key}"`

I added a extend segment flag. I removed the updates to the transcriptions completely. Since the segments chagen from rel2 train to rel3 train change and do not occur anymore.

datasets/tedlium2.py
```diff
@@ -214,3 +214,63 @@ def load_stm_data(self, stm_file):
...
         return data
+
+
+class CreateTEDLIUM2BlissCorpusJobV2(CreateTEDLIUM2BlissCorpusJob):
+    """
+    It has the same logic as CreateTEDLIUM2BlissCorpusJob but the corpus name is distinct for each partition
+    """
+
+    def __init__(self, corpus_folders):
+        super().__init__(corpus_folders)
```
michelwi, line 225: Not needed if nothing is changed. By default the parent `__init__` will be used anyway.

```suggestion
```

michelwi (commented):
One could also move much of the duplicated code into a separate function (that only exists in the base class and then the `make_corpus` function would call this one and set the names accordingly.

We should probably add a deprecation notice to the old class.

Should we also "move" the old class to deprecated?

datasets/tedlium2.py
```diff
@@ -214,3 +214,66 @@ def load_stm_data(self, stm_file):
...
                 data.append([rec_name, channel, spk_name, start, end, gender, text])
         return data
+
+
+class CreateTEDLIUM2BlissCorpusJobV2(CreateTEDLIUM2BlissCorpusJob):
+    """
+    It has the same logic as CreateTEDLIUM2BlissCorpusJob but the corpus name is distinct for each partition
+    """
+
+    def __init__(self, corpus_folders):
```
Atticus1806, line 224:
```suggestion
    def __init__(self, corpus_folders: Dict[str, tk.Path]):
```
Also please add the import at the top please, can't properly in review. Also documentation (even though its not in the other job) would be nice.

datasets/tedlium2.py
```diff
@@ -214,3 +214,66 @@ def load_stm_data(self, stm_file):
...
+                if rec_name != last_rec_name:
+                    if recording:
+                        c.add_recording(recording)
+                    recording = corpus.Recording()
+                    recording.name = rec_name
+                    recording.audio = os.path.join(audio_dir, "%s.sph" % rec_name)
+                    last_rec_name = rec_name
+                    seg_id = 1
+
+                #This solved the TedLium known problem of additional space before apostrof
```
Atticus1806, line 264: Isn't it "Apostrophe"?

datasets/tedlium2.py
Marvin84: I will delete the `init` definition as suggested by @michelwi

Atticus1806:
@michelwi what happens with the apptek hashes here?

datasets/tedlium2.py
```diff
@@ -214,3 +214,64 @@ def load_stm_data(self, stm_file):
...
+                    if recording:
+                        c.add_recording(recording)
+                    recording = corpus.Recording()
+                    recording.name = rec_name
+                    recording.audio = os.path.join(audio_dir, "%s.sph" % rec_name)
+                    last_rec_name = rec_name
+                    seg_id = 1
+
+                # This solved the TedLium known problem of additional space before apostrophe
+                normalized_text = text.replace(" '", "'")
```
JackTemaki, line 263: @Marvin84 the text processing is altered in the `make_stm`, this is where you would have to change it

JackTemaki (changes requested):

michelwi:
> @michelwi what happens with the apptek hashes here?

our recipes depend on https://github.com/rwth-i6/i6_core/pull/505. If you rebase/merge the current main head the problem should go away.

datasets/tedlium2.py
```diff
@@ -214,3 +214,115 @@ def load_stm_data(self, stm_file):
...
+                        next_seg = None if idx == len(data) - 1 else data[idx + 1]
+                        data[idx][3], data[idx][4] = extend_segment_time(data[idx], pre_seg, next_seg, 0.15, 0.1)
+
+                for seg in data:
+                    f.write(" ".join(seg) + "\n")
+
+            f.close()
+            shutil.move("%s.stm" % corpus_key, self.out_stm_files[corpus_key].get_path())
+
+    def make_corpus(self):
```
JackTemaki, line 279: Do you make any changes in this function? If not it can be removed.

JackTemaki (changes requested):

datasets/tedlium2.py
Marvin84: Yes, this function now does the name distinction for each partition, the original reason why this PR was opened :-)

Marvin84:
@JackTemaki @michelwi @Atticus1806 @christophmluscher I would like to work on the `i6_experiments` changes for using the new version, could I get this merged please?

JackTemaki (changes requested):
As discussed offline, the apostrophe merging is now missing

PR #491, opened 2024-03-28
Add PickOptimalParametersJob

JackTemaki:
Job that takes pickleable parameters and a `tk.Variable` value, and just returns the paramaters corresponding to the highest or lowest value.

### DIFF
--- a/tools/parameter_tuning.py
+++ b/tools/parameter_tuning.py
@@ -1,0 +1,47 @@
+from sisyphus import Job, Task, tk
+from typing import Any, List, Tuple
+
+import numpy as np
+
+from i6_core.util import instanciate_delayed
+
+
+class PickOptimalParametersJob(Job):
+    """
+    Pick a set of optimal pickleable parameters based on their assigned (dynamic) score value.
+    """
+
+    def __init__(self, parameters: List[Tuple[Any]], values: List[tk.Variable], mode="minimize"):
+        """
+        :param parameters: list of tuples of parameters, must be pickleable
+        :param values: list of tk.Variables containing int or float, used to determine the best
+            set of parameters. Some calculations might be done using DelayedOps math.
+        :param mode: "minimize" or "maximize"
+        """
+        assert len(parameters) == len(values)
+        for param in parameters[1:]:
+            assert len(param) == len(parameters[0]), "all entries should have the same number of parameters"
+        assert mode in ["minimize", "maximize"]
+        self.parameters = parameters
+        self.values = values
+        self.mode = mode
+        self.num_values = len(values)
+        self.num_parameters = len(parameters[0])
+
+        self.out_optimal_parameters = [self.output_var("param_%i" % i, pickle=True) for i in range(self.num_parameters)]
+
+    def tasks(self):
+        yield Task("run", mini_task=True)
+
+    def run(self):
+        values = instanciate_delayed(self.values)
+
+        if self.mode == "minimize":
+            index = np.argmin(values)
+        else:
+            index = np.argmax(values)
+
+        best_parameters = self.parameters[index]
+
+        for i, param in enumerate(best_parameters):
+            self.out_optimal_parameters[i].set(param)
--- a/tools/parameter_tuning.py
+++ b/tools/parameter_tuning.py
@@ -1,5 +1,5 @@
 from sisyphus import Job, Task, tk
-from typing import Any, Sequence
+from typing import Any, Literal, Sequence, Union
 
 import numpy as np
 
@@ -14,11 +14,17 @@ class PickOptimalParametersJob(Job):
     Can be used to e.g. pick best lm-scale and prior scale to a corresponding ScliteJob.out_wer.
     """
 
-    def __init__(self, parameters: Sequence[Sequence[Any]], values: Sequence[tk.Variable], mode):
+    def __init__(
+        self,
+        parameters: Sequence[Sequence[Any]],
+        values: Sequence[tk.Variable],
+        mode: Union[Literal["maximize"], Literal["minimize"]],
+    ):
         """
         :param parameters: a row / column matrix of parameters, where the best row should be selected
-        :param values: list of tk.Variables containing int or float, used to determine the best
-            row of parameters. Some calculations might be done using DelayedOps math beforehand.
+        :param values: sequence of tk.Variables with the same length as row, containing int or float.
+            Used to determine the best row of parameters.
+            Some calculations might be done using DelayedOps math beforehand.
         :param mode: "minimize" or "maximize"
         """
         assert len(parameters) == len(values)
--- a/tools/parameter_tuning.py
+++ b/tools/parameter_tuning.py
@@ -28,7 +28,6 @@ class PickOptimalParametersJob(Job):
         self.parameters = parameters
         self.values = values
         self.mode = mode
-        self.num_values = len(values)
         self.num_parameters = len(parameters[0])
 
         self.out_optimal_parameters = [self.output_var("param_%i" % i, pickle=True) for i in range(self.num_parameters)]
--- a/tools/parameter_tuning.py
+++ b/tools/parameter_tuning.py
@@ -16,6 +16,7 @@ class GetOptimalParametersAsVariableJob(Job):
 
     def __init__(
         self,
+        *,
         parameters: Sequence[Sequence[Any]],
         values: Sequence[tk.Variable],
         mode: Union[Literal["maximize"], Literal["minimize"]],
--- a/tools/parameter_tuning.py
+++ b/tools/parameter_tuning.py
@@ -22,10 +22,14 @@ class GetOptimalParametersAsVariableJob(Job):
         mode: Union[Literal["maximize"], Literal["minimize"]],
     ):
         """
-        :param parameters: a row / column matrix of parameters, where the best row should be selected
-        :param values: sequence of tk.Variables with the same length as row, containing int or float.
-            Used to determine the best row of parameters.
-            Some calculations might be done using DelayedOps math beforehand.
+        :param parameters: parameters[best_idx] will be written to self.out_optimal_parameters
+             as Sisyphus output variables.
+             parameters[best_idx] (and thus self.out_optimal_parameters) is a sequence of fixed length,
+             to allow to index into it.
+             Thus, len(parameters[i]) must be the same for all i.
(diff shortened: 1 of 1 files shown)


Atticus1806 (approved):

tools/parameter_tuning.py
```diff
@@ -0,0 +1,47 @@
...
+class PickOptimalParametersJob(Job):
+    """
+    Pick a set of optimal pickleable parameters based on their assigned (dynamic) score value.
+    """
+
+    def __init__(self, parameters: List[Tuple[Any]], values: List[tk.Variable], mode="minimize"):
+        """
+        :param parameters: list of tuples of parameters, must be pickleable
+        :param values: list of tk.Variables containing int or float, used to determine the best
+            set of parameters. Some calculations might be done using DelayedOps math.
```
albertz, line 18: I don't exactly understand the meaning of `parameters` and `values` from this.

From the code, I see that `parameters` and `values` must have the same length. This should also be in the documentation.
Then I see that all tuples in `parameters` must have the same length. This should also be in the documentation.

From the code, I understand that you `values` is used to determine the the best index, and then `parameters[best_index]` is written to the output. From that logic, it seems a bit arbitrary that `parameters[i]` must be always a list and always of the same length, but this is a restriction due to how Sisyphus works. This should also be explained in the documentation.

tools/parameter_tuning.py
```diff
@@ -0,0 +1,47 @@
...
+
+
+class PickOptimalParametersJob(Job):
+    """
+    Pick a set of optimal pickleable parameters based on their assigned (dynamic) score value.
+    """
+
+    def __init__(self, parameters: List[Tuple[Any]], values: List[tk.Variable], mode="minimize"):
+        """
+        :param parameters: list of tuples of parameters, must be pickleable
```
SimBe195, line 16: I'm not really getting the reasoning for the `tuple` structure of the parameters. Why shouldn't it be possible to pass a list of dicts, lists or whatever else the user prefers as long as it's pickleable? I would make this more generic. In this case you could get rid of the `self.num_parameters`, the length check and only define one output var instead of a list.

tools/parameter_tuning.py
```diff
@@ -0,0 +1,47 @@
...
+        :param mode: "minimize" or "maximize"
+        """
+        assert len(parameters) == len(values)
+        for param in parameters[1:]:
+            assert len(param) == len(parameters[0]), "all entries should have the same number of parameters"
+        assert mode in ["minimize", "maximize"]
+        self.parameters = parameters
+        self.values = values
+        self.mode = mode
+        self.num_values = len(values)
```
SimBe195, line 28:
```suggestion
```
Unused?

SimBe195 (changes requested):

albertz (changes requested):
Ok to merge after the doc/types fixes, and maybe the name (see below).

This jobs seems very specific to me. Why do you need the output parameters as variables? More generic would be like a `ArgmaxJob` or so which given a sequence of values, it writes the argmax to an output variable. And then a `GatherJob`, which gathers from a sequence of arbitrary structure, given the index. But here, multiple variants would make sense to me: Hardlink/copy the input path if the inputs to `GatherJob` are `Path` objects, or writing output variables as you do. The `Path` variant would seem more useful to me. And I would make it support any arbitrary structure (via `tree` or so), not just supporting tuple.

But anyway, if you need this specific variant here, I don't have objections in merging it. However, I wonder a bit about changing the name of the job, considering that it is so specific. Specifically in that it writes output variables, which was not what I expected initially.

tools/parameter_tuning.py
```diff
@@ -0,0 +1,47 @@
...
+
+from i6_core.util import instanciate_delayed
+
+
+class PickOptimalParametersJob(Job):
+    """
+    Pick a set of optimal pickleable parameters based on their assigned (dynamic) score value.
+    """
+
+    def __init__(self, parameters: List[Tuple[Any]], values: List[tk.Variable], mode="minimize"):
```
albertz, line 14: I would also remove the default for `mode`. It's very arbitrary to have "minimum" being the default here.

JackTemaki:
> Ok to merge after the doc/types fixes, and maybe the name (see below).
> 
> This jobs seems very specific to me. Why do you need the output parameters as variables? More generic would be like a `ArgmaxJob` or so which given a sequence of values, it writes the argmax to an output variable. And then a `GatherJob`, which gathers from a sequence of arbitrary structure, given the index. But here, multiple variants would make sense to me: Hardlink/copy the input path if the inputs to `GatherJob` are `Path` objects, or writing output variables as you do. The `Path` variant would seem more useful to me. And I would make it support any arbitrary structure (via `tree` or so), not just supporting tuple.
> 
> But anyway, if you need this specific variant here, I don't have objections in merging it. However, I wonder a bit about changing the name of the job, considering that it is so specific. Specifically in that it writes output variables, which was not what I expected initially.

It did not test it explicitly, but you already can put any data structure you want, as the tk.Variable gets created in pickle mode. You should also be able to add paths if you want. I had the sequence of sequence already before enabling that, so maybe I should just remove the nested sequence so that it is clear you can put anything. I think that might make it more obvious.

Elementary things as ArgMax and Gather should then better be implemented on the level of the delayed_ops...

JackTemaki:
> > Ok to merge after the doc/types fixes, and maybe the name (see below).
> > This jobs seems very specific to me. Why do you need the output parameters as variables? More generic would be like a `ArgmaxJob` or so which given a sequence of values, it writes the argmax to an output variable. And then a `GatherJob`, which gathers from a sequence of arbitrary structure, given the index. But here, multiple variants would make sense to me: Hardlink/copy the input path if the inputs to `GatherJob` are `Path` objects, or writing output variables as you do. The `Path` variant would seem more useful to me. And I would make it support any arbitrary structure (via `tree` or so), not just supporting tuple.
> > But anyway, if you need this specific variant here, I don't have objections in merging it. However, I wonder a bit about changing the name of the job, considering that it is so specific. Specifically in that it writes output variables, which was not what I expected initially.
> 
> It did not test it explicitly, but you already can put any data structure you want, as the tk.Variable gets created in pickle mode. You should also be able to add paths if you want. I had the sequence of sequence already before enabling that, so maybe I should just remove the nested sequence so that it is clear you can put anything. I think that might make it more obvious.
> 
> Elementary things as ArgMax and Gather should then better be implemented on the level of the delayed_ops...

Ah, I see now while this works (not have sequence of sequence) this is not good for the use case I want, because I want to have the individual values directly. I will keep it like it is to not cause delays.

tools/parameter_tuning.py
JackTemaki: > but this is a restriction due to how Sisyphus works

The use case we have also does not makes sense otherwise, so that is okay.

albertz:
I think you require the same structure (or tuple of same length) because you want to access the individual outputs on the setup level from the Sisyphus job. That's what I mean when I write "this is a restriction due to how Sisyphus works".

That's ok. I just was confused about this and only understood it after checking the actual code. My comment on this is just that it should be explained in the documentation, why you have this constraint.

tools/parameter_tuning.py
JackTemaki: If everything is pickled you can only unpickle it "within" the next job, but we want to handle each individual parameter in the manager itself.

JackTemaki: if you want that behavior, you can still add a dict as single entry in the (formerly tuple) sequence.

JackTemaki:
@albertz suggestions for the name? I can not come up with one I like more in this usecase scenario.

tools/parameter_tuning.py
```diff
@@ -0,0 +1,55 @@
+from sisyphus import Job, Task, tk
+from typing import Any, Literal, Sequence, Union
+
+import numpy as np
+
+from i6_core.util import instanciate_delayed
+
+
+class PickOptimalParametersJob(Job):
```
albertz, line 9:
```suggestion
class WriteBestParameterVariablesJob(Job):
```
?

My main wish would be to have "Variable" somewhere inside the name (or sth similar to reflect that this writes Sisyphus variables as output). Because this was the main thing I got confused about.

tools/parameter_tuning.py
```diff
@@ -0,0 +1,55 @@
...
+    Each optimal parameter is outputted individually to be accessible in the Sisyphus manager.
+
+    Can be used to e.g. pick best lm-scale and prior scale to a corresponding ScliteJob.out_wer.
+    """
+
+    def __init__(
+        self,
+        parameters: Sequence[Sequence[Any]],
+        values: Sequence[tk.Variable],
+        mode: Union[Literal["maximize"], Literal["minimize"]],
```
albertz, line 21: I like explicit keyword args.
```suggestion
        *,
        mode: Union[Literal["maximize"], Literal["minimize"]],
```
Or maybe put the `*,` even earlier.

Marvin84 (approved):
I took a quick look and it seems fine to me

JackTemaki:
@albertz @SimBe195  I think all issues are addressed now...

SimBe195 (approved):

tools/parameter_tuning.py
```diff
@@ -0,0 +1,56 @@
...
+        *,
+        parameters: Sequence[Sequence[Any]],
+        values: Sequence[tk.Variable],
+        mode: Union[Literal["maximize"], Literal["minimize"]],
+    ):
+        """
+        :param parameters: a row / column matrix of parameters, where the best row should be selected
+        :param values: sequence of tk.Variables with the same length as row, containing int or float.
+            Used to determine the best row of parameters.
+            Some calculations might be done using DelayedOps math beforehand.
```
albertz, line 28: I think calling this a matrix is confusing. Also, it does not explain my aspect why it must be a sequence.
```suggestion
        :param parameters: parameters[best_idx] will be written to self.out_optimal_parameters
           as Sisyphus output variables.
           parameters[best_idx] (and thus self.out_optimal_parameters) is a sequence of fixed length,
           to allow to index into it.
           Thus, len(parameters[i]) must be the same for all i.
        :param values: best_idx = argmax(values) or argmin(values).
            Must have len(values) == len(parameters).
            Some calculations might be done using DelayedOps math beforehand.
```

JackTemaki: For me this docstring is not really helpful, but I also do not want to delay so I will just commit

albertz (approved):

PR #492, opened 2024-04-02
Add BlissLexiconToLineBasedLexiconJob

JackTemaki:
converts bliss lexicon to a line based lexicon used for Flashlight/ Torchaudio CTC decoder.


michelwi:
Is this a duplicate of [BlissLexiconToG2PLexiconJob](https://github.com/rwth-i6/i6_core/blob/main/g2p/convert.py#L15)?

JackTemaki:
I will wait until I can confirm my WERs are the same, and then we can close this and use the existing job.

PR #496, opened 2024-04-12
Add functionality to merge subcorpora and recordings for MergeCorporaJob

Icemole:
This new merge strategy flattens at the top level subcorpora, while merging recordings and segments with the same name.

Besides, I took the liberty to briefly comment what the rest of the merge strategies did.

If you want, before dumping I could sort the segments by start/end so that the final corpus is better structured.

Edit: my use case is splitting a big corpus into smaller subcorpora through `SegmentCorpusJob`/`FilterCorpusBySegmentsJob`, doing some work with the smaller subcorpora (that I haven't been able to parallelize, thus I split in smaller subcorpora), and then merging again through `MergeCorporaJob`. Then, I'd like the updated corpus to have the same shape as the original corpus, but none of the merging strategies for `MergeCorporaJob` does what I want. Regardless of whether my specific use case is valid or not, I think the functionality is useful.

Edit: `CONCATENATE` doesn't work because it duplicates the speakers inside the subcorpora, and doesn't merge the recordings.

### DIFF
--- a/corpus/transform.py
+++ b/corpus/transform.py
@@ -360,7 +360,7 @@ class MergeCorporaJob(Job):
         to_add_spk = {spk.name: spk for spk in to_add.top_level_speakers()}
         for name, spk in to_add_spk.items():
             if name in base_spk:
-                assert spk == base_spk[name], (
+                assert spk.__dict__ == base_spk[name].__dict__, (
                     f"Found same speaker {name} with different attributes in the base/to be added corpus.\n"
                     f"Speaker attributes (base corpus): {base_spk[name]}.\n"
                     f"Speaker attributes (corpus to be added): {spk}."
@@ -403,7 +403,7 @@ class MergeCorporaJob(Job):
         to_add_spk = {spk.name: spk for spk in to_add.speakers}
         for name, spk in to_add_spk.items():
             if name in base_spk:
-                assert spk == base_spk[name], (
+                assert spk.__dict__ == base_spk[name].__dict__, (
                     f"Found same speaker {name} with different attributes in the base/to be added corpus.\n"
                     f"Speaker attributes (base corpus): {base_spk[name]}.\n"
                     f"Speaker attributes (corpus to be added): {spk}."
--- a/corpus/transform.py
+++ b/corpus/transform.py
@@ -430,8 +430,6 @@ class MergeCorporaJob(Job):
                     f"Affected segment name: {name}.\n"
                     f"Affected recording (base): {base_seg[name].recording.fullname()}.\n"
                     f"Affected recording (to be added): {seg.recording.fullname()}.\n"
-                    f"Affected corpus (base): {base_seg[name].recording.corpus.fullname()}.\n"
-                    f"Affected corpus (to be added): {seg.recording.corpus.fullname()}.\n"
                 )
             base.add_segment(seg)
 
--- a/corpus/transform.py
+++ b/corpus/transform.py
@@ -297,7 +297,7 @@ class MergeStrategy(enum.Enum):
     SUBCORPORA = 0  # Merge as subcorpora of a common corpus.
     FLAT = 1  # Flatten corpus structure by ignoring subcorpora.
     CONCATENATE = 2  # Concatenate all subcorpora, recordings and speakers.
-    # Like SUBCORPORA, but flatten each top level subcorpora, and merge subcorpora and recordings with the same name.
+    # Flatten each top level subcorpora, and merge subcorpora and recordings with the same name.
     MERGE_SUBCORPORA_AND_RECORDINGS = 3
 
 
--- a/corpus/transform.py
+++ b/corpus/transform.py
@@ -366,9 +366,7 @@ class MergeCorporaJob(Job):
                                 rec_name_to_rec[rec.name].add_segment(seg)
             else:
                 assert False, "invalid merge strategy"
-        # import pdb
 
-        # pdb.set_trace()
         merged_corpus.dump(self.out_merged_corpus.get_path())
 
 
--- a/corpus/transform.py
+++ b/corpus/transform.py
@@ -415,16 +415,24 @@ class MergeCorporaJob(Job):
         base_seg = {seg.name: seg for seg in base.segments}
         add_seg = {seg.name: seg for seg in to_add.segments}
         for name, seg in add_seg.items():
-            assert name not in base_seg, (
-                "The same segment exists in two corpora. "
-                "The MergeCorporaJob can't handle this situation, even if both segments are equal.\n"
-                "Please filter either of the two segments in the corpora to be merged. "
-                f"Affected segment name: {name}.\n"
-                f"Affected recording (base): {base_seg[name].recording.fullname()}.\n"
-                f"Affected recording (to be added): {seg.recording.fullname()}.\n"
-                f"Affected corpus (base): {base_seg[name].recording.corpus.fullname()}.\n"
-                f"Affected corpus (to be added): {seg.recording.corpus.fullname()}.\n"
-            )
+            if name in base_seg:
+                # Only compare segments with respect to their actual segment data.
+                # Don't include the recording to which they belong, since it's bound to change.
+                # Not only because it's a dynamic object, but also because it comes from a completely different corpus.
+                # In practice, the equality check would yield False when comparing segments with recording objects.
+                seg_without_rec = {key: value for key, value in seg.__dict__.items() if key != "recording"}
+                base_seg_without_rec = {
+                    key: value for key, value in base_seg[name].__dict__.items() if key != "recording"
+                }
+                assert seg_without_rec == base_seg_without_rec, (
+                    "The same segment exists in two corpora with different parameters. "
+                    "Please filter either of the two segments in the corpora to be merged. "
+                    f"Affected segment name: {name}.\n"
+                    f"Affected recording (base): {base_seg[name].recording.fullname()}.\n"
+                    f"Affected recording (to be added): {seg.recording.fullname()}.\n"
+                    f"Affected corpus (base): {base_seg[name].recording.corpus.fullname()}.\n"
+                    f"Affected corpus (to be added): {seg.recording.corpus.fullname()}.\n"
+                )
             base.add_segment(seg)
 
 
--- a/corpus/transform.py
+++ b/corpus/transform.py
@@ -326,7 +326,7 @@ class MergeCorporaJob(Job):
         merged_corpus.name = self.name
         if self.merge_strategy == MergeStrategy.MERGE_SUBCORPORA_AND_RECORDINGS:
             sc_name_to_sc: Dict[str, corpus.Corpus] = {}
-            rec_name_to_rec: Dict[str, corpus.Recording] = {}
+            rec_name_to_sc_rec: Dict[str, Dict[str, corpus.Recording]] = collections.defaultdict(dict)
         for corpus_path in self.bliss_corpora:
             c = corpus.Corpus()
             c.load(tk.uncached_path(corpus_path))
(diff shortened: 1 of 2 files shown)


corpus/transform.py
```diff
@@ -337,9 +343,32 @@ def run(self):
...
+                        else:
+                            # The recording had already been added to some subcorpus.
+                            # Gracefully merge the segments from both recordings.
+                            for seg in rec.segments:
+                                rec_name_to_rec[rec.name].add_segment(seg)
             else:
                 assert False, "invalid merge strategy"
+        # import pdb
 
+        # pdb.set_trace()
```
michelwi, line 371:
```suggestion

```
remove debugging code

corpus/transform.py
```diff
@@ -293,9 +294,11 @@ def info(self):
...
 
 class MergeStrategy(enum.Enum):
-    SUBCORPORA = 0
-    FLAT = 1
-    CONCATENATE = 2
+    SUBCORPORA = 0  # Merge as subcorpora of a common corpus.
+    FLAT = 1  # Flatten corpus structure by ignoring subcorpora.
+    CONCATENATE = 2  # Concatenate all subcorpora, recordings and speakers.
+    # Like SUBCORPORA, but flatten each top level subcorpora, and merge subcorpora and recordings with the same name.
+    MERGE_SUBCORPORA_AND_RECORDINGS = 3
```
michelwi, line 301: I don't think it is "like SUBCORPORA" as in SUBCORPORA we add one layer of subcorpus-ness and here we do merging
```suggestion
    MERGE_SUBCORPORA_AND_RECORDINGS = 3  # Merge subcorpora and recordings with the same name.
```

michelwi (changes requested):
The implementation is incorrect in the following cases:
* nested subcorpora are not handled
```
# file1
corpus/subcorpusA/subcorpusB1/recording1/segmenta
corpus/subcorpusA/subcorpusB1/recording1/segmentb
corpus/subcorpusA/subcorpusB2/recording1/segmenta
corpus/subcorpusA/subcorpusB2/recording1/segmentb

# file2
corpus/subcorpusA/subcorpusB1/recording1/segmentc
corpus/subcorpusA/subcorpusB1/recording1/segmentd
corpus/subcorpusA/subcorpusB2/recording1/segmentc
corpus/subcorpusA/subcorpusB2/recording1/segmentd

# merged
corpus/subcorpusA/recording1/segmenta
corpus/subcorpusA/recording1/segmentb
corpus/subcorpusA/recording1/segmenta
corpus/subcorpusA/recording1/segmentb
corpus/subcorpusA/recording1/segmentc
corpus/subcorpusA/recording1/segmentd
corpus/subcorpusA/recording1/segmentc
corpus/subcorpusA/recording1/segmentd
```
* recordings in different subcorpora but with the same name will be merged
```
# file1
corpus/subcorpusA1/recording1/segmenta
corpus/subcorpusA2/recording1/segmenta

# file2
corpus/subcorpusA1/recording1/segmentb
corpus/subcorpusA2/recording1/segmentb

# merged
corpus/subcorpusA1/recording1/segmenta
corpus/subcorpusA1/recording1/segmenta
corpus/subcorpusA1/recording1/segmentb
corpus/subcorpusA1/recording1/segmentb
corpus/subcorpusA2/  # empty subcorpus
```

Maybe we should add some test cases for this Job to verify what happens is as we expect

Icemole:
> The implementation is incorrect in the following cases:
> 
>     * nested subcorpora are not handled
> 
> 
> ```
> # file1
> corpus/subcorpusA/subcorpusB1/recording1/segmenta
> corpus/subcorpusA/subcorpusB1/recording1/segmentb
> corpus/subcorpusA/subcorpusB2/recording1/segmenta
> corpus/subcorpusA/subcorpusB2/recording1/segmentb
> 
> # file2
> corpus/subcorpusA/subcorpusB1/recording1/segmentc
> corpus/subcorpusA/subcorpusB1/recording1/segmentd
> corpus/subcorpusA/subcorpusB2/recording1/segmentc
> corpus/subcorpusA/subcorpusB2/recording1/segmentd
> 
> # merged
> corpus/subcorpusA/recording1/segmenta
> corpus/subcorpusA/recording1/segmentb
> corpus/subcorpusA/recording1/segmenta
> corpus/subcorpusA/recording1/segmentb
> corpus/subcorpusA/recording1/segmentc
> corpus/subcorpusA/recording1/segmentd
> corpus/subcorpusA/recording1/segmentc
> corpus/subcorpusA/recording1/segmentd
> ```
> 
>     * recordings in different subcorpora but with the same name will be merged
> 
> 
> ```
> # file1
> corpus/subcorpusA1/recording1/segmenta
> corpus/subcorpusA2/recording1/segmenta
> 
> # file2
> corpus/subcorpusA1/recording1/segmentb
> corpus/subcorpusA2/recording1/segmentb
> 
> # merged
> corpus/subcorpusA1/recording1/segmenta
> corpus/subcorpusA1/recording1/segmenta
> corpus/subcorpusA1/recording1/segmentb
> corpus/subcorpusA1/recording1/segmentb
> corpus/subcorpusA2/  # empty subcorpus
> ```
> 
> Maybe we should add some test cases for this Job to verify what happens is as we expect

I see. You're right, the comparison shouldn't be so shallow when comparing recordings, and instead differentiate among subcorpora names.

I'll try to come up with some tests for the new functionality.

Icemole:
You were correct, there are some errors that my test (still WIP) caught. I will try to push a definitive version by tomorrow.

michelwi:
here is my shower thoughts implementation of a `MERGE_RECURSIVE` strategy

```
    @staticmethod
    def merge_corpora(base: corpus.Corpus, toadd: corpus.Corpus):
        # Speaker
        base_spk = {spk.name: spk for spk in base.top_level_speakers()}
        add_spk = {spk.name: spk for spk in toadd.top_level_speakers()}
        for name, spk in add_spk.items():
            if name in base_spk:
                continue  # TODO
            else:
                base.add_speaker(spk)

        # Subcorpora
        base_sc = {sc.name: sc for sc in base.top_level_subcorpora()}
        add_sc = {sc.name: sc for sc in toadd.top_level_subcorpora()}
        for name, sc in add_sc.items():
            if name in base_sc:
                MergeCorporaJob.merge_corpora(base_sc[name], sc)
            else:
                base.add_subcorpus(sc)

        # Recordings
        base_rec = {rec.name: rec for rec in base.top_level_recordings()}
        add_rec = {rec.name: rec for rec in toadd.top_level_recordings()}
        for name, rec in add_rec.items():
            if name in base_rec:
                MergeCorporaJob.merge_recordings(base_rec[name], rec)
            else:
                base.add_recording(rec)

    @staticmethod
    def merge_recordings(base: corpus.Recording, toadd: corpus.Recording):
        assert base.audio == toadd.audio, "Same recording name points do different audio files"
        # Speaker
        base_spk = {spk.name: spk for spk in base.speakers}
        add_spk = {spk.name: spk for spk in toadd.speakers}
        for name, spk in add_spk.items():
            if name in base_spk:
                continue  # TODO
            else:
                base.speakers[name] = spk

        # Segments
        base_seg = {seg.name: seg for seg in base.segments}
        add_seg = {seg.name: seg for seg in toadd.segments}
        for name, seg in add_seg.items():
            assert name not in base_seg, "same segment in two corpora"
            base.add_segment(seg)
```

Icemole:
@michelwi thank you for your suggestion. My implementation here only tries to merge the top level subcorpora and flatten the corpus structure to have only top level subcorpora (by taking all recordings from the innermost subcorpora and bringing them to the first corpus). However, its behavior might be too confusing, and we could think about your `MERGE_RECURSIVE` strategy which makes sense as well.

michelwi:
I thought your intention was to exactly revert the `SegmentCorpusJob` / `FilterCorpusBySegmentsJob` combo. Then you would have to take care of nested subcorpora etc as well.

Icemole:
I see. I think your implementation addresses this pretty well. I'll try it out today.

Icemole:
I followed the `pytest` wiki on how to skip certain functions from being run, but I'm struggling to make it work in practice. Does anyone know what's happening? Locally my single test passes:

```
platform linux -- Python 3.10.12, pytest-7.4.2, pluggy-1.3.0
...
collected 1 item                                                                                                                                                                               

tests/job_tests/corpus/test_transform.py .                                                                                                                                               [100%]

====================================================================================== 1 passed in 1.20s =======================================================================================
```

michelwi:
> Does anyone know what's happening? Locally my single test passes:

Pytest: `Dict[str,Any]` vs `dict[str,Any]`
the latter is only allowed in python>=3.9 and here the pipeline runs in 3.8 

AppTek hashes: merge the current main head, as the apptek recipes depend on https://github.com/rwth-i6/i6_core/pull/505

Icemole:
Thanks, I think the changes are now ok.

I basically implemented what you said, which was fine. I added more details on the checks to add duplicated segments and speakers (which you might find strange because I use `__dict__` to compare both speakers, although this seems to work in my implementation), but I practically didn't change anything else.

corpus/transform.py
```diff
@@ -337,11 +339,94 @@ def run(self):
...
+            else:
+                base.speakers[name] = spk
+
+        # Merge the segments.
+        base_seg = {seg.name: seg for seg in base.segments}
+        add_seg = {seg.name: seg for seg in to_add.segments}
+        for name, seg in add_seg.items():
+            assert name not in base_seg, (
+                "The same segment exists in two corpora. "
+                "The MergeCorporaJob can't handle this situation, even if both segments are equal.\n"
```
michelwi, line 420: > even if both segments are equal.

mh.. should we implement this?

tests/job_tests/corpus/test_transform.py
```diff
@@ -0,0 +1,257 @@
...
+            bliss_corpora=[Path(c1_path), Path(c2_path)],
+            name="custom_name_for_merged_corpus",
+            merge_strategy=MergeStrategy.MERGE_RECURSIVE,
+        )
+        merge_corpora_job.out_merged_corpus = Path(merged_corpus_path)
+        merge_corpora_job.run()
+
+        merged_corpus = libcorpus.Corpus()
+        merged_corpus.load(merged_corpus_path)
+        expected_merged_corpus_structure = {
```
michelwi, line 222: since you already have everything written up (Good job writing a test case 👍 ) would it be easy to just add the expected structures for the other merging methods as well and have full coverage of the job?

michelwi (approved):

corpus/transform.py
Icemole: I can also assert that `seg.__dict__` == `base_seg[name].__dict__`, just as I did for the speakers.

Icemole: I've implemented the dict check. I've made sure to remove the recording, since two segments would always be different if we compared them along with their associated recordings (mutable object).

tests/job_tests/corpus/test_transform.py
Icemole: Done :)

Icemole:
@michelwi I've requested your review for the tests and the newly added segment equality check.

corpus/transform.py
```diff
@@ -337,11 +339,102 @@ def run(self):
...
+                    key: value for key, value in base_seg[name].__dict__.items() if key != "recording"
+                }
+                assert seg_without_rec == base_seg_without_rec, (
+                    "The same segment exists in two corpora with different parameters. "
+                    "Please filter either of the two segments in the corpora to be merged. "
+                    f"Affected segment name: {name}.\n"
+                    f"Affected recording (base): {base_seg[name].recording.fullname()}.\n"
+                    f"Affected recording (to be added): {seg.recording.fullname()}.\n"
+                    f"Affected corpus (base): {base_seg[name].recording.corpus.fullname()}.\n"
+                    f"Affected corpus (to be added): {seg.recording.corpus.fullname()}.\n"
```
michelwi, line 434: I thought I had commented on that before, but I can't find it any more, so sorry if duplicate:

The lines with affected corpus are redundant as `recording.fullname()` will just return `recording.corpus.fullname() + '/' + recording.name` so all information is already contained in the recording print.
And one could even just print `seg.fullname()` and `base_seg[name].fullname()`

Besides, by construction of the merge algorithm, both fullnames should coincide.

michelwi (approved):

hannah220 (approved):

Icemole:
If there is anything else I should do for this PR before merging, please let me know.

PR #497, opened 2024-04-12
Remove empty recordings in filtering jobs

Icemole:
This PR adds behavior to remove recordings that are left empty due to the filtering performed. It affects the `FilterCorpusBySegmentDurationJob` and the `FilterCorpusRemoveUnknownWordSegmentsJob`.

### DIFF
--- a/corpus/filter.py
+++ b/corpus/filter.py
@@ -363,7 +363,7 @@ class FilterCorpusBySegmentDurationJob(Job):
         bliss_corpus: Path,
         min_duration: float = 0.1,
         max_duration: float = 120.0,
-        delete_empty_recordings: bool = True,
+        delete_empty_recordings: bool = False,
     ):
         """
         :param bliss_corpus: path of the corpus file
--- a/corpus/filter.py
+++ b/corpus/filter.py
@@ -23,6 +23,24 @@ from sisyphus import *
 Path = setup_path(__package__)
 
 
+def _delete_empty_recordings(corpus: corpus.Corpus, removed_recordings_file: str):
+    """
+    Deletes all recordings that are empty after the filtering done by some of the jobs in this file.
+
+    :param c: Corpus for which to delete the empty recordings.
+    :param removed_recordings_file: File in which to dump all recordings that have been deleted.
+    """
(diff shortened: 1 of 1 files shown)


JackTemaki (approved):
If you want you could move the code into a separate private function in the file outside of the jobs, just to save on redundant code. Otherwise ok

corpus/filter.py
```diff
@@ -321,21 +325,50 @@ def unknown_filter(corpus: corpus.Corpus, recording: corpus.Recording, segment:
...
+    """
+
+    __sis_hash_exclude__ = {"delete_empty_recordings": False}
+
+    def __init__(
+        self,
+        bliss_corpus: Path,
+        min_duration: float = 0.1,
+        max_duration: float = 120.0,
+        delete_empty_recordings: bool = True,
```
michelwi, line 356:
```suggestion
        delete_empty_recordings: bool = False,
```
while i agree that `True` should be used by default, if we change the actual default value to `True`, then this breaks the hashes of all existing setups that used this job and did not specify a value.

michelwi (changes requested):

Icemole:
[This chunk of code](https://github.com/rwth-i6/i6_core/blob/249d34f6236160cfa5b62c64d30652626e6b3a95/corpus/filter.py#L255) is also a bit redundant now, but if I adapted it to the current structure there would be two loops over all_recordings. The theoretical time complexity is the same, but practically we need more time to retrieve from main memory and so on... Let me know if you want this shortened as well.

corpus/filter.py
Icemole: Thanks, I meant to set to False for hash compatibility purposes but this slipped past me.

michelwi:
> This chunk of code is also a bit redundant now

I vote for readability / consistence over minuscule time improvements.

michelwi (approved):

corpus/filter.py
michelwi: As you see, the AppTek hashes pipeline had also identified this problem. 👍

Icemole:
Would this be safe for merging now?

PR #501, opened 2024-04-24
TextDictToTextLinesJob

albertz:


michelwi (commented):
code lgtm.

But is `corpus/convert.py` the correct place for this or would users rather expect this unter `text` (because it processes text files) or `returnn` (because mostly used for returnn search output)?

albertz:
I also don't really care too much, so suggest sth, and I will move it there.

albertz:
Note that I will also work on another job, similar to `CorpusToStmJob` but operating on such text-dict format, so for that job (`TextDictToStmJob` or so), it will be the same question where to put it.

michelwi:
> I also don't really care too much, so suggest sth, and I will move it there.

I also don't have a strong opinion. maybe someone else has, or we can also leave it.

> (`TextDictToStmJob` or so), 

ToStm Jobs could even live next to recognition/scoring... 🤷 

or we do `text/convert.py` or even a dedicated `format_conversion` package?

michelwi:
Failed pipeline(s) is due to a bug in bitbucket where it sometimes fails to clone a repo from github. It is not a real problem.

JackTemaki:
`text/convert.py` would be fine for me, just not `corpus/conversion.py` because ideally (which is the case for 90% of the jobs, and 100% in corpus/conversion.py so far) they are all related to "bliss xml" corpus files.

albertz:
Ok I moved it to `text/convert.py`.

JackTemaki (approved):

michelwi (approved):

PR #503, opened 2024-04-24
SearchWordsDummyTimesToCTMJob

albertz:


returnn/search.py
```diff
@@ -490,6 +490,74 @@ def run(self):
...
+        self.filter_tags = filter_tags
+
+        self.out_ctm_file = self.output_path("search.ctm")
+
+    def tasks(self):
+        yield Task("run", mini_task=True)
+
+    def run(self):
+        # nan/inf should not be needed, but avoids errors at this point and will print an error below,
+        # that we don't expect an N-best list here.
```
michelwi, line 518: where exactly would this be printed?

michelwi (approved):

returnn/search.py
albertz: Well, in the line `words = text.split()`, it would throw an error, it will print sth like `split` is not a member of <whatever object type there was>. But I can add an extra `assert isinstance(text, str)` if you want (although we also don't have that for `SearchWordsToCTMJob`.

albertz: What I mean is, I think this exception you will see later will be easier to understand, in the sense that you see more easily what the problem is, compared to the exception you would see here in `eval`, which might be quite confusing when you stumble upon it.

michelwi: I don't think this would be necessary. I was just looking for the indicated error message 

> will print an error below, that we don't expect an N-best list

and was wondering if I missed anyting

albertz:
While testing this, I noticed, I think sclite requires that the sequences are in the same order as in the STM file. The code here just uses the same order as the search output, which is often a different order (e.g. reversely ordered by seq len). So, this is not useful in this form. There needs to be another option which defines the seq order. E.g. via an existing reference text-dict file (e.g. via `CorpusToTextDictJob`) or maybe the STM itself.

JackTemaki:
> While testing this, I noticed, I think sclite requires that the sequences are in the same order as in the STM file.

Doesn't the `sort_files` flag of ScliteJob solve this?

albertz:
> Doesn't the `sort_files` flag of ScliteJob solve this?

Ah maybe it does, I did not knew about this option. Will try soon. But I think this option `seq_order_file` might anyway be useful. But then I make it optional here.

sarahberanek:
> > Doesn't the `sort_files` flag of ScliteJob solve this?
> 
> Ah maybe it does, I did not knew about this option. Will try soon. But I think this option `seq_order_file` might anyway be useful. But then I make it optional here.

If you are using `hubscr.pl` from sclite as part of the pipeline, the sorting will be done automatically. Thats at least how it is done in  the `ApptekScorer` from `apptek_asr` to score.

albertz:
Ok good to know.

So, what about the PR here now? Code is good now? Can be merged?

Judyxujj (approved):

Atticus1806:
Code looks good to me, I am just wondering: Do we maybe want to have an if check for the `assert isinstance(d, dict), "only search output file with dict format is supported"` and `raise NotImplementedError`? Then we can maybe extend this job more freely in the future. But I dont have a strong opinion on this.

albertz:
> Code looks good to me, I am just wondering: Do we maybe want to have an if check for the `assert isinstance(d, dict), "only search output file with dict format is supported"` and `raise NotImplementedError`? Then we can maybe extend this job more freely in the future. But I dont have a strong opinion on this.

I don't think the type of exception (`AssertionError` vs `NotImplementedError`) is really critical here, and you save one line of code with `assert`.

PR #504, opened 2024-04-24
TextDictToStmJob

albertz:


text/convert.py
```diff
@@ -34,3 +37,125 @@ def run(self):
...
+    def __init__(
+        self,
+        text_dict: Path,
+        *,
+        exclude_non_speech: bool = True,
+        non_speech_tokens: Optional[List[str]] = None,
+        remove_punctuation: bool = True,
+        punctuation_tokens: Optional[Union[str, List[str]]] = None,
+        fix_whitespace: bool = True,
+        name: str = "",
```
michelwi, line 57: I don't think this name is ever used except to change the filename (on disk) of the output file.. which anyway should be irrelevant.
-> remove?

text/convert.py
```diff
@@ -34,3 +37,125 @@ def run(self):
...
+        tag_mapping: Sequence[Tuple[Tuple[str, str, str], Dict[int, Path]]] = (),
+    ):
+        """
+        :param text_dict: e.g. via :class:`CorpusToTextDictJob`
+        :param exclude_non_speech: non speech tokens should be removed
+        :param non_speech_tokens: defines the list of non speech tokens
+        :param remove_punctuation: should punctuation be removed
+        :param punctuation_tokens: defines list/string of punctuation tokens
+        :param fix_whitespace: should white space be fixed.
+            !!!be aware that the corpus loading already fixes white space!!!
```
michelwi, line 67: is this comment relevant for text dict or copy-pasted from the bliss job?
should we remove this parameter entirely if the `eval` call anyway corrects all whitespace errors?

text/convert.py
```diff
@@ -34,3 +37,125 @@ def run(self):
...
+    """
+
+    def __init__(
+        self,
+        text_dict: Path,
+        *,
+        exclude_non_speech: bool = True,
+        non_speech_tokens: Optional[List[str]] = None,
+        remove_punctuation: bool = True,
+        punctuation_tokens: Optional[Union[str, List[str]]] = None,
```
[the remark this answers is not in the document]
michelwi: we could even rename it `XXX_tokens_to_remove` to make this more clear.

text/convert.py
albertz: I copy-pasted the `CorpusToStmJob`. I personally don't use any of these options (i.e. I just use the defaults) and I did not really try to understand what they all do, i.e. which of those options are really necessary, and in what situations they make sense.

However, it seems to me that all of the options would apply to the text-dict file just in the same way as it would apply to the Bliss file, i.e. I don't see that there is anything specific about Bliss here, so I thought it makes sense to just have exactly the same logic.

Regarding this specific comment, I don't really understand what it means exactly, even for the `CorpusToStmJob`. I would assume it very much depends on your pipeline whether white space is already fixed. I don't know what "corpus loading" here refers to. I also don't exactly know what it means that "white space is fixed", i.e. what are the requirements for Sclite in the end. But again, whatever all this means, I don't think this is specific to Bliss, so this comment would apply to text-dict just as well, or not?

text/convert.py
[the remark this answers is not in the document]
albertz: I don't know. I did not think too much about it. If the options don't make sense, why do we have them in `CorpusToStmJob`?

I thought it would be good to have the options also consistent to `CorpusToStmJob` if possible, so it's simple to switch.

text/convert.py
albertz: Maybe it influences how it is shown in sclite? I don't really know.

But I checked, and I don't see any usage of `name` in `CorpusToStmJob` across i6_experiments (if PyCharm search is reliable here...), so maybe it's really rarely (or never) used, so then yea, we could remove this.

text/convert.py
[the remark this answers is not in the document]
albertz: Ok I also search for `exclude_non_speech` and `remove_punctuation` in i6_experiments and I also never see that they are set. So yea, maybe we can just remove them.

I see `non_speech_tokens` (and maybe also `punctuation_tokens`) being used a few times. So the question would be whether we change the name now and break consistency, or not.

text/convert.py
albertz: The `eval` call would not touch the content of a string. If there are leading/trailing whitespaces (or other kind of whitespaces within?) the string, `eval` would not affect that. So this option here definitely has an effect.

text/convert.py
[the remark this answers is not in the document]
albertz: I simplified that now. I removed `exclude_non_speech` and `remove_punctuation` and renamed the others to `remove_XXX`.

text/convert.py
albertz: So, what to do here? Just remove the comment? Or just leave it?

michelwi: I guess they are referring to [this line](https://github.com/rwth-i6/i6_core/blob/de835b85a513a279a131d855ed12754643ae73e9/lib/corpus.py#L126) in the corpus lib, which has the same code as in this job.

Therefore if one converted Bliss corpora to stm, this option should be a redundant.

Here we don't have this, therefore I would say remove the comment and keep the option.
```suggestion
```

text/convert.py
```diff
@@ -34,3 +37,119 @@ def run(self):
...
+        self.out_stm_path = self.output_path("corpus.stm")
+
+    def tasks(self):
+        yield Task("run", mini_task=True)
+
+    def run(self):
+        # nan/inf should not be needed, but avoids errors at this point and will print an error below,
+        # that we don't expect an N-best list here.
+        c = eval(uopen(self.text_dict, "rt").read(), {"nan": float("nan"), "inf": float("inf")})
+        assert isinstance(c, dict)
```
Atticus1806, line 88: Similar comment as in the other PR, would this make sense here too?

albertz: Not sure. Even the job name indicates that this is for the text-dict format.

Atticus1806: Dont have a strong opinion on this, was just thinking if at some point we wanted to extend if its easier one way around. But we can keep it as is.

Atticus1806 (approved):

michelwi (approved):

PR #505, opened 2024-04-26
SpellingConversionJob: allow Path as input otherwise breaks hashes.

sarahberanek:
If the spelling conversion mapping file is created by another job we can not just parse the string with `get_path() `without breaking hashes in our `apptek_asr` pipeline.
Therefore adding the option to give the file as a `tk.Path.`

### DIFF
--- a/lexicon/conversion.py
+++ b/lexicon/conversion.py
@@ -349,10 +349,7 @@ class SpellingConversionJob(Job):
         self.set_vis_name("Convert Between Regional Orth Spellings")
 
         self.bliss_lexicon = bliss_lexicon
-        if isinstance(orth_mapping_file, tk.Path):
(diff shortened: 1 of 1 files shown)


lexicon/conversion.py
```diff
@@ -348,7 +349,10 @@ def __init__(
         self.set_vis_name("Convert Between Regional Orth Spellings")
 
         self.bliss_lexicon = bliss_lexicon
-        self.orth_mapping_file = orth_mapping_file
```
Icemole, line 351: This would incur in the same issue as you're seeing right now, because this is done by the manager. This if should be done on the `run()` function instead.

Icemole (changes requested):

Icemole (approved):

kuacakuaca (approved):

PR #507, opened 2024-04-29
Add job to compute the context label from alignment

Judyxujj:


mm/context_label.py
```diff
@@ -0,0 +1,161 @@
...
+
+import numpy as np
+
+from i6_core.lib.hdf import get_returnn_simple_hdf_writer
+from i6_core.lib.rasr_cache import FileArchive
+from sisyphus import Job, Task, tk, gs
+from apptek_asr.artefacts.factory import AbstractArtefactRepository
+
+
+class GetContextLabelFromDenseTyingJob(Job):
```
Marvin84, line 12: Since you are also reading the center phoneme information, I would not call the job `GetContextLabel`, maybe `GetPhonemeLabelsFromNoTyingDense`? Also I think `DenseTying` is not clear, since it hints to some kind of tying, whereas here is about no tying at all but complete enumeration.

mm/context_label.py
```diff
@@ -0,0 +1,161 @@
...
+        for line in open(dense_tying_path.get_path(), "rt"):
+            if not line.startswith("#"):
+                num_classes = max(num_classes, int(line.strip().split()[1]))
+
+        state_tying = dict((k, int(v)) for l in open(dense_tying_path.get_path()) for k, v in [l.strip().split()[0:2]])
+
+        return state_tying, num_classes
+
+    @classmethod
+    def get_target_labels_from_dense(cls, dense_label: int, hmm_state: int, n_contexts: int) -> Tuple[int, int, int]:
```
Marvin84, line 64: The implementation here should reflect the exact implementation in RASR, IMO, where options such as use_boundary_classes and use_word_end_classes are also considered.

mm/context_label.py
```diff
@@ -0,0 +1,161 @@
...
+        """ """
+
+        futureLabel = np.mod(dense_label, n_contexts)
+        popFutureLabel = np.floor_divide(dense_label, n_contexts)
+
+        pastLabel = np.mod(popFutureLabel, n_contexts)
+        centerState = np.floor_divide(popFutureLabel, n_contexts)
+        centerState = np.floor_divide(centerState - hmm_state, 3)
```
Marvin84, line 72: This is not correct. Please see also the comment above.

Once the identity of right and left are popped, one is left with the center phoneme state information. Depending on how you dumped the `state-tying` options in `RASR`, you would do further steps to pop also word end class and hmm state info.
My suggestion here is that the class has a dictionary that has the label info, such as whether end of word or boundary classes are used, and what is the number of hmm state per phoneme. And then to apply the exact inverse function with respect to the implementation of RASR [here](https://github.com/rwth-i6/rasr/blob/master/src/Am/ClassicStateTying.cc#L272)

Marvin84 (changes requested):
Please see the single comment.

mm/context_label.py
```diff
@@ -0,0 +1,132 @@
...
+
+    def tasks(self):
+        yield Task("run", resume="run", rqmt=self.rqmt)
+
+    @classmethod
+    def get_tying_and_num_classes(cls, dense_tying_path: tk.Path) -> Tuple[Dict, int]:
+        num_classes = 0
+        for line in open(dense_tying_path.get_path(), "rt"):
+            if not line.startswith("#"):
+                num_classes = max(num_classes, int(line.strip().split()[1]))
```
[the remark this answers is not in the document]
michelwi: especially for dense tying, would the number of classes not be much smaller than the maximum value? i.e.
```
num_classes = len(state_tying.values())
```

mm/context_label.py
```diff
@@ -0,0 +1,132 @@
...
+
+        for file in alignment_cache.ft:
+            info = alignment_cache.ft[file]
+            if info.name.endswith(".attribs"):
+                continue
+
+            alignment = alignment_cache.read(file, "align")
+            aligned_allophones = ["%s.%d" % (alignment_cache.allophones[t[1]], t[2]) for t in alignment]
+            dense_targets = [dense_tying[allo] for allo in aligned_allophones]
+            hmm_state_ids = [alignment[i][2] for i in range(len(alignment))]
```
michelwi, line 96: more pythonic?
```suggestion
            hmm_state_ids = [align[2] for align in alignment]
```

mm/context_label.py
michelwi: or just parse the string representation of the allophone?

mm/context_label.py
```diff
@@ -0,0 +1,132 @@
...
+        return futureLabel, centerState, pastLabel
+
+    def run(self):
+        returnn_root = None if self.returnn_root is None else self.returnn_root.get_path()
+        SimpleHDFWriter = get_returnn_simple_hdf_writer(returnn_root)
+        out_hdf_left_context = SimpleHDFWriter(filename=self.out_hdf_left_context, dim=1)
+        out_hdf_right_context = SimpleHDFWriter(filename=self.out_hdf_right_context, dim=1)
+        out_hdf_center_context = SimpleHDFWriter(filename=self.out_hdf_center_context, dim=1)
+
+        dense_tying, _ = self.get_tying_and_num_classes(self.dense_tying_path)
```
michelwi, line 83: if num classes is not used, why compute it?

michelwi (changes requested):

mm/context_label.py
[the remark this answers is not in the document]
Marvin84: > Why only dense tying, given an alignment we can always get left/center/right context.
You would need to infer the identity in this case from the allophone strings, and then you would need to know what type of labeling you would do for the phonemes (depending on their position within the lexicon).
Here, the idea is to have this type of info already consistent with the RASR labeling logic and therefore available state-tyings. The labeling that gives easily access to the identity of left, center, right with all the different flavors for the center (phoneme? phoneme state? word end distinction) is in fact no tying dense.

Marvin84 (approved):
Looks good to me!

JackTemaki:
The code looks ok on the surface after a quick glance, but as I am not familiar with the logic I would prefer if others approve that have more knowledge of this.

mm/context_label.py
```diff
@@ -0,0 +1,182 @@
...
+            for k, g in itertools.groupby(dense_targets):
+                seg_len = len(list(g))
+                f, c, l = self.get_target_labels_from_dense(k, self.dense_label_info)
+
+                past_label_strings = past_label_strings + [l] * seg_len
+                center_state_strings = center_state_strings + [c] * seg_len
+                future_label_strings = future_label_strings + [f] * seg_len
+
+            out_hdf_left_context.insert_batch(
+                inputs=np.array(past_label_strings).reshape(1, -1, 1),
```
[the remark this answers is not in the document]
Judyxujj: The `sparse` is added as an option the same way as in  `RasrAlignmentDumpHDFJob`

mm/context_label.py
```diff
@@ -0,0 +1,203 @@
...
+        self.out_hdf_right_context = self.output_path("right_context.hdf")
+        self.out_hdf_center_context = self.output_path("center_context.hdf")
+
+        self.rqmt = {"cpu": 1, "mem": 8, "time": 0.5}
+
+    def tasks(self):
+        yield Task("run", resume="run", rqmt=self.rqmt)
+
+    @classmethod
+    def get_tying(cls, dense_tying_path: tk.Path) -> Tuple[Dict, int]:
```
albertz, line 71:
```suggestion
    def get_tying(cls, dense_tying_path: tk.Path) -> Dict[str, int]:
```
?

albertz, line 71: I would also add some documentation what exactly the returned dict is representing. I guess "state repr -> state idx" or sth like that?

mm/context_label.py
```diff
@@ -0,0 +1,203 @@
...
+        if dense_label_info.use_boundary_classes:
+            boundary_class = np.mod(center_state, num_boundary_classes)
+            center_state = np.floor_divide(center_state, num_boundary_classes)
+
+        hmm_state_class = np.mod(center_state, dense_label_info.num_hmm_states_per_phon)
+        center_label = np.floor_divide(center_state, dense_label_info.num_hmm_states_per_phon)
+
+        return future_label, center_label, past_label
+
+    def sanity_check(self, max_class_index, dense_label_info):
```
albertz, line 103:
```suggestion
    def sanity_check(self, max_class_index: int, dense_label_info: DenseLabelInfo):
```
?

mm/context_label.py
```diff
@@ -0,0 +1,203 @@
...
+        if dense_label_info.use_word_end_classes:
+            word_end_class = np.mod(center_state, num_word_end_classes)
+            center_state = np.floor_divide(center_state, num_word_end_classes)
+
+        if dense_label_info.use_boundary_classes:
+            boundary_class = np.mod(center_state, num_boundary_classes)
+            center_state = np.floor_divide(center_state, num_boundary_classes)
+
+        hmm_state_class = np.mod(center_state, dense_label_info.num_hmm_states_per_phon)
+        center_label = np.floor_divide(center_state, dense_label_info.num_hmm_states_per_phon)
```
albertz, line 99: Why do you use `np.mod` and `np.floor_divice` and not just `%` and `//`? I think the latter is simpler and cleaner and better.

```suggestion
        future_label = dense_label % dense_label_info.n_contexts
        pop_future_label = dense_label // dense_label_info.n_contexts

        past_label = pop_future_label % dense_label_info.n_contexts
        center_state = pop_future_label // dense_label_info.n_contexts

        if dense_label_info.use_word_end_classes:
            word_end_class = center_state % num_word_end_classes
            center_state = center_state // num_word_end_classes

        if dense_label_info.use_boundary_classes:
            boundary_class = center_state % num_boundary_classes
            center_state = center_state // num_boundary_classes

        hmm_state_class = center_state % dense_label_info.num_hmm_states_per_phon
        center_label = center_state // dense_label_info.num_hmm_states_per_phon
```

mm/context_label.py
```diff
@@ -69,6 +69,9 @@ def tasks(self):
 
     @classmethod
     def get_tying(cls, dense_tying_path: tk.Path) -> Dict[str, int]:
+        """
+        reads state tying file and return dict "state repr -> state idx"
```
albertz, line 73:
```suggestion
        reads state tying file
        
        :return: dict state repr -> state idx
```

albertz (approved):
After my final small comment, looks ok from my side.

mm/context_label.py
```diff
@@ -0,0 +1,210 @@
...
+
+        self.rqmt = {"cpu": 1, "mem": 8, "time": 0.5}
+
+    def tasks(self):
+        yield Task("run", resume="run", rqmt=self.rqmt)
+
+    @classmethod
+    def get_tying(cls, dense_tying_path: tk.Path) -> Dict[str, int]:
+        """
+        reads state tying file and return dict "state repr -> state idx"
```
albertz, line 73:
```suggestion
        reads state tying file
        
        :return: dict state repr -> state idx
```

michelwi (approved):
I am fine with the code. We could still discuss if the sparse option should be true by default, or if your use case actually needs the dense representation. (?)

mm/context_label.py
```diff
@@ -0,0 +1,212 @@
...
+    ):
+        """
+        Get past/center/future context label of alignment by calculating back labels from dense tying and write the
+        labels into hdf file.
+        (C.f. NoStateTyingDense in rasr
+        https://github.com/rwth-i6/rasr/blob/a942e3940c30eeba900c873f3bfb3f48d5b39ddb/src/Am/ClassicStateTying.cc#L272)
+
+        :param alignment_cache_path: path to alginment cache
+        :param allophone_path: path to allohone
+        :param dense_tying_path: path to denser tying file
```
kuacakuaca, line 46: typo: denser -> dense

PR #508, opened 2024-05-06
remove cleanup_old_models from disallowed args

JackTemaki:
After working for it a longer time, and also as request from @Atticus1806, we figured it should be absolutely okay to force cleanup_old_models to be hashed.


JackTemaki:
Ooops, this was already merged, not sure why this was still open for me

PR #509, opened 2024-05-20
Add plots to AlignmentJob

Icemole:
Plots have only been added to `i6_core.corpus.filter.FilterSegmentsByAlignmentConfidenceJob`. However, it's useful to see the alignment score plot before actually filtering it. This PR adds optional (but recommended, given the default value of the `plot_alignment_scores` parameter) functionality to `i6_core.mm.AlignmentJob`.

Edit: the plotting is improved with respect to the former job in the sense that the former job only handled positive numbers between 0 and 200, while the current code can handle all ranges of numbers. This is useful, for instance, when aligning with a neural-network-based feature scorer that outputs log-probabilities as alignment scores.

Note that even if the functionality of the previous jobs changes, the practical output result is the same: whoever ignored plots because they used the job prior to this PR can ignore them, and whoever uses the plots from now on won't suffer any consequence.

I've also made it so that hashes don't change either.

### DIFF
--- a/mm/alignment.py
+++ b/mm/alignment.py
@@ -157,9 +157,7 @@ class AlignmentJob(rasr.RasrCommand, Job):
         logging.info(
             f"Max {np_alignment_scores.max()}; min {np_alignment_scores.min()}; median {np.median(np_alignment_scores)}"
         )
-        logging.info(
-            f"Total number of segments: {np_alignment_scores.size}; 90-th percentile: {higher_percentile.size}"
-        )
+        logging.info(f"Total number of segments: {np_alignment_scores.size}; 90-th percentile: {higher_percentile}")
 
         # Plot the data.
         matplotlib.use("Agg")
--- a/mm/alignment.py
+++ b/mm/alignment.py
@@ -20,7 +20,7 @@ class AlignmentJob(rasr.RasrCommand, Job):
     Align a dataset with the given feature scorer.
     """
 
-    __sis_hash_exclude__ = {"plot_alignment_scores": True}
+    __sis_hash_exclude__ = {"plot_alignment_scores": False}
 
     def __init__(
         self,
@@ -33,7 +33,7 @@ class AlignmentJob(rasr.RasrCommand, Job):
         rtf=1.0,
         extra_config=None,
         extra_post_config=None,
-        plot_alignment_scores=True,
+        plot_alignment_scores=False,
     ):
         """
         :param rasr.crp.CommonRasrParameters crp:
@@ -46,6 +46,7 @@ class AlignmentJob(rasr.RasrCommand, Job):
         :param extra_config:
         :param extra_post_config:
         :param plot_alignment_scores: Whether to plot the alignment scores (normalized over time) or not.
+            The recommended value is `True`. The default value is `False` for retrocompatibility purposes.
         """
         assert isinstance(feature_scorer, rasr.FeatureScorer)
 
--- a/mm/alignment.py
+++ b/mm/alignment.py
@@ -153,13 +153,16 @@ class AlignmentJob(rasr.RasrCommand, Job):
             del document
 
         np_alignment_scores = np.asarray(alignment_scores)
+        higher_percentile = np.percentile(np_alignment_scores, 90)  # There can be huge outliers.
         logging.info(
-            f"Max {np_alignment_scores.max()}; Min {np_alignment_scores.min()}; Median {np.median(np_alignment_scores)}"
+            f"Max {np_alignment_scores.max()}; min {np_alignment_scores.min()}; median {np.median(np_alignment_scores)}"
+        )
+        logging.info(
+            f"Total number of segments: {np_alignment_scores.size}; 90-th percentile: {higher_percentile.size}"
         )
 
         # Plot the data.
         matplotlib.use("Agg")
-        higher_percentile = np.percentile(np_alignment_scores, 90)  # There can be huge outliers.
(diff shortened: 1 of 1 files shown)


mm/alignment.py
```diff
@@ -121,6 +135,38 @@ def run(self, task_id):
...
+            document = ET.parse(util.uopen(file_path))
+            _seg_list = document.findall(".//segment")
+            for seg in _seg_list:
+                avg = seg.find(".//score/avg")
+                alignment_scores.append(float(avg.text))
+            del document
+
+        np_alignment_scores = np.asarray(alignment_scores)
+        logging.info(
+            f"Max {np_alignment_scores.max()}; Min {np_alignment_scores.min()}; Median {np.median(np_alignment_scores)}"
```
kuacakuaca, line 157: maybe also add the total number of segments/alignment scores and the number of segments above the 90 percentile, since those are not taken for plotting?

Icemole: Sure, I can do that.

I wonder if it would be interesting to add an output file consisting of all segments along with their estimated percentile.

mm/alignment.py
```diff
@@ -15,6 +16,12 @@
 
 
 class AlignmentJob(rasr.RasrCommand, Job):
+    """
+    Align a dataset with the given feature scorer.
+    """
+
+    __sis_hash_exclude__ = {"plot_alignment_scores": True}
```
michelwi, line 23: A very good idea to add plotting to the AlignmentJob, which I would endorse.

I my opinion the `__sis_hash_exclude__` should be with the value that recovers the old behavior. And then also the default value should be the one with the default behavior. Otherwise some jobs that have been run with a previous commit would now have the output but not have the file created.

michelwi (changes requested):
request changes for `higher_percentile.size` and `tk.uncached_path`, others discussion welcome.

mm/alignment.py
Icemole: I thought about it, but I didn't consider it an issue. Yes, the functionality (i.e. spawned tasks) changes, but either way the user wasn't expecting to use the plots, so nothing really changed from the point of view of the user, I think.

If you insist on this, I will revert to the default behavior.

mm/alignment.py
```diff
@@ -121,6 +135,41 @@ def run(self, task_id):
...
+        alignment_scores = []
+        for log_file in self.out_log_file.values():
+            logging.info("Reading: {}".format(log_file))
+            file_path = tk.uncached_path(log_file)
+            document = ET.parse(util.uopen(file_path))
+            _seg_list = document.findall(".//segment")
+            for seg in _seg_list:
+                avg = seg.find(".//score/avg")
+                alignment_scores.append(float(avg.text))
+            del document
```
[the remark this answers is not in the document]
Icemole: There seems to be [`ElementTree.iterparse`](https://docs.python.org/3/library/xml.etree.elementtree.html#xml.etree.ElementTree.iterparse). I will try to implement this.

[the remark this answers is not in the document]
Icemole: `ElementTree.iterparse()` doesn't seem to have any context manager functionality. Besides:

> [The resulting `ElementTree.iterparse` object] has a `root` attribute that references the root element of the resulting XML tree once source is fully read.

So I don't think we can rely on this function to minimize memory footprint.

There's [`Element.clear()`](https://docs.python.org/3/library/xml.etree.elementtree.html#xml.etree.ElementTree.Element.clear) which we can use instead of `del document`:

```suggestion
            document = ET.iterparse(util.uopen(file_path))
            root = document.getroot()
            _seg_list = root.findall(".//segment")
            # ...
            root.clear()
```

But I find it to be very similar, if not the same, as `del document`. And I'm not sure it actually works.

---

Update after some researching: some users have used the `iterparse()` + `Element.clear()` combo that I mentioned above in different ways:
1. Clear element by element. [Source](https://stackoverflow.com/questions/14924200/loading-huge-xml-files-and-dealing-with-memoryerror).
2. Clearing element by element (solution above) doesn't work?? So clear the root element every time??? [Source](https://stackoverflow.com/questions/10074200/should-memory-usage-increase-when-using-elementtree-iterparse-when-clearing).

The second alternative didn't seem very intuitive for me, since the documentation says:
> `clear()`
Resets an element. This function removes all subelements, clears all attributes, and sets the text and tail attributes to None.

[Source](https://docs.python.org/3/library/xml.etree.elementtree.html#xml.etree.ElementTree.Element.clear).

Although I did find this answer in Stack Overflow, which reinforces my idea that there should be an emphasis on the docs on **current subelements**:

> the "root.clear()" strategy in the ElementTree documentation only removes fully parsed children of the root

[Source](https://stackoverflow.com/a/44509632).

I guess all subelements of a partial root doesn't involve modifying the future root, but the documentation is very scarce about this topic. If you want I can implement it like this, but as I said we're still actively managing memory through `Element.clear()`.

Let me know your thoughts.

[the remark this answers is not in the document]
michelwi: > Let me know your thoughts.

you did some great researching and in the end I think having only a single statement is more nice that having an even more complicated implementation that only "hides" the explicit memory management.. ok, lets keep it until there are better suggestions.

mm/alignment.py
michelwi: > If you insist on this

lets say I have a weak preference. But I will be ok if others agree with you

Icemole: Meh, I can live with this. An approval in the hand is worth two in the bush :woman_shrugging:

michelwi (approved):
I shall approve then

curufinwe (approved):

PR #510, opened 2024-05-22
Pass input_names and output_names directly in TorchOnnxExportJob

kuacakuaca:
https://github.com/rwth-i6/returnn/issues/1517

### DIFF
--- a/returnn/compile.py
+++ b/returnn/compile.py
@@ -242,9 +242,8 @@ class TorchOnnxExportJob(Job):
 
         self.returnn_config = returnn_config
         self.checkpoint = checkpoint
-
-        input_names = list(returnn_config.config["extern_data"].keys()) if input_names is None else input_names
-        output_names = list(returnn_config.config["model_outputs"].keys()) if output_names is None else output_names
+        input_names = list(returnn_config.config["extern_data"].keys()) if ("extern_data" in returnn_config.config and input_names is None) else input_names
+        output_names = list(returnn_config.config["model_outputs"].keys()) if ("model_outputs" in returnn_config.config and output_names is None) else output_names
         self.input_names = input_names
         self.output_names = output_names
 
--- a/returnn/compile.py
+++ b/returnn/compile.py
@@ -232,17 +232,22 @@ class TorchOnnxExportJob(Job):
         :param returnn_config: RETURNN config object
         :param checkpoint: Path to the checkpoint for export
         :param device: target device for graph creation
-        :param input_names: sequence of model input names
+        :param input_names: sequence of model input names.
+            If not specified, will automatically determine from `extern_data` when available in `returnn_config.config`.
         :param output_names: sequence of model output names.
-            If not specified, will automatically determine from model_outputs when available in returnn_config.config.
+            If not specified, will automatically determine from `model_outputs` when available in `returnn_config.config`.
         :param returnn_python_exe: file path to the executable for running returnn (python binary or .sh)
         :param returnn_root: file path to the RETURNN repository root folder
         """
 
         self.returnn_config = returnn_config
         self.checkpoint = checkpoint
+
+        input_names = list(returnn_config.config["extern_data"].keys()) if input_names is None else input_names
+        output_names = list(returnn_config.config["model_outputs"].keys()) if output_names is None else output_names
         self.input_names = input_names
         self.output_names = output_names
+
         self.device = device
         self.returnn_python_exe = util.get_returnn_python_exe(returnn_python_exe)
         self.returnn_root = util.get_returnn_root(returnn_root)
--- a/returnn/compile.py
+++ b/returnn/compile.py
@@ -233,7 +233,8 @@ class TorchOnnxExportJob(Job):
         :param checkpoint: Path to the checkpoint for export
         :param device: target device for graph creation
         :param input_names: sequence of model input names
-        :param output_names: sequence of model output names
+        :param output_names: sequence of model output names.
+            If not specified, will automatically determine from model_outputs when available in returnn_config.config.
         :param returnn_python_exe: file path to the executable for running returnn (python binary or .sh)
         :param returnn_root: file path to the RETURNN repository root folder
         """
--- a/returnn/compile.py
+++ b/returnn/compile.py
@@ -232,7 +232,7 @@ class TorchOnnxExportJob(Job):
         :param returnn_config: RETURNN config object
         :param checkpoint: Path to the checkpoint for export
         :param device: target device for graph creation
-        :param input_names: sequence of model output names
+        :param input_names: sequence of model input names
         :param output_names: sequence of model output names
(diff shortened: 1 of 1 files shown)


returnn/compile.py
```diff
@@ -228,12 +232,16 @@ def __init__(
...
         :param device: target device for graph creation
+        :param input_names: sequence of model output names
+        :param output_names: sequence of model output names
         :param returnn_python_exe: file path to the executable for running returnn (python binary or .sh)
         :param returnn_root: file path to the RETURNN repository root folder
         """
 
         self.returnn_config = returnn_config
         self.checkpoint = checkpoint
+        self.input_names = input_names
```
albertz, line 243: You miss the logic I stated in https://github.com/rwth-i6/returnn/issues/1517:
```suggestion
        input_names = list(returnn_config.config["extern_data"].keys()) if ("extern_data" in returnn_config.config and input_names is None) else input_names
        output_names = list(returnn_config.config["model_outputs"].keys()) if ("model_outputs" in returnn_config.config and output_names is None) else output_names
        self.input_names = input_names
```

returnn/compile.py
```diff
@@ -228,12 +232,16 @@ def __init__(
         :param returnn_config: RETURNN config object
         :param checkpoint: Path to the checkpoint for export
         :param device: target device for graph creation
+        :param input_names: sequence of model output names
```
albertz, line 235:
```suggestion
        :param input_names: sequence of model input names
```

albertz: Actually, also should explain what happens if you don't pass it.
```suggestion
        :param input_names: sequence of model input names.
            If not specified, will automatically determine from extern_data when available in returnn_config.config.
```

returnn/compile.py
```diff
@@ -228,12 +232,16 @@ def __init__(
...
+        :param output_names: sequence of model output names
```
albertz, line 236:
```suggestion
        :param output_names: sequence of model output names.
            If not specified, will automatically determine from model_outputs when available in returnn_config.config.
```

returnn/compile.py
kuacakuaca: why can't we leave them `None`?

albertz: You can, but you should somewhere have the logic from the code I posted.

kuacakuaca: Now I get it. thanks!

returnn/compile.py
```diff
@@ -228,12 +232,17 @@ def __init__(
         :param returnn_config: RETURNN config object
         :param checkpoint: Path to the checkpoint for export
         :param device: target device for graph creation
+        :param input_names: sequence of model input names
```
albertz, line 235:
```suggestion
        :param input_names: sequence of model input names.
            If not specified, will automatically determine from extern_data when available in returnn_config.config.
```

returnn/compile.py
```diff
@@ -228,12 +232,22 @@ def __init__(
...
+            If not specified, will automatically determine from `model_outputs` when available in `returnn_config.config`.
         :param returnn_python_exe: file path to the executable for running returnn (python binary or .sh)
         :param returnn_root: file path to the RETURNN repository root folder
         """
 
         self.returnn_config = returnn_config
         self.checkpoint = checkpoint
+
+        input_names = list(returnn_config.config["extern_data"].keys()) if input_names is None else input_names
+        output_names = list(returnn_config.config["model_outputs"].keys()) if output_names is None else output_names
```
albertz, line 247: It's still not exactly what I wrote. It should be:
```suggestion
        input_names = list(returnn_config.config["extern_data"].keys()) if ("extern_data" in returnn_config.config and input_names is None) else input_names
        output_names = list(returnn_config.config["model_outputs"].keys()) if ("model_outputs" in returnn_config.config and output_names is None) else output_names
```

albertz:
Needs some black reformatting now. Also not sure if the docstring is maybe too long. Then needs some line break there as well.

But otherwise looks good to me.

returnn/compile.py
```diff
@@ -228,12 +232,29 @@ def __init__(
         :param returnn_config: RETURNN config object
         :param checkpoint: Path to the checkpoint for export
         :param device: target device for graph creation
+        :param input_names: sequence of model input names.
```
michelwi, line 235: the order of parameters in function definition should match the order of the docstring (currently `device` is off)

returnn/compile.py
```diff
@@ -228,12 +232,29 @@ def __init__(
...
+            if ("extern_data" in returnn_config.config and input_names is None)
+            else input_names
+        )
+        output_names = (
+            list(returnn_config.config["model_outputs"].keys())
+            if ("model_outputs" in returnn_config.config and output_names is None)
+            else output_names
+        )
+        self.input_names = input_names
+        self.output_names = output_names
```
michelwi, line 256: Is it valid to have only one out of `input_names`, `output_names` set? (e.g. user misconfiguration, or the config contains `extern_data` but not `model_outputs`).

Should we assert that both have the same length?

albertz: > Is it valid to have only one out of `input_names`, `output_names` set?

You mean for the job? Or for the script? But both would be valid. If one is not set, it would be automatically determined.

> (e.g. user misconfiguration, or the config contains `extern_data` but not `model_outputs`).

This would anyway then be an error in the script, because it anyway needs them to define the shape/dims.

> Should we assert that both have the same length?

You mean the input_names len = output_names len? But this doesn't need to be the case?

Or you mean that input_names matches extern_data keys, and also output_names matches model_outputs? That is already checked by the script.

returnn/compile.py
```diff
@@ -228,12 +232,29 @@ def __init__(
...
+        :param output_names: sequence of model output names.
+            If not specified, will automatically determine from `model_outputs` when available in `returnn_config.config`.
         :param returnn_python_exe: file path to the executable for running returnn (python binary or .sh)
         :param returnn_root: file path to the RETURNN repository root folder
         """
 
         self.returnn_config = returnn_config
         self.checkpoint = checkpoint
+        input_names = (
+            list(returnn_config.config["extern_data"].keys())
```
albertz, line 246:
```suggestion
            # Get the list here, because ReturnnConfig serialization might potentially reorder via sort_config=True.
            list(returnn_config.config["extern_data"].keys())
```

returnn/compile.py
michelwi: > You mean for the job? Or for the script?

I meant for the script.

> If one is not set, it would be automatically determined.

ok, this answers my question, thanks.

michelwi (approved):

albertz (approved):

NeoLegends:
After this PR I'm having some issues w/ dynamic axes (e.g. `data:size1`) not being specified by the job, but being inferred by RETURNN. I think these should be added to the list of inferred input and output names, too.

PR #511, opened 2024-05-28
Add type hints to returnn/compile.py

Icemole:
The requirements were hardcoded to None, so it was a bit hard to achieve something from this job other than for it to be a mini task.

### DIFF
--- a/returnn/compile.py
+++ b/returnn/compile.py
@@ -32,7 +32,7 @@ class CompileTFGraphJob(Job):
         epoch: Optional[Union[int, tk.Variable]] = None,
         verbosity: int = 4,
         device: Optional[str] = None,
-        summaries_tensor_name=None,
+        summaries_tensor_name: Optional[str] = None,
         output_format: str = "meta",
         returnn_python_exe: Optional[tk.Path] = None,
         returnn_root: Optional[tk.Path] = None,
--- a/returnn/compile.py
+++ b/returnn/compile.py
@@ -25,37 +25,37 @@ class CompileTFGraphJob(Job):
 
     def __init__(
         self,
-        returnn_config,
-        train=0,
-        eval=0,
-        search=0,
-        epoch=None,
-        verbosity=4,
-        device=None,
+        returnn_config: Union[ReturnnConfig, tk.Path, str],
+        train: int = 0,
+        eval: int = 0,
+        search: int = 0,
+        epoch: Optional[Union[int, tk.Variable]] = None,
+        verbosity: int = 4,
+        device: Optional[str] = None,
         summaries_tensor_name=None,
-        output_format="meta",
-        returnn_python_exe=None,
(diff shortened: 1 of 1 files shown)


Atticus1806:
Good Idea but I think you need to add a hash exclude similar to the other two for the default in order for it not to break hashes.
Also thanks for the typing refactoring :)

kuacakuaca (approved):

kuacakuaca (changes requested):
request changes because of hash break

returnn/compile.py
```diff
@@ -2,62 +2,60 @@
...
-        verbosity=4,
-        device=None,
+        returnn_config: Union[ReturnnConfig, tk.Path, str],
+        train: int = 0,
+        eval: int = 0,
+        search: int = 0,
+        epoch: Optional[Union[int, tk.Variable]] = None,
+        verbosity: int = 4,
+        device: Optional[str] = None,
         summaries_tensor_name=None,
```
michelwi, line 35:
```suggestion
        summaries_tensor_name: str =None,
```
I think this should be str?

returnn/compile.py
```diff
@@ -2,62 +2,60 @@
...
-        output_format="meta",
-        returnn_python_exe=None,
-        returnn_root=None,
-        rec_step_by_step=None,
-        rec_json_info=False,
+        output_format: str = "meta",
+        returnn_python_exe: Optional[tk.Path] = None,
+        returnn_root: Optional[tk.Path] = None,
+        rec_step_by_step: Optional[tk.Path] = None,
```
michelwi, line 39:
```suggestion
        rec_step_by_step: Optional[str] = None,
```
I think the old typing was correct

Icemole:
@Atticus1806 @michelwi @kuacakuaca I've already fixed this, thanks for warning me.

I've also done a bit more type refactoring since you liked it :)

returnn/compile.py
Icemole: I guess so, but I didn't see any example of this.

I'll just set it as string :woman_shrugging:

returnn/compile.py
Icemole: Definitely. Copy paste issue :sweat:

Atticus1806 (approved):

curufinwe:
Iirc having rqmt as a parameter is discouraged. You can just update the resulting job object to set new requirements. This is done in many other jobs like that.

Icemole:
@curufinwe suggested me to just set the created job's requirements manually through `job.rqmt = ...`. Would that be something you'd agree on? If so, maybe this PR could be just about adding some typing hints and cleaning up the code a bit ^^

Icemole:
I've just reverted the code and changed the title accordingly :slightly_smiling_face:

michelwi (approved):

PR #512, opened 2024-06-07
Add more options for hugging face download job

mmz33:


datasets/huggingface.py
```diff
@@ -8,6 +8,8 @@
 
 from i6_core.util import instanciate_delayed
 
+from datasets import load_dataset, Split
```
albertz, line 11: I would not do the import here. It should only be done inside the job.

datasets/huggingface.py
```diff
@@ -24,13 +24,19 @@ class DownloadAndPrepareHuggingFaceDatasetJob(Job):
     https://github.com/huggingface/datasets/issues/4179
     """
 
+    __sis_hash_exclude__ = {"split": None, "token": None}
+
+    from datasets import Split
```
albertz, line 29: You should not import sth in the global scope but inside the job only (i.e. only inside `run`).

datasets/huggingface.py
dthulke: It needs to be here to be able to define the type hint, right?

datasets/huggingface.py
mmz33: but then how to import the class type `Split`?

dthulke (approved):
LGTM

datasets/huggingface.py
albertz: Ah for type hinting, you would do sth like:
```python
from typing import TYPE_CHECKING

if TYPE_CHECKING:
    from datasets import Split
```
But this is not in the class scope but in global scope.

datasets/huggingface.py
```diff
@@ -24,13 +24,19 @@ class DownloadAndPrepareHuggingFaceDatasetJob(Job):
...
+
     def __init__(
         self,
         path: Union[str, DelayedBase],
         name: Optional[str] = None,
         *,
         data_files: Optional[Any] = None,
         revision: Optional[str] = None,
+        split: Optional[Union[str, Split]] = None,
```
albertz, line 38: Is that really what you would use in a Sisyphus config, this custom `Split` object? Can't this be just a string?

datasets/huggingface.py
albertz: But see my other comment below: Do you really want that? I think `split` would be specified as string usually, or not?

mmz33: Ok done.

datasets/huggingface.py
mmz33: i am not sure if we would really use the `Split` object. I guess in most cases we will use a string. But i just added that so it is consistent to the source code documentation.

datasets/huggingface.py
mmz33: Sorry I pushed before seeing your message. I think we would use a string usually. Thus, I will remove this import then.

datasets/huggingface.py
mmz33: Ok the import is removed now.

datasets/huggingface.py
```diff
@@ -89,5 +98,7 @@ def hash(cls, kwargs):
             "name": kwargs["name"],
             "data_files": kwargs["data_files"],
             "revision": kwargs["revision"],
+            "split": kwargs["split"],
+            "token": kwargs["token"],
```
albertz, line 102: Ah I realize, we do custom hashing. The sis hash exclude is ignored then. 

And this here is wrong. It will change the hash.

datasets/huggingface.py
```diff
@@ -24,13 +24,17 @@ class DownloadAndPrepareHuggingFaceDatasetJob(Job):
     https://github.com/huggingface/datasets/issues/4179
     """
 
+    __sis_hash_exclude__ = {"split": None, "token": None}
```
albertz, line 27: Remove this.

albertz (changes requested):

datasets/huggingface.py
mmz33: No wait. It is called via `super().hash(d)` or? I think it still use the `__sis_hash_exclude__`.

datasets/huggingface.py
```diff
@@ -24,13 +24,17 @@ class DownloadAndPrepareHuggingFaceDatasetJob(Job):
...
+
```
albertz, line 28:
```suggestion
```

albertz: Oh sorry, you are right.

datasets/huggingface.py
albertz: Oh right. Ok then.

albertz (approved):

PR #515, opened 2024-06-20
ExtractSentencePieceVocabJob

albertz:


JackTemaki (approved):

Atticus1806 (approved):

text/label/sentencepiece/vocab.py
```diff
@@ -0,0 +1,29 @@
...
+from sisyphus import Job, Task, tk
+
+
+class ExtractSentencePieceVocabJob(Job):
+    """
+    Extract the vocab from a sentence piece model (SPM)
+    """
+
+    def __init__(self, model: tk.Path):
+        super().__init__()
```
michelwi, line 14: We never call `Job.__init__`. In fact this is called already in the `Jobs` metaclass `__call__` (https://github.com/rwth-i6/sisyphus/blob/1be2f3c87e67c2c5142223ace435139fe39bca94/sisyphus/job.py#L120) via `Job._sis_init` (https://github.com/rwth-i6/sisyphus/blob/1be2f3c87e67c2c5142223ace435139fe39bca94/sisyphus/job.py#L230)

Not sure if it has side effects when you call it here as well.

text/label/sentencepiece/vocab.py
```diff
@@ -0,0 +1,29 @@
...
+        self.model = model
+        self.out_vocab = self.output_path("spm.vocab")
```
michelwi, line 16: we usually have a newline between job inputs and outputs.

text/label/sentencepiece/vocab.py
albertz: The `self.__init__` you reference will call `ExtractSentencePieceVocabJob.__init__`, not `Job.__init__`.

If you say we never call `Job.__init__`, I would say this is actually a bug then in most other jobs. Although, actually there is no `Job.__init__`, i.e. the `super().__init__()` here will just call `object.__init__()`, so it has no effect. But it's still good practice to anyway call it when you derive from some other class.

michelwi: yes, sorry, you are right.

PR #516, opened 2024-06-21
CorpusObject: add parameters to init

Icemole:
I find it much better to provide the parameters to an object via `__init__` than to provide them by creating the object and explicitly setting the actual attribute. This helps for instance when I have to create a bunch of `CorpusObject`s in a list comprehension, which is similar to my use case.

If there was any explicit reason why these were not in the `__init__` function please let me know.

### DIFF
--- a/meta/system.py
+++ b/meta/system.py
@@ -40,10 +40,10 @@ class CorpusObject(tk.Object):
         duration: Optional[float] = None,
     ):
         """
-        self.corpus_file: Bliss corpus xml
-        self.audio_dir: Audio directory if paths are relative (usually not needed)
-        self.audio_format: Format type of the audio files, see e.g. get_input_node_type()
-        self.duration: Duration of the corpus, is used to determine job time
+        :param corpus_file: Bliss corpus xml
+        :param audio_dir: Audio directory if paths are relative (usually not needed)
+        :param audio_format: Format type of the audio files, see e.g. get_input_node_type()
+        :param duration: Duration of the corpus, is used to determine job time
         """
         self.corpus_file = corpus_file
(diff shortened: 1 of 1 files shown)


JackTemaki:
Is there anything that speaks against making this a `dataclass` instead?

Icemole:
> Is there anything that speaks against making this a dataclass instead?

That's also a good idea :) the only thing I'm worried about is inheriting from a non-`dataclass` like `tk.Object`. Otherwise, nothing at all!

JackTemaki:
The inheritance can be removed, it plays no role at all.

meta/system.py
```diff
@@ -32,11 +32,23 @@ class CorpusObject(tk.Object):
...
+        corpus_file: Optional[tk.Path] = None,
+        audio_dir: Optional[tk.Path] = None,
+        audio_format: Optional[str] = None,
+        duration: Optional[float] = None,
+    ):
+        """
+        self.corpus_file: Bliss corpus xml
+        self.audio_dir: Audio directory if paths are relative (usually not needed)
+        self.audio_format: Format type of the audio files, see e.g. get_input_node_type()
+        self.duration: Duration of the corpus, is used to determine job time
```
albertz, line 46: The docstring format is wrong.

Icemole:
> The inheritance can be removed, it plays no role at all as far as I can see. At least in i6_experiments there is not a single check for isinstance(tk.Object) and it does exactly nothing.

Thanks for the feedback. I'll change it and leave it throughout the weekend, and if nobody complains we could merge this next week.

michelwi (approved):
I was worried this might change hashes, but apparently it does not.

PR #518, opened 2024-06-24
requirements: add dm-tree

albertz:
`i6_core/tests/job_tests/returnn/test_dump.py` currently fails: https://github.com/rwth-i6/i6_core/actions/runs/9643202359/job/26592670818?pr=517
```
...
File "/home/runner/work/i6_core/i6_core/returnn/returnn/__main__.py", line 218, in init_data
    line: train_data, extra_train = load_data(config, train_cache_bytes, "train")
    locals:
      train_data = <global> None
      extra_train = <not found>
      load_data = <global> <function load_data at 0x7fa3660c3af0>
      config = <global> <returnn.config.Config object at 0x7fa366322c40>
      train_cache_bytes = <local> [107](https://github.com/rwth-i6/i6_core/actions/runs/9643202359/job/26592670818?pr=517#step:7:108)37418241
  File "/home/runner/work/i6_core/i6_core/returnn/returnn/__main__.py", line 190, in load_data
    line: data = init_dataset(kwargs)
    locals:
      data = <not found>
      init_dataset = <global> <function init_dataset at 0x7fa3663020d0>
      kwargs = <local> {'name': 'train', 'class': 'DummyDataset', 'input_dim': 3, 'output_dim': 4, 'num_seqs': 2, 'chunking_variance': 0}, len = 6
  File "/home/runner/work/i6_core/i6_core/returnn/returnn/datasets/basic.py", line 1467, in init_dataset
    line: kwargs, cached_files = cache.handle_cached_files_in_config(kwargs)
    locals:
      kwargs = <local> {'name': 'train', 'input_dim': 3, 'output_dim': 4, 'num_seqs': 2, 'chunking_variance': 0}
      cached_files = <not found>
      cache = <local> <returnn.util.file_cache.FileCache object at 0x7fa365cad040>
      cache.handle_cached_files_in_config = <local> <bound method FileCache.handle_cached_files_in_config of <returnn.util.file_cache.FileCache object at 0x7fa365cad040>>
  File "/home/runner/work/i6_core/i6_core/returnn/returnn/util/file_cache.py", line 273, in FileCache.handle_cached_files_in_config
    line: import tree
    locals:
      tree = <not found>
ModuleNotFoundError: No module named 'tree'
```

`tree` is needed now because of https://github.com/rwth-i6/returnn/pull/1546. (I did not really think about this while merging this PR. But maybe it's ok?)

### DIFF
--- a/.github/workflows/job_tests.yml
+++ b/.github/workflows/job_tests.yml
@@ -35,6 +35,7 @@ jobs:
           pip install pytest
           sudo apt update && sudo apt install -y libsndfile1
           pip install -r i6_core/requirements.txt
+          pip install -r returnn/requirements.txt
       - name: Test Jobs
         run: |
           set -e
--- a/requirements.txt
+++ b/requirements.txt
@@ -3,4 +3,4 @@ h5py
 matplotlib
 numpy
 soundfile
-
+dm-tree
--- a/requirements.txt
+++ b/requirements.txt
@@ -3,4 +3,4 @@ h5py
 matplotlib
 numpy
 soundfile
-dm-tree
+


NeoLegends:
We can make RETURNN caching dependent on having `tree` installed. If a user uses `CachedFile` but does not have `tree`, training setup will fail eventually, because of the paths wrapped in the `CachedFile` dataclass, so there isn't a big worry here I assume? If `tree` is not installed, we log a warning and continue normally, hoping there are no cached files anywhere in the config.

michelwi:
`tree` is not per so a requirement of i6_core but of returnn. And there it is correctly listed under [requirements.txt](https://github.com/rwth-i6/returnn/blob/3c5999824ea3cca6e3bd3766d23680c9ae3fcaea/requirements.txt#L5).

I would propose to not add tree to the requirements of i6_core but instead update the test case(s) that use returnn to also install the requirements of returnn.

albertz:
> We can make RETURNN caching dependent on having `tree` installed.

`tree` is already needed to just iterate through the config:
```python
def handle_cached_files_in_config(self, config: Any) -> Tuple[Any, List[str]]:
    """
    :param config: some config, e.g. dict, or any nested structure
    :return: modified config, where all :class:`CachedFile` instances are replaced by the cached file path,
        and the list of cached files which are used.
    """
    import tree

    res_files = []

    def _handle_value(value):
        if isinstance(value, CachedFile):
            res = self.get_file(value.filename)
            res_files.append(res)
            return res
        return value

    return tree.map_structure(_handle_value, config), res_files
```
So when we do not have `tree`, we cannot easily iterate through the dict (or other nested structures).

NeoLegends:
> So when we do not have tree, we cannot easily iterate through the dict (or other nested structures).

that‘s my point. If there is no tree, we don‘t iterate the config and hope there are no files to be cached in there.

albertz:
> `tree` is not per so a requirement of i6_core but of returnn. And there it is correctly listed under [requirements.txt](https://github.com/rwth-i6/returnn/blob/3c5999824ea3cca6e3bd3766d23680c9ae3fcaea/requirements.txt#L5).

Well even though it is listed there, it was optional, and I think many people don't have it installed, and the RETURNN PR https://github.com/rwth-i6/returnn/pull/1546 now probably breaks a lot of existing setups.

albertz:
> that‘s my point. If there is no tree, we don‘t iterate the config and hope there are no files to be cached in there.

No, definitely not. We should never make the behavior dependent on what is installed. That can cause very weird behavior.

michelwi (changes requested):
> > `tree` is not per so a requirement of i6_core but of returnn. And there it is correctly listed under [requirements.txt](https://github.com/rwth-i6/returnn/blob/3c5999824ea3cca6e3bd3766d23680c9ae3fcaea/requirements.txt#L5).
> 
> Well even though it is listed there, it was optional, and I think many people don't have it installed, and the RETURNN PR [rwth-i6/returnn#1546](https://github.com/rwth-i6/returnn/pull/1546) now probably breaks a lot of existing setups.

that is true and might be an issue. But for _this_ PR I would still conceptually not change i6_core/requirements.txt but add after https://github.com/rwth-i6/i6_core/blob/d1c5bc80e9d264abb76150dad52c8bc7df3e2dd4/.github/workflows/job_tests.yml#L37
another one 
```
pip install -r returnn/requirements.txt 
```

michelwi (approved):

NeoLegends (approved):

PR #520, opened 2024-06-25
Update ComputePerplexityJob with additional optional outputs

JJorgeDSIC:
Update the `ComputePerplexityJob` to get the new PPL combinations from RASR's log, these are:

- PPL (as before)
- PPL without unknowns
- PPL without considering sentence-end token.
- PPL without unknowns and without considering sentence-end token.
- Other stats such as num unknowns and unknown ratio.

### DIFF
--- a/lm/perplexity.py
+++ b/lm/perplexity.py
@@ -65,19 +65,31 @@ class ComputePerplexityJob(rasr.RasrCommand, Job):
             root = ET.parse(f)
         ppl = root.find(".//perplexity")
         self.perplexity.set(float(ppl.text))
+
+        #Retrocompatibility checks
         ppl_wo_eos = root.find(".//perplexity-without-eos")
-        self.perplexity_without_eos.set(float(ppl_wo_eos.text))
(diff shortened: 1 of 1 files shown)


lm/perplexity.py
```diff
@@ -33,7 +33,13 @@ def __init__(
 
         self.log_file = self.log_file_output_path("compute_ppl", crp, False)
         self.score_file = self.output_path("word.scores")
-        self.perplexity = self.output_var("perplexity")
+        self.num_tokens = self.output_var("num-tokens")
```
michelwi, line 36: Once the Job finished, all output variables should be set. 

In the code below there can be the case, that some of these variables remain unset. I propose either
* add a fallback value here 
* or add else cases below that set some sentinel values

(e.g. 0 or inf, whatever reasonable)

michelwi (changes requested):

michelwi (approved):

Icemole (approved):

PR #521, opened 2024-06-28
docstrings for AlignSplitAccumulateSequence

### LINKED ISSUE #354: Extra Requirements in GMM Trainings
After looking at some crashed jobs I found some lets say unusual values for requirements like GB in my GMM jobs (10, 7, 5). 
Short investigation found that the parameters `"*_extra_rqmt"` do not replace the requirement, but rather add the requirements to the basic rqmt (c.f. `udpate_rqmt` in mm_sequence.py).
After talking to @JackTemaki we conlcuded that this is somewhat surprising and usually done differently, where handin rqmt  to a job in the construction would replace the rqmt not add it. 

In general this is no big deal, but for me the values in the baseline rather look like intended replacement values and not added values I just wanted to bring this up, because we might be able to save resources in the future with reducing some of them.

JackTemaki:
Had this started at some point, fixes #354

### DIFF
--- a/meta/mm_sequence.py
+++ b/meta/mm_sequence.py
@@ -51,26 +51,33 @@ class AlignSplitAccumulateSequence:
             - "align"
 
             An action can be written as e.g. "align!" to indicate that this alignment output
-            should be marked as output, meaning it will stored in `self.selected_alignment`
+            should be marked as output, meaning it will be stored in `self.selected_alignment`
             and get the keep_value for "selected" (or the Sisyphus default if not defined)
-        :param FlowNetwork feature_flow:
-        :param initial_mixtures:
-        :param initial_alignment:
-        :param str parallelization:
-        :param dict seq_extra_args:
-        :param dict align_extra_args:
-        :param dict split_extra_args:
-        :param dict accumulate_extra_args:
-        :param dict[str] align_extra_rqmt:
-        :param dict[str] split_extra_rqmt:
-        :param dict[str] accumulate_extra_rqmt:
+        :param FlowNetwork feature_flow: flow network for feature extraction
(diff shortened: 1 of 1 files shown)


meta/mm_sequence.py
```diff
@@ -51,26 +51,33 @@ def __init__(
...
-        :param dict accumulate_extra_args:
-        :param dict[str] align_extra_rqmt:
-        :param dict[str] split_extra_rqmt:
-        :param dict[str] accumulate_extra_rqmt:
+        :param FlowNetwork feature_flow: flow network for feature extraction
+        :param initial_mixtures: initialize the sequence with given mixtures.
+            Can be None only if initial_alignment is given.
+        :param initial_alignment: initialize the sequence with given alignment.
+            can be None only if initial_mixtures are given.
+        :param str parallelization: Unused
```
michelwi, line 61: lol.. you are right.. then we can remove the assert below?

meta/mm_sequence.py
```diff
@@ -51,26 +51,33 @@ def __init__(
...
-        :param dict split_extra_args:
-        :param dict accumulate_extra_args:
-        :param dict[str] align_extra_rqmt:
-        :param dict[str] split_extra_rqmt:
-        :param dict[str] accumulate_extra_rqmt:
+        :param FlowNetwork feature_flow: flow network for feature extraction
+        :param initial_mixtures: initialize the sequence with given mixtures.
+            Can be None only if initial_alignment is given.
+        :param initial_alignment: initialize the sequence with given alignment.
+            can be None only if initial_mixtures are given.
```
michelwi, line 60: should we care about proper typing?

michelwi (approved):
Thanks for better documentation.

christophmluscher (approved):

PR #522, opened 2024-07-03
ExtractSeqLensJob

albertz:


returnn/dataset.py
```diff
@@ -149,3 +149,110 @@ def run(self):
                 hdf_writer.insert_batch(numpy.asarray([[speaker_index]], dtype="int32"), [1], [segment_name])
 
         hdf_writer.close()
+
+
+class ExtractSeqLensJob(Job):
+    """
+    Extracts sequence lengths from a dataset for one specific key.
+
+    Also see: :class:`i6_experiments.users.schmitt.corpus.statistics.GetSeqLenFileJob`
```
JackTemaki, line 158: I am not sure I understand why this is referenced here, this job is a local user job which can be changed or removed at any point, and it works quite different.

returnn/dataset.py
```diff
@@ -149,3 +149,110 @@ def run(self):
...
+            # so merge dataset and post_dataset now.
+            dataset_dict.update(self.post_dataset)
+        config.config["dataset"] = dataset_dict
+        config.write(self.out_returnn_config_file.get_path())
+
+    def run(self):
+        """run"""
+        import tempfile
+        import shutil
+        from returnn.config import set_global_config, Config
```
JackTemaki, line 223: For this there should be an option to specify Returnn explicitly. See: https://github.com/rwth-i6/i6_core/blob/main/lib/hdf.py#L28

returnn/dataset.py
```diff
@@ -149,3 +149,110 @@ def run(self):
...
+        *,
+        key: str,
+        format: str,
+        returnn_config: Optional[ReturnnConfig] = None,
+    ):
+        """
+        :param dataset: dict for :func:`returnn.datasets.init_dataset`
+        :param post_dataset: extension of the dataset dict, which is not hashed
+        :param key: e.g. "data", "classes" or whatever the dataset provides
+        :param format: "py" or "txt"
```
JackTemaki, line 175: maybe output_format is a more descriptive name? At least I did not understand what this is about without looking at the actual code.

Also maybe a better explanation what py or txt means here would be helpful, but this is optional. I guess most people that want to use this job know what it looks like from working with other jobs. I was just surprised that for txt it does store the seq name, while for py it does.

JackTemaki (changes requested):

returnn/dataset.py
albertz: I had this in my own local copy. I just forgot to remove this.

albertz, line 158:
```suggestion
```

returnn/dataset.py
```diff
@@ -149,3 +149,117 @@ def run(self):
...
+        self.out_file = self.output_path(f"seq_lens.{output_format}")
+
+        self.rqmt = {"gpu": 0, "cpu": 1, "mem": 4, "time": 1}
+
+    @classmethod
+    def hash(cls, parsed_args):
+        """hash"""
+        parsed_args = parsed_args.copy()
+        parsed_args.pop("post_dataset")
+        if not parsed_args["returnn_config"]:
```
Icemole, line 200: Maybe add this to `Job.__sis_hash_exclude__` for more clarity?

returnn/dataset.py
```diff
@@ -149,3 +149,117 @@ def run(self):
...
+        config = Config()
+        config.load_file(self.out_returnn_config_file.get_path())
+        set_global_config(config)
+
+        dataset_dict = config.typed_value("dataset")
+        assert isinstance(dataset_dict, dict)
+        dataset = init_dataset(dataset_dict)
+        dataset.init_seq_order(epoch=1)
+
+        with tempfile.NamedTemporaryFile("w") as tmp_file:
```
Icemole, line 242: Just wondering, is there any issue with directly writing to the specific file?

Icemole, line 242: Maybe we can give the user the option to gzip this, in some occasions it might grow too large.

returnn/dataset.py
albertz: I don't think that's really necessary here. In my opinion, it doesn't really make it more clear. It's anyway not strictly necessary that I do this here at all. I just found it a bit cleaner.

So, if people don't like this here, I can also just remove it.

returnn/dataset.py
albertz: Yes. It is more stress on the fileserver where you have the `output` dir, which is likely on NFS. It will be much more syscalls, smaller writes to it. Instead, the copy in the end are only very few syscalls.

albertz: Yea sure, but is this necessary for this PR now?

returnn/dataset.py
JackTemaki: @Icemole hash_exclude is only if you add new parameters to an existing job, but never for parameters you do not want to hash at all.

JackTemaki (approved):

returnn/dataset.py
JackTemaki: It would be a nice feature for larger datasets, but I would be okay with adding that later.

returnn/dataset.py
michelwi: > hash_exclude is only if you add new parameters to an existing job, but never for parameters you do not want to hash at all.

That is true, but programmatically these two lines are equivalent to
```
__sis_hash_exclude__ = {"returnn_config": None}
```
(i.e. it is only not hashed, if it evaluates to False - which is exactly if it is `None`)

Which then raises the question why we have it at all. Hashing `None` here would be perfectly fine.

JackTemaki: Oh yes thats right, it should be removed completely as it has no effect at all.

returnn/dataset.py
```diff
@@ -149,3 +149,117 @@ def run(self):
...
+
+        self.rqmt = {"gpu": 0, "cpu": 1, "mem": 4, "time": 1}
+
+    @classmethod
+    def hash(cls, parsed_args):
+        """hash"""
+        parsed_args = parsed_args.copy()
+        parsed_args.pop("post_dataset")
+        if not parsed_args["returnn_config"]:
+            parsed_args.pop("returnn_config")
```
albertz, line 201:
```suggestion
```

michelwi (approved):

returnn/dataset.py
Icemole: Thanks for the feedback! It makes sense.

PR #523, opened 2024-07-03
Add ComputeTSEJob

SimBe195:
Adds a new `ComputeTSEJob` that can compute the TSE value of some alignment compared to a reference. The outputs of the job are the overall TSE value as well as detailed shift-counts of word-start and word-end frames as dicts and as histogram plots.

The job allows alignments with different amounts of subsampling and it's also possible to pass a function that specifies a seq-tag transformation to match up the alignment seq-tags with the reference seq-tags.


mm/alignment.py
```diff
@@ -413,3 +417,246 @@ def run(self):
                         total_score += float(e.find("./score/total").text)
 
                 f.write("%f\n" % (total_score / total_frames))
+
+
+class ComputeTSEJob(Job):
+    """
+    Compute TSE of some alignment compared to a reference
```
Marvin84, line 424: I would underline that this is a word-level TSE and underline how this is computed in case of HMM topology with silence and blank-based models.
Maybe with the reference to our HMM vs CTC paper?

SimBe195: Yes, thanks, I updated the docstring and made the description more detailed.

JackTemaki (commented):
Looks good from the technical side, as I cannot comment on the algorithmic side, I will pass the responsibility of a second approval to someone else :)

Icemole (changes requested):

SimBe195:
Thanks @Icemole, I incorporated your suggestions now.

mm/alignment.py
```diff
@@ -413,3 +417,261 @@ def run(self):
...
+            ), f"Found different number of word starts ({len(hyp_word_starts)}) than word ends ({len(hyp_word_ends)}) in reference. Something seems to be broken."
+
+            if len(hyp_word_starts) != len(ref_word_starts):
+                logging.warning(
+                    f"Sequence {hyp_seq_tag} ({idx} / {len(file_list)}:\n    Discarded because the number of words in alignment ({len(hyp_word_starts)}) does not equal the number of words in reference ({len(ref_word_starts)})."
+                )
+                discarded_seqs += 1
+                continue
+
+            # Sometimes different feature extraction or subsampling may produce mismatched lengths that are different by one frame, so cut off at the shorter length
```
Icemole, line 581: I just noticed that the comment explicitly states "one frame". Should we check that here? So adding `if hyp_seq_len - 1 <= ref_seq_len <= hyp_seq_len + 1` do what's below. I'm not sure how you considered this originally, maybe this check isn't needed.

mm/alignment.py
```diff
@@ -413,3 +417,261 @@ def run(self):
...
+            shorter_seq_length = min(hyp_seq_length, ref_seq_length)
+            hyp_word_ends[-1] = min(hyp_word_ends[-1], shorter_seq_length)
+            ref_word_ends[-1] = min(ref_word_ends[-1], shorter_seq_length)
+
+            assert all(start < shorter_seq_length for start in hyp_word_starts)
```
Icemole, line 586: Would this be equivalent to the four assertions?
```suggestion
            assert hyp_word_ends[-1] <= hyp_word_starts[-1] and hyp_word_ends[-1] < hyp_word_ends[-2]
            assert ref_word_ends[-1] <= ref_word_starts[-1] and ref_word_ends[-1] < ref_word_ends[-2]
```
In principle you're only changing the last item, so as long as the last item still has a relative order with respect to the second-to-last item (and is still logically correct with respect to the word start, in the sense that `start < end`), you're guaranteed that the relative order holds for every other item in the list.

> you're guaranteed that the relative order holds for every other item in the list

That should be what `_compute_word_boundaries` already guarantees: it iterates in an orderly manner for the sequence of allophones, and only inserts in an orderly manner the timestamps in `word_starts`, `word_ends` such that the ordering is guaranteed inside each list.

mm/alignment.py
SimBe195: Thinking about this, it can actually be more than one frame once the subsampling factor comes into play. For example if one sequence has length 14 with subsampling 1 then a model with subsampling 4 would give an alignment of length 3 or 4 depending on floor/ceil division which would be upsampled to 12 or 16. So I wouldn't check a specific maximum frame difference.

mm/alignment.py
SimBe195: I modified the cut-off logic a bit using the logic with iterating from the back of the lists, now everything should be consistent and the assertions are not needed anymore.

PR #524, opened 2024-07-04
Add round_factor to BlissToPcmHDFJob

michelwi:
I had the following problem:

`RoundingScheme.rasr_compatible` calculates the same start and end samples as rasr would for a given sampling rate. But when we change the sampling rate of the audio (e.g. upsample 8kHz to 16kHz), while keeping the sampling rate for the alignment, then the number of samples could again be off again. 

e.g. 
```
 8kHz:  80120 Samples -> 1000 Frames
16kHz: 160241 Samples -> 1001 Frames
```

To compensate for this I introduce a `round_factor` that indicates to round in a way so that the effective sampling rate was reduced by this factor.

This does make the code harder to read and probably not many actual dound_factors will be used, so alternatively I could just add a new RoundingScheme "rasr_compatible_upsample_2" which would then not interfere with the existing code but be less generic. What do you thingk?

### DIFF
--- a/returnn/hdf.py
+++ b/returnn/hdf.py
@@ -225,7 +225,11 @@ class BlissToPcmHDFJob(Job):
         start_and_duration = auto()
         rasr_compatible = auto()
 
-    __sis_hash_exclude__ = {"multi_channel_strategy": BaseStrategy(), "rounding": RoundingScheme.start_and_duration}
+    __sis_hash_exclude__ = {
+        "multi_channel_strategy": BaseStrategy(),
+        "rounding": RoundingScheme.start_and_duration,
+        "round_factor": 1.0,
+    }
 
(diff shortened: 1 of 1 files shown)


curufinwe:
Did you check if your logic is correct also for non-integer rounding factors? I.e. is the use of double math.ceil OK in that regime too? (For integer rounding factor it does nothing).

michelwi:
> Did you check if your logic is correct also for non-integer rounding factors? I.e. is the use of double math.ceil OK in that regime too? (For integer rounding factor it does nothing).

Not really. Maybe I just get rid of the extra floor/ceil and only allow integers.

curufinwe (approved):

christophmluscher (approved):

PR #525, opened 2024-07-08
Adds dataset parameter to class:`DiscountNgramsJob`

christophmluscher:
Adds a parameter to the class:`DiscountNgramsJob`. This allows the discounting to be optimized for a specific dataset.

### DIFF
--- a/lm/srilm.py
+++ b/lm/srilm.py
@@ -109,7 +109,7 @@ class DiscountNgramsJob(Job):
         count_exe: tk.Path,
         *,
         vocab: Optional[tk.Path] = None,
-        data_for_opt: Optional[tk.Path] = None,
+        data_for_optimization: Optional[tk.Path] = None,
         extra_discount_args: Optional[List[str]] = None,
         use_modified_srilm: bool = False,
         cpu_rqmt: int = 1,
@@ -121,7 +121,7 @@ class DiscountNgramsJob(Job):
         :param counts: file with the ngram counts, see :class:`CountNgramsJob.out_counts`.
         :param count_exe: path to the binary.
         :param vocab: vocabulary file for the discounting.
-        :param data_for_opt: the discounting will be optimized on this dataset.
+        :param data_for_optimization: the discounting will be optimized on this dataset.
         :param extra_discount_args: additional arguments for the discounting step.
         :param use_modified_srilm: Use the i6 modified SRILM version by Sundermeyer.
                                    The SRILM binary ngram-count was modified.
@@ -133,7 +133,7 @@ class DiscountNgramsJob(Job):
         self.ngram_order = ngram_order
         self.counts = counts
         self.vocab = vocab
-        self.data_for_opt = data_for_opt
+        self.data_for_opt = data_for_optimization
         self.discount_args = extra_discount_args or []
         self.use_modified_srilm = use_modified_srilm
 
--- a/lm/srilm.py
+++ b/lm/srilm.py
@@ -133,7 +133,7 @@ class DiscountNgramsJob(Job):
         self.ngram_order = ngram_order
         self.counts = counts
         self.vocab = vocab
-        self.data_for_opt = data_for_optimization
+        self.data_for_optimization= data_for_optimization
         self.discount_args = extra_discount_args or []
         self.use_modified_srilm = use_modified_srilm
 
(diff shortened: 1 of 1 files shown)


lm/srilm.py
```diff
@@ -100,13 +100,16 @@ class DiscountNgramsJob(Job):
...
+    __sis_hash_exclude__ = {"data_for_optimization": None}
+
     def __init__(
         self,
         ngram_order: int,
         counts: tk.Path,
         count_exe: tk.Path,
         *,
         vocab: Optional[tk.Path] = None,
+        data_for_opt: Optional[tk.Path] = None,
```
michelwi, line 112: In the optimal case the key of `__sis_hash_exclude__` would match the newly added parameter.

I would prefer "optimization" over "opt"

lm/srilm.py
```diff
@@ -410,7 +415,7 @@ def get_ppl(self):
                     if ln == "sentences,":
                         self.out_num_sentences.set(int(line[idx - 1]))
                     if ln == "words,":
-                        self.out_num_words.set(int(line[idx - 1]))
+                        self.out_num_words.set(float(line[idx - 1]))
```
michelwi, line 418: why would there ever be a non-integer number of words?

michelwi (changes requested):

lm/srilm.py
christophmluscher: yes good point ^^

lm/srilm.py
christophmluscher: I now have large corpora where the count switches to a scientific representation within the ppl log.

curufinwe (approved):

lm/srilm.py
michelwi: But 
```
>>> int(1e6)
1000000
````
that is fine.. or is the number larger than the maximum int?

Comment (christophmluscher) [reply lm/srilm.py:418]: yes it was quite large and the job crashed complaining about the int conversion.. I cant find the crashed job anymore.. would probably need some digging..

Comment (michelwi) [reply lm/srilm.py:418]: There is no problem with large ints, but int can indeed not parse strings with scientific notation
```
>>> int(2**1024-1)
179769313486231590772930519078902473361797697894230657273430081157732675805500963132708477322407536021120113879871393357658789768814416622492847430639474124377767893424865485276302219601246094119453082952085005768838150682342462881473913110540827237163350510684586298239947245938479716304835356329624224137215
>>> int(1e6)
1000000
>>> int("1e6")
  File "<stdin>", line 1, in <module>
ValueError: invalid literal for int() with base 10: '1e6'
>>> float("1e6")
1000000.0
>>> int(float("1e6"))
1000000
>>> int(eval("1e6"))
1000000
```
maybe we could do
```suggestion
                        self.out_num_words.set(int(float(line[idx - 1])))
```
instead?

lm/srilm.py
```diff
@@ -129,6 +133,7 @@ def __init__(
         self.ngram_order = ngram_order
         self.counts = counts
         self.vocab = vocab
+        self.data_for_opt = data_for_optimization
```
michelwi, line 136:
```suggestion
        self.data_for_optimization= data_for_optimization
```
just for consistent naming

lm/srilm.py
```diff
@@ -153,10 +158,10 @@ def create_files(self):
...
         ]
         if self.vocab is not None:
-            cmd += [
-                f"  -vocab {self.vocab.get_cached_path()} \\\n",
-            ]
-        cmd += ["  -kn discounts\\\n"] if not self.use_modified_srilm else [f"  -multi-kn-file discounts\\\n"]
+            cmd.append(f"  -vocab {self.vocab.get_cached_path()} \\\n")
+        cmd += ["  -kn discounts\\\n"] if not self.use_modified_srilm else [f"  -multi-kn-file discounts \\\n"]
+        if self.data_for_opt is not None:
+            cmd.append(f"  -optimize-discounts {self.data_for_opt.get_cached_path()} \\\n")
```
michelwi, line 164:
```suggestion
        if self.data_for_optimization is not None:
            cmd.append(f"  -optimize-discounts {self.data_for_optimization.get_cached_path()} \\\n")
```

michelwi (approved):

michelwi:
Bitbucket currently has [an incident](https://bitbucket.status.atlassian.com/incidents/qmrwh87h86sj), lets try rerunning the test once this is resolved.

PR #526, opened 2024-07-08
Add number of labels calculator functions to DenseLabelInfo

NeoLegends:
To reduce repetition.

### DIFF
--- a/mm/context_label.py
+++ b/mm/context_label.py
@@ -24,6 +24,35 @@ class DenseLabelInfo:
     use_boundary_classes: bool
     num_hmm_states_per_phon: int
 
+    def num_center_labels(self) -> int:
+        """
+        :return: number of center labels
+        """
+        return self.n_contexts * self.num_hmm_states_per_phon * self._factor()
+
+    def num_diphone_labels(self) -> int:
+        """
+        :return: number of diphone labels
+        """
+        return self.num_center_labels() * self.n_contexts
+
+    def num_triphone_labels(self) -> int:
+        """
+        :return: number of triphone labels
+        """
+        return self.num_center_labels() * (self.n_contexts**2)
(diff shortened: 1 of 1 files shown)


Marvin84 (approved):

mm/context_label.py
```diff
@@ -24,6 +24,35 @@ class DenseLabelInfo:
...
+        return self.num_center_labels() * (self.n_contexts**2)
+
+    def _factor(self) -> int:
+        """
+        :return: num labels factor introduced by word-end/word-boundary augmentation
+        """
+        if self.use_word_end_classes:
+            return 2
+        elif self.use_boundary_classes:
+            return 4
```
michelwi, line 52: Is this correct? e.g. if I use `use_boundary_classes=True` does this mean I have to set `use_word_end_classes = False`?
How does this interact with the implementation in `get_target_labels_from_dense`, where both can be set independently?
 If yes, then I would expand on that in the docstring and maybe even add asserts against wrong configuration or make this an enum instead. Otherwise maybe 
```suggestion
        if self.use_boundary_classes:
            return 4
        elif self.use_word_end_classes:
            return 2
```
would be better? I think boundary_classes implies word_end_classes.

Marvin84: > use_boundary_classes=True does this mean I have to set use_word_end_classes = False

Yes, these are mutually exclusive. 
boundary_class: distinction between within word, end of word, start of word and none, for total of 4 classes
word_end_class: start of word together within word and none are now in one class, and end of word in another class, for total of 2 classes

> asserts against wrong configuration

Yes, good point.
There is an assertion in `rasr` for this, also at recipe level we always make sure they are not both set to True

michelwi (approved):

PR #527, opened 2024-07-09
Fix plot() task in FairseqHydraTrainingJob

AndreasPlt:
Currently, the task `plot()` in the `FairseqHydraTrainingJob` fails  (as noted by @vieting off-line) due to an incorrect parsing of the hydra log file. I therefore corrected the parsing of this file to read the loss and LR values from the right lines.

### DIFF
--- a/fairseq/training.py
+++ b/fairseq/training.py
@@ -343,8 +343,8 @@ class FairseqHydraTrainingJob(Job):
                     i = 0
                     while i < len(lines):
                         line = lines[i]
-                        if "begin validation on" in line or "end of epoch" in line:
-                            epoch_dict = eval(lines[i + 1][lines[i + 1].index("{") :])
+                        if "[train][INFO]" in line or "[valid][INFO]" in line:
+                            epoch_dict = eval(line[line.index("{") :])
                             try:
                                 epoch = int(epoch_dict["epoch"])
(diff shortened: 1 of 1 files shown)


vieting:
@Judyxujj did you also have failing plot tasks? It does fail for the logs of wav2vec 2.0 pre-training. However, I'm not familiar enough with different fairseq setups to know whether this proposed change will break the task for the logs of other trainings.

vieting:
@michelwi can you tell why the AppTek hashes test fails? It can hardly be related to the proposed changes here, right?

michelwi:
> @michelwi can you tell why the AppTek hashes test fails? It can hardly be related to the proposed changes here, right?

This is a fork of i6_core and not a branch, therefore it does not have access to our API key to trigger the pipeline and this step immediately fails.

vieting:
> This is a fork of i6_core and not a branch, therefore it does not have access to our API key to trigger the pipeline and this step immediately fails.

I see, thanks. Then we can just ignore this, changes inside the plot task can anyway never affect the hashes.

vieting:
@Judyxujj could you please have a look at this PR? :)

Judyxujj (approved):

vieting (approved):

PR #528, opened 2024-07-18
Add `PlotAlignmentsJob`

Icemole:
~~The clip isn't beneficial when one wants to see the plot as a whole.~~

Edit: the description above is obsolete. The new PR description is: added job to visualize the alignment scores.


michelwi:
But without the clip, the plots can be illegible.. e.g. if all relevant data is between 0 and 10 and there is one outlier at 10^5, then one would not see anything in the resulting plot.

Icemole:
... which is what just happened to me. Great. I'm not sure what to do here. Clip by `3 * stddev`? 99th percentile? Allow user customization of the cut? The point is that I would like to see most of the plot, but the 90th percentile isn't enough for my needs.

Icemole:
Actually clipping by `median + 3 * stddev` would be better than clipping by `mean + 3 * stddev`.

Icemole:
I'm thinking of adding a `PlotAlignmentsJob` that allows for more customizability than a task in the `AlignmentJob`.

Icemole:
Should we remove this parameter from the hashes? I don't want to modify the parameter and then having to rerun the whole job. Is there any way to only have to run a given task after a parameter change?

michelwi:
> Should we remove this parameter from the hashes? 

it changes the outputs of the job.. 🤷 

> I don't want to modify the parameter and then having to rerun the whole job.

yes, but you have to, if you want the parameter to take effect

> Is there any way to only have to run a given task after a parameter change?

I don't think so, the hash is on job-level

Icemole:
I wouldn't like having to align everything every time I want to slightly modify the resulting plot. I think the best course of action involves adding a new job `PlotAlignmentsJob` which would be highly customizable. I'll try to do this.

sarahberanek:
In my opinion one should always look at all data - probably in log - to see in outliers and then zoom into to the region of interest. So I would never think that one plot is enough.

mm/alignment.py
```diff
@@ -391,6 +391,97 @@ def hash(cls, kwargs):
...
+        yield Task("plot", resume="plot", rqmt=self.rqmt)
+
+    def plot(self):
+        import numpy as np
+        import matplotlib
+        import matplotlib.pyplot as plt
+
+        # Parse the files and search for the average alignment score values (normalized over time).
+        alignment_scores = []
+        for log_file in self.out_log_file.values():
```
michelwi, line 453: this should go over `alignment_log_file` (a single file, so no loop even required). But even better, make the parameter to this job a `List[tk.Path]` and then loop over the list.

mm/alignment.py
```diff
@@ -391,6 +391,97 @@ def hash(cls, kwargs):
...
+        logging.info("STATS:")
+        logging.info(f"Max: {np_alignment_scores.max()}")
+        logging.info(f"Min: {min_value}")
+        logging.info(f"Median {np.median(np_alignment_scores)}")
+        logging.info(f"95-th percentile: {p95_value}")
+        logging.info(f"Total number of segments: {np_alignment_scores.size}")
+
+        # Plot the data.
+        matplotlib.use("Agg")
+        clip_low = self.clip_low or self.clip_percentile_high or min_value
```
michelwi, line 475: If at all should be `self.clip_percentile_low` but both do not exist..

mm/alignment.py
```diff
@@ -391,6 +391,97 @@ def hash(cls, kwargs):
...
+            By default clips at the 95th percentile (for removing huge outliers).
+        :param num_bins: Number of histogram bins. By default `50`.
+        """
+        self.alignment_log_file = alignment_log_file
+        clip_low_dict = {"clip_low": clip_low, "clip_percentile_low": clip_percentile_low}
+        clip_low_diff_none_values = [lower_val for lower_val in clip_low_dict.values() if lower_val is not None]
+        assert len(clip_low_diff_none_values) <= 1, (
+            f"More than one value for the low clip explicitly given (different from None): {clip_low_dict.items()}. "
+            "Please only give one value."
+        )
```
michelwi, line 430: That sounds like `clip_low is None != clip_percentile_low is None` in very complicated.

But should we even disallow both? we could just take whatever is lower/higher (i mean more restrictive) and go with it.

mm/alignment.py
```diff
@@ -391,6 +391,97 @@ def hash(cls, kwargs):
...
+            document = ET.parse(util.uopen(file_path))
+            _seg_list = document.findall(".//segment")
+            for seg in _seg_list:
+                avg = seg.find(".//score/avg")
+                alignment_scores.append(float(avg.text))
+            del document
+
+        np_alignment_scores = np.asarray(alignment_scores)
+        min_value = np_alignment_scores.min()
+        p95_value = np.percentile(np_alignment_scores, 95)
```
michelwi, line 465: why p95? for symmetry reasons we should take max

Icemole: Just for outlier check. But you're right, if the user wants to remove outliers they should add `clip_percentile_high=95`.

mm/alignment.py
Icemole: Done, thanks for the suggestion.

mm/alignment.py
Icemole: They do now exist, thanks for the heads-up.

mm/alignment.py
Icemole: Done, thanks for the suggestion.

mm/alignment.py
```diff
@@ -551,9 +659,10 @@ def run(self) -> None:
                 self.hyp_silence_phone,
                 self.hyp_upsample_factor,
             )
-            assert len(hyp_word_starts) == len(
-                hyp_word_ends
-            ), f"Found different number of word starts ({len(hyp_word_starts)}) than word ends ({len(hyp_word_ends)}). Something seems to be broken."
+            assert len(hyp_word_starts) == len(hyp_word_ends), (
```
michelwi, line 662: Please take care that your ide does not reformat lines that do not belong to the PR.

e.g. set "format only changed lines" or select black from environment as your formatting provider as opposed to black that ships with the extension or ruff.

mm/alignment.py
```diff
@@ -391,6 +391,114 @@ def hash(cls, kwargs):
...
+            clip_value_low = np.percentile(self.clip_percentile_low or 0)
+            clip_low = self.clip_low if self.clip_low > clip_value_low else clip_value_low
+        else:
+            clip_low = self.clip_low or self.clip_percentile_low or min_value
+        if self.clip_high is not None and self.clip_percentile_high is not None:
+            # Plot the most restrictive (lowest).
+            clip_value_high = np.percentile(self.clip_percentile_high or 0)
+            clip_high = self.clip_high if self.clip_high < clip_value_high else clip_value_high
+        else:
+            clip_high = self.clip_high or self.clip_percentile_high or max_value
```
michelwi, line 491:
```suggestion
        clip_low = max(self.clip_low, np.percentile(self.clip_percentile_low), min_value)
        clip_high = min(self.clip_high, np.percentile(self.clip_percentile_high), max_value)
```
the most restrictive shall prevail

mm/alignment.py
```diff
@@ -391,6 +391,114 @@ def hash(cls, kwargs):
...
+    and :func:`i6_core.corpus.filter.FilterSegmentsByAlignmentConfidenceJob.plot`.
+    """
+
+    def __init__(
+        self,
+        alignment_log_files: List[tk.Path],
+        clip_low: Optional[int] = None,
+        clip_high: Optional[int] = None,
+        clip_percentile_low: Optional[int] = None,
+        clip_percentile_high: Optional[int] = None,
```
michelwi, line 407:
```suggestion
        clip_low: float = float("-inf"),
        clip_high: float = float("inf"),
        clip_percentile_low: float = 0.0,
        clip_percentile_high: float = 100.0,
```
* float over int
* we can already put non-restricting default values

mm/alignment.py
```diff
@@ -391,6 +391,114 @@ def hash(cls, kwargs):
...
+        :param zoom_x_max: Maximum X value in which to zoom in on the plot. If `None`, won't zoom in.
+        :param zoom_y_min: Minimum Y value in which to zoom in on the plot. If `None`, won't zoom in.
+        :param zoom_y_max: Maximum Y value in which to zoom in on the plot. If `None`, won't zoom in.
+        :param num_bins: Number of histogram bins. By default `50`.
+        """
+        self.alignment_log_files = alignment_log_files
+        self.clip_low = clip_low
+        self.clip_high = clip_high
+        self.clip_percentile_low = clip_percentile_low
+        self.clip_percentile_high = clip_percentile_high
```
michelwi, line 438:
```suggestion
        assert clip_low < clip_high
        self.clip_percentile_low = clip_percentile_low
        assert 0.0 <= clip_percentile_low <= 100.0, "percentile should be between 0 and 100"
        self.clip_percentile_high = clip_percentile_high
        assert 0.0 <= clip_percentile_high <= 100.0, "percentile should be between 0 and 100"
        assert clip_percentile_low < clip_percentile_high
```
add some reasonable assertions

mm/alignment.py
```diff
@@ -391,6 +391,114 @@ def hash(cls, kwargs):
...
+        matplotlib.use("Agg")
+        if self.clip_low is not None and self.clip_percentile_low is not None:
+            # Plot the most restrictive (highest).
+            clip_value_low = np.percentile(self.clip_percentile_low or 0)
+            clip_low = self.clip_low if self.clip_low > clip_value_low else clip_value_low
+        else:
+            clip_low = self.clip_low or self.clip_percentile_low or min_value
+        if self.clip_high is not None and self.clip_percentile_high is not None:
+            # Plot the most restrictive (lowest).
+            clip_value_high = np.percentile(self.clip_percentile_high or 0)
```
michelwi, line 488: `or 0` is actually wrong for percentile_high (so request changes), but see my other suggestions.

michelwi (changes requested):

mm/alignment.py
Icemole: Thanks for the suggestion.

Icemole:
>>[Icemole](https://github.com/Icemole) requested a review from [michelwi](https://github.com/michelwi) https://github.com/rwth-i6/i6_core/pull/528#event-13742469432
>
>From my last review there is still open to revert the two changes in the existing code. This is probably also why the black step in the pipeline fails. Otherwise lgtm.

I changed it because I saw it was a line 150 characters long, which is over the 120 character threshold. Let me know if you still want me to revert it to its original state.

michelwi (approved):
No, if this change makes the line length complicit and does not break the tests, then we can keep it. :)

mm/alignment.py
```diff
@@ -391,6 +391,108 @@ def hash(cls, kwargs):
...
+        zoom_x_max: Optional[float] = None,
+        zoom_y_min: Optional[float] = None,
+        zoom_y_max: Optional[float] = None,
+        num_bins: int = 50,
+    ):
+        """
+        :param alignment_log_files: Alignment log files from the alignment job.
+        :param clip_low: Number symbolizing the absolute number at which the plot will be clipped to the left.
+            If given along with any other value restricting the minimum, the most restrictive value will prevail.
+            By default clips at the minimum value. Has priority over :param:`clip_percentile_low`.
```
hannah220, line 418: `clip_low = max(self.clip_low, np.percentile(self.clip_percentile_low), min_value)`
this looks like neither `clip_low` nor `clip_percentile_low` have priority

Icemole: Done. I think it's now better expressed. Thanks for the heads up!

hannah220 (approved):

PR #529, opened 2024-07-22
Also infer dynamic axes for every model input and output

NeoLegends:
I think this could be a fix for https://github.com/rwth-i6/i6_core/pull/510#issuecomment-2242807757 and https://bitbucket.org/omnifluent/apptek_asr/pull-requests/1173/overview.

### DIFF
--- a/returnn/compile.py
+++ b/returnn/compile.py
@@ -275,12 +275,14 @@ class TorchOnnxExportJob(Job):
         # Pass the tensor names here because ReturnnConfig serialization might
         # potentially reorder via `sort_config=True`, and the order matters for
         # output tensor <-> tensor name association.
-        input_names = self.input_names or self.collect_tensor_names(
-            str(self.returnn_root), self.returnn_config.config.get("extern_data", {})
-        )
-        output_names = self.output_names or self.collect_tensor_names(
-            str(self.returnn_root), self.returnn_config.config.get("model_outputs", {})
-        )
+        input_names = self.input_names
+        if input_names is None and "extern_data" in self.returnn_config.config:
+            input_names = self.collect_tensor_names(str(self.returnn_root), self.returnn_config.config["extern_data"])
+        output_names = self.output_names
+        if output_names is None and "model_outputs" in self.returnn_config.config:
+            output_names = self.collect_tensor_names(
+                str(self.returnn_root), self.returnn_config.config["model_outputs"]
+            )
         if input_names:
             cmd += ["--input_names", ",".join(input_names)]
         if output_names:
--- a/returnn/compile.py
+++ b/returnn/compile.py
@@ -277,11 +277,15 @@ class TorchOnnxExportJob(Job):
         # output tensor <-> tensor name association.
         input_names = self.input_names
         if input_names is None and "extern_data" in self.returnn_config.config:
-            input_names = self.collect_tensor_names(str(self.returnn_root), self.returnn_config.config["extern_data"])
+            input_names = self.collect_tensor_names(
+                str(self.returnn_root),
+                self.returnn_config.config["extern_data"],
+            )
         output_names = self.output_names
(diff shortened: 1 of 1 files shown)


albertz (changes requested):
I think you should not hardcode `...:size1` here. You can just check the entry in `extern_data`/`model_outputs`, it will tell you what dynamic dims are there.

NeoLegends:
How so? This is not within RETURNN (where this info might be available), and there are cases where there's simply `extern_data = {"data": {"dim": 60}}` and no more info.

albertz:
> there are cases where there's simply `extern_data = {"data": {"dim": 60}}`

It applies `extern_data_template_from_config_opts`. This will call `Tensor(**opts)`. I.e. for your example, `Tensor(dim=60)`. It follows the defaults from `Tensor`.

returnn/compile.py
```diff
@@ -281,10 +271,44 @@ def run(self):
...
+    def collect_tensor_names(returnn_root: str, data_dict: Dict[str, Any]) -> List[str]:
+        if returnn_root not in sys.path:
+            sys.path.append(returnn_root)
+
+        try:
+            from returnn.tensor import Tensor
+        except ImportError:
+            # For backwards compatibility assume dynamic time axis (i.e. axis 1 in
+            # BTF order) if we cannot import the Tensor class.
+            return [name for k in data_dict.keys() for name in [k, f"{k}:size1"]]
```
albertz, line 302: I don't think the behavior of this function should depend on some import error.
```suggestion
        from returnn.tensor import Tensor
```

returnn/compile.py
```diff
@@ -281,10 +271,44 @@ def run(self):
...
+        names = []
+        for name, opts in data_dict.items():
+            names.append(name)
+            tensor = Tensor(name=name, **opts)
+            for i, dim in enumerate(tensor.shape):
+                # `None` means the dim is dynamic
+                if dim is None:
+                    # i+1 because the batch dim is implicit in RETURNN and for
+                    # historical reasons not part of `Tensor.dims`
+                    names.append(f"{name}:size{i + 1}")
```
albertz, line 313:
```suggestion
            for i, dim in enumerate(tensor.dims):
                # We need seq lengths if there is a dyn size which is not a scalar.
                if dim.dyn_size_ext and self.dyn_size_ext.dims:
                    names.append(f"{name}:size{i}")
```

returnn/compile.py
```diff
@@ -281,10 +271,37 @@ def run(self):
...
+        input_names = self.input_names or self.collect_tensor_names(
+            str(self.returnn_root), self.returnn_config.config.get("extern_data", {})
+        )
+        output_names = self.output_names or self.collect_tensor_names(
+            str(self.returnn_root), self.returnn_config.config.get("model_outputs", {})
+        )
+        if input_names:
+            cmd += ["--input_names", ",".join(input_names)]
+        if output_names:
+            cmd += ["--output_names", ",".join(output_names)]
```
albertz, line 287: It's not exactly the same logic as before.
```suggestion
        input_names = self.input_names
        if input_names is None and "extern_data" in self.returnn_config.config:
            input_names = self.collect_tensor_names(
                str(self.returnn_root), self.returnn_config.config["extern_data"]
            )
        output_names = self.output_names
        if output_names is None and "model_outputs" in self.returnn_config.config:
            output_names = self.collect_tensor_names(
                str(self.returnn_root), self.returnn_config.config["model_outputs"]
            )
        if input_names is not None:
            cmd += ["--input_names", ",".join(input_names)]
        if output_names is not None:
            cmd += ["--output_names", ",".join(output_names)]
```

NeoLegends: Ah, yes. If the arrays are empty but passed to the job constructor the behavior differs. Will fix.

albertz (approved):

michelwi (approved):

PR #531, opened 2024-08-02
Add `GetOovRateJob`

Icemole:
Adds the `GetOovRateJob`, which computes the OOV rate of a text file with respect to a given vocabulary.

This is motivated by the lack of jobs that perform this basic operation, and by the complexity of manually having to tinker with the outputs of already existing tools in order to get this number.

If there was a job that already did this, please let me know.

### DIFF
--- a/lm/vocabulary.py
+++ b/lm/vocabulary.py
@@ -3,6 +3,7 @@ __all__ = [
     "LmIndexVocabularyFromLexiconJob",
     "VocabularyFromLmJob",
     "VocabularyFromTextJob",
+    "GetOovRateJob",
 ]
 from sisyphus import Job, Task, tk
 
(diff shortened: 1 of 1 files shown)


lm/vocabulary.py
```diff
@@ -195,3 +196,38 @@ def run(self):
...
+        self.rqmt = {"cpu": 1, "mem": 5.0, "time": 2.0}
+
+    def tasks(self):
+        yield Task("run", resume="run", rqmt=self.rqmt)
+
+    def run(self):
+        vocab = set()
+        with uopen(self.vocab_file.get_path(), "rt") as in_vocab:
+            for word in in_vocab:
+                vocab.add(word.strip())
```
NeoLegends, line 221:
```suggestion
        with uopen(self.vocab_file.get_path(), "rt") as in_vocab:
            vocab = set(word.strip() for word in in_vocab)
```

NeoLegends (approved):
Nice! Small nit, feel free to ignore, nothing major.

lm/vocabulary.py
```diff
@@ -195,3 +196,38 @@ def run(self):
...
+    def __init__(self, text_file: tk.Path, vocab_file: tk.Path):
+        self.text_file = text_file
+        self.vocab_file = vocab_file
+
+        self.out_oov_rate = self.output_var("oov_rate")
+
+        self.rqmt = {"cpu": 1, "mem": 5.0, "time": 2.0}
+
+    def tasks(self):
+        yield Task("run", resume="run", rqmt=self.rqmt)
```
NeoLegends, line 215: Oh, I do feel like this should be a `mini_task`?

michelwi:
> If there was a job that already did this, please let me know.

is `apptek_asr.text.stats.ExtractOovWordsFromTextJob` not doing the same and more?

Icemole:
>> If there was a job that already did this, please let me know.
>
> is apptek_asr.text.stats.ExtractOovWordsFromTextJob not doing the same and more?

Yes, I was told about this after the PR was created :)

I think it's still nice to have this on i6_core for everyone to use, but feel free to close/disregard.

lm/vocabulary.py
Icemole: There are potentially many words to be loaded in the vocabulary (I'm not sure 5 GB are enough for 250-500k vocab size), and the job would take a relatively long time with any medium/large-sized text file. I interpret a mini task to be a task with minimal computation, like writing a small script to execute later.

Because of this, I'd tend to leave this as a cluster task and not as a mini task. If you still have issues with this, let me know and we can discuss this further.

Icemole: That makes sense, for some reason my gut feeling was that it would have been much higher. I'll lower the amount to 1-2 GB. I still feel like it's not suitable to be a mini task, even with the small memory footprint.

NeoLegends: Have you tried out running one of these tasks w/ a 500k vocab?

Icemole:
Closing, see comment above.

PR #536, opened 2024-08-16
Extends ShuffleAndSplitJob: allows gzip files as input and output

christophmluscher:
Extends the functionality of the `ShuffleAndSlpitJob` to process gzipped file input and output a gzipped file allowing for a wider use of the job.

### DIFF
--- a/corpus/segments.py
+++ b/corpus/segments.py
@@ -176,7 +176,7 @@ class ShuffleAndSplitSegmentsJob(Job):
         self.shuffle_seed = shuffle_seed
 
         self.out_segments = {
-            k: self.output_path("%s.segments" % k if not zip_output else "%s.segments.gz" % k)
+            k: self.output_path(f"{k}.segments{'.gz' if zip_output else ''}")
             for k in self.split.keys()
         }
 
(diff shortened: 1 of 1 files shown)


NeoLegends (approved):

NeoLegends:
Nit: I think we often call the init flag `compressed` rather than `zip_output`.

corpus/segments.py
```diff
@@ -172,13 +175,16 @@ def __init__(self, segment_file, split=None, shuffle=True, shuffle_seed=0x3C5EA3
         self.shuffle = shuffle
         self.shuffle_seed = shuffle_seed
 
-        self.out_segments = {k: self.output_path("%s.segments" % k) for k in self.split.keys()}
+        self.out_segments = {
+            k: self.output_path("%s.segments" % k if not zip_output else "%s.segments.gz" % k)
```
michelwi, line 179:
```suggestion
            k: self.output_path(f"{k}.segments{'.gz' if zip_output else ''}")
```
shorter + use f-string ?

michelwi (approved):
just one optional comment for style.. otherwise lgtm

christophmluscher:
> Nit: I think we often call the init flag `compressed` rather than `zip_output`.

sadly I found both..

PR #538, opened 2024-08-18
Fix dumped HDF permissions

michelwi:
The permission number is interpreted bitwise and needs to be given in octal.

Decimal 644 is interpreted as 1204, which is entirely not desired.

Kudos to @Max-Ryujin for spotting the problem and testing the fix.


JackTemaki (approved):
Sorry, I missed this one somehow.

Atticus1806 (approved):

PR #540, opened 2024-09-03
Strip newlines when parsing `TextDict`s to avoid `OverflowError`

### LINKED ISSUE #539: Parsing a large TextDict fails on recent python versions
E.g. for LS960 you get:

```
OverflowError                             
Traceback (most recent call last)
File /usr/local/lib/python3.10/dist-packages/sisyphus/task.py:188, in Task.run(self=<Task 'run' job=Job<alias/datasets/LibriSpeech/t...ext/convert/TextDictToTextLinesJob.xMTMuHiJ4xBa>>, task_id=1, resume_job=False, logging_thread=<LoggingThread(Thread-2, started daemon 140001375073856)>)
    186             logging.info("Starting subtask for arg id: %d args: %s" % (arg_id, str(args)))
    187             logging.info("-" * 60)
--> 188             f(*args)
        f = <bound method TextDictToTextLinesJob.run of Job<alias/datasets/LibriSpeech/train_other_960_corpus_text_lines work/i6_core/text/convert/TextDictToTextLinesJob.xMTMuHiJ4xBa>>
        args = []
    189 except sp.CalledProcessError as e:
    190     if e.returncode == 137:
    191         # TODO move this into engine class

File recipe/i6_core/text/convert.py:33, in TextDictToTextLinesJob.run(self=Job<alias/datasets/LibriSpeech/train_other_960_c...text/convert/TextDictToTextLinesJob.xMTMuHiJ4xBa>)
     30 def run(self):
     31     # nan/inf should not be needed, but avoids errors at this point and will print an error below,
     32     # that we don't expect an N-best list here.
---> 33     d = eval(uopen(self.text_dict, "rt").read(), {"nan": float("nan"), "inf": float("inf")})
        {"nan": float("nan"), "inf": float("inf")} = {'nan': nan, 'inf': inf}
        float("nan") = nan
        float("inf") = inf
        self = Job<alias/datasets/LibriSpeech/train_other_960_corpus_text_lines work/i6_core/text/convert/TextDictToTextLinesJob.xMTMuHiJ4xBa>
        self.text_dict = <Path work/i6_core/corpus/convert/CorpusToTextDictJob.JIQTGMdLEmbz/output/text_dictionary.py.gz>
     34     assert isinstance(d, dict)  # seq_tag -> text
     36     with uopen(self.out_text_lines, "wt") as out:

OverflowError: line number table is too long
```

Working on a fix.

NeoLegends:
Closes #539

### DIFF
--- a/mm/alignment.py
+++ b/mm/alignment.py
@@ -1,4 +1,10 @@
-__all__ = ["AlignmentJob", "DumpAlignmentJob", "AMScoresFromAlignmentLogJob", "ComputeTimeStampErrorJob"]
+__all__ = [
+    "AlignmentJob",
+    "DumpAlignmentJob",
+    "PlotAlignmentJob",
+    "AMScoresFromAlignmentLogJob",
+    "ComputeTimeStampErrorJob",
+]
 
 import logging
 import math
@@ -477,16 +483,16 @@ class PlotAlignmentJob(Job):
 
         # Plot the data.
         matplotlib.use("Agg")
-        clip_low = max(self.clip_low, np.percentile(self.clip_percentile_low), min_value)
-        clip_high = min(self.clip_high, np.percentile(self.clip_percentile_high), max_value)
+        clip_low = max(self.clip_low, np.percentile(np_alignment_scores, self.clip_percentile_low), min_value)
+        clip_high = min(self.clip_high, np.percentile(np_alignment_scores, self.clip_percentile_high), max_value)
         np.clip(np_alignment_scores, clip_low, clip_high, out=np_alignment_scores)
         plt.hist(np_alignment_scores, bins=self.num_bins)
         plt.xlabel("Average Maximum-Likelihood Score")
         plt.ylabel("Number of Segments")
         plt.title("Histogram of Alignment Scores")
         plt.gca().set_xlim(left=self.zoom_x_min, right=self.zoom_x_max)
-        plt.gca().set_ylim(left=self.zoom_y_min, right=self.zoom_y_max)
-        plt.savefig(fname=self.out_plot_avg.get_path())
+        plt.gca().set_ylim(bottom=self.zoom_y_min, top=self.zoom_y_max)
+        plt.savefig(fname=self.out_plot.get_path())
 
 
 class AMScoresFromAlignmentLogJob(Job):
--- a/returnn/search.py
+++ b/returnn/search.py
@@ -446,7 +446,12 @@ class SearchWordsToCTMJob(Job):
         corpus.load(self.bliss_corpus.get_path())
         d = util.parse_text_dict(self.recog_words_file)
         with util.uopen(self.out_ctm_file.get_path(), "wt") as out:
-            out.write(";; <name> <track> <start> <duration> <word> <confidence> [<n-best>]\n")
+            # Do not print optional [n-best] header, some downstream evaluation pipelines
+            # use the number of headers for validation. Since we do not print n-best-list
+            # information this validation fails and discards the entire search outputs.
+            #
+            # See https://github.com/rwth-i6/i6_core/pull/542.
+            out.write(";; <name> <track> <start> <duration> <word> <confidence>\n")
             for seg in corpus.segments():
                 seg_start = 0.0 if seg.start == float("inf") else seg.start
                 seg_end = 0.0 if seg.end == float("inf") else seg.end
@@ -535,7 +540,12 @@ class SearchWordsDummyTimesToCTMJob(Job):
         else:
             seq_order = d.keys()
         with util.uopen(self.out_ctm_file.get_path(), "wt") as out:
-            out.write(";; <name> <track> <start> <duration> <word> <confidence> [<n-best>]\n")
+            # Do not print optional [n-best] header, some downstream evaluation pipelines
+            # use the number of headers for validation. Since we do not print n-best-list
+            # information this validation fails and discards the entire search outputs.
+            #
+            # See https://github.com/rwth-i6/i6_core/pull/542.
+            out.write(";; <name> <track> <start> <duration> <word> <confidence>\n")
             for seg_fullname in seq_order:
                 assert isinstance(
                     seg_fullname, str


util.py
```diff
@@ -383,3 +383,24 @@ def update_nested_dict(dict1: Dict[str, Any], dict2: Dict[str, Any]):
         else:
             dict1[k] = v
     return dict1
+
+
+def parse_text_dict(path: Union[str, tk.Path]) -> Dict[str, str]:
+    """
+    Loads the text dict at :param:`path` making sure not to trigger line counter overflow.
```
albertz, line 390: Can you expand this explanation?
* Put a ref to the issue here: https://github.com/rwth-i6/i6_core/issues/539
* Also quote the exception you got: `OverflowError: line number table is too long`
* Explain the exception. What does it mean? Why does this function fix it? I don't really understand what the exception means. My guess would be that the dict becomes too big. But then this function here would have the same problem, because as I understand it, you intend to create just the same dict, just in a different way?

util.py
```diff
@@ -383,3 +383,24 @@ def update_nested_dict(dict1: Dict[str, Any], dict2: Dict[str, Any]):
...
+    """
+
+    with uopen(path, "rt") as text_dict_file:
+        txt = text_dict_file.read()
```
albertz, line 394: I wonder a bit: as I understand it, your problem is that the dict is really huge (how huge? how many entries? how big is the file?). But then, isn't it a problem to load it all into memory? Also via the following code (`strip`, `splitlines`, etc), you are even creating multiple copies of the data in memory.

util.py
```diff
@@ -383,3 +383,24 @@ def update_nested_dict(dict1: Dict[str, Any], dict2: Dict[str, Any]):
...
+
+    # remove leading and trailing dict brackets
+    txt = txt.strip().strip("{}").strip()
```
albertz, line 397: This is too sloppy for me. You expect a very certain structure. Do that more explicitly, checking for that:
```suggestion
    # remove leading and trailing dict brackets
    txt = txt.strip()
    assert txt[:1] == "{" and txt[-1:] == "}"
    txt = txt[1:-1]
```

util.py
```diff
@@ -383,3 +383,24 @@ def update_nested_dict(dict1: Dict[str, Any], dict2: Dict[str, Any]):
...
+
+    lines = txt.splitlines()
+    result = {
+        k: v
+        # parse chunkwise to avoid line counter overflow when the text dict is very large
+        for chunk in chunks(lines, max(1, len(lines) // 1000))
+        for k, v in eval("\n".join(["{", *chunk, "}"]), {"nan": float("nan"), "inf": float("inf")}).items()
```
albertz, line 404: This code does not work in all cases. E.g. consider a dict formatted as this:
```python
{
"a":
42
}
```
(Yea, this probably will not happen when it was written with some of our tools, but who knows, maybe it comes from elsewhere, and/or there was some `black` on the generated output, or whatever.)
This should at least be discussed in a code comment.

albertz: Note, in RETURNN, we use `returnn.util.literal_py_to_pickle.literal_eval`, which solves this problem in another way, which should be faster and more generic.

util.py
NeoLegends: It is my understanding that the memory itself is not an issue, but rather that the file in my case is about 280k lines long, which is where python's line counter overflows for some reason. This is why I'm not worried about memory consumption itself in this module, but rather just the way the string is parsed.

util.py
NeoLegends: > This should at least be discussed in a code comment.

Yes, I'll add one.

> Note, in RETURNN, we use `returnn.util.literal_py_to_pickle.literal_eval`, which would solve this problem in another way, which should be faster and more generic.

That implementation contains a native C++ module, I don't think this should be made part of `i6_core`.

----

Why did we end up with an implementation based on Python literals anyway, and why haven't we used a proper data format (e.g. JSON) from the get go for this? All this custom code would never have had to be written and the issue here would likely have never occured in the first place with a proper data exchange format.

NeoLegends:
I've noticed you can also simply strip any newline from the text dict before parsing, this also avoids the error. Newer python versions than 3.10 also don't seem to run into this error anymore.

util.py
albertz: What is the Python line counter? There are no "lines" in a dict? Or you mean for Python source code?

But anyway, you should add this explanation to the code, so that you can understand why we have this code there.

util.py
albertz: > native C++ module

Well, you could argue, the Python stdlib is full of native code... I don't really see being "native" a problem here. The main problem is maybe that then RETURNN becomes a dependency for those specific jobs. That's maybe unwanted. But I'm not sure if it's really so much a problem. But then, maybe we can also use this code here for now and keep that in mind. (Maybe add it in a comment, that this is an alternative.)

> Why did we end up with an implementation based on Python literals anyway, and why haven't we used a proper data format (e.g. JSON) from the get go for this? 

My assumption was, it's faster and simpler. Those assumptions turned out to be wrong though, so yes, JSON would have been better.

Python literals are also more flexible and generic though, but in many cases (like here), this is not really needed, and/or you can work around that. But JSON is really restricted. I think it also does not support inf/nan, so it doesn't properly work for N-best lists (rarely, but sometimes we get inf in the score; it would be annoying if it then rarely crashes, and/or you need a stupid workaround just to clip it to 1e30 or so).

albertz:
> I've noticed you can also simply strip any newline from the text dict before parsing,

But this can break the code even more, and even silently? The other complaint I had before would at least fail with an error if the chunking is incorrect. This stripping can corrupt the data silently without error. I definitely would not do that.

Example:
```python
{
"a":
"""
a
b
c
"""
}
```

curufinwe (approved):

albertz (changes requested):
I still would somehow change this. As said, this can silently corrupt the data without you noticing it.

curufinwe:
But how should this be changed? I don't see a good method to fix this. The "correct" fix would be to change the output format to something like json that can be parsed independently of the number of lines, but that's a much bigger change. And yes, while the changes suggested by @NeoLegends might not work in every case, I would argue that the cases where it does not work should not really occur in the data the search outputs.

albertz:
Well, I already use those jobs whenever I have this format, not just when it comes from the old TF based RETURNN search code. E.g. my new PT-based RETURNN pipeline also uses this format, but there I have other custom code to write the data into this format.

Another way to fix this is to have custom parser code for this, which properly parses this. E.g. there is `ast.literal_eval`, and there is my native code for `literal_eval_to_pickle`, and there is also more out there, which can parse such subsets of Python.

But even when fixing this, this whole format is still problematic: When your text dicts are really so large, they will consume a lot of memory, as they are fully loaded into memory, and due to Python overhead for every object, and then also memory fragmentation, this can consume a lot of memory. For example, when I created 128 hypotheses per each sequence of some of the eval corpora from Librispeech (e.g. dev-other), that already consumed about 20 GB or so (I forgot the exact number), and 256 hypotheses was already too much, i.e. all such jobs were killed by the Linux OOM killer (most of those jobs were run via the local engine).

So, when you reach such sizes, I wonder if this is really the right format. With custom parser code, which would not load all into memory, you could still handle it, but using some other format maybe just makes more sense. E.g., whenever I create hypotheses on the Librispeech train corpora (as in #539), I just use the HDF format instead.

I would prefer not to add such a hack, with partly incorrect behavior, to avoid any potential problems in the future.

curufinwe:
Given that there is no activity here for the last 6 months I would suggest to close the PR. @NeoLegends any objections?

PR #542, opened 2024-09-11
Drop optional CTM n-best header

NeoLegends:
This optional header fields leads to issues in GLM mapping, resulting in high (almost 100 percent) deletion rates, because all lines with fewer than 7 columns are dropped from the file.

### DIFF
--- a/returnn/search.py
+++ b/returnn/search.py
@@ -449,7 +449,7 @@ class SearchWordsToCTMJob(Job):
         d = eval(util.uopen(self.recog_words_file.get_path(), "rt").read())
         assert isinstance(d, dict), "only search output file with dict format is supported"
         with util.uopen(self.out_ctm_file.get_path(), "wt") as out:
-            out.write(";; <name> <track> <start> <duration> <word> <confidence> [<n-best>]\n")
+            out.write(";; <name> <track> <start> <duration> <word> <confidence>\n")
             for seg in corpus.segments():
                 seg_start = 0.0 if seg.start == float("inf") else seg.start
                 seg_end = 0.0 if seg.end == float("inf") else seg.end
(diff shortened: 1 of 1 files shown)


michelwi (approved):
I am fine with dropping the field in the header, but the actual bug, I think, lies in our (AppTek) scoring pipeline and we are diskussing how to best fix it atm.

albertz:
Just to clarify: Can you explain why this was not an issue so far? Because we usually don't do GLM mapping?

returnn/search.py
```diff
@@ -449,7 +449,7 @@ def run(self):
         d = eval(util.uopen(self.recog_words_file.get_path(), "rt").read())
         assert isinstance(d, dict), "only search output file with dict format is supported"
         with util.uopen(self.out_ctm_file.get_path(), "wt") as out:
-            out.write(";; <name> <track> <start> <duration> <word> <confidence> [<n-best>]\n")
+            out.write(";; <name> <track> <start> <duration> <word> <confidence>\n")
```
albertz, line 452: Can you add a comment explaining that the number of columns matter here, and for what (maybe also giving a ref to the PR here)?

michelwi:
> Just to clarify: Can you explain why this was not an issue so far? Because we usually don't do GLM mapping?

The "bug" is - I think - in the AppTek scoring pipeline where we recently started to use the number of fields in the header to discard "invalid" lines in the CTM which do not have the correct number of fields.

So far this was not an issue because no one would use the ctm outputs of this job with the AppTek scoring pipeline and the AppTek asr pipeline would not print this optional field in the header.

albertz:
Ok, then I would mention exactly that in a code comment. But in any case, I think the code change is otherwise ok then.

albertz (approved):

PR #543, opened 2024-09-23
Add `f32` support when reading RASR caches

Icemole:
This is especially useful when reading caches that only have one value per element, for instance caches from `EnergyJob`s.

### DIFF
--- a/lib/rasr_cache.py
+++ b/lib/rasr_cache.py
@@ -928,7 +928,7 @@ class MixtureSet:
             self.densities[n, 1] = cov_idx
 
         self.num_mixtures = self.read_u32()
-        self.mixtures: List[Optional[Tuple[List[int], List[float]]]] = [None] * self.num_mixtures  # nopep8
+        self.mixtures: List[Optional[Tuple[List[int], List[float]]]] = [None] * self.num_mixtures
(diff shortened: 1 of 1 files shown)


albertz (approved):

NeoLegends (approved):

Icemole:
Since black was failing for some issue that I hadn't created (at least locally it said so), I also took the liberty to add better (and shorter) typing to the variables instead of having the typing in a long comment.

lib/rasr_cache.py
```diff
@@ -923,9 +928,7 @@ def __init__(self, filename):
             self.densities[n, 1] = cov_idx
 
         self.num_mixtures = self.read_u32()
-        self.mixtures = [
-            None
-        ] * self.num_mixtures  # type: typing.List[typing.Optional[typing.Tuple[typing.List[int],typing.List[float]]]]  # nopep8
+        self.mixtures: List[Optional[Tuple[List[int], List[float]]]] = [None] * self.num_mixtures  # nopep8
```
albertz, line 931:
```suggestion
        self.mixtures: List[Optional[Tuple[List[int], List[float]]]] = [None] * self.num_mixtures
```

michelwi (approved):

PR #544, opened 2024-09-24
Add V2 version of the Corpus class

Icemole:
The class `CorpusV2` allows `O(1)` access to any segment or recording.

It might be a naive implementation, but I think a concept similar to this should be enough for our needs of accessing any segment in `O(subcorpus_depth)`, which is practically `O(1)` in our cases.

It's a draft so any feedback and changes will be appreciated.

### DIFF
--- a/lib/corpus.py
+++ b/lib/corpus.py
@@ -356,28 +356,45 @@ class CorpusV2(Corpus):
 
     def get_segment_by_full_name(self, name: str) -> Optional[Segment]:
         """
-        :return: the segment specified by its full name
+        Obtains a segment from the corpus given its full name in the corpus.
+
+        :param name: The full name of the segment.
+        :return: The segment to be searched for, or `None` if not found.
+        """
+        split_segment_name = name.split("/")
+        potential_corpus_name = split_segment_name[0]
+        if self.name == potential_corpus_name:
+            # The name was the own corpus'. This can happen the first iteration when giving the full segment name,
+            # for instance 'base_corpus/subcorpus1/subcorpus2/recording/segment'.
+            # Get rid of the first part for the search.
+            name = split_segment_name[1:]
+
+        return self._get_segment_by_full_name(name)
+
+    def _get_segment_by_full_name(self, name: str) -> Optional[Segment]:
+        """
+        :param name: The name of the segment to be searched for, relative to the current corpus.
+        :return: The segment whose name coincides with :param:`name` relative to the current corpus,
+            or `None` if not found.
         """
         if name == "":
             # Found nothing.
             return None
 
-        if name in self.segments:
-            return self.segments[name]
+        # Base case: the recording name comes first, and the segment name appears immediately afterwards.
+        recording_name = name.split("/")[0]
+        segment_name = name[len(f"{recording_name}/") :]
+        if recording_name in self.recordings:
+            return self.recordings[recording_name][segment_name]
         else:
-            subcorpus_name = name.split("/")[0]
-            segment_name_from_subcorpus = name[len(f"{subcorpus_name}/") :]
-            if self.name == subcorpus_name:
-                # The name was the own corpus'. This can happen when giving the full segment name.
-                # Ignore the former part.
-                subcorpus_name = name.split("/")[0]
-                segment_name_from_subcorpus = name[len(f"{subcorpus_name}/") :]
-
+            # Recursive case: look one level deeper to the indicated subcorpus.
+            subcorpus_name = recording_name
+            segment_full_name_from_subcorpus = segment_name
             assert subcorpus_name in self.subcorpora, (
                 f"Subcorpus '{subcorpus_name}' required for accessing segment '{name}' "
                 "not found in the list of subcorpora: {list(self.subcorpora.keys())}."
             )
-            return self.subcorpora[subcorpus_name].get_segment_by_full_name(segment_name_from_subcorpus)
+            return self.subcorpora[subcorpus_name].get_segment_by_full_name(segment_full_name_from_subcorpus)
 
     def get_recording_by_full_name(self, name: str) -> Optional[RecordingV2]:
         """
--- a/lib/corpus.py
+++ b/lib/corpus.py
@@ -60,7 +60,7 @@ class CorpusParser(sax.handler.ContentHandler):
             assert isinstance(
                 e, (Corpus, CorpusV2)
             ), "<subcorpus> may only occur within a <corpus> or <subcorpus> element"
-            subcorpus = ()
+            subcorpus = type(e)()
             subcorpus.name = attrs["name"]
             subcorpus.parent_corpus = e
             if isinstance(e, Corpus):
--- a/lib/corpus.py
+++ b/lib/corpus.py
@@ -165,8 +165,8 @@ class Corpus(NamedEntity, CorpusSection):
         for r in self.recordings:
             yield from r.segments
 
-    for sc in self.subcorpora:
-        yield from sc.segments()
+        for sc in self.subcorpora:
+            yield from sc.segments()
 
     def get_segment_by_name(self, name: str) -> Segment:
         """
--- a/lib/corpus.py
+++ b/lib/corpus.py
@@ -331,12 +331,10 @@ class CorpusV2(Corpus):
         """
         :return: the segment specified by its name
         """
-        for seg in self.segments():
-            if seg.name == name:
-                return seg
-        assert False, f"Segment '{name}' was not found in corpus"
+        assert name in self.segments, f"Segment '{name}' was not found in corpus"
+        return self.segments[name]
 
-     def get_segment_by_full_name(self, name: str) -> Optional[Segment]:
+    def get_segment_by_full_name(self, name: str) -> Optional[Segment]:
         """
         :return: the segment specified by its full name
         """
@@ -349,9 +347,19 @@ class CorpusV2(Corpus):
         else:
             subcorpus_name = name.split("/")[0]
             segment_name_from_subcorpus = name[len(f"{subcorpus_name}/"):]
+            if self.name == subcorpus_name:
+                # The name was the own corpus'. This can happen when giving the full segment name.
+                # Ignore the former part.
+                subcorpus_name = name.split("/")[0]
+                segment_name_from_subcorpus = name[len(f"{subcorpus_name}/"):]
+
+            assert subcorpus_name in self.subcorpora, (
+                f"Subcorpus '{subcorpus_name}' required for accessing segment '{name}' "
+                "not found in the list of subcorpora: {list(self.subcorpora.keys())}."
+            )
             return self.subcorpora[subcorpus_name].get_segment_by_full_name(segment_name_from_subcorpus)
 
-    def get_recording_by_full_name(self, name: str) -> Optional[Segment]:
+    def get_recording_by_full_name(self, name: str) -> Optional[RecordingV2]:
         """
         :return: the segment specified by its full name
         """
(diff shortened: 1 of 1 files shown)


lib/corpus.py
```diff
@@ -163,7 +163,7 @@ def segments(self) -> Iterable[Segment]:
         """
         for r in self.recordings:
             yield from r.segments
-        for sc in self.subcorpora:
+    for sc in self.subcorpora:
```
michelwi:
```suggestion
        for sc in self.subcorpora:
```
this change would be a typo, right?

lib/corpus.py
```diff
@@ -301,6 +301,154 @@ def _dump_internal(self, out: TextIO, indentation: str = ""):
...
+    attribute is set. Corpora with include statements can be read but are written back as a single file.
+
+    The difference with respect to :class:`i6_core.lib.corpus.Corpus` is that in this class we can access
+    any recording or segment in practically O(1).
+    """
+
+    def __init__(self):
+        super().__init__()
+
+        self.parent_corpus: Optional[Corpus] = None
```
michelwi, line 318:
```suggestion
```
this already exists from the parent class

lib/corpus.py
```diff
@@ -301,6 +301,154 @@ def _dump_internal(self, out: TextIO, indentation: str = ""):
...
+
+        self.subcorpora: Dict[str, Corpus] = {}
+        self.recordings: Dict[str, RecordingV2] = {}
```
michelwi, line 321: how about we add the new Dict members under a different name and then add a property like
```
@property
def subcorpora(self):
    return list(self._subcorpora.values())

@subcorpora.setter
def subcorpora(self, values):
    for sc in values:
        self._subcorpora[sc.name] = sc
```
so that most previous code would also work with V2. Also many functions below would then not need to be re-implemented.

lib/corpus.py
```diff
@@ -301,6 +301,154 @@ def _dump_internal(self, out: TextIO, indentation: str = ""):
...
+        if name == "":
+            # Found nothing.
+            return None
+
+        if name in self.segments:
+            return self.segments[name]
+        else:
+            subcorpus_name = name.split("/")[0]
+            segment_name_from_subcorpus = name[len(f"{subcorpus_name}/"):]
+            return self.subcorpora[subcorpus_name].get_segment_by_full_name(segment_name_from_subcorpus)
```
michelwi: I do not like that the functions `get_segment_by_{full_,}name` differ in the efficiency of the implementation and also in the behavior if a segment is not found. 
Also the `get_segment_by_full_name` is recursive and then the `name` that is passed is not the full_name any more.

lib/corpus.py
```diff
@@ -301,6 +301,154 @@ def _dump_internal(self, out: TextIO, indentation: str = ""):
...
+        """
+        if name == "":
+            # Found nothing.
+            return None
+
+        if name in self.segments:
+            return self.segments[name]
+        else:
+            subcorpus_name = name.split("/")[0]
+            segment_name_from_subcorpus = name[len(f"{subcorpus_name}/"):]
```
michelwi: we can already determine the location of the segment within the subcorpora/recordings by `len(name.split("/"))` so we could indeed loop over the name and use e.g.
```
L = len(name.split('/'))
active_element = self
for i,n in enumerate(name.split('/')):
    if L-i > 2:
        active_element = active_element.subcorpora[n]
    elif L-i > 1:
        active_element = active_element.recordings[n]
    else:
        return  active_element.segments[n]
```
to directly navigate to the segment.

But a recursive implementation is also fine (an maybe cleaner == nicer) if done nicely :)

lib/corpus.py
```diff
@@ -301,6 +301,154 @@ def _dump_internal(self, out: TextIO, indentation: str = ""):
...
+
+     def get_segment_by_full_name(self, name: str) -> Optional[Segment]:
+        """
+        :return: the segment specified by its full name
+        """
+        if name == "":
+            # Found nothing.
+            return None
+
+        if name in self.segments:
```
michelwi, line 347: is `self.segments` not the function defined a few lines above? Be careful and consistent in the naming!

lib/corpus.py
```diff
@@ -301,6 +301,154 @@ def _dump_internal(self, out: TextIO, indentation: str = ""):
...
+    def get_recording_by_full_name(self, name: str) -> Optional[Segment]:
+        """
+        :return: the segment specified by its full name
+        """
+        if name == "":
+            # Found nothing.
+            return None
+
+        if name in self.segments:
+            return self.segments[name]
```
michelwi: get *recording* is the functions name

lib/corpus.py
```diff
@@ -338,6 +486,20 @@ def add_segment(self, segment: Segment):
         self.segments.append(segment)
 
 
+class RecordingV2(Recording):
```
michelwi, line 509: also the dump function has to be overridden because
```
        for s in self.segments:
            s.dump(out, indentation + "  ")
```
if `self.segments` is a dict now, then `s` will be the keys.

michelwi (changes requested):

lib/corpus.py
Icemole: Yes, already broke the pipeline twice and already fixed :)

lib/corpus.py
```diff
@@ -301,6 +303,172 @@ def _dump_internal(self, out: TextIO, indentation: str = ""):
             out.write("%s</subcorpus>\n" % (indentation,))
 
 
+class CorpusV2(Corpus):
```
albertz, line 306: Does it make sense to derive from `Corpus`?

lib/corpus.py
```diff
@@ -301,6 +303,172 @@ def _dump_internal(self, out: TextIO, indentation: str = ""):
...
+    any recording or segment in practically O(1).
+    """
+
+    def __init__(self):
+        super().__init__()
+
+        self.parent_corpus: Optional[Corpus] = None
+
+        self.subcorpora: Dict[str, Corpus] = {}
+        self.recordings: Dict[str, RecordingV2] = {}
```
albertz, line 321: You must not redefine the types of attribs in a subclass in an incompatible way. (The only thing which is allowed is to specify/restrict them more, e.g. changing from `Dict[str, Any]` to `Dict[str, int]` in a subclass is ok.)

albertz (changes requested):

lib/corpus.py
Icemole: Then I shall not inherit from `Corpus`.

lib/corpus.py
Icemole: If your comment below is valid, then I shall not inherit from `Corpus`.

lib/corpus.py
albertz: No, that's not necessarily true. You could also just rename them. And then maybe define `subcorpora`/`recordings` as property.

albertz: That's basically also what @michelwi said.

lib/corpus.py
Icemole: I like better the recursive implementation, but let me know if you prefer the iterative one.

lib/corpus.py
Icemole: In my view, we shouldn't restrain ourselves to using V1 code anymore, since it leads the user to keep using retrocompatible code instead of striving for optimal performance. If the user needs a `Corpus`, then they should build a `Corpus`, and should only build a `CorpusV2` if they know what they're doing.

If you still think that it would be a nice addition, please let me know, we can debate it.

Anyway, I expect that the `CorpusV2` class makes sense only in very specific scenarios, such as when having to access segments in O(1). Otherwise, the `Corpus` class still makes sense as a repository of subcorpora, recordings and segments, since the functionality in `V2` won't need to be used by many users.

lib/corpus.py
albertz: Iterative is always to be preferred for Python.

Icemole:
As I stated in a comment above, note that the CorpusV2 has some memory redundancy from the dictionary indexed by name and the object in memory having the name as well. So it's a bit of a tradeoff, more memory for less access time.

I still expect the `Corpus` class to be useful as a repository of subcorpora, recordings and segments, for users who don't want fast access to segments and want to, for instance, filter all segments according to a criterion or iterate over all recordings (which are pretty common use cases, at least from my side).

lib/corpus.py
michelwi: > In my view, we shouldn't restrain ourselves to using V1 code anymore

This would be another argument for not inheriting from Corpus. Maybe a Corpus-like implementation is not even necessary. As you said 

> I expect that the CorpusV2 class makes sense only in very specific scenarios,

Maybe for these scenarios a simple data structure or even just a 
```
segment_map = { s.fullname(): s for s in some_corpus.all_segments() }
```
would be sufficient to fulfill their performance needs.

Icemole: Yes, that makes a lot of sense. This implementation was originally thought in order to get the orthography of a given segment in O(1), which was the issue we were facing. @michelwi do you think returning a segment map from full names to segments would do the trick for your specific use case (I don't know all the details)?

@albertz and the rest of the reviewers, what's your take on this? Do you think the functionality here is too redundant? Our use case could be covered by a small function (or two, if you want to get recordings in O(1) as well) in the `Corpus` class.

Icemole:
Copy-pasting from a comment above:

> Maybe for these scenarios a simple data structure or even just a `segment_map = { s.fullname(): s for s in some_corpus.all_segments() }` would be sufficient to fulfill their performance needs.

Yes, that makes a lot of sense. This implementation was originally thought in order to get the orthography of a given segment in O(1), which was the issue we were facing. @michelwi do you think returning a segment map from full names to segments would do the trick for your specific use case (I don't know all the details)?

@albertz and the rest of the reviewers, what's your take on this? Do you think the functionality here is too redundant? Our use case could be covered by a small function (or two, if you want to get recordings in O(1) as well) in the Corpus class.

lib/corpus.py
albertz: About the memory redundancy, I don't think there is really any problem. It will only have duplicated some of the pointers. You probably won't even notice this.

I haven't really checked too much, but if you can just put your extension into the `Corpus` class itself, or rewrite `Corpus` such that it stays compatible, I think I would prefer that over having a `CorpusV2`.

lib/corpus.py
```diff
@@ -56,45 +57,63 @@ def startElement(self, name: str, attrs: Dict[str, str]):
             assert len(self.elements) == 1, "<corpus> may only occur as the root element"
             e.name = attrs["name"]
         elif name == "subcorpus":
-            assert isinstance(e, Corpus), "<subcorpus> may only occur within a <corpus> or <subcorpus> element"
-            subcorpus = Corpus()
+            assert isinstance(
+                e, (Corpus, CorpusV2)
```
albertz, line 61: I thought it's removed now? Are those changes here still up-to-date?

lib/corpus.py
```diff
@@ -300,6 +299,18 @@ def _dump_internal(self, out: TextIO, indentation: str = ""):
...
+        """
+        :return: Mapping from segment fullnames to actual segments.
+        """
+        return {seg.fullname(): seg for seg in self.segments()}
+
+    def get_recording_mapping(self) -> Dict[str, Recording]:
+        """
+        :return: Mapping from recording fullnames to actual recordings.
+        """
+        return {rec.fullname(): rec for rec in self.recordings}
```
michelwi, line 312:
```suggestion
        return {rec.fullname(): rec for rec in self.all_recordings()}
```
in case of (nested) sub-corpora the current implementation will only contain the top-level recordings while my proposed change would also descend into sub-corpora and get all recordings.

Icemole:
@albertz I cannot merge while you're requesting changes. Could you please review the PR whenever you have time? Thanks!

albertz (approved):

PR #546, opened 2024-10-07
Allow DelayedBase parameter in WriteToTextFileJob

BenediktConze:
Properly handle the content parameter of `WriteToTextFileJob` if it derives from `DelayedBase`.

### DIFF
--- a/text/processing.py
+++ b/text/processing.py
@@ -282,7 +282,7 @@ class WriteToTextFileJob(Job):
 
     def __init__(self, content, out_name: str = "file.txt"):
         """
-        :param list|dict|str content: input which will be written into a text file
+        :param list|dict|str|DelayedBase content: input which will be written into a text file
         :param out_name: user specific file name for the output file
         """
         self.content = content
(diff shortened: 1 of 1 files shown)


michelwi:
Instead of adding another case, could we use
```
content = instanciate_delayed(self.content)
```
and then list and dicts of DelayedBase are also immediately supported?

BenediktConze:
Good point, I changed it.

michelwi (approved):

text/processing.py
```diff
@@ -282,7 +282,7 @@ class WriteToTextFileJob(Job):
 
     def __init__(self, content, out_name: str = "file.txt"):
         """
-        :param list|dict|str content: input which will be written into a text file
+        :param list|dict|str|DelayedBase content: input which will be written into a text file
```
Atticus1806, line 285: Since you are touching the code, could you move the typing to the header (like for `out_name`)?

Atticus1806 (approved):

PR #548, opened 2024-10-11
Jobs for Search Errors

Marvin84:
These two jobs are implemented by @curufinwe for calculating search errors, once the `rasr` has [this commit](https://github.com/rwth-i6/rasr/commit/f0c58edf89bf4f4705acd743db0fd56562d0dce4).

The merging should be straight-forward.

### DIFF
--- a/recognition/advanced_tree_search.py
+++ b/recognition/advanced_tree_search.py
@@ -2,6 +2,7 @@ __all__ = [
     "AdvancedTreeSearchLmImageAndGlobalCacheJob",
     "AdvancedTreeSearchJob",
     "AdvancedTreeSearchWithRescoringJob",
+    "RescoreLatticeCacheJob",
     "BidirectionalAdvancedTreeSearchJob",
     "BuildGlobalCacheJob",
 ]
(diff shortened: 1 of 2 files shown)


recognition/advanced_tree_search.py
```diff
@@ -536,6 +537,151 @@ def create_config(
...
+        pruning_threshold=16.0,
+        history_limit=0,
+        rescoring_lookahead_scale=1.0,
+        rtf: float = 1.0,
+        cpu: int = 1,
+        mem: float = 4.0,
+        use_gpu: bool = False,
+        extra_config=None,
+        extra_post_config=None,
+    ):
```
michelwi, line 556: as typical in @curufinwes Jobs, a docstring is missing xP
type hints would also be appreciated

recognition/advanced_tree_search.py
```diff
@@ -536,6 +537,151 @@ def create_config(
...
+        self.set_vis_name("Rescore Lattice")
+
+        kwargs = locals()
+        del kwargs["self"]
+
+        (
+            self.config,
+            self.post_config,
+        ) = RescoreLatticeCacheJob.create_config(**kwargs)
+        self.input_lattice_cache = lattice_cache
```
michelwi, line 566: This Job assumes a single lattice cache as input but produces multiple caches as output.
Would the application then be to provide the bundle file for the parameter `lattice_cache`? 
And this is implemented because lattice processing would be done with fewer nodes than lattice generation?

recognition/scoring.py
```diff
@@ -664,3 +665,52 @@ def calc_wer(self):
         os.chdir(old_dir)
 
         return wer
+
+
+class IntersectStmCtm(Job):
```
Atticus1806, line 670: Could you please add a docstring what this job does?

Atticus1806 (changes requested):
I did not verify the RASR config, but else looks good (only typing/documentation missing)

recognition/advanced_tree_search.py
```diff
@@ -2,6 +2,7 @@
     "AdvancedTreeSearchLmImageAndGlobalCacheJob",
     "AdvancedTreeSearchJob",
     "AdvancedTreeSearchWithRescoringJob",
+    "RescoreLatticeCacheJob",
```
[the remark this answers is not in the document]
Marvin84: Might be :) I can change that

recognition/scoring.py
```diff
@@ -664,3 +665,52 @@ def calc_wer(self):
...
+                        ctm_lines[recording].extend(segment_comments)
+                        segment_comments = []
+                    ctm_lines[recording].append(line)
+
+        common_recordings = set(stm_lines.keys()).intersection(set(ctm_lines.keys()))
+
+        for path, lines in [(self.out_stm, stm_lines), (self.out_ctm, ctm_lines)]:
+            with uopen(path, "wt") as out:
+                out.write("".join(lines[None]))
+                del lines[None]
```
michelwi, line 714: why `del lines[None]`? I assume because `None` would be in the set `common_recordings` and in the for-loop below we would also iterate again over it and write the header-lines a second time.

But now we still do iterate over it and when `None` shows up, the defaultdict nature of `lines` prevents a crash and leads to `out.write("".join([])` which if fine I guess.
Maybe semantically more nice would be to remove `None` from the set of `common_recordings` e.g.
```
common_recordings = set(stm_lines.keys()).intersection(set(ctm_lines.keys())) - {None}
```

I leave this open.

recognition/scoring.py
```diff
@@ -664,3 +665,52 @@ def calc_wer(self):
         os.chdir(old_dir)
 
         return wer
+
+
+class IntersectStmCtm(Job):
+    def __init__(self, stm_path: tk.Path, ctm_path: tk.Path):
```
michelwi, line 671: doc str pls

michelwi:
> I did not verify the RASR config, but else looks good (only typing/documentation missing)

RASR config looks good to me. The job rescores a lattice with a different LM and applies additional pruning as far as I can see.

recognition/advanced_tree_search.py
```diff
@@ -880,3 +881,148 @@ def create_config(cls, crp, extra_config, extra_post_config, **kwargs):
...
+        self.input_lattice_cache = lattice_cache
+        self.exe = self.select_exe(crp.flf_tool_exe, "flf-tool")
+        self.concurrent = crp.concurrent
+        self.use_gpu = use_gpu
+
+        self.out_log_file = self.log_file_output_path("rescoring", crp, True)
+        self.out_single_lattice_caches = dict(
+            (task_id, self.output_path("lattice.cache.%d" % task_id, cached=True))
+            for task_id in range(1, crp.concurrent + 1)
+        )
```
NeoLegends, line 921: Nit: I'd prefer this to be 
```suggestion
        self.out_single_lattice_caches = {
            task_id: self.output_path("lattice.cache.%d" % task_id, cached=True)
            for task_id in range(1, crp.concurrent + 1)
        }
```

NurAd-Din:
Hey there! I am also interested in this job could we please get it merged soon? Thank you!

curufinwe (approved):

PR #549, opened 2024-10-16
ReturnnTrainingJob info, more robust LR log loading

albertz:
In my case, I had some np.float64(...) entries in it (for the effective LR).
Not sure how I got that (maybe due to new PyTorch/Numpy), and I already pushed some RETURNN change to avoid this in the future, but I think we can anyway just handle that here as well.

### DIFF
--- a/returnn/training.py
+++ b/returnn/training.py
@@ -287,6 +287,8 @@ class ReturnnTrainingJob(Job):
         return run_cmd
 
     def info(self):
+        import numpy as np
+
         def try_load_lr_log(file_path: str) -> Optional[dict]:
             # Used in parsing the learning rates
             @dataclass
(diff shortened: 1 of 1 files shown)


returnn/training.py
```diff
@@ -287,6 +287,8 @@ def _get_run_cmd(self):
         return run_cmd
 
     def info(self):
+        import numpy as np
```
michelwi, line 290: what was the reasoning behind doing a local import vs module level?

michelwi (approved):

returnn/training.py
albertz: * I like to keep things local to see where they are used. That's a thing you should do in general, for all kinds of things, about the structure of code (e.g. local variables, etc.).
* It's not good practice to put too much things into the global namespace.
* It could increase the module load time when `numpy` was not imported yet.

albertz: Note, the only (valid) argument to put the import at module level is that you see all the dependencies of the module more easily. But I'm not sure this is so important here.

JackTemaki: It comes very surprising to me to break with the convention now, but I do not care too much. 

>  the only

Definitely not the only in general, maybe the only one here. Once you use it more often you save lines of code, definitely a benefit.

JackTemaki (approved):

PR #550, opened 2024-10-18
use tobytes instead of tostring

Stefanwuu:
tostring in array.array has been replaced by tobytes and deprecated since Python 3.2 and is already removed in Python 3.9. I therefore propose to change the code to avoid this problem.

### DIFF
--- a/lib/rasr_cache.py
+++ b/lib/rasr_cache.py
@@ -453,7 +453,7 @@ class FileArchive:
             a = array.array("b")
             a.fromfile(self.f, comp)
             # unpack
-            b = zlib.decompress(a.tostring(), 15 + 32)
+            b = zlib.decompress(a.tobytes(), 15 + 32)
             # substitute self.f by an anonymous memmap file object
             # restore original file handle after we're done
             backup_f = self.f
@@ -808,7 +808,7 @@ class MixtureSet:
         """
         a = array.array("b")
         a.fromfile(self.f, l)
-        return a.tostring().decode(enc)
+        return a.tobytes().decode(enc)
 
     def read_f32(self):
         """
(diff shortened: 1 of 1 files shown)


lib/rasr_cache.py
```diff
@@ -808,7 +808,7 @@ def read_str(self, l, enc="ascii"):
         """
         a = array.array("b")
         a.fromfile(self.f, l)
-        return a.tostring().decode(enc)
+        return a.tobytes().decode(enc)
```
michelwi, line 811: based on the fact that here is a `.decode(enc)` following the `tostring()` call, I assume the result already was a byte-string and we do not need to change anything else?

michelwi (approved):
Failing test is because this PR comes from a fork rather than a branch and can be ignored since lib does not contain pipeline code.

lib/rasr_cache.py
Stefanwuu: The thing is tostring is removed for Python >=3.9, which means calling it directly results in an error. I assume since we are gradually shifting to higher Python versions and Python <3.2 is hardly used, it might make sense to change? But we can of course also just require people using lower Python version when calling the rasr_cache, not a problem for me.

Stefanwuu: Just for clarifying: tostring and tobytes are absolutely the same thing with just different names, see [here](https://docs.python.org/3/library/array.html#array.array.tobytes) I don't know why they did it

michelwi: > and we do not need to change anything else?

I meant "change additional code because instead of a str we now get a bytes object" (which is not the case). I am totally in favor of the change.

JackTemaki (approved):

PR #551, opened 2024-10-18
more consistent LR file reading

albertz:
Follow-up to #549. Also make it more consistent.

### DIFF
--- a/returnn/training.py
+++ b/returnn/training.py
@@ -17,6 +17,7 @@ import sys
 import os
 import shutil
 import subprocess as sp
+import numpy as np
 from typing import Dict, Sequence, Iterable, List, Optional, Union
 
 from sisyphus import *
@@ -287,8 +288,6 @@ class ReturnnTrainingJob(Job):
         return run_cmd
 
     def info(self):
-        import numpy as np
-
         def try_load_lr_log(file_path: str) -> Optional[dict]:
             # Used in parsing the learning rates
             @dataclass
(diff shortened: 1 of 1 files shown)


michelwi (approved):

JackTemaki:
Wouldn't it make sense to define this only once globally, instead of 3 times at different positions if it is about "consistency"?

albertz:
> Wouldn't it make sense to define this only once globally, instead of 3 times at different positions if it is about "consistency"?

`EpochData` is different. It would require some more rewriting of the code, which I wanted to avoid now.

returnn/training.py
```diff
@@ -298,7 +297,10 @@ class EpochData:
 
             try:
                 with open(file_path, "rt") as file:
-                    return eval(file.read().strip(), {"EpochData": EpochData, "np": np})
+                    return eval(
+                        file.read().strip(),
+                        {"EpochData": EpochData, "nan": float("nan"), "inf": float("inf"), "np": np},
```
NeoLegends, line 302: Is it ok to just have the `np` key here as this is the common name under which numpy is imported, or should we also add `"numpy": np`?

NeoLegends (approved):
Just one small comment, otherwise LGTM.

returnn/training.py
albertz: It's mandatory, because this is how Numpy `float64.__repr__` prints it:
```
In [1]: import numpy

In [2]: numpy.float64(1.23)
Out[2]: np.float64(1.23)
```

NeoLegends: Ahh, I see. Then the other variant also won‘t be necessary.

albertz: (Btw, e.g. `numpy.array(1.23)` has just the repr `array(1.23)`, so maybe adding `array` might also make sense, but not here, or not now, as we never have arrays in the LR file.)

PR #554, opened 2024-11-14
Fix FairseqHydraPretrainJob for better start_checkpoint behavior

AndreasPlt:
Currently, when `fairseq_hydra_config["checkpoint"]["restore_file"]` is given as a parameter for the pretraining, this makes resuming jobs much more difficult since it will always restart from the given file - even though the model might have been trained for several more epochs already. Since changing the `restore_file` parameter to the new checkpoint  would change the job parameter again, the job would also not be able continue training again.

I therefore propose to handle the `"restore_file"` parameter in the job individually and not leaving it to fairseq. The idea is to save the given `"restore_file"` as a job attribute to later move it to `"output/checkpoints/checkpoint_last.pt"` if the latter does not exist yet, which is also the default parameter inside fairseq. That way, when training has been already done for several epochs, the job can still continue from `checkpoint_last.pt`, and use the given checkpoint if the training hasn't already ran earlier.

### DIFF
--- a/fairseq/training.py
+++ b/fairseq/training.py
@@ -38,6 +38,13 @@ class FairseqHydraConfig:
         self.package_name = package_name
         self.check_consistency()
 
+    @property
+    def data(self):
+        """
+        Get the underlying data of the FairseqHydraConfig
+        """
+        return self.config_dict
+
     def write(self, path: str):
         config_dict = self.config_dict.copy()
         config_dict = util.update_nested_dict(config_dict, self.post_config_dict)
@@ -170,6 +177,24 @@ class FairseqHydraTrainingJob(Job):
         kwargs = locals()
         del kwargs["self"]
 
+        # check for start checkpoint
+        #if (
+        #    fairseq_hydra_config.data.get("checkpoint", {}).get("restore_file", None) is not None
+        #    and fairseq_hydra_config.data["checkpoint"]["restore_file"] != "checkpoint_last.pt"
+        #    and os.path.exists(os.path.join(self.output_path("checkpoints", directory=True), "checkpoint_last.pt"))
+        #):
+        #    # start_checkpoint provided but checkpoint_last.pt exists: start_checkpoint will be ignored
+        #    print(
+        #        "Warning: start_checkpoint will be ignored as checkpoint_last.pt exists in output directory"
+        #    )
+        #    fairseq_hydra_config.data["checkpoint"]["restore_file"] = "checkpoint_last.pt"
+
+        
+        self.start_checkpoint = None
+        if fairseq_hydra_config.data.get("checkpoint", {}).get("restore_file") is not None:
+            self.start_checkpoint = fairseq_hydra_config.data["checkpoint"]["restore_file"]
+            fairseq_hydra_config.data["checkpoint"]["restore_file"] = "checkpoint_last.pt"
+
         self.command_line_args = command_line_args or []
         stored_epochs = list(range(save_interval, max_epoch, save_interval)) + [max_epoch]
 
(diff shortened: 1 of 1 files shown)


fairseq/training.py
```diff
@@ -170,6 +177,12 @@ def __init__(
         kwargs = locals()
         del kwargs["self"]
 
+        # save start checkpoint and rename to checkpoint_latest
+        self.start_checkpoint = None
+        if fairseq_hydra_config.data.get("checkpoint", {}).get("restore_file") is not None:
+            self.start_checkpoint = fairseq_hydra_config.data["checkpoint"]["restore_file"]
+            fairseq_hydra_config.data["checkpoint"]["restore_file"] = "checkpoint_last.pt"
```
michelwi, line 184: You said the default value is already `"checkpoint_last.pt"` right? then we could simplify to
```suggestion
        self.start_checkpoint = fairseq_hydra_config.data.get("checkpoint", {}).pop("restore_file")
```
(pop removes the key/value pair and returns None if they key does not exist.)

fairseq/training.py
```diff
@@ -231,9 +244,28 @@ def create_fairseq_hydra_config(cls, fairseq_hydra_config, max_epoch, max_update
...
+        if not os.path.exists(os.path.join(self.out_checkpoint_dir.get_path(), "checkpoint_last.pt")):
+            print(f"Linking {start_checkpoint} to {self.out_checkpoint_dir.get_path()}")
+            os.symlink(
+                start_checkpoint,
+                os.path.join(self.out_checkpoint_dir.get_path(), "checkpoint_last.pt")
+            )
+            os.symlink(
+                start_checkpoint,
+                os.path.join(self.out_checkpoint_dir.get_path(), os.path.basename(start_checkpoint))
+            )
```
michelwi, line 264: is this needed?

fairseq/training.py
```diff
@@ -231,9 +244,28 @@ def create_fairseq_hydra_config(cls, fairseq_hydra_config, max_epoch, max_update
...
         return res
+    
+    def _fairseq_prepare_checkpoint(self, start_checkpoint):
+        # rename the start checkpoint to checkpoint_last.pt if it is not None and checkpoint_last.pt does not exist
+        if start_checkpoint is None:
+            print("No start checkpoint provided")
+            return
+        if not os.path.exists(start_checkpoint):
+            raise FileNotFoundError(f"Start checkpoint {start_checkpoint} does not exist")
+        if not os.path.exists(os.path.join(self.out_checkpoint_dir.get_path(), "checkpoint_last.pt")):
```
michelwi, line 255: I don't think `checkpoint_last.pt` can ever exist here, because we call this function only once in the `create_files` Task.
When we resume a training (i.e. Task `run`) later, this function is not called again.
((but we can still check 🤷 ))

fairseq/training.py
AndreasPlt: Yep, good point!

fairseq/training.py
AndreasPlt: Thanks for that hint, I didn't think about that `create_files` will be only called once. However, I would still argue for leaving that check in there to avoid the file being rewritten and therefore losing the progress of previous trainings in the form of the `checkpoint_last.pt` file. IMO it doesn't hurt code comprehension/readability and only makes the intention of the code more explicit

michelwi (approved):

fairseq/training.py
```diff
@@ -38,6 +38,13 @@ def __init__(
         self.package_name = package_name
         self.check_consistency()
 
+    @property
+    def data(self):
+        """
+        Get the underlying data of the FairseqHydraConfig
+        """
+        return self.config_dict
+
```
vieting, line 47: I guess this is not really needed. Do you think it improves the code quality/readability?

fairseq/training.py
vieting: I wonder if this can potentially break things depending on `start_checkpoint`. If its basename has a special meaning to fairseq (similar to `checkpoint_last.pt`), this could maybe result in unexpected behavior, right?

vieting (commented):
Looks good to me in general, just two nitpicks.

fairseq/training.py
AndreasPlt: In my opinion, accessing the actual config data in the `fairseq_hydra_config` with `fairseq_hydra_config.data` is more readable and clearer than `fairseq_hydra_config.config_dict`. However, it is not strictly necessary and would still work with the latter option, so in the end its just up to taste

vieting:
Btw, please apply black to your code, the test failed. The AppTek test probably only fails because the PR is created from a fork that has no permission to trigger the test.

michelwi:
> Btw, please apply black to your code, the test failed. The AppTek test probably only fails because the PR is created from a fork that has no permission to trigger the test.

Correct, we don't use this job, so we do not care.

fairseq/training.py
vieting: We could maybe just assign a fixed name here like e.g. `checkpoint_initial.pt` (if that has no special role so far). Since it's a symlink, it's still very easy to see the original filename if that's of interest.

AndreasPlt:
I recently found out that there is a fairseq parameter `checkpoint.continue_once` that essentially does the same as my proposed modification to the job, but inside of fairseq itself (see also [here](https://github.com/facebookresearch/fairseq/blob/main/fairseq/dataclass/configs.py#L654)). I therefore decided to close this PR with a reference to this fairseq parameter, if this functionality should be required by someone else in the future.

PR #555, opened 2024-11-14
Sentence Length Histogram Job

christophmluscher:
Job to generate a histogram of the sentence lengths

### DIFF
--- a/text/info.py
+++ b/text/info.py
@@ -8,24 +8,34 @@ from sisyphus import Job, Task, tk
 
 
 class SentenceLengthHistogramJob(Job):
-    def __init__(self, corpus_path: tk.Path, max_chars: Optional[int] = None, max_words: Optional[int] = None):
+    def __init__(
+        self,
+        input_text: tk.Path,
+        max_chars: Optional[int] = None,
+        max_words: Optional[int] = None,
+        full_histogram: bool = True,
+    ):
         """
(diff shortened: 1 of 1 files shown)


text/info.py
```diff
@@ -0,0 +1,90 @@
...
+
+                num_words = len(line.split())
+                num_chars = len(line)
+
+                counter_words[num_words] += 1
+                counter_chars[num_chars] += 1
+
+                if (self.max_words is not None or self.max_chars is not None) and (
+                    num_words > self.max_words or num_chars > self.max_chars
+                ):
```
michelwi, line 65: if one out of `self.max_words` and `self.max_chars` is not None but the other is, then the first set of parentheses evaluates to True and the second one is evaluated where `some_number > None` will lead to an error. Suggest to transpose the parentheses:
```suggestion
                if (self.max_words is not None and num_words > self.max_words) or (
                    self.max_chars is not None and num_chars > self.max_chars
                ):
```

michelwi (changes requested):

michelwi (approved):

JackTemaki (approved):
I feel the job is very specific, and maybe user would want to adjust the plot (e.g. different bin sizes). One extension idea: you could print the removed lines in a different color or so.

But technically it seems fine, so I guess I can approve.

christophmluscher:
I would let future users extend this if they want further functionality :D

PR #556, opened 2024-11-14
Text File Splitting Job

christophmluscher:
Split a text file into several text files.

### DIFF
--- a/text/info.py
+++ b/text/info.py
@@ -103,7 +103,7 @@ class CountLinesJob(Job):
         yield Task("run", rqmt=self.rqmt)
 
     def run(self):
-        zcat_cmd = ["zcat", "-f", "ted2_transcripts.gz"]
+        zcat_cmd = ["zcat", "-f", self.input_text.get_path()]
         zcat_res = subprocess.run(zcat_cmd, check=True, capture_output=True)
 
         wc_cmd = ["wc", "-l"]
--- a/text/info.py
+++ b/text/info.py
@@ -1,6 +1,7 @@
 __all__ = ["SentenceLengthHistogramJob"]
 
 from collections import Counter
+import subprocess
 from typing import Optional
 
 from i6_core.util import uopen
@@ -88,3 +89,24 @@ class SentenceLengthHistogramJob(Job):
         ax.plot(x, y, "s")
 
         fig.savefig(fig_path, bbox_inches="tight")
+
+
+class CountLinesJob(Job):
+    def __init__(self, input_text: tk.Path):
+        self.input_text = input_text
+
+        self.out_num_sentences = self.output_var("num_sentences.txt")
+
+        self.rqmt = {"cpu": 1, "mem": 1, "time": 1}
+
+    def tasks(self):
+        yield Task("run", rqmt=self.rqmt)
+
+    def run(self):
+        zcat_cmd = ["zcat", "-f", "ted2_transcripts.gz"]
+        zcat_res = subprocess.run(zcat_cmd, check=True, capture_output=True)
+
+        wc_cmd = ["wc", "-l"]
+        wc_res = subprocess.run(wc_cmd, input=zcat_res.stdout, check=True, capture_output=True)
+
+        self.out_num_sentences.set(int(wc_res.stdout.decode("utf-8").strip()))
--- a/text/info.py
+++ b/text/info.py
@@ -1,4 +1,4 @@
-__all__ = ["SentenceLengthHistogramJob"]
+__all__ = ["SentenceLengthHistogramJob", "CountLinesJob"]
 
 from collections import Counter
 import subprocess
--- a/text/info.py
+++ b/text/info.py
@@ -1,0 +1,90 @@
+__all__ = ["SentenceLengthHistogramJob"]
+
+from collections import Counter
+from typing import Optional
+
+from i6_core.util import uopen
+from sisyphus import Job, Task, tk
+
+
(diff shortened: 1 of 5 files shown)


text/processing.py
```diff
@@ -305,3 +310,85 @@ def run(self):
...
+            ]
+            subprocess.run(split_cmd, check=True)
+
+            for file_id in range(1, self.num_output_files + 1):
+                file_path = f"split.{file_id:04}.txt"
+                assert os.path.isfile(file_path) and os.path.getsize(file_path) > 0
+
+            if self.in_text_file.get_path().endswith(".gz"):
+                os.remove(text_file)
+
```
michelwi, line 382:
```suggestion
```
I think the TemporaryDirectory will delete itself after the block exists?

michelwi (approved):
just a little nitpick comment :)

text/processing.py
```diff
@@ -305,3 +310,82 @@ def run(self):
...
             else:
                 raise NotImplementedError
+
+
+class SplitTextFileJob(Job):
+    def __init__(
+        self,
+        text_file: tk.Path,
+        num_lines_per_split: int,
+        num_text_file_lines: int,
```
JackTemaki, line 320: In the old MT recipes we would compute such things automatically in the Job, even with another job in the init so that we could determine rqmt automatically (I am also quite sure exactly this kind of Job existed there.) I am not a fan at all of providing a meta information about the input, which can be computed by a simple command.

JackTemaki (changes requested):
The `num_text_file_lines` parameter is unnecessary and should be removed.

text/processing.py
christophmluscher: But then I would need to run the command in the manager since the value would need to be computed in the `__init__` and given the dataset sizes I am working with this becomes quite slow..

christophmluscher: offline discussion result: make argument optional and have it computed in the job if not provided

JackTemaki: Sorry no, this is not what I meant, we had a "line count" job for that, it would definitely not run in the manager.

Also for me it is fine to count within the run.

christophmluscher: I need within the `__init__` to get the number of output files. so in run is to late

christophmluscher: new discussion result: add CountLinesJob and execute within the `__init__`

text/info.py
```diff
@@ -88,3 +89,24 @@ def _plot(counter: Counter, fig_path: str):
...
+
+        self.out_num_sentences = self.output_var("num_sentences.txt")
+
+        self.rqmt = {"cpu": 1, "mem": 1, "time": 1}
+
+    def tasks(self):
+        yield Task("run", rqmt=self.rqmt)
+
+    def run(self):
+        zcat_cmd = ["zcat", "-f", "ted2_transcripts.gz"]
```
JackTemaki, line 106: I think this should rather be the input text, otherwise the job would be oddly specific ;)

text/processing.py
christophmluscher: we had a miscommunication 😅 

`CountLinesJob` added but removed from `SplitTextFileJob`

text/processing.py
```diff
@@ -305,3 +311,91 @@ def run(self):
...
+        self.num_lines_per_split = num_lines_per_split
+        self.zip_output = zip_output
+
+        if num_text_file_lines is not None:
+            self.num_text_file_lines = num_text_file_lines
+
+            self.num_output_files = self.num_text_file_lines // num_lines_per_split + int(
+                bool(self.num_text_file_lines % num_lines_per_split)
+            )
+        else:
```
JackTemaki: I now understood that this way is not possible, because the outputs can not be dynamic. I would leave this  else branch as "not implemented yet".

JackTemaki (changes requested):

text/info.py
christophmluscher: hahaha ups that is from my testing xD

JackTemaki:
We usually do not use .txt for sisyphus variable files, I did a small change there.

JackTemaki (approved):

michelwi (commented):
I am of a mind to request changes but.. ok lets say "mildly strong suggest changes"

text/processing.py
```diff
@@ -305,3 +311,85 @@ def run(self):
                     f.write(f"{line}\n")
             else:
                 raise NotImplementedError
+
+
+class SplitTextFileJob(Job):
```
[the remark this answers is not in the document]
christophmluscher: Differences are basically:
- split is faster than python implementation
- Can read/write gzip files
- uses temp dir

text/processing.py
```diff
@@ -305,3 +311,85 @@ def run(self):
...
+        self.num_lines_per_split = num_lines_per_split
+        self.num_text_file_lines = num_text_file_lines
+        self.zip_output = zip_output
+
+        if num_text_file_lines is not None:
+            self.num_output_files = self.num_text_file_lines // self.num_lines_per_split + int(
+                bool(self.num_text_file_lines % self.num_lines_per_split)
+            )
+        else:
+            raise NotImplementedError
```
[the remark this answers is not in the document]
christophmluscher: @JackTemaki preferred to have the variable optional. I dont have a strong opinion on this.. My use case is covered.. :D

yeah quite a bunch of useless commits.. but we squash in the end.. so no effort from my side to clean up

PR #557, opened 2024-11-19
Fix DumpAlignmentJob flow

Icemole:
The flow needed the parameters 'id' and 'TASK'.

Extract of the `dump.flow` used by the job:
```
  <node filter="generic-cache" id="$(id)" name="alignment-cache" path=".../alignment.cache.$(TASK)"/>
[...]
  <node file="alignment.cache.$(TASK)" filter="speech-alignment-dump" id="$(id)" name="dumper"/>
```


michelwi (approved):
Parameter `"TASK"` is not an explicit dependence here, but is _almost always_ used by `original_alignment` and `new_alignment`.

christophmluscher (approved):

PR #558, opened 2024-11-19
Add job to plot alignment phonemes

Icemole:
`PlotViterbiAlignmentJob`: job to plot the Viterbi alignment. The point of this job is to check for a specific issue in the final alignments.

---

~~`DumpSegmentTextAlignmentJob`: job to dump text/alignment pairs for a specific corpus and set of alignment files. The point of this job is to review in a fast way the text against the corresponding alignments.~~

~~The output is a compressed text file of the form:~~
```
<segment-id>
<segment-text>
<id-0> <timestamp-start-0> <timestamp-end-0> <alignment-0> <weight-0>
...
```

~~Feel free to propose any other format in the comments.~~

Edit 2025-05-12: `DumpSegmentTextAlignmentJob` removed. The RASR `archiver` binary can do the same.


returnn/hdf.py
```diff
@@ -181,7 +187,10 @@ def __init__(
         self.alignment = alignment
         self.rasr_exe = rasr.RasrCommand.select_exe(crp.nn_trainer_exe, "nn-trainer")
         self.feature_flow = ReturnnRasrTrainingJob.create_flow(feature_flow, alignment)
-        (self.rasr_config, self.rasr_post_config,) = ReturnnRasrTrainingJob.create_config(
+        (
```
Icemole, line 190: This is the byproduct of running `ruff format`. I hope this is ok.

DanEnergetics:
I prefer the second to last or last dumping format that you suggested. They would make it easiest to get a quick grasp on the duration of certain phonemes for me.

mm/alignment.py
```diff
@@ -789,3 +792,176 @@ def plot(self):
...
+        ax.set_xlabel("Frame")
+        ax.xaxis.set_label_coords(0.98, -0.03)
+        ax.set_xbound(0, num_alignments - 1)
+        ax.set_ybound(-0.5, max_timestamp - 0.5)
+
+        ax.set_yticks(np.arange(max_timestamp))
+        ax.set_yticklabels(allophone_sequence)
+
+        ax.imshow(viterbi_matrix, cmap="Blues", interpolation="nearest", aspect="auto", origin="lower")
+        return fig
```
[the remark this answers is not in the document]
Icemole: Adding an `Optional[tk.Path]` as the corpus file, and if provided modifying the png's title, can be done!

mm/alignment.py
```diff
@@ -789,3 +792,176 @@ def plot(self):
...
+                    f"{segment_id_to_text[segment_id]}{self.csv_separator}"
+                    f"{segment_id_to_alignment[segment_id]}{self.csv_separator}\n"
+                )
+
+
+class PlotViterbiAlignmentJob(Job):
+    """
+    Plots the alignments of each segment in the specified alignment files.
+
+    Original job author: dmann.
```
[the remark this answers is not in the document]
Icemole: As you prefer :) then your contribution will remain in our hearts haha

mm/alignment.py
```diff
@@ -789,3 +792,160 @@ def plot(self):
...
+
+class DumpSegmentTextAlignmentJob(Job):
+    """
+    Dumps all text and alignments for the given corpus and alignment files in a human-readable csv format.
+    """
+
+    def __init__(
+        self,
+        corpus_file: tk.Path,
+        alignment_files: Iterable[tk.Path],
```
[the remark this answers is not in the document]
michelwi: It is unclear which file format is expected. maybe the name `alignment_caches` would make it more clear that rasr caches are expected. otherwise add a bit more info to docstr pls.

Icemole:
> I prefer the second to last or last dumping format that you suggested. They would make it easiest to get a quick grasp on the duration of certain phonemes for me.

Maybe we can specify a default mapping (csv file?) and an optional custom mapping to be provided by the user.

mm/alignment.py
```diff
@@ -789,3 +792,176 @@ def plot(self):
...
+        fig, ax = plt.subplots(figsize=(10, 10))
+        ax.set_xlabel("Frame")
+        ax.xaxis.set_label_coords(0.98, -0.03)
+        ax.set_xbound(0, num_alignments - 1)
+        ax.set_ybound(-0.5, max_timestamp - 0.5)
+
+        ax.set_yticks(np.arange(max_timestamp))
+        ax.set_yticklabels(allophone_sequence)
+
+        ax.imshow(viterbi_matrix, cmap="Blues", interpolation="nearest", aspect="auto", origin="lower")
```
michelwi, line 978:
```suggestion
        ax.imshow(viterbi_matrix, cmap="Blues", interpolation="none", aspect="auto", origin="lower")
```
I think we should turn off interpolation for more "correctness". Would this be a problem?

mm/alignment.py
```diff
@@ -789,3 +792,176 @@ def plot(self):
...
+        :return: Matrix corresponding to the Viterbi alignment.
+        """
+        num_alignments = len(label_idx_seq)
+        max_timestamp = max(label_idx_seq) + 1
+        viterbi_matrix = np.zeros((max_timestamp, num_alignments), dtype=np.float32)
+        for t, idx in enumerate(label_idx_seq):
+            viterbi_matrix[idx, t] = 1.0
+        return viterbi_matrix
+
+    def plot(self, viterbi_matrix: np.array, allophone_sequence: List[str]):
```
michelwi: The typing indicates that the function does not return anything, but this is not the case.

mm/alignment.py
Icemole: I tested it and there's no problem at all :woman_shrugging:

mm/alignment.py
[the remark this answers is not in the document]
Icemole: I just modified the variables to be `alignment_caches`. That makes sense.

Icemole:
After the PR was created, I was told that the RASR archiver can do the same as the `DumpSegmentTextAlignmentJob` in this PR. I'd propose to remove the job. The most innovative job from this PR (and the one I use the most as of today) is still `PlotViterbiAlignmentJob`.

mm/alignment.py
```diff
@@ -819,3 +842,278 @@ def run(self):
...
+
+        ax.set_yticks(np.arange(max_timestamp))
+        ax.set_yticklabels(allophone_sequence)
+
+        ax.set_title(title)
+
+        ax.imshow(viterbi_matrix, cmap="Blues", interpolation="none", aspect="auto", origin="lower")
+
+        # The plot will be purposefully divided into subdirectories.
+        os.makedirs(os.path.dirname(os.path.join(self.out_plot_dir.get_path(), file_name)), exist_ok=True)
```
kuacakuaca, line 1057: isn't `os.path.dirname(os.path.join(self.out_plot_dir.get_path(), file_name)` exactly `os.path.join(self.out_plot_dir.get_path())`?

mm/alignment.py
```diff
@@ -819,3 +842,278 @@ def run(self):
...
+
+    def make_viterbi_matrix(self, label_idx_seq: np.array) -> np.array:
+        """
+        :return: Matrix corresponding to the Viterbi alignment.
+        """
+        num_alignments = len(label_idx_seq)
+        max_timestamp = max(label_idx_seq) + 1
+        viterbi_matrix = np.zeros((max_timestamp, num_alignments), dtype=np.float32)
+        for t, idx in enumerate(label_idx_seq):
+            viterbi_matrix[idx, t] = 1.0
```
kuacakuaca, line 1025: shouldn't it be `viterbi_matrix[t, idx]` if rows represent time steps?

Icemole: I've studied the code further. The thing is that we're using the Y axis for timestamps because we want to represent the phonemes on the Y axis. Then, this parameter represents the alignment indices, for instance in an alignment of `np.array(["a", "a", "b", "b", "b", "c", ...])`, `label_indices` would be `[0, 0, 1, 1, 1, 2, ...]` (see `extract_phoneme_sequence`), so the maximum value of the Y axis is actually the number of phonemes in the alignment.

I've updated the variable names to clarify better this.

Icemole: I believe the variables are now better represented, and the code is better explained. @kuacakuaca could you please check again? Thanks!

kuacakuaca (approved):

Icemole:
Thanks for the reviews!

PR #559, opened 2024-11-29
Fix compute ppl job

curufinwe:
Currently any change to the crp object will change the hash of ComputePerplexityJob. Even changes to the acoustic model for example. Typically RASR jobs implement their own hash function to hash the generated config, not the input crp. This was overlooked for the ComputePerplexityJob. To avoid breaking hashes I add a new version with the fixed behavior.


Icemole (approved):
Looks good :+1:

michelwi (approved):

sarahberanek (approved):
This was so annoying! I didnt realize thats whats happening!
Good change!
Can we do anything about neural LM PPLs?

PR #560, opened 2024-12-03
Fix newline in Bliss corpus orth

Icemole:
The actual functionality of the code seems ambiguous with respect to the explanatory comment some lines above.

In order to keep the information from the writing, the newline should be substituted by a space, and not by an empty string. Example:

```
HELLO
WORLD
```

Current functionality: `HELLOWORLD`.

My proposal is consistent with the human way of reading: `HELLO WORLD`.

### DIFF
--- a/lib/corpus.py
+++ b/lib/corpus.py
@@ -123,7 +123,7 @@ class CorpusParser(sax.handler.ContentHandler):
             # writing, thus we remove multiple spaces and newlines
             text = self.chars.strip()
             text = re.sub(" +", " ", text)
-            text = re.sub("\n", "", text)
+            text = re.sub("\n", " ", text)
             e.orth = text
         elif isinstance(e, Speaker) and name != "speaker-description":
             # we allow all sorts of elements within a speaker description


kuacakuaca (approved):

albertz (commented):
I wonder if this can break anything else? Should we maybe have a `strip()` at the end?

michelwi:
How does rasr handle this? I think this lib should follow rasr behavior even if it is not how we expect or want it.

christophmluscher (changes requested):
I agree with @michelwi this would break older setups. Maybe introduce `V2`?

Atticus1806 (changes requested):
I agree with the other two, this should not change old existing corpora.

Icemole:
Hi all, thank you for your answers. I see that there's a lot of (understandable) reservations towards changing this.

For further context, my initial (and only, so far) use case was being able to split words separated by `\n` in the transcription, but then I realized that I could do that when creating the first corpus to begin with, for which I had the control.

Again, thank you all for your feedback. From my side this can be declined so as not to cause further confusion, and apologies for any already caused.

I'll leave the thread open in case the V2 idea has settled on your minds and your heart is now burning with desire of implementing this :smile:

michelwi:
FYI, I gave it a quick test, and it seems to me that RASR separates words on newlines. So the current implementation is indeed a bug.

Icemole:
It would then be interesting to see a `CorpusV2` implementation some day. But integrating this in all our jobs would be a very complicated task...

PR #564, opened 2024-12-09
Add left/right context orth

Icemole:
This PR adds support for the `<left-context-orth>` and `<right-context-orth>`, which specify optional alignment tokens for RASR alignment.

Moreover, this PR also adds parameters to the segment constructor.

### DIFF
--- a/lib/corpus.py
+++ b/lib/corpus.py
@@ -116,7 +116,7 @@ class CorpusParser(sax.handler.ContentHandler):
     def endElement(self, name: str):
         e = self.elements[-1]
 
-        if name == "orth":
+        if name == "orth" or name == "left-context-orth" or name == "right-context-orth":
             assert isinstance(e, Segment)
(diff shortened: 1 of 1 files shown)


michelwi (approved):

albertz (approved):
See my comment. Otherwise ok.

lib/corpus.py
```diff
@@ -116,15 +116,15 @@ def startElement(self, name: str, attrs: Dict[str, str]):
     def endElement(self, name: str):
         e = self.elements[-1]
 
-        if name == "orth":
+        if name == "orth" or name == "left-context-orth" or name == "right-context-orth":
```
curufinwe, line 119:
```suggestion
        if name in {"orth", "left-context-orth", "right-context-orth"}:
```

curufinwe (changes requested):

curufinwe (approved):

PR #565, opened 2024-12-10
Add `full_orth` function

Icemole:
Function to get the full orthographhy from the segment.

This function should be used from now on instead of accessing `orth` directly. We can also name the function `orth()` to minimize the impact of the change.

### DIFF
--- a/lib/corpus.py
+++ b/lib/corpus.py
@@ -394,14 +394,7 @@ class Segment(NamedEntity):
         """
         :return: Left context orth (if any) + orth + right context orth (if any).
         """
-        text = ""
-        if self.left_context_orth is not None:
-            text += self.left_context_orth.strip() + " "
-        text += self.orth
-        if self.right_context_orth is not None:
-            text += " " + self.right_context_orth.strip()
-
-        return text
+        return " ".join([s for s in [self.left_context_orth, self.orth, self.right_context_orth] if s])
 
     def fullname(self) -> str:
         return self.recording.fullname() + "/" + self.name
--- a/lib/corpus.py
+++ b/lib/corpus.py
@@ -383,7 +383,7 @@ class Segment(NamedEntity):
         self.start = start
         self.end = end
         self.track = track
-        self._orth = orth
+        self.orth = orth
         self.left_context_orth = left_context_orth
         self.right_context_orth = right_context_orth
         self.speaker_name = speaker_name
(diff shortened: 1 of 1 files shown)


Icemole:
Question: should we add this as a property of `orth` instead? But then we'd have to create a `self._orth` variable, which implies also doing some kind of setter:

```
@property
def orth(self):
    # Same as current full_orth()

@orth.setter
def orth(self, text):
    self._orth = text
```

But then this would be super misleading:

```
s = Segment()
s.left_context_orth = "Hello"
s.orth = "world"
print(s.orth)  # 'Hello world', instead of the potentially expected 'world'.
```

lib/corpus.py
```diff
@@ -383,13 +383,26 @@ def __init__(
         self.start = start
         self.end = end
         self.track = track
-        self.orth = orth
+        self._orth = orth
```
albertz, line 386: I con'g understand this change?

Icemole: Typo from a previous version in which I used a property. Sorry about that, will revert back.

lib/corpus.py
```diff
@@ -390,6 +390,19 @@ def __init__(
...
+        :return: Left context orth (if any) + orth + right context orth (if any).
+        """
+        text = ""
+        if self.left_context_orth is not None:
+            text += self.left_context_orth.strip() + " "
+        text += self.orth
+        if self.right_context_orth is not None:
+            text += " " + self.right_context_orth.strip()
+
+        return text
```
albertz, line 404: Can there be a case where `self.orth == ""`?
```suggestion
        return " ".join([s for s in [self.left_context_orth, self.orth, self.right_context_orth] if s])
```

michelwi: Sure, orth can even be None.

Icemole: > Can there be a case where self.orth == ""?

There can definitely be such a case, it totally depends on the user. Although it's not normal, and implies that there might be something wrong with the segment (e.g. TN removed all of the data).

Your code is more dense and meaningful, I'll use your approach, which also works nicely when everything's `None`.

albertz:
> Question: should we add this as a property of `orth` instead?

No.

albertz (approved):

michelwi:
> Question: should we add this as a property of `orth` instead?

No.. as you already noticed this would be a terrible interface. 

Also all Jobs that interact with the lib would need special logic to handle the optionality of the *_context_orth.

michelwi (approved):

PR #567, opened 2025-01-08
`BlissToPcmHDFJob`: change `round_factor` typing from `int` to `float`

Icemole:
Right now the typing of `round_factor` in `BlissToPcmHDFJob` implies that the original sampling rate is a divisor of the new sampling rate. For instance, 8kHz -> 16kHz conversion is supported, but not 16kHz -> 44.1kHz.

Moreover, downsampling would not be supported either, since 16kHz -> 8kHz results in `round_factor = 0.5`.

This PR aims to fix this issue by simply treating `round_factor` as a floating point number. I haven't seen anything in the code that could interfere with this change. Do you see anything against this change?

### DIFF
--- a/returnn/hdf.py
+++ b/returnn/hdf.py
@@ -229,7 +229,7 @@ class BlissToPcmHDFJob(Job):
     __sis_hash_exclude__ = {
         "multi_channel_strategy": BaseStrategy(),
         "rounding": RoundingScheme.start_and_duration,
-        "round_factor": 1,
+        "round_factor": 1.0,
     }
 
     def __init__(
@@ -240,7 +240,7 @@ class BlissToPcmHDFJob(Job):
         multi_channel_strategy: BaseStrategy = BaseStrategy(),
         returnn_root: Optional[tk.Path] = None,
         rounding: RoundingScheme = RoundingScheme.start_and_duration,
-        round_factor: int = 1,
+        round_factor: float = 1.0,
     ):
         """
 


michelwi:
> Do you see anything against this change?

First and foremost:
```
                        start = int(segment.start * audio.samplerate / self.round_factor) * self.round_factor
                        duration = (
                            int((segment.end - segment.start) * audio.samplerate / self.round_factor)
                            * self.round_factor
                        )
```
start and duration are not ints any more if `self.round_factor` is a float now, this is a technical problem.

Then you start adding more `int(..)`, `ceil(..)`, or `floor(..)` calls and we should start wondering which rounding scheme is the correct one in all of the cases.. since - remember - if we are off by one sample some times, we might be off by one frame a few times and we might be off one chunk eventually..

So I request you think about it some more :)

michelwi (changes requested):
* add rounding to start/duration computation
* think about which rounding is correct in which case

Icemole:
Thank you for the feedback, Willi. I didn't realize at first that `seek()` and `read()` should be provided ints (of course they should be, given the nature of the audio samples). I initially thought it was an easy change, but the implementation would now be much more difficult, since we should not only test the result, but also dive into RASR code so that both approaches match :thinking: 

I'll close this PR since we don't need this for now. Luckily we only upsample...

PS: I've just seen the changes in https://github.com/rwth-i6/i6_core/pull/524/commits (the original implementation of `round_factor`) and I saw that this was already discussed.

PR #568, opened 2025-01-10
Add target_sampling_rate parameter to BlissToPcmHDFJob

michelwi:
Instead of calling a bunch of `BlissFfmpegJob` before `BlissToPcmHDFJob` we now can resample the audio data on the fly before writing to the HDF file. 
Useful for multi-bandwidth training.

Also a bit of cleanup in the Job as I was on it anyway.

### DIFF
--- a/requirements.txt
+++ b/requirements.txt
@@ -1,5 +1,6 @@
 black==22.3.0
 h5py
+librosa
 matplotlib
 numpy
 soundfile
--- a/returnn/hdf.py
+++ b/returnn/hdf.py
@@ -4,6 +4,7 @@ from dataclasses import dataclass
 from enum import Enum, auto
 import glob
 import math
+import librosa
 import numpy as np
 import os
 import shutil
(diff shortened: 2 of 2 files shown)


michelwi:
huh.. are we not having librosa in i6_core?

JackTemaki:
> huh.. are we not having librosa in i6_core?

No, I think so far no core job used it.

michelwi:
> No, I think so far no core job used it.

yes, grep told me.. I would like to change my question to "Is is available in your apptainer images / systems / etc, or would I break someones things if I merged this?" 
Otherwise I could just add it to the requirements and have it available in the test pipeline.

Icemole (approved):
Looks good. Do you know how it interacts with the `round_factor` parameter? Should we assert, or at least warn, that only one of the two should be set at a given point?

michelwi:
> Do you know how it interacts with the `round_factor` parameter?

technically both are solving the same problem in two different ways:
* if the audio was resampled externally, `round_factor` makes this Job read the resampled audio file as if the timestamps were calculated with the original sampling rate.
* with resampling the audio file *is* read at the "original" sampling rate and resampled internally.

Now if you had say 48kHz data, externally resampled it to 8kHz and then internally upsampled to 16kHz again, it would make sense to set both parameters; but I admit this is a contrived scenario and in practice we would only use one of those parameters.

curufinwe (approved):

PR #569, opened 2025-01-10
Add new functionalities to BPE lexicon creation

Stefanwuu:
In order to use this job in more complicated scenarios, I added some functionalities to CreateBPELexiconJob. The original functionalities and hashes should stay unchanged.

### DIFF
--- a/lexicon/bpe.py
+++ b/lexicon/bpe.py
@@ -1,6 +1,7 @@
 __all__ = ["CreateBPELexiconJob"]
 
 import subprocess as sp
+import logging
 import os
 import sys
 from typing import List, Optional, Set, Union
@@ -19,6 +20,8 @@ class CreateBPELexiconJob(Job):
     This job is still in experimental state, and only tested with Flashlight BPE decoding
     """
 
+    __sis_hash_exclude__ = {"skip_unk_lemmas": False, "add_all_bpe_phonemes": True, "additional_words": None}
+
     def __init__(
         self,
         base_lexicon_path: tk.Path,
@@ -28,6 +31,9 @@ class CreateBPELexiconJob(Job):
         unk_label: str = "UNK",
         vocab_blacklist: Optional[Union[List[str], Set[str]]] = None,
         keep_special_lemmas: bool = True,
+        skip_unk_lemmas: bool = False,
+        add_all_bpe_phonemes: bool = True,
+        additional_words: Optional[tk.Path] = None,
     ):
         """
         :param base_lexicon_path: base lexicon (can be phoneme based) to take the lemmas from
@@ -41,6 +47,12 @@ class CreateBPELexiconJob(Job):
             usually yes for RASR search and no for Flashlight search.
             The phonemes of the special lemmas will also be kept, therefore
             make sure there is no overlap with the BPE vocab.
+        :param skip_unk_lemmas: whether simply skip lemmas out of the BPE vocab
+            useful if you set vocab_blacklist
+        :param add_all_bpe_phonemes: If set to True, all BPE vocab will be added to lexicon phonemes,
+            otherwise, only phonemes appear in lexicon lemma will be added to the lexicon.
+        :param additional_words: Aside from vocab specified in base_lexicon, we might want to convert some other words,
+            e.g. untranslatable words by a g2p model in case of g2p-augmented lexicon
         """
         self.base_lexicon_path = base_lexicon_path
         self.bpe_codes = bpe_codes
@@ -53,6 +65,9 @@ class CreateBPELexiconJob(Job):
             # convert list to set for faster "in" check
             self.vocab_blacklist = set(vocab_blacklist)
         self.keep_special_lemmas = keep_special_lemmas
+        self.skip_unk_lemmas = skip_unk_lemmas
+        self.add_all_bpe_phonemes = add_all_bpe_phonemes
+        self.additional_words = additional_words
 
         self.out_lexicon = self.output_path("lexicon.xml.gz", cached=True)
 
@@ -85,18 +100,31 @@ class CreateBPELexiconJob(Job):
                     vocab_file.write(symbol + " -1\n")
                     symbol = symbol.replace(".", "_")
                     vocab.add(symbol)
-                    lexicon.add_phoneme(symbol.replace(".", "_"))
+                    if self.add_all_bpe_phonemes:
+                        lexicon.add_phoneme(symbol.replace(".", "_"))
 
         return vocab, lexicon
+
+    def _fill_additional_words(self):
+        additional_words_list = set()
+        if self.additional_words is not None:
+            with util.uopen(self.additional_words.get_path(), "rt") as f:
+                for line in f:
+                    line = line.strip()
+                    additional_words_list.update(line)
+        return sorted(additional_words_list)
 
     def run(self):
         base_lexicon = Lexicon()
         base_lexicon.load(self.base_lexicon_path)
 
         lm_tokens, special_lemmas = self._fill_lm_tokens(base_lexicon)
+        additional_words_list = self._fill_additional_words()
 
         with util.uopen("words", "wt") as f:
             for t in lm_tokens:
+                f.write(f"{t}\n")
+            for t in additional_words_list:
                 f.write(f"{t}\n")
 
         vocab, lexicon = self._fill_vocab_and_lexicon()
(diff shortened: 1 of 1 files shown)


lexicon/bpe.py
```diff
@@ -127,15 +155,32 @@ def run(self):
...
+                used_vocab.update(set(bpe_pron.split()))
                 lexicon.add_lemma(Lemma([orth], [bpe_pron.replace(".", "_")], lemma.synt, lemma.eval))
 
+        for word in additional_words_list:
+            bpe_pron = " ".join([token if token in vocab else self.unk_label for token in w2b[word].split()])
+            if self.skip_unk_lemmas and self.unk_label in bpe_pron.split():
+                logging.info(f"Lemma {word} is skipped due to unknown BPE vocab.")
+                continue
+            used_vocab.update(set(bpe_pron.split()))
+            lexicon.add_lemma(Lemma([word], [bpe_pron.replace(".", "_")]))
```
albertz, line 178: Can't you share the code?

lexicon/bpe.py
```diff
@@ -85,14 +100,28 @@ def _fill_vocab_and_lexicon(self):
...
+                    line = line.strip()
+                    additional_words_list.add(line)
+        return sorted(additional_words_list)
+
     def run(self):
         base_lexicon = Lexicon()
         base_lexicon.load(self.base_lexicon_path)
 
+        additional_words_list = self._fill_additional_words()
+        print(additional_words_list)
```
albertz, line 122:
```suggestion
```

lexicon/bpe.py
```diff
@@ -41,6 +47,12 @@ def __init__(
             usually yes for RASR search and no for Flashlight search.
             The phonemes of the special lemmas will also be kept, therefore
             make sure there is no overlap with the BPE vocab.
+        :param skip_unk_lemmas: whether simply skip lemmas out of the BPE vocab
+            useful if you set vocab_blacklist
+        :param add_all_bpe_phonemes: If set to True, all BPE vocab will be added to lexicon phonemes,
+            otherwise, only phonemes appear in lexicon lemma will be added to the lexicon.
```
curufinwe, line 53:
```suggestion
        :param add_all_bpe_phonemes: If set to True, all BPE tokens will be added to lexicon as phonemes,
            otherwise, only tokens that appear in the base lexicon will be added to the output lexicon.
```

lexicon/bpe.py
```diff
@@ -41,6 +47,12 @@ def __init__(
...
+        :param additional_words: Aside from vocab specified in base_lexicon, we might want to convert some other words,
+            e.g. untranslatable words by a g2p model in case of g2p-augmented lexicon
```
curufinwe, line 55:
```suggestion
        :param additional_words: Aside from the vocabulary specified in base_lexicon, we might want to convert some other words,
            e.g. untranslatable words by a g2p model in case of g2p-augmented lexicon
```

lexicon/bpe.py
```diff
@@ -129,13 +157,22 @@ def run(self):
...
         w2b = {w: b for w, b in zip(lm_tokens, bpe_tokens)}
 
+        used_vocab = set()
         for lemma in base_lexicon.lemmata:
             if lemma.special:
                 continue
             for orth in lemma.orth:
                 bpe_pron = " ".join([token if token in vocab else self.unk_label for token in w2b[orth].split()])
+                if self.skip_unk_lemmas and self.unk_label in bpe_pron.split():
+                    logging.info(f"Lemma {orth} is skipped due to unknown BPE vocab.")
```
curufinwe, line 167:
```suggestion
                    logging.info(f"Lemma {orth} is skipped due to use of the BPE token for <unknown>.")
```

lexicon/bpe.py
```diff
@@ -85,14 +100,27 @@ def _fill_vocab_and_lexicon(self):
...
         return vocab, lexicon
 
+    def _fill_additional_words(self):
+        additional_words_list = set()
+        if self.additional_words is not None:
+            with util.uopen(self.additional_words.get_path(), "rt") as f:
+                for line in f:
+                    line = line.strip()
+                    additional_words_list.add(line)
+        return sorted(additional_words_list)
```
curufinwe, line 115:
```suggestion
        if self.additional_words is not None:
            with util.uopen(self.additional_words.get_path(), "rt") as f:
                res = {line.strip() for line in f}
        else:
            res = set()
        return sorted(res)
```

lexicon/bpe.py
```diff
@@ -85,14 +100,27 @@ def _fill_vocab_and_lexicon(self):
                     vocab_file.write(symbol + " -1\n")
                     symbol = symbol.replace(".", "_")
                     vocab.add(symbol)
-                    lexicon.add_phoneme(symbol.replace(".", "_"))
+                    if self.add_all_bpe_phonemes:
+                        lexicon.add_phoneme(symbol.replace(".", "_"))
 
         return vocab, lexicon
 
+    def _fill_additional_words(self):
```
michelwi, line 108:
```suggestion
    def _get_additional_words(self):
```
A function that returns the thing it is named after I would call "get" whereas "fill" implies that there is some existing container in the class that is filled inplace

EDIT: nevermind.. the above functions also already are called "fill" and returns stuff.. then better stay consistent.

lexicon/bpe.py
michelwi:
```suggestion
        if self.additional_words is not None:
            with util.uopen(self.additional_words.get_path(), "rt") as f:
                return sorted({line.strip() for line in f})
        return []
```
even more simpler

lexicon/bpe.py
```diff
@@ -85,14 +100,27 @@ def _fill_vocab_and_lexicon(self):
...
+                    additional_words_list.add(line)
+        return sorted(additional_words_list)
+
     def run(self):
         base_lexicon = Lexicon()
         base_lexicon.load(self.base_lexicon_path)
 
+        additional_words_list = self._fill_additional_words()
+        for w in additional_words_list:
+            base_lexicon.add_lemma(Lemma([w], None))  # add empty lemmata with only orth for additional words
```
michelwi, line 123: what is the benefit of adding lemmata without pronunciation? they can never be recognized and in training I guess it will be as is the word was deleted from the transcript?

lexicon/bpe.py
michelwi: `self.unk_label in bpe_pron.split()` means there is some string in the word that cannot be represented with the current BPE vocab (e.g. greek letters with an all latin vocab), right?
```suggestion
                    logging.info(f"Lemma {orth} is skipped because it cannot be represented with the BPE vocab.")
```

lexicon/bpe.py
```diff
@@ -28,6 +31,9 @@ def __init__(
         unk_label: str = "UNK",
         vocab_blacklist: Optional[Union[List[str], Set[str]]] = None,
         keep_special_lemmas: bool = True,
+        skip_unk_lemmas: bool = False,
+        add_all_bpe_phonemes: bool = True,
+        additional_words: Optional[tk.Path] = None,
```
michelwi, line 36: We could do it in this job, but my feeling is that adding a bunch of words with empty pronunciation is a separate task and I would rather have a `CreateEmptyPronunciationLexiconJob` or `CreateOrthOnlyLexiconJob` and then use the `MergeLexiconJob` to combine those.

lexicon/bpe.py
Stefanwuu: Yeah, we normally filter out non-word(like [noise]) from the text corpora to train BPE, and when we try to convert non-word lemmata in the base phoneme lexicon, they would be unknown to the BPE vocab due to the existence of "[" and "]". 
Btw, this PR is now NOT used in Apptek BPE pipeline and will only be a feature for i6 people. I will still try to get it merged, but don't worry that it would slow down our process in Apptek.

lexicon/bpe.py
Stefanwuu: Later, we have a for-loop(line 161) going through all lemma orthography in base_lexicon and writing their BPE-form "pronunciation" to the final lexicon. So it would not hurt if it's empty in the base_lexicon now. I did an additional for-loop over additional_word_list before, but there was a suggested change regarding code re-usage:)

michelwi: Ah, right, I missed that 👍

lexicon/bpe.py
```diff
@@ -129,13 +156,22 @@ def run(self):
...
         for lemma in base_lexicon.lemmata:
             if lemma.special:
                 continue
             for orth in lemma.orth:
                 bpe_pron = " ".join([token if token in vocab else self.unk_label for token in w2b[orth].split()])
+                if self.skip_unk_lemmas and self.unk_label in bpe_pron.split():
+                    logging.info(f"Lemma {orth} is skipped due to use of the BPE token for <unknown>.")
+                    continue
+                used_vocab.update(set(bpe_pron.split()))
                 lexicon.add_lemma(Lemma([orth], [bpe_pron.replace(".", "_")], lemma.synt, lemma.eval))
```
hannah220, line 170: why `if not`? Shouldn't we add when phoneme when `self.add_all_bpe_phonemes` is true?

PR #570, opened 2025-01-13
Add GetLongestAllophoneFileJob

Icemole:
Adds job to get the biggest allophone file out of those provided as parameter.

### DIFF
--- a/mm/alignment.py
+++ b/mm/alignment.py
@@ -4,8 +4,10 @@ __all__ = [
     "PlotAlignmentJob",
     "AMScoresFromAlignmentLogJob",
     "ComputeTimeStampErrorJob",
+    "GetBiggestAllophoneFileJob",
 ]
 
+import itertools
 import logging
 import math
 import os
@@ -14,6 +16,7 @@ import statistics
 import xml.etree.ElementTree as ET
 from typing import Callable, Counter, List, Optional, Tuple, Union
 
+import regex as re
 from sisyphus import *
 
 Path = setup_path(__package__)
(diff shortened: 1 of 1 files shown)


mm/alignment.py
```diff
@@ -789,3 +792,48 @@ def plot(self):
...
+        self.out_biggest_allophone_file = self.output_path("allophone_file.txt")
+
+        self.rqmt = {"cpu": 1, "mem": 1.0, "time": 1.0}
+
+    def tasks(self):
+        yield Task("run", resume="run", rqmt=self.rqmt)
+
+    def run(self):
+        allophone_files = [util.uopen(allophone_file.get_path(), "rt") for allophone_file in self.allophone_files]
+        allophone_header_regex = re.compile(r"# Number of allophones: (\d+)")
```
michelwi, line 817: pipeline doesn't seem to like `regex`, maybe just use `line.split()[4]` instead?

michelwi (approved):
I would maybe rebrand "biggest" to "longest", but that is only semantics, the code looks good to me.

mm/alignment.py
```diff
@@ -789,3 +791,53 @@ def plot(self):
...
+    def tasks(self):
+        yield Task("run", resume="run", rqmt=self.rqmt)
+
+    def run(self):
+        allophone_files = [util.uopen(allophone_file.get_path(), "rt") for allophone_file in self.allophone_files]
+        highest_num_allo, highest_num_allo_idx = -1, 0
+        for lines in itertools.zip_longest(*allophone_files):
+            if highest_num_allo == -1:
+                for i, line in enumerate(lines):
+                    assert line.startswith(
```
curufinwe, line 820: If you look at the code parsing allophone files in RASR (https://github.com/rwth-i6/rasr/blob/e08c2f925e3df633be16125671590e7def50f1e1/src/Am/ClassicStateModel.cc#L58) you can see that lines starting with # are completely ignored. Thus an allophone file that does not have this header is also valid. Thus I would suggest you adapt your parsing logic to exclude lines with comments at all and only compare the sequence of allophones in each file.

curufinwe (changes requested):

mm/alignment.py
Icemole: Done

mm/alignment.py
```diff
@@ -789,3 +791,38 @@ def plot(self):
...
+
+    def tasks(self):
+        yield Task("run", resume="run", rqmt=self.rqmt)
+
+    def run(self):
+        allophone_files = [util.uopen(allophone_file.get_path(), "rt") for allophone_file in self.allophone_files]
+        with open(self.out_longest_allophone_file.get_path(), "wt") as f:
+            for i, lines in enumerate(itertools.zip_longest(*allophone_files)):
+                # Keep removing redundant elements until a relevant one is found.
+                while lines[0] is None:
```
curufinwe, line 819: I'm not sure if this code would work with arbitrary comments, I think it would be better (and certainly easier to read) if you just apply a filter function on the open files, e.g.:
```
filtered_allophone_files = [filter(lambda line: not line.startswith("#"), f) for f in allophone_files]
```

curufinwe (changes requested):

mm/alignment.py
Icemole: It makes sense, thank you for the suggestion.

curufinwe (approved):

PR #577, opened 2025-01-29
SpellingConversionJob: clean lexicon if multiple mappings have been applied

sarahberanek:
For our **EN_US** lexicon we used this job to map all **UK** spellings to **US** spellings:
For example:
```
<lemma>
<orth>backpedaled</orth>
<orth>backpedalled</orth>
<phon>b æ k p ɛ d ə l d</phon>
<synt>
<tok>backpedalled</tok>
</synt>
</lemma>
```
so far so good, but now for the **EN_GB** lexicon we use the **EN_US** lexicon and map it all to **UK** spelling again, which resulted in this redundant, a bit ugly, but not incorrect lemma:
```
<lemma>
<orth>backpedalled</orth>
<orth>backpedaled</orth>
<orth>backpedalled</orth>
<phon>b æ k p ɛ d ə l d</phon>
<synt>
<tok>backpedalled</tok>
</synt>
</lemma>
```
with my code fixes here the final lemma looks like this, which is much nicer:
```
<lemma>
<orth>backpedalled</orth>
<orth>backpedaled</orth>
<phon>b æ k p ɛ d ə l d</phon>
</lemma>
```

I am very sure I am the first person who ever encountered that :-)

### DIFF
--- a/lexicon/conversion.py
+++ b/lexicon/conversion.py
@@ -506,9 +506,7 @@ class SpellingConversionJob(Job):
             elif source_lemma:
                 # Remove target orth if already present at a later position
                 # and insert it at the first position again
-                for orth in source_lemma.orth.copy():
-                    if orth == target_orth:
-                        source_lemma.orth.remove(orth)
+                source_lemma.orth = [orth for orth in source_lemma.orth if orth != target_orth]
                 source_lemma.orth.insert(0, target_orth)
                 if not source_lemma.synt:
                     source_lemma.synt = source_orth.split()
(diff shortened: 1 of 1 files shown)


sarahberanek:
@michelwi any idea why the apptek hashes break?

Icemole (approved):

sarahberanek:
> It would be great in the future to add functionality to `Lemma` that automatically handles this. Fixing this in a lexicon job doesn't prevent it from appearing in the rest of the jobs.

Thats actually a good idea. Thanks for the approval. I merge.

PR #580, opened 2025-02-05
Add DumpRecordingAudiosJob and compute_rec_duration

Icemole:
Added functionality:
- `compute_rec_duration`: computes the duration of a given recording. Currently accepts mp3, wav or aac (will convert to wav on the fly).
- `DumpRecordingAudiosJob`: dumps all recordings of a (list of) corpus into a file, one audio per line. Optionally dumps another file with `<audio>\t<duration>` per line.


Icemole:
I would like to add a `get_audio_to_duration_mapping` helper function, but I can't do it in `i6_core/lib/corpus.py` because there would be a circular import. If you have any suggestions on where to place the function, please let me know. For now I'll remove it.

corpus/info.py
```diff
@@ -0,0 +1,57 @@
...
+                audio_to_duration[r.audio] = duration
+
+        with uopen(self.out_audio_list.get_path(), "wt") as f:
+            for audio in audio_to_duration.keys():
+                f.write(f"{audio}\n")
+
+        if self.dump_durations:
+            with uopen(self.out_audio_durations.get_path(), "wt") as f:
+                for audio, duration in audio_to_duration.items():
+                    f.write(f"{audio}\t{duration}")
```
hannah220, line 57: You need newline at the end: 
`f.write(f"{audio}\t{duration}\n")`

Icemole: Thank you! Indeed this could have been catastrophic :sweat_smile:

I've fixed the feedback :)

corpus/info.py
```diff
@@ -0,0 +1,57 @@
...
+    """
+
+    def __init__(
+        self, corpus_files: Union[tk.Path, List[tk.Path]], dump_durations: bool = False, zip_output: bool = False
+    ):
+        r"""
+        :param corpus_file: Corpus file from which to obtain the audio list.
+        :param dump_durations: Whether to dump the durations of the audios along with the audio list.
+            The `out_audio_durations` output file will contain an audio/duration pair,
+            separated by :param:`duration_file_delimiter`.
```
hannah220, line 24: It's always separated by tab, not by `duration_file_delimiter`

lib/audio.py
```diff
@@ -0,0 +1,43 @@
...
+
+
+import subprocess as sp
+import tempfile
+
+try:
+    from mutagen.mp3 import MP3
+except ImportError:
+    # No mutagen module. mp3 conversion will not be supported.
+    pass
```
michelwi, line 11: instead you could also move the import into the function where it is only called if needed

lib/audio.py
```diff
@@ -0,0 +1,43 @@
...
+    """
+    if rec_audio.endswith("mp3"):
+        try:
+            audio_length = MP3(rec_audio).info.length
+        except NameError as exception:
+            exception.msg = (
+                "The 'mutagen' module is required to calculate the duration of the mp3 file "
+                f"{rec_audio}, but it's not installed in your system."
+            )
+            raise exception
```
michelwi, line 30: If the module does not exist, we can still get the result in the same way as with `.aac` files: use ffmpec to convert to wav
If you want this can be a fallback. or even the default implementation for mp3's so we do not add a new dependency.

corpus/info.py
michelwi, line 1: Oftentimes I am annoyed if I look for a Job that "takes a corpus and changes it somehow", because we have `corpus.transform` and `corpus.convert`.
Now here `corpus.info` is introduced while `corpus.stats` already exists. Would you hazard this Job may also go to the existing namespace? `DumpRecordingAudiosJob` is not really ant statistic.. but neither are `ExtractOovWordsFromCorpusJob` and `CountCorpusWordFrequenciesJob`..

Icemole: Yes, I'm also super bothered by that. Because `DumpRecordingAudiosJob` shouldn't go into `stats.py` since as you said it's not a statistic. That's why I created `info.py`.

I'll add it into `stats.py` for the sake of completing the PR, but to be fair I'd rather move everything into `info.py`, since a statistic is a piece of information, but a piece of information is not a statistic :slightly_smiling_face: 

Maybe in some other PR I'll get crazy enough to do what we proposed here, that is, refactoring all common jobs into common locations...

michelwi: Changing the location of existing jobs unfortunately changes is hash. Therefore, although i agree with your sentiment, I made the suggestion that is lower friction now..

lib/audio.py
Icemole: ~~I'm a bit hesitant because the `compute_rec_duration` should be called many times repeatedly, usually `num_recordings_in_corpus` calls, and the `import` overhead might be relatively significant then.~~

Interesting read [here](https://stackoverflow.com/a/32151422/5747616). It seems that importing inside a function doesn't impact the overall performance. However, importing a general namespace hurts performance a bit. I'll definitely take this into account :)

lib/audio.py
Icemole: I wouldn't want `ffmpeg` to be the default implementation because it usually takes a very long time to convert the audio file. Opening and reading the file's metadata will be much faster than converting to wav.

I can definitely add it as a fallback though :)

Icemole: Done! I also caught the situation where `ffmpeg` is not found in the system. Do we assume in general that `ffmpeg` is available?

lib/audio.py
```diff
@@ -0,0 +1,77 @@
...
+    Computes the duration of a given recording in seconds.
+
+    :param str rec_audio: Path of the recording.
+        The accepted formats are mp3, wav, aac or any audio format parseable by soundfile.
+    """
+    if rec_audio.endswith("mp3"):
+        try:
+            from mutagen.mp3 import MP3
+
+            audio_length = MP3(rec_audio).info.length
```
michelwi, line 45: I think we can return this directly
```suggestion
            return MP3(rec_audio).info.length
```

lib/audio.py
```diff
@@ -0,0 +1,77 @@
...
+            except FileNotFoundError as exception:
+                # ffmpeg doesn't exist either.
+                exception.msg = (
+                    f"mutagen python module not found, required to calculate duration of mp3 file.\n"
+                    "ffmpeg binary not found either, so mp3 -> wav conversion can't succeed.\n"
+                    "Please either install the mutagen python module or the ffmpeg binary.\n"
+                    f"Can't read duration of mp3 file: {rec_audio}."
+                )
+                raise exception
+        return audio_length
```
michelwi, line 62: not needed if suggestion above is accepted
```suggestion
```

corpus/stats.py
```diff
@@ -99,3 +102,48 @@ def run(self):
...
+                c = libcorpus.Corpus()
+                c.load(corpus_file.get_path())
+
+                for r in c.all_recordings():
+                    f.write(f"{r.audio}\n")
+                    if self.dump_durations:
+                        f_dur.write(f"{r.audio}\t{compute_rec_duration(r.audio)}\n")
+
+        if self.dump_durations:
+            f_dur.close()
```
michelwi, line 149: > Manually open/close file

ah.. I don't find that elegant. let me instead suggest
```suggestion
        with uopen(self.out_audio_list.get_path(), "wt") as f, uopen(self.out_audio_durations.get_path(), "wt") if self.dump_durations else nullcontext() as f_dur:
            for corpus_file in self.corpus_files:
                c = libcorpus.Corpus()
                c.load(corpus_file.get_path())

                for r in c.all_recordings():
                    f.write(f"{r.audio}\n")
                    if self.dump_durations:
                        f_dur.write(f"{r.audio}\t{compute_rec_duration(r.audio)}\n")
```
add `from contextlib import nullcontext` plus let black do its magic.

(kudos to [stackoverflow](https://stackoverflow.com/questions/27803059/conditional-with-statement-in-python))

Icemole: That's wonderful! Thank you for the suggestion! :star_struck:

Icemole:
@michelwi there's an issue with the option you proposed. It doesn't work for python 3.8, which is what's being run in the repo. Take a look at the following example:
```
# test.py
with (
        open("f", "r") as f,
        open("f", "r") if True else open("f", "r") as f1
):
        pass
```
And in the command line:
```
$ touch f
$ pyenv local system  # 3.10
$ python3 test.py
### Program ended successfully
$ pyenv local 3.8
$ python3 test.py
  File "test.py", line 2
    open("f", "r") as f,
                     ^
SyntaxError: invalid syntax
```
This is why the pipeline's breaking. It would seem that the `with` expression was bugged (or "functionality worked as intended") in version 3.8 since it doesn't admit any newlines. For more reading, see [this insightful stack overflow post](https://stackoverflow.com/q/68924790/5747616) and ["What's new in Python 3.10"](https://docs.python.org/3.10/whatsnew/3.10.html#parenthesized-context-managers).

I'll try to find a workaround.

lib/audio.py
```diff
@@ -0,0 +1,76 @@
...
+                    f"mutagen python module not found, required to calculate duration of mp3 file.\n"
+                    "ffmpeg binary not found either, so mp3 -> wav conversion can't succeed.\n"
+                    "Please either install the mutagen python module or the ffmpeg binary.\n"
+                    f"Can't read duration of mp3 file: {rec_audio}."
+                )
+                raise exception
+    elif rec_audio.endswith("aac"):
+        # The aac format has unreliable timestamps. Convert to wav through ffmpeg for an accurate measurement.
+        try:
+            _compute_rec_duration_convert_to_wav(rec_audio)
```
michelwi, line 65:
```suggestion
           return  _compute_rec_duration_convert_to_wav(rec_audio)
```

michelwi (changes requested):
now return missing

Icemole:
I've now double-checked the code, and I think everything's correct :)

lib/audio.py
```diff
@@ -0,0 +1,74 @@
...
+
+def compute_rec_duration(audio_file: str) -> float:
+    """
+    Computes the duration of a given recording in seconds.
+
+    :param str audio_file: Path of the recording.
+        The accepted formats are mp3, wav, aac or any audio format parseable by soundfile.
+    """
+    if audio_file.endswith("mp3"):
+        try:
```
sarahberanek, line 42: what about .flac or .opus ?
I have encountered those in mtp work - not sure if this would happen in apptek resources.

corpus/stats.py
```diff
@@ -99,3 +103,44 @@ def run(self):
...
+        with uopen(self.out_audio_list.get_path(), "wt") as f, uopen(
+            self.out_audio_durations.get_path(), "wt"
+        ) if self.dump_durations else nullcontext() as f_dur:  # fmt: off
+            for corpus_file in self.corpus_files:
+                c = libcorpus.Corpus()
+                c.load(corpus_file.get_path())
+
+                for r in c.all_recordings():
+                    f.write(f"{r.audio}\n")
+                    if self.dump_durations:
```
hannah220, line 145: As I understood, you don't need `if self.dump_durations:`, because f_dur will be nullcontext in that case and you can "write" it anyways.

Icemole: I thought `nullcontext()` was to clean up further the code. Does `nullcontext()` also provide a "dummy" file handler?

lib/audio.py
Icemole: This could be added in the future whenever needed. I haven't tested these.

Do you have code to get the durations from those?

corpus/stats.py
Icemole:
```
>>> from contextlib import nullcontext
>>> with nullcontext() as f_dur:
...     f_dur.write("asdf")
...     
  File "<python-input-2>", line 2, in <module>
    f_dur.write("asdf")
    ^^^^^^^^^^^
AttributeError: 'NoneType' object has no attribute 'write'
>>> 
```
`nullcontext()` seems to map to `None`. This is only added for cleaner code, but it doesn't map to a "dummy"/empty file handler.

hannah220 (approved):

lib/audio.py
sarahberanek: I just checked I even encountered mp4 and m4a formats.
I used ffmpeg to convert them. Not sure about duration, but I bet then your normal code here would work too.

Icemole: Maybe you can create another PR with the changes?

Do you always use `ffmpeg`? In many cases we can do "better". For example, as far as I know, the `mutagen` module reads the duration from the MP3 header/metadata.

PR #581, opened 2025-02-06
Improve memory footprint of ComputeNgramLmPerplexityJob .get_ppl

michelwi:
The current implementation constructs a list of the entire files content in memory just to (almost) discard it immediately.

We found that this leads to high memory usage and can slow down the cluster head node (as this is a mini_task). Using a deque instead reduces the memory footprint drastically.

### DIFF
--- a/lm/srilm.py
+++ b/lm/srilm.py
@@ -12,6 +12,7 @@ import os
 import shutil
 import subprocess
 
+from collections import deque
 from enum import Enum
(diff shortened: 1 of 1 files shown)


DanEnergetics (approved):

jtrmal:
One of the weirdest uses of dequeueue I have ever seen but LGTM

jtrmal (approved):
LGTM

PR #584, opened 2025-02-21
Add `reformat_orth` parameter to `Corpus.load()`

Icemole:
Counter-proposal to https://github.com/rwth-i6/i6_core/pull/560.

Allow the user to set whether the "nicer Bliss corpus formatting" should be applied or not. By default it's applied so that the default behavior is not overridden. I still believe that the user should have the capability to control this behavior, hence the reason for this PR.

Many less reviewers this time since it's a much less breaking change :)

### DIFF
--- a/lib/corpus.py
+++ b/lib/corpus.py
@@ -42,13 +42,24 @@ class CorpusParser(sax.handler.ContentHandler):
     is currently beeing read.
     """
 
-    def __init__(self, corpus: Corpus, path: str):
+    def __init__(self, corpus: Corpus, path: str, pretty_format_orth: bool = True):
+        """
+        :param corpus: Corpus to be parsed.
+        :param path: Path of the parent corpus (needed for include statements).
+        :param pretty_format_orth: Whether to do some processing of the text
+            that goes into the orth tag to get a nicer formating.
+            If `True`, removes multiline content in the orth tag, leading/trailing spaces,
+            and multiple spaces inside the text.
+
+            Defaults to `True`.
+        """
         super().__init__()
 
         self.elements: List[NamedEntity] = [
             corpus
-        ]  # stack of objects to store the element of the corpus that is beeing read
-        self.path = path  # path of the parent corpus (needed for include statements)
+        ]  # stack of objects to store the element of the corpus that is being read
+        self.path = path
+        self.pretty_format_orth = pretty_format_orth
         self.chars = ""  # buffer for character events, it is reset whenever a new element starts
 
     def startElement(self, name: str, attrs: Dict[str, str]):
@@ -118,12 +129,14 @@ class CorpusParser(sax.handler.ContentHandler):
 
         if name in {"orth", "left-context-orth", "right-context-orth"}:
             assert isinstance(e, Segment)
-            # we do some processing of the text that goes into the orth tag to get a nicer formating, some corpora may have
-            # multiline content in the orth tag, but to keep it that way might not be consistent with the indentation during
-            # writing, thus we remove multiple spaces and newlines
-            text = self.chars.strip()
-            text = re.sub(" +", " ", text)
-            text = re.sub("\n", "", text)
+            if self.pretty_format_orth:
+                # we do some processing of the text that goes into the orth tag to get a nicer formating,
+                # some corpora may have multiline content in the orth tag,
+                # but to keep it that way might not be consistent with the indentation during writing,
+                # thus we remove multiple spaces and newlines
+                text = self.chars.strip()
+                text = re.sub(" +", " ", text)
+                text = re.sub("\n", "", text)
             setattr(e, name.replace("-", "_"), text)
         elif isinstance(e, Speaker) and name != "speaker-description":
             # we allow all sorts of elements within a speaker description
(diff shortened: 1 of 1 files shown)


lib/corpus.py
```diff
@@ -42,13 +42,24 @@ class CorpusParser(sax.handler.ContentHandler):
     is currently beeing read.
     """
 
-    def __init__(self, corpus: Corpus, path: str):
+    def __init__(self, corpus: Corpus, path: str, pretty_format_orth: bool = True):
```
michelwi, line 45: I would call it strip_orth to be more descriptive.. (the prettiness is arguable).

I request changes as to the default should preserve old behavior and therefore be false.

michelwi (changes requested):

lib/corpus.py
albertz, line 45:
```suggestion
    def __init__(self, corpus: Corpus, path: str, *, pretty_format_orth: bool = True):
```

lib/corpus.py
```diff
@@ -266,14 +280,20 @@ def filter_segments(self, filter_function: FilterFunction):
         for sc in self.subcorpora:
             sc.filter_segments(filter_function)
 
-    def load(self, path: str):
+    def load(self, path: str, pretty_format_orth: bool = True):
```
albertz, line 283:
```suggestion
    def load(self, path: str, *, pretty_format_orth: bool = True):
```

lib/corpus.py
albertz: I agree on using a better name.

But the default would be True? The current code does this stripping.

michelwi: > But the default would be True? The current code does this stripping.

Sorry, right.. I missed that and only saw the insertions.

curufinwe: My proposal for the name of the flag would be `reformat_orth`. Functionality itself is OK I guess.

Icemole:
Thank you everyone for your feedback!

PS: I've eventually named the variable `reformat_orth`, I hope @michelwi is fine with this decision.

michelwi (approved):

Icemole:
I noticed that spaces in `<orth> %s </orth>` kept accumulating, so I kept the `strip()` as a minimal formatting to be done always.

Icemole:
Yes, I definitely want to use it, but I'm still debating on the implementation.

For context, this example is one of many that we have available with newlines:
```
        <orth>   I've been happy, I've been cheerful,
my emotions are flying all over the place.   </orth>
```

Note that the newline acts as a space, while there's no extra space. We could then replace `"\n"` with `" \n"` in our corpus parsing, but we'd ideally like to be as faithful as possible to the actual raw corpus. Then again, the most sane approach is to replace `"\n"` with `" \n"` in our use case and lose a bit of rawness instead of having to implement a `replace_orth` in all corpus loading jobs... It feels overkill, and I'm already suffering the consequences trying to catch all `load()` that we have :cry: 

Let me discuss this with the team in case we have some more use cases and come back to the PR in a few hours. Let's not merge this for now.

albertz:
> ... in our use case ...

But that's exactly the point I was raising here: You have a specific use case in mind, but this PR here is not exactly what you actually prefer for this use case? So why to implement a feature which is only used by you for now, but is not really what you actually want?

Icemole:
I've discussed it with the team, and we'd like to keep the functionality if it's fine with you. We agreed that we would only implement it whenever needed, so in principle it would only be used for limited use cases, but we're losing potentially relevant information that could be useful to any other team member by ignoring newlines.

As for the PR itself, I'm debugging an issue where a normal orth loaded as `<orth> bla bla </orth>` is dumped with three spaces as `<orth>   bla bla   </orth>`. It shouldn't affect the final `orth` since I now unconditionally use the `strip()`, but it's bothering me because I feel like I've missed something. Let me check this.

Icemole:
> >  ... in our use case ...
> 
> But that's exactly the point I was raising here: You have a specific use case in mind, but this PR here is not exactly what you actually prefer for this use case? So why to implement a feature which is only used by you for now, but is not really what you actually want?

I just saw your comment, it slipped through. I don't disagree with your statements, and I'm not against other alternatives, but I just want to get this done in any possible way, and I see that there are many alternatives that will all bring us to the outcome of having no formatting applied.

Before implementing this, I thought about other alternatives, like:
1. Creating a `CorpusNoFormat` (example name) class that doesn't apply any reformatting at all, while also removing ambiguous spaces when dumping the final xml file (`<orth> %s </orth>` instead of `<orth>%s</orth>`) so that the content in the xml file truly reflects what the corpus wants it to reflect.
2. Implement our own corpus version and not pollute this repository's. If you think that will make it better, I'm also up for it, considering that nobody else has complained about this for all these years.

However, I believe that the alternative I'm providing in this PR is the easiest one that's able to be integrated into the `Corpus` class. And it's a nice solution to our issues, just like any other that I proposed above.

albertz:
I'm not exactly sure I understand the problem and use case anyway.

To clarify:

(1) If we correctly follow [the XML spec](https://www.w3.org/TR/xml/#sec-line-ends), it means the newline should be preserved, right?

(2) RASR parses newline as word separator, just like other whitespace. I assume it also collapses multiple whitespace into a single word separator (there are no empty words anyway)? It's not really inconsistent to the XML spec to handle both newline and other whitespaces as word separator.

(3) Here in i6_core, we also collapse, but newlines are not converted into whitespace, but instead they are simply removed. So this indeed is not correct.

Some followup questions:

Do we have any case where we would want to differentiate between whitespace and newline?

How do we get the newline in there in the first place? Does it belong there or not (specifically for the case that we want to differ between newline and whitespace; but maybe we never want that)? If it sometimes does not belong there, but still is formatted that way, that seems to be a bug, and that should be fixed?

Note how Python `str.split` behaves:
```python
In [1]: "hello\nworld".split()
Out[1]: ['hello', 'world']

In [2]: "hello \n world".split()
Out[2]: ['hello', 'world']
```
So I think the `reformat_orth` option makes sense. At least Python `str.split` behaves as RASR then.

Issue #585, opened 2025-02-26
Testing the existence of a file in a wrong directory


jtrmal:
https://github.com/rwth-i6/i6_core/blob/25db6956ba1ab67ca947d84cc96017694c269c65/text/processing.py#L380

I believe the string needs to contain `{tmp_dir}`, i.e.
```
file_path = f"{tmp_dir}/split.{file_id:04}.txt"
```

I tested it just now and the SplitTextFileJob  works (was crashing on the assert before)

michelwi:
Yes. pls submit your fix as PR.

albertz:
Fixed via #586

PR #586, opened 2025-02-26
use proper directory when checking if file exists

jtrmal:
Should address #585

### DIFF
--- a/text/processing.py
+++ b/text/processing.py
@@ -377,7 +377,7 @@ class SplitTextFileJob(Job):
             subprocess.run(split_cmd, check=True)
 
             for file_id in range(1, self.num_output_files + 1):
-                file_path = f"split.{file_id:04}.txt"
+                file_path = f"{tmp_dir}/split.{file_id:04}.txt"
                 assert os.path.isfile(file_path) and os.path.getsize(file_path) > 0
 
             if self.zip_output:


michelwi (approved):

jtrmal:
Not sure what to do about the "apptek hashes" test

michelwi (changes requested):
There are several issues with the recent update:
1. When we add new parameters to a Job, then there are new inputs which affects the hash of all existing instances of the Job. To remedy this the new parameter should have a default value that preserves the old behavior (this is only sort of the case) and the Job should get a `__sis_hash_exclude__` like so:
```
class SplitTextFileJob(Job):
    __sis_hash_exclude__ = {"max_output_files": None}
```
2. Docstring for the new parameter is missing
3. The behavior feels a bit strange to me, that when we want to specify the number of files to get, we would first have to provide some guesses or arbitrarily large numbers for `num_lines_per_split` and `num_text_file_lines` (e.g. 1 line per split and 1M lines) to just then limit the number of files by `max_output_files`. I would propose to instead make all config options Optional, then check if what was provided is sufficient to run the Job and if it is not contradictory. And then run with what you got.

Lets remind us on the implementation in case we have `num_text_file_lines=10`, `num_lines_per_split=4`, and `max_output_files=3`. 
In case we use the `-l` flag we get the file sizes 4, 4, and 2
In case we use the `-o` flag we get the file sizes 3, 4, and 3

which is different. Which is a problem if we wanted to reproduce old setups. So (at least) if the new parameter is not provided, then the old `-l` flag should still be used.
Also, always using the number of output files makes it harder to implement a case where it is unknown.

michelwi:
> Not sure what to do about the "apptek hashes" test

The pipeline requires some API token that is stored in a repository secret. This PR is made from a fork which does not know the secret. The only way to have the test pass is to create a branch in the rwth-i6/i6_core repo.
But I can also manually trigger a pipeline. and write the result here.

And I can tell with certainty, that the first version would have passed (since only changes in the run function were made, which cannot change hashes) and the second version would fail (because of missing `__sis_hash_exclude__`)

jtrmal:
I guess I can revert the new changes, if you want, because I fully understand that depending on the point of view the changes muddle the semantics of the class/job

michelwi:
If you want in a fix quickly than we can merge the first commit fast.

Otherwise or additionally I'd be happy working with you in bringing the updated functionality into a mergable state. :)

jtrmal:
Ad 3) thats correct, unfortunately I have statistics (number of lines) gathered before a stage that removes some lines from the script. Which is a logical error in the script, I do realize that...   :/ 
Maybe it would be more useful to implement automatic determination of the number of lines (the "NotImplemented" exception on line 346) -- that would deal with my point a)
I would still like to be  able to specify the max number of splits, I don't see an alternative effective solution to respect limits from slurm than providing in from the top caller

michelwi:
>  implement automatic determination of the number of lines

the problem is less determining the number of lines at runtime, but figuring out how many outputs there will be before the inputs are ready. I.e. if your script (I assume sisyphus config) happened to remove one more line from the input to reduce the number of outputs, this can only be knowable once the Job has been executed and that would then change the computational graph.

So in that case we need to define the output independent of the number of files generated. E.g. one single folder is the output, this folder contains an arbitrary but fixed number of files. But I hazard that the remainder of the pipeline does not like not knowing the number of output files, so a full redesign would be in order... :/

michelwi (approved):

albertz (approved):

PR #588, opened 2025-03-06
add __repr__ for corpus objects

vieting:
Recently, I was playing around with segmentations and I found it useful to have a better representation of the corpus objects (mainly `Segment`, but also `Recording`) when printing them for testing. It's only a small change, but maybe someone else could also like it in the future.

### DIFF
--- a/lib/corpus.py
+++ b/lib/corpus.py
@@ -463,7 +463,7 @@ class Segment(NamedEntity):
             out.write("</segment>\n")
 
     def __repr__(self):
-        return f"{self.__class__.__name__}({self.fullname()}:{self.start:.02f}-{self.end:.02f})"
+        return f"<{self.__class__.__name__} {self.fullname()} {self.start:.02f}-{self.end:.02f})>"
 
 
 class Speaker(NamedEntity):
--- a/lib/corpus.py
+++ b/lib/corpus.py
@@ -26,7 +26,10 @@ class NamedEntity:
         self.name: Optional[str] = None
 
     def __repr__(self):
-        return f"{self.__class__.__name__}({self.fullname()})"
+        if self.name is None:
+            return f"<{self.__class__.__name__}>"
+        else:
+            return f"<{self.__class__.__name__} {self.name}>"
 
 
 class CorpusSection:
@@ -344,6 +347,9 @@ class Corpus(NamedEntity, CorpusSection):
         """
         return {rec.fullname(): rec for rec in self.all_recordings()}
 
+    def __repr__(self):
+        return f"<{self.__class__.__name__} {self.fullname()}>"
+
 
 class Recording(NamedEntity, CorpusSection):
     def __init__(self):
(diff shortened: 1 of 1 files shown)


lib/corpus.py
```diff
@@ -25,6 +25,9 @@ def __init__(self):
         super().__init__()
         self.name: Optional[str] = None
 
+    def __repr__(self):
+        return f"{self.__class__.__name__}({self.fullname()})"
```
albertz, line 29: I don't understand. There is no `fullname` method here? Maybe you want to add that as an abstract base method?

albertz, line 29: There is the convention that either what you return by `__repr__` is an `eval`-able string, i.e. a valid Python expression which would result in the object. If the format is like `${classname}(${args})`, this is the case. However, even though it looks like valid Python, it's not correct here.

Or, if you don't intend to return a valid Python expression, then the convention is to print it deliberately not using valid Python syntax, e.g. like `<${classname} ${args}>`.

lib/corpus.py
```diff
@@ -459,6 +462,9 @@ def dump(self, out: TextIO, indentation: str = ""):
         else:
             out.write("</segment>\n")
 
+    def __repr__(self):
+        return f"{self.__class__.__name__}({self.fullname()}:{self.start:.02f}-{self.end:.02f})"
```
albertz, line 466:
```suggestion
        return f"<{self.__class__.__name__} {self.fullname()}:{self.start:.02f}-{self.end:.02f}>"
```

michelwi: Is the colon character disallowed in segment names? Otherwise or even then the notation might lead to confusion as to what is part of the name and what are the timestamps.

```suggestion
        return f"<{self.__class__.__name__} {self.fullname()} {self.start:.02f}-{self.end:.02f})>"
```
or maybe even 
```suggestion
        return f"<{self.__class__.__name__} {self.fullname()} start={self.start:.02f} end={self.end:.02f})>"
```

lib/corpus.py
michelwi: The class `Speaker` inherits from `NamedEntity` and does not have a fullname - implementation wise and also conceptually. 

> add that as an abstract base method?

or even a concrete method that returns `self.name or ""` and then all Corpus/Subcorpus/Recordings/Segments already have their own implementations that give the full name.

vieting: Right, I missed that. So should we just do `self.name or ""` here? I could then also add `__repr__` for `Corpus` and `Recording` using `fullname` there.

vieting: I see, thanks for the hint! I'll wait for a conclusion in the other threads to resolve this.

lib/corpus.py
vieting: Thanks for the suggestions. I applied the one with spaces since I like that most. Feel free to comment if you disagree.

lib/corpus.py
michelwi: I have no preference as long as it is correct in all cases :)

albertz (approved):

PR #589, opened 2025-03-12
Add `reformat_orth` to `MapRecordingsJob`

Icemole:
I'd need to access `reformat_orth` when loading a corpus via `MapRecordingsJob`.

I hope I don't need it anymore, but I can't guarantee that we'll need it in the future. Maybe we can use this PR as a base for future courses of action when developing new jobs, or needing the `reformat_orth` in old jobs.

### DIFF
--- a/corpus/transform.py
+++ b/corpus/transform.py
@@ -640,13 +640,27 @@ class MapRecordingsJob(Job):
     Applies a function to all recordings of a given corpus.
     """
 
-    def __init__(self, bliss_corpus: tk.Path, recording_callable: Callable[[corpus.Recording], corpus.Recording]):
+    __sis_hash_exclude__ = {"reformat_orth": True}
+
+    def __init__(
+        self,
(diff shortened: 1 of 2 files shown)


corpus/transform.py
```diff
@@ -640,13 +640,25 @@ class MapRecordingsJob(Job):
...
     """
 
-    def __init__(self, bliss_corpus: tk.Path, recording_callable: Callable[[corpus.Recording], corpus.Recording]):
+    __sis_hash_exclude__ = {"corpus_load_reformat_orth": True}
+
+    def __init__(
+        self,
+        bliss_corpus: tk.Path,
+        recording_callable: Callable[[corpus.Recording], corpus.Recording],
+        corpus_load_reformat_orth: bool = True,
```
albertz, line 649:
```suggestion
        *,
        corpus_load_reformat_orth: bool = True,
```

michelwi (approved):

PR #590, opened 2025-03-20
Fix an issue in ConcatenateJob with command line too long if too many files

jtrmal:
if there is too many files ("too many" depends on the lenght of the paths and name itself), the original command fails as there is build-in limit on the length of program arguments in bas (IIRC around 13k bytes).
This change fixes this issue.

### DIFF
--- a/text/processing.py
+++ b/text/processing.py
@@ -141,9 +141,7 @@ class ConcatenateJob(Job):
         assert isinstance(text_files, list)
 
         for input_file in text_files:
-            assert isinstance(input_file, Path) or isinstance(
-                input_file, str
-            ), "input to Concatenate is not a valid path"
+            assert isinstance(input_file, (Path, str)), "input to Concatenate is not a valid path"
 
         self.text_files = text_files
         self.zip_out = zip_out
@@ -168,7 +166,7 @@ class ConcatenateJob(Job):
 
         with util.uopen(self.out, "wb") as out_file:
             for f in self.f_list:
-                print(f)
+                logging.info(f)
                 with util.uopen(f, "rb") as in_file:
                     shutil.copyfileobj(in_file, out_file)
 
--- a/text/processing.py
+++ b/text/processing.py
@@ -159,13 +159,13 @@ class ConcatenateJob(Job):
         yield Task("run", rqmt={"mem": 3, "time": 3})
 
     def run(self):
-        self.f_list = [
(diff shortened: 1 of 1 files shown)


text/processing.py
```diff
@@ -159,14 +161,16 @@ def tasks(self):
         yield Task("run", rqmt={"mem": 3, "time": 3})
 
     def run(self):
-        self.f_list = " ".join(
+        self.f_list = [
```
jtrmal, line 164: I now see an issue in this -- the variable is originally different (string instead of lists).
I can change that at an expense of extra memory. But do we even need that variable?

text/processing.py
```diff
@@ -140,8 +140,10 @@ def __init__(self, text_files: List[Path], zip_out: bool = True, out_name: str =
 
         assert isinstance(text_files, list)
 
-        for input in text_files:
-            assert isinstance(input, Path) or isinstance(input, str), "input to Concatenate is not a valid path"
+        for input_file in text_files:
```
jtrmal, line 143: This was just to fix complaints about shadowing function input()

text/processing.py
```diff
@@ -140,8 +140,10 @@ def __init__(self, text_files: List[Path], zip_out: bool = True, out_name: str =
...
+            assert isinstance(input_file, Path) or isinstance(
+                input_file, str
+            ), "input to Concatenate is not a valid path"
```
curufinwe, line 146:
```suggestion
            assert isinstance(input_file, (Path, str)), "input to Concatenate is not a valid path"
```

text/processing.py
curufinwe: I would argue we don't need it as an instanciated list or string. It can also be a generator expression that we use in the loop below.

text/processing.py
```diff
@@ -159,14 +161,16 @@ def tasks(self):
...
-        )
-        if self.zip_out:
-            self.sh("zcat -f {f_list} | gzip > {out}")
-        else:
-            self.sh("zcat -f {f_list} > {out}")
+        ]
+
+        with util.uopen(self.out, "wb") as out_file:
+            for f in self.f_list:
+                print(f)
```
curufinwe, line 171:
```suggestion
                logging.info(f)
```

curufinwe (changes requested):

curufinwe (approved):

PR #591, opened 2025-04-02
Extend FilterCorpusRemoveUnknownWordSegmentsJob with new parameters segment_oov_tolerance and recording_oov_tolerance

kuacakuaca:
Extend FilterCorpusRemoveUnknownWordSegmentsJob to allow setting a OOV rate tolerance value for both segments and recordings

### DIFF
--- a/corpus/filter.py
+++ b/corpus/filter.py
@@ -269,7 +269,12 @@ class FilterCorpusRemoveUnknownWordSegmentsJob(Job):
     Filter segments of a bliss corpus if there are unknowns with respect to a given lexicon
     """
 
-    __sis_hash_exclude__ = {"all_unknown": True, "delete_empty_recordings": False}
+    __sis_hash_exclude__ = {
+        "all_unknown": True,
+        "delete_empty_recordings": False,
+        "segment_oov_tolerance": 0.0,
+        "recording_oov_tolerance": 1.0,
+    }
 
     def __init__(
         self,
@@ -278,6 +283,8 @@ class FilterCorpusRemoveUnknownWordSegmentsJob(Job):
         case_sensitive: bool = False,
         all_unknown: bool = True,
         delete_empty_recordings: bool = False,
+        segment_oov_tolerance: float = 0.0,
+        recording_oov_tolerance: float = 1.0,
     ):
         """
         :param bliss_corpus:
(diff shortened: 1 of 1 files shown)


corpus/filter.py
```diff
@@ -278,19 +283,25 @@ def __init__(
         case_sensitive: bool = False,
         all_unknown: bool = True,
         delete_empty_recordings: bool = False,
+        segment_oov_tolerance: float = 0.0,
```
michelwi, line 286: I am thinking about how `all_unknown` should interact with `segment_oov_tolerance` because they control the same aspect with `all_unknown=True -> segment_oov_tolerance=0.99999` (discard only if all words are unknown) and `all_unknown=False-> segment_oov_tolerance=0.0` (discard if already a single word is unknown).

Because in the current implementation someone could use 
```
my_job = FilterCorpusRemoveUnknownWordSegmentsJob(some_corpus, segment_oov_tolerance=0.5)
```
and the tolecance would not be respected, because `all_unknown` defaults to `True` and is stronger than the tolerance.

### Suggestion 1:
just keep the `all_unknown` parameter and elevate the typing to `float` and use it instead of `segment_oov_tolerance`.. (Note, that booleans are also ints (T/F == 1/0) and that ints are also floats.. i.e. `1.2 * True == 1.2` is valid)
```
return num_oov_words <= len(words) * self.all_unknown
```
We would have to be careful that the corner cases of `True == 1.0` and `False == 0.0` still work correctly.

### Suggestion 2:
change the typing of all_unknown and segment_oov_tolerance to Optional and default to `None`. Internally we only keep one floating point parameter and we make sure that the value is sane.
```
assert all_unknown is None or segment_oov_tolerance is None, "do not specify all_unknown and segment_oov_tolerance at the same time"
if segment_oov_tolerance is not None:
    self segment_oov_tolerance = segment_oov_tolerance
elif all_unknown is None or all_unknown:  # nothing specified or True
    self segment_oov_tolerance = 0.999999
else:
    self segment_oov_tolerance = 0.0
```

kuacakuaca: thanks for reviewing! I prefer the second suggestion. Just one question, 0.999999 is only safe if the number of words is not larger than 10**6 which is likely to hold, do you want to add an assertion to ensure that?

corpus/filter.py
```diff
@@ -373,6 +375,15 @@ def unknown_filter(corpus: corpus.Corpus, recording: corpus.Recording, segment:
 
         c.dump(self.out_corpus.get_path())
 
+    @classmethod
+    def hash(cls, kwargs):
+        kwargs_copy = dict(**kwargs)
+
+        if kwargs_copy["all_unknown"] is True:
+            kwargs_copy["all_unknown"] = None
```
michelwi, line 383: how about
```suggestion
        if "all_unknown" in kwargs_copy and kwargs_copy["all_unknown"] is True:
            del kwargs_copy["all_unknown"]
```
?

corpus/filter.py
```diff
@@ -330,19 +347,43 @@ def unknown_filter(corpus: corpus.Corpus, recording: corpus.Recording, segment:
...
         c.filter_segments(unknown_filter)
 
         if self.delete_empty_recordings:
             # Remove the recordings without segments due to the filtering.
             _delete_empty_recordings(c, self.out_removed_recordings.get_path())
 
+        if self.recording_oov_threshold < 1.0:
+            recordings_to_be_removed = []
+            for r in c.all_recordings():
+                num_seg = num_segments_per_recording[r.fullname()]
```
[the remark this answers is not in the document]
michelwi: I think the order is wrong
```suggestion
            for r in c.all_recordings():
                num_seg = num_segments_per_recording[r.fullname()]
                if num_seg == 0:
                    continue
```
otherwise `num_seg` is not defined yet, right?

curufinwe (approved):

corpus/filter.py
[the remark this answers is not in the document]
kuacakuaca: yes, so I didn't take it.

michelwi (approved):

PR #592, opened 2025-04-17
Extend alignment_flow with left/right context orth

hannah220:


mm/flow.py
```diff
@@ -37,11 +37,14 @@ def linear_segmentation_flow(feature_energy_net, alignment_cache=None):
...
-def alignment_flow(feature_net, alignment_cache_path=None):
+def alignment_flow(feature_net, alignment_cache_path=None, add_left_right_context_orth=False):
     assert "features" in feature_net.get_output_ports()
     net = FlowNetwork()
     net.add_output("alignments")
-    net.add_param(["id", "orthography", "TASK"])
+    if add_left_right_context_orth:
+        net.add_param(["id", "orthography", "left-context-orthography", "right-context-orthography", "TASK"])
+    else:
+        net.add_param(["id", "orthography", "TASK"])
```
michelwi, line 47: does the order matter? Otherwise I would suggest to factor out:
```suggestion
    net.add_param(["id", "orthography", "TASK"])
    if add_left_right_context_orth:
        net.add_param(["left-context-orthography", "right-context-orthography"])
```

mm/flow.py
```diff
@@ -50,11 +53,26 @@ def alignment_flow(feature_net, alignment_cache_path=None):
     aggregate = net.add_node("generic-aggregation-vector-f32", "aggregate")
     net.link(mapping[feature_net.get_output_links("features").pop()], aggregate)
 
-    alignment = net.add_node(
-        "speech-alignment",
-        "alignment",
-        {"id": "$(id)", "orthography": "$(orthography)"},
-    )
+    if add_left_right_context_orth:
```
michelwi, line 56: similarly to the factoring out above one could always add the base form of the alignment node and then if add_left_right_context_orth update the node.. but I don't know from the top of my head how to do it.

Probably something like `net.nodes[alignment].update({"left-context-orthography": "$(left-context-orthography)","right-context-orthography": "$(right-context-orthography)"})`

michelwi (commented):
I think the code is correct and works as intenden.. just some (optional) stylistic suggestions.

curufinwe:
I have nothing to add to what @michelwi commented on.

Icemole (approved):

mm/flow.py
```diff
@@ -53,8 +55,18 @@ def alignment_flow(feature_net, alignment_cache_path=None):
...
         "speech-alignment",
         "alignment",
-        {"id": "$(id)", "orthography": "$(orthography)"},
+        {
+            "id": "$(id)",
+            "orthography": "$(orthography)",
+        },
     )
+    if add_left_right_context_orth:
+        net.nodes["alignment"].update(
```
curufinwe, line 64:
```suggestion
        net.nodes[alignment].update(
```
`add_node` might rename your node if a node with the same name already exists. Thus you should always use the name returned by `add_node` when modifying the node afterwards.

curufinwe (changes requested):

curufinwe (approved):

Icemole:
Functionality works just fine!

Icemole:
I believe the feedback proposed by @michelwi has also been addressed. Should we merge?

PR #593, opened 2025-04-23
`BlissToPcmHDFJob`: allow writing compressed audio to HDF

NeoLegends:
HDF is a better container format than .ogg.zip, but currently the job only allows storing uncompressed audio. When the datasets get large, compressed audio is easier to handle.

### DIFF
--- a/returnn/hdf.py
+++ b/returnn/hdf.py
@@ -4,6 +4,7 @@ from dataclasses import dataclass
 from enum import Enum, auto
 import glob
 import math
+import io
 import librosa
 import numpy as np
 import os
@@ -209,7 +210,9 @@ class ReturnnRasrDumpHDFJob(ReturnnDumpHDFJob):
 class BlissToPcmHDFJob(Job):
     """
     Gets audio files from a Bliss corpus and stores them as HDF file
-    compatible with the RETURNN HDFDataset
+    compatible with the RETURNN HDFDataset.
+
+    Contrary to the job name, can also write compressed audio to the HDF to save space.
     """
 
     class BaseStrategy:
@@ -232,6 +235,7 @@ class BlissToPcmHDFJob(Job):
         "rounding": RoundingScheme.start_and_duration,
         "round_factor": 1,
         "target_sampling_rate": None,
+        "compress_format": None,
     }
 
     def __init__(
@@ -244,6 +248,7 @@ class BlissToPcmHDFJob(Job):
         rounding: RoundingScheme = RoundingScheme.start_and_duration,
         round_factor: int = 1,
         target_sampling_rate: Optional[int] = None,
+        compress_format: Optional[str] = None,
     ):
         """
 
@@ -260,9 +265,12 @@ class BlissToPcmHDFJob(Job):
             rasr_compatible will round up the start time and round down the end time
         :param round_factor: do the rounding based on a sampling rate that is scaled down by this factor
         :param target_sampling_rate: desired sampling rate for the HDF, data will be resampled to this rate if needed
+        :param compress_format: If set, do not write PCM data but compressed audio instead.
+            This is a string that is passed to soundfile as the format argument.
+            The audio can be decompressed downstream via e.g. soundfile
         """
         self.set_vis_name("Dump audio to HDF")
-        assert output_dtype in ["float64", "float32", "int32", "int16"]
+        assert output_dtype in ("float64", "float32", "int32", "int16")
 
         self.bliss_corpus = bliss_corpus
         self.segment_file = segment_file
@@ -272,6 +280,7 @@ class BlissToPcmHDFJob(Job):
         self.rounding = rounding
         self.round_factor = round_factor
         self.target_sampling_rate = target_sampling_rate
+        self.compress_format = compress_format
 
         self.out_hdf = self.output_path("audio.hdf")
 
@@ -333,6 +342,13 @@ class BlissToPcmHDFJob(Job):
                         target_sr=sr,
                         axis=0,
                     ).astype(self.output_dtype)
+
+                # compress if necessary
+                if self.compress_format is not None:
+                    out_sr = self.target_sampling_rate if self.target_sampling_rate else audio.samplerate
+                    buf = io.BytesIO()
+                    sf.write(buf, data, out_sr, format=self.compress_format)
+                    data = np.array([v for v in buf.getvalue()], dtype=np.uint8)
 
                 # add audio to hdf
                 out_hdf.insert_batch(


returnn/hdf.py
```diff
@@ -260,9 +265,12 @@ def __init__(
             rasr_compatible will round up the start time and round down the end time
         :param round_factor: do the rounding based on a sampling rate that is scaled down by this factor
         :param target_sampling_rate: desired sampling rate for the HDF, data will be resampled to this rate if needed
+        :param compress_format: If set, do not write PCM data but compressed audio instead.
+            This is a string that is passed to soundfile as the format argument.
+            The audio can be decompressed downstream via e.g. soundfile
```
albertz, line 270: Can you add some examples?

albertz:
I wonder, how do you handle it on RETURNN side? Via `PostprocessingDataset`?

(Note, I think that CuDNN (or some other CUDA lib?) actually provides some functions to do some decoding of popular formats on GPU. Maybe MP3. Not exactly sure. Also not sure what PyTorch exposes.)

albertz:
> HDF is a better container format than .ogg.zip

That's definitely true. But note that the way we use HDF also has also issues. See:
https://github.com/rwth-i6/returnn/issues/1716
https://github.com/rwth-i6/returnn/issues/1669

returnn/hdf.py
```diff
@@ -260,9 +265,12 @@ def __init__(
...
         """
         self.set_vis_name("Dump audio to HDF")
-        assert output_dtype in ["float64", "float32", "int32", "int16"]
+        assert output_dtype in ("float64", "float32", "int32", "int16")
```
michelwi, line 273: why tuple instead of list but not set?

returnn/hdf.py
```diff
@@ -334,6 +343,13 @@ def run(self):
                         axis=0,
                     ).astype(self.output_dtype)
 
+                # compress if necessary
+                if self.compress_format is not None:
+                    out_sr = self.target_sampling_rate if self.target_sampling_rate else audio.samplerate
+                    buf = io.BytesIO()
+                    sf.write(buf, data, out_sr, format=self.compress_format)
+                    data = np.array([v for v in buf.getvalue()], dtype=np.uint8)
```
michelwi, line 351: is it required to use `uint8` here?
Would this ignore the `output_dtype` defined above, or would it rather be e.g. float data compressed and written in raw uint8 numbers?

returnn/hdf.py
NeoLegends: I remember vaguely somewhere that Python can optimize the tuples a bit because they are not mutable, that's why I like using them in these cases.

returnn/hdf.py
NeoLegends: It is going to be the latter. After compression, the data is going to be bytes not matter what because it is stored in some sort of format (e.g. ogg). Making _that_ dtype configurable does not make much sense, I think. The output dtype still makes a difference because it limits the resolution of the audio samples before compression. After decompression you can again use any dtype you like, I don't think the compressed formats specify whether they store "floating point" or "int16" samples.

Does this answer your question?

michelwi: Yes, my question was answered, but then I went into a different direction researching what the correct dtype for byte arrays was (there is void, but I guess that is for byte-string-arrays?) and found `numpy.frombuffer` (would that be applicable and more efficient here?) and now I am wondering if this array of data is even meaningful in the context of the returnn hdf file format i.e. when we do 
```
                out_hdf.insert_batch(
                    inputs=data.reshape(1, -1, 1),
                    seq_len=[data.shape[0]],
                    seq_tag=[segment.fullname()],
                )
```
is the seq_len not completely wrong now? should we put the original seq_len?

returnn/hdf.py
albertz: With Python 3.9.7 on Mac:
```
In [4]: def a(s):
   ...:     return s in ["float64", "float32", "int32", "int16"]
   ...: 

In [5]: def b(s):
   ...:     return s in ("float64", "float32", "int32", "int16")
   ...: 

In [6]: def c(s):
   ...:     return s in {"float64", "float32", "int32", "int16"}
   ...: 

In [7]: %timeit a("int16")
70.1 ns ± 0.529 ns per loop (mean ± std. dev. of 7 runs, 10000000 loops each)

In [8]: %timeit b("int16")
68.1 ns ± 1.24 ns per loop (mean ± std. dev. of 7 runs, 10000000 loops each)

In [9]: %timeit c("int16")
46.7 ns ± 0.208 ns per loop (mean ± std. dev. of 7 runs, 10000000 loops each)

In [10]: %timeit a("complex32")
72 ns ± 0.496 ns per loop (mean ± std. dev. of 7 runs, 10000000 loops each)

In [11]: %timeit b("complex32")
69.8 ns ± 1.39 ns per loop (mean ± std. dev. of 7 runs, 10000000 loops each)

In [12]: %timeit c("complex32")
45.9 ns ± 0.446 ns per loop (mean ± std. dev. of 7 runs, 10000000 loops each)
```
Same with Python 3.13.0:
```
In [5]:  %timeit a("int16")
43.9 ns ± 0.349 ns per loop (mean ± std. dev. of 7 runs, 10,000,000 loops each)

In [6]: %timeit b("int16")
43.3 ns ± 0.0945 ns per loop (mean ± std. dev. of 7 runs, 10,000,000 loops each)

In [7]: %timeit c("int16")
30.9 ns ± 0.124 ns per loop (mean ± std. dev. of 7 runs, 10,000,000 loops each)
```
(I actually did not expect this. I was expecting, for such small sized containers, the overhead of a hashmap would dominate. I'm pretty sure that would be the case for C/C++.)

returnn/hdf.py
albertz: No, `seq_len` is correct. `seq_len` is if you insert a batch with shape [B,T,D], we want that `max(seq_len) = T`.

NeoLegends: > array of data is even meaningful in the context of the returnn hdf file format

To clarify: the data inside the HDF is mostly useless for training without a postprocessing step in between. This postprocessing step will go from "sequence of bytes that represents audio data in some arbitrary, possibly compressed container format" to "raw audio frames/PCM". As part of that, the seq length will go from "number of bytes that you need to decompress the audio in the container format" to the "correct" one for training, i.e. "number of raw PCM audio frames".

In the HDF the correct `seq_len` to use is "number of bytes that represent the audio in the given container format" because that number is needed to fetch the correct bytes from the HDF at runtime.

NeoLegends:
> I wonder, how do you handle it on RETURNN side? Via PostprocessingDataset?

Yes, that was the idea :)

> That's definitely true. But note that the way we use HDF also has some issues. See: [...]

Yeah, in these regimes I am going to use DistributeFilesDataset to reduce the number of HDFs I need to load at a time.

returnn/hdf.py
michelwi: ok, thanks for the detailed explanation. I think I now understand this implementation.

Now my only reservation is: "Isn't that a bit hacky and error prone? Should we not rather define a different dataset format that transparently handles this for the user?"

Imagine you have two data sources and only one of it is compressed... That will be a postprocessing function  / dataset  integration nightmare

NeoLegends: I think the required effort/overhead is the same if you have to use a different dataset type for one part of the data vs. having to wrap a part of the data in `PostprocessingDataset`. The only difference I could see is outfitting `HDFDataset` with transparent decompression functionality that it can apply entirely by itself by looking at e.g. an attribute in the HDF that says whether the data is compressed or raw audio.

NeoLegends: @albertz WDYT about a separate dataset implementation that uses the same disk format as the classic HDFDataset, but then transparently handles compressed audio by means of additional data keys that tell the dataset whether the data is audio and then whether it is compressed? I think importantly, this implementation would have/need to be backwards compatible (wrt. the disk format) so that existing HDF artefacts can be reused/used together without any special logic? Alternatively, the existing HDFDataset could be extended to support this use case. WDYT is the better option?

NeoLegends:
Hm, yeah. Good question. I now think that backwards compatibility/integration compatibility with existing data matters maybe a bit more than I thought it would. I think actually having the HDF data loading pipeline be transparent about whether the audio is compressed or not would be a net benefit to the user.

returnn/hdf.py
albertz: I would not do any of that and just use it as you originally intended with `PostprocessingDataset`. I don't really see the big advantage. It sounds more like overengineering, and producing further problems (more code, more complexity in RETURNN).

NeoLegends:
For my work on EVDS I'm now using a simplified (wrt. configurability) but more robust and much faster version of this job. 

That version might be a bit overkill for i6 use cases though, do we want to merge it instead of my changes here? I think it can be made just as flexible as the current i6_core version.

```python
class BlissToPcmHDFJob2(Job):
    """
    Gets audio files from a Bliss corpus and stores them as HDF file compatible with
    the RETURNN HDFDataset.

    Contrary to the job name, can also write compressed audio to the HDF to save
    space.

    More robust version of the `i6_core` version of the job, suitable for processing
    EVDS audio. This job is also faster in the case of a corpus split into many
    sub-corpora because it will parallelize the computation. To reduce disk pressure
    it will still load every segment just once.

    On EVDS this job manages an xRTF of about 5000 using an rqmt of 16 cores.

    Only supports RASR-specific feature rounding and the sum strategy (i.e. just mix down
    all channels into one) for multi-channel audio.
    """

    def __init__(
        self,
        bliss_corpus: Path,
        splits: Sequence[Path],
        *,
        compress_format: Optional[str] = None,
        compress_factor: Optional[float] = 0.7,
        target_sampling_rate: int = 16000,
        returnn_root: Path,
        concurrent: int = 1,
    ):
        self.splits = splits
        self.compress_format = compress_format
        self.compress_factor = compress_factor
        self.target_sampling_rate = target_sampling_rate
        self.returnn_root = returnn_root
        assert concurrent > 0
        self.concurrent = concurrent

        self.out_hdfs = [self.output_path(f"{i:0d}.hdf") for i in range(len(splits))]

        num_cores = 16
        mem_per_core = 1
        self.rqmt = {"cpu": num_cores, "mem": num_cores * mem_per_core + 2, "time": 48}

    def tasks(self):
        yield Task("run", rqmt=self.rqmt, args=range(self.concurrent))

    def run(self, index: int):
        out_hdfs = list(chunks(self.out_hdfs, self.concurrent))[index]
        splits = list(chunks(self.splits, self.concurrent))[index]
        assert len(out_hdfs) == len(splits)

        _logging.info(f"Writing audio to {out_hdfs} from {splits}.")

        SimpleHDFWriter = get_returnn_simple_hdf_writer(self.returnn_root.get_path())
        hdf_writers = [SimpleHDFWriter(filename=out_hdf.get_path(), dim=1) for out_hdf in out_hdfs]

        segment_whitelists = []
        for split in splits:
            with uopen(split, "rt") as f:
                segments_whitelist = {line.strip() for line in f if len(line.strip()) > 0}
            segment_whitelists.append(segments_whitelist)
        assert len(segment_whitelists) == len(out_hdfs)
        all_whitelists = set.union(*segment_whitelists)

        c = corpus.Corpus()
        c.load(self.bliss_corpus.get_path())

        pool = multiprocessing.Pool(max(self.rqmt["cpu"] - 1, 1))
        apply_func = partial(
            self._process_seq,
            compress_format=self.compress_format,
            compress_factor=self.compress_factor,
            round_factor=1,
            target_sampling_rate=self.target_sampling_rate,
        )
        recs = (
            # We send every segment of the recording grouped together with the audio to 
            # avoid having to read the audio file more than once.
            #
            # We only pass down the required audio/segment metadata instead of the whole
            # `i6_core.lib.corpus.Segment` object to avoid excessive pickling overhead,
            # stalling the worker processes (if we don't).
            (rec.audio, segments)
            for rec in c.all_recordings()
            if len(
                segments := [
                    (full_name, segment.start, segment.end)
                    for segment in rec.segments
                    if (full_name := segment.fullname()) in all_whitelists
                ]
            )
            > 0
        )
        for results in pool.imap_unordered(apply_func, recs, chunksize=8):
            for (segment_name, data), (hdf_writer, segments_whitelist) in itertools.product(
                results, zip(hdf_writers, segment_whitelists)
            ):
                if segment_name not in segments_whitelist:
                    continue
                _logging.info(f"Writing {segment_name} to {hdf_writer.filename}.")
                hdf_writer.insert_batch(
                    inputs=data.reshape(1, -1, 1),
                    seq_len=[data.shape[0]],
                    seq_tag=[segment_name],
                )
        for hdf_writer in hdf_writers:
            hdf_writer.close()

    @staticmethod
    def _process_seq(
        recording: Tuple[str, Sequence[Tuple[str, float, float]]],
        *,
        compress_format: Optional[str],
        compress_factor: Optional[float],
        round_factor: int,
        target_sampling_rate: int,
    ) -> List[Tuple[str, np.ndarray]]:
        max_tries = 3
        audio_file, segments = recording
        for i in range(max_tries):
            try:
                return BlissToPcmHDFJob2.__process_seq(
                    audio_file,
                    segments,
                    compress_format=compress_format,
                    compress_factor=compress_factor,
                    round_factor=round_factor,
                    target_sampling_rate=target_sampling_rate,
                )
            except Exception as exc:
                if i < max_tries - 1:
                    _logging.warning(f"Failed to process audio {audio_file}: {exc}")
                else:
                    _logging.warning(f"Failed to process audio {audio_file} after {max_tries} attempts. Skipping.")
        return []

    @staticmethod
    def __process_seq(
        audio_file: str,
        segments: Sequence[Tuple[str, float, float]],
        *,
        compress_format: Optional[str],
        compress_factor: Optional[float],
        round_factor: int,
        target_sampling_rate: int,
    ) -> List[Tuple[str, np.ndarray]]:
        _logging.info(f"Processing {audio_file} with {len(segments)} segments.")
        results = []
        with tempfile.TemporaryDirectory() as tmp_dir:
            # We preprocess all data with ffmpeg because it is more robust to different
            # audio formats like gsm-ms (which soundfile cannot read) or mp3 (standard format,
            # but soundfile cannot seek). We also resample it at the same time and merge channels.
            #
            # We still use soundfile to extract the segments because we want to slice the audio
            # data on a frame-by-frame basis (to match with RASR), while ffmpeg supports only
            # temporal slices.
            audio_file_out = os.path.join(tmp_dir, os.path.basename(audio_file) + ".wav")
            subprocess.check_call(
                [
                    "ffmpeg",
                    "-y",
                    "-i",
                    audio_file,
                    "-ac",
                    "1",
                    "-af",
                    "aresample=resampler=soxr",
                    "-ar",
                    str(target_sampling_rate),
                    audio_file_out,
                ],
                stdout=subprocess.DEVNULL,
                stderr=subprocess.DEVNULL,
            )
            with sf.SoundFile(audio_file_out) as audio:
                for segment_name, start, end in segments:
                    # use rasr rounding
                    start = math.ceil(start * audio.samplerate / round_factor) * round_factor
                    duration = math.floor(end * audio.samplerate / round_factor) * round_factor - start

                    audio.seek(start)
                    data = audio.read(duration, always_2d=True, dtype=np.int16)
                    data = np.sum(data, axis=-1, dtype=data.dtype)

                    if compress_format is not None:
                        assert compress_factor is not None
                        buf = io.BytesIO()
                        sf.write(
                            buf,
                            data,
                            target_sampling_rate,
                            format=compress_format,
                            compression_level=compress_factor,
                        )
                        data = np.array(list(buf.getvalue()), dtype=np.uint8)

                    results.append((segment_name, data))

        return results

```

albertz:
> I'm now using a simplified (wrt. configurability) but more robust and much faster version of this job.

Can you summarize how it's different? Why is it faster? What do you do differently? And why can't this be done in `BlissToPcmHDFJob`?

Btw, for me, it's still an open question whether there are actually any advantages with the HDF dataset over the ogg.zip (w.r.t. runtime speed, and runtime memory consumption; or what else would be relevant?). But that doesn't mean we should not have this in i6_core anyway.

NeoLegends:
When you apply the existing job on a corpus that is split into many subcorpora and there are recordings in the corpus that have many segments in them, the jobs will end up loading the recording audio multiple times (instead of just once) from the FS. Depending on the corpus this can really increase the I/O load the jobs will cause.

My version of the job can process multiple subcorpora at a time (it's given not just one but all the segment filter lists) and so it can "optimize" I/O by structuring processing around recorings and not segments. It only loads every audio file once and then distributes the resulting segments into their final HDFs afterwards. With some added multiprocessing the reduced I/O pressure makes the new version faster than the existing version, even though there is now only one job instance per bliss corpus instead of one job instance per split.

It also pre-processes every audio file with FFmpeg, which allows using the job on e.g. .mp3 or gsm-ms audio. soundfile either cannot read these formats (gsm-ms) or cannot seek in them (mp3).

> What do you do differently?

I don't have all the different rounding variants implemented right now, as you see. But this can be implemented.

> And why can't this be done in BlissToPcmHDFJob?

The outputs of the job change. Now there is more than one HDF file per job. We could shoehorn this into the existing job, but it's probably cleaner to just have a V2 version.

albertz:
So, if you anyway don't plan to use the newly proposed feature for `BlissToPcmHDFJob`, but instead you plan to use `BlissToPcmHDFJobV2` for that instead, then I don't think it makes sense to introduce these changes to `BlissToPcmHDFJob`, but instead to directly add `BlissToPcmHDFJobV2`. Maybe make a separate PR to have this cleaner.

Also, what you explained here about the difference to V1, this should ideally all go into the docstring of the V2 job.

Also, maybe rename the job then anyway? `BlissToHDFJobV2`?

Also, regarding having Oggs in HDF vs ogg.zip, I think this comparison should still be done, for HDFDataset+PostprocessingDataset vs OggZipDataset, w.r.t. startup time, memory usage, speed.

NeoLegends:
Closing this PR for now, changes will be added as separate job.

PR #596, opened 2025-05-06
FilterCorpusBySegmentsJob: allow compressed segment list

NeoLegends:


corpus/filter.py
```diff
@@ -242,7 +242,7 @@ def tasks(self):
     def run(self):
         segments = []
         for seg in self.segment_file_list:
-            with open(tk.uncached_path(seg)) as f:
+            with uopen(tk.uncached_path(seg)) as f:
```
albertz, line 245: There is a difference: `open` will always try to open in text mode. `uopen` is somewhat unspecific for that without options, so it could be that you get text mode (`readlines()` returns `list[str]`) or bytes (`readlines()` returns `list[bytes]`). I don't think you want to allow for both. So maybe just do:
```suggestion
            with uopen(tk.uncached_path(seg), "rt") as f:
```

corpus/filter.py
```diff
@@ -242,7 +242,7 @@ def tasks(self):
     def run(self):
         segments = []
         for seg in self.segment_file_list:
-            with open(tk.uncached_path(seg)) as f:
+            with uopen(tk.uncached_path(seg), "rt") as f:
```
DanEnergetics, line 245: Just plugging in `seg` should be enough since `tk.uncached_path` is [called internally](https://github.com/rwth-i6/i6_core/blob/77b765eaa112924d131c4a71be374b6339d8e0cb/util.py#L148)
```suggestion
            with uopen(seg, "rt") as f:
```

albertz (approved):

DanEnergetics (approved):

PR #599, opened 2025-05-08
Add ApplySentencepieceToTextJob

Stefanwuu:
Similar to BPE, I add a job to apply sentencepiece model to text.

### DIFF
--- a/text/label/sentencepiece/apply.py
+++ b/text/label/sentencepiece/apply.py
@@ -4,7 +4,7 @@ import subprocess as sp
 import tempfile
 import os
 
-from sisyphus import *
+from sisyphus import Job, Task, tk
 
 import i6_core.util as util
 
--- a/text/label/sentencepiece/apply.py
+++ b/text/label/sentencepiece/apply.py
@@ -68,8 +68,9 @@ class ApplySentencepieceToTextJob(Job):
             tmp_infile = os.path.join(tmp, "in_text.txt")
             tmp_outfile = os.path.join(tmp, "out_text.txt")
             # normalize text format
-            with util.uopen(tmp_infile, "wt") as out:
-                sp.call(["zcat", "-f", input_file], stdout=out)
+            with util.uopen(input_file, "rt") as in_file, open(tmp_infile, "wt") as out:
+                for line in in_file:
+                    out.write(line)
 
             with util.uopen(tmp_infile, "rt") as fin, util.uopen(tmp_outfile, "wt") as fout:
                 for line in fin:
--- a/text/label/sentencepiece/apply.py
+++ b/text/label/sentencepiece/apply.py
@@ -64,10 +64,13 @@ class ApplySentencepieceToTextJob(Job):
         spm = sentencepiece.SentencePieceProcessor(model_file=self.sentencepiece_model.get_path())
         unk_id = spm.unk_id()
         with tempfile.TemporaryDirectory(prefix=gs.TMP_PREFIX) as tmp:
+            input_file = self.text_file.get_path()
             tmp_infile = os.path.join(tmp, "in_text.txt")
             tmp_outfile = os.path.join(tmp, "out_text.txt")
-            # normalize text format
-            shutil.copy(self.text_file.get_path(), tmp_infile)
+            # normalize text format(gz or txt)
+            with util.uopen(input_file, "rt") as in_file, open(tmp_infile, "wt") as out:
+                for line in in_file:
+                    out.write(line)
 
             with util.uopen(tmp_infile, "rt") as fin, util.uopen(tmp_outfile, "wt") as fout:
                 for line in fin:
@@ -79,7 +82,8 @@ class ApplySentencepieceToTextJob(Job):
                             pieces = ["<unk>" if x == unk_id else y for x, y in zip(pieces_id, pieces)]
                     fout.write(" ".join(pieces) + "\n")
 
-            shutil.copy(tmp_outfile, self.out_sentencepiece_text.get_path())
+            with util.uopen(tmp_outfile, "rt") as fin, util.uopen(self.out_sentencepiece_text, "wt") as fout:
+                shutil.copyfileobj(fin, fout) 
 
     @classmethod
     def hash(cls, parsed_args):
--- a/text/label/sentencepiece/apply.py
+++ b/text/label/sentencepiece/apply.py
@@ -77,7 +77,7 @@ class ApplySentencepieceToTextJob(Job):
                         pieces_id = spm.encode(line.rstrip("\n"))
                         assert len(pieces_id) == len(pieces)
                         if unk_id in pieces_id:
-                            pieces[:] = ["<unk>" if x == unk_id else y for x, y in zip(pieces_id, pieces)]
+                            pieces = ["<unk>" if x == unk_id else y for x, y in zip(pieces_id, pieces)]
                     fout.write(" ".join(pieces) + "\n")
 
             shutil.copy(tmp_outfile, self.out_sentencepiece_text.get_path())
--- a/text/label/sentencepiece/apply.py
+++ b/text/label/sentencepiece/apply.py
@@ -1,0 +1,102 @@
+import logging
+import shutil
+import subprocess as sp
+import tempfile
+import os
+
+from sisyphus import *
+
+import i6_core.util as util
+
+try:
+    import sentencepiece
+except ImportError:
+    if not hasattr(gs, "WARNING_NO_SENTENCEPIECE") or gs.WARNING_NO_SENTENCEPIECE is True:
+        logging.warning(
+            "The package 'sentencepiece' is not installed in the manager python env. Please make sure it is installed "
+            "in the python environment running the Sisyphus worker. To suppress this warning set "
+            "'WARNING_NO_SENTENCEPIECE=False' in the settings.py"
+        )
+
+class ApplySentencepieceToTextJob(Job):
+    """
+    Apply sentencepiece model on a text file, basically a wrapper for spm.encode
+    """
+
+    __sis_hash_exclude__ = {"gzip_output": False}
+
+    def __init__(
+        self,
+        text_file: tk.Path,
+        sentencepiece_model: tk.Path,
+        map_unk: bool = False,
+        buffer_size: int = 32768,
+        gzip_output: bool = False,
+        mini_task=True,
+    ):
+        """
+        :param text_file: words text file to convert to sentencepiece
+        :param sentencepiece_model: path to the trained sentencepiece model
+        :param map_unk: when encoding string to string, spm won't map oov symbol to <unk> but keep it as is. 
+            This option forces the oov labels to be <unk> by cecking encoded indices.
+        :param gzip_output: use gzip on the output text
+        :param mini_task: if the Job should run locally, e.g. only a small (<1M lines) text should be processed
+        """
+        self.text_file = text_file
+        self.sentencepiece_model = sentencepiece_model
+        self.map_unk = map_unk
+        self.gzip_output = gzip_output
+        self.buffer_size = buffer_size
+
+        self.out_sentencepiece_text = self.output_path("words_to_sentencepiece.txt.gz" if gzip_output else "words_to_sentencepiece.txt")
+
+        self.mini_task = mini_task
+        self.rqmt = {"cpu": 1, "mem": 2, "time": 2}
+
+    def tasks(self):
+        if self.mini_task:
+            yield Task("run", mini_task=True)
+        else:
+            yield Task("run", rqmt=self.rqmt)
+
+    def run(self):
+        import sentencepiece
+        spm = sentencepiece.SentencePieceProcessor(model_file=self.sentencepiece_model.get_path())
+        unk_id = spm.unk_id()
+        with tempfile.TemporaryDirectory(prefix=gs.TMP_PREFIX) as tmp:
+            input_file = self.text_file.get_path()
+            tmp_infile = os.path.join(tmp, "in_text.txt")
+            tmp_outfile = os.path.join(tmp, "out_text.txt")
+            # normalize text format
+            with util.uopen(tmp_infile, "wt") as out:
+                sp.call(["zcat", "-f", input_file], stdout=out)
+
+            with util.uopen(tmp_infile, "rt") as fin, \
+                 util.uopen(tmp_outfile, "wt") as fout:
+                buf = []
+                for lineno, line in enumerate(fin, 1):
+                    pieces = spm.encode(line.rstrip("\n"), out_type=str)
+                    if self.map_unk:
+                        pieces_id = spm.encode(line.rstrip("\n"))
+                        assert len(pieces_id) == len(pieces)
+                        if unk_id in pieces_id:
+                            pieces[:] = ["<unk>" if x == unk_id else y for x, y in zip(pieces_id, pieces)]
+                    buf.append(" ".join(pieces) + "\n")
+                    # Flush periodically to avoid large memory spikes
+                    if lineno % self.buffer_size == 0:
+                        fout.writelines(buf)
+                        buf.clear()
+                if buf:  # write any remainder
+                    fout.writelines(buf)
+
+            if self.gzip_output:
+                with util.uopen(tmp_outfile, "rt") as fin, util.uopen(self.out_sentencepiece_text, "wb") as fout:
+                    sp.call(["gzip"], stdin=fin, stdout=fout)
+            else:
+                shutil.copy(tmp_outfile, self.out_sentencepiece_text.get_path())
+
+    @classmethod
+    def hash(cls, parsed_args):
+        del parsed_args["mini_task"]
+        return super().hash(parsed_args)
+
--- a/text/label/sentencepiece/apply.py
+++ b/text/label/sentencepiece/apply.py
@@ -42,11 +42,12 @@ class ApplySentencepieceToTextJob(Job):
         self.text_file = text_file
         self.sentencepiece_model = sentencepiece_model
         self.enable_unk = enable_unk
-        self.rqmt: Optional[Dict[str, Any]] = {"cpu": 1, "mem": 2, "time": 2}
 
         self.out_sentencepiece_text = self.output_path(
             "words_to_sentencepiece.txt.gz" if gzip_output else "words_to_sentencepiece.txt"
         )
+
+        self.rqmt: Optional[Dict[str, Any]] = {"cpu": 1, "mem": 2.0, "time": 2.0}
 
     def tasks(self):
         yield Task("run", rqmt=self.rqmt, mini_task=self.rqmt is None)
--- a/text/label/sentencepiece/apply.py
+++ b/text/label/sentencepiece/apply.py
@@ -63,24 +63,18 @@ class ApplySentencepieceToTextJob(Job):
 
         spm = sentencepiece.SentencePieceProcessor(model_file=self.sentencepiece_model.get_path())
         unk_id = spm.unk_id()
-        with tempfile.TemporaryDirectory(prefix=gs.TMP_PREFIX) as tmp:
-            # Note: Important to overtake potential file extensions from the input/output filenames, e.g. ".gz".
-            tmp_infile = os.path.join(tmp, "in_text." + os.path.basename(self.text_file.get_path()))
-            tmp_outfile = os.path.join(tmp, "out_text." + os.path.basename(self.out_sentencepiece_text.get_path()))
 
-            shutil.copy(self.text_file.get_path(), tmp_infile)
-
-            with util.uopen(tmp_infile, "rt") as fin, util.uopen(tmp_outfile, "wt") as fout:
-                for line in fin:
-                    pieces = spm.encode(line.rstrip("\n"), out_type=str)
-                    if self.map_unk:
-                        pieces_id = spm.encode(line.rstrip("\n"))
-                        assert len(pieces_id) == len(pieces)
-                        if unk_id in pieces_id:
-                            pieces = ["<unk>" if x == unk_id else y for x, y in zip(pieces_id, pieces)]
-                    fout.write(" ".join(pieces) + "\n")
-
-            shutil.copy(tmp_outfile, self.out_sentencepiece_text.get_path())
+        with util.uopen(self.text_file.get_path(), "rt") as fin, util.uopen(
+            self.out_sentencepiece_text.get_path(), "wt"
+        ) as fout:
+            for line in fin:
+                pieces = spm.encode(line.rstrip("\n"), out_type=str)
+                if self.map_unk:
+                    pieces_id = spm.encode(line.rstrip("\n"))
+                    assert len(pieces_id) == len(pieces)
+                    if unk_id in pieces_id:
+                        pieces = ["<unk>" if x == unk_id else y for x, y in zip(pieces_id, pieces)]
+                fout.write(" ".join(pieces) + "\n")
 
     @classmethod
     def hash(cls, parsed_args):
--- a/text/label/sentencepiece/apply.py
+++ b/text/label/sentencepiece/apply.py
@@ -82,8 +82,7 @@ class ApplySentencepieceToTextJob(Job):
                             pieces[:] = ["<unk>" if x == unk_id else y for x, y in zip(pieces_id, pieces)]
                     fout.write(" ".join(pieces) + "\n")
 
-            with util.uopen(tmp_outfile, "rt") as fin, util.uopen(self.out_sentencepiece_text, "wt") as fout:
-                shutil.copyfileobj(fin, fout)
+            shutil.copy(tmp_outfile, self.out_sentencepiece_text.get_path())
 
     @classmethod
     def hash(cls, parsed_args):
--- a/text/label/sentencepiece/apply.py
+++ b/text/label/sentencepiece/apply.py
@@ -23,30 +23,28 @@ class ApplySentencepieceToTextJob(Job):
     """
     Apply sentencepiece model on a text file, basically a wrapper for spm.encode
     """
-
-    __sis_hash_exclude__ = {"gzip_output": False}
-
     def __init__(
         self,
+        *,
         text_file: tk.Path,
         sentencepiece_model: tk.Path,
         map_unk: bool = False,
         buffer_size: int = 32768,
         gzip_output: bool = False,
-        mini_task=True,
+        mini_task: bool =True,
     ):
         """
         :param text_file: words text file to convert to sentencepiece
         :param sentencepiece_model: path to the trained sentencepiece model
         :param map_unk: when encoding string to string, spm won't map oov symbol to <unk> but keep it as is.
             This option forces the oov labels to be <unk> by cecking encoded indices.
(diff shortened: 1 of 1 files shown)


albertz:
I simply use `SearchOutputRawReplaceJob(output, [(" ", ""), ("▁", " ")])` for that. That is specific for our text-dict format. For other formats (e.g. just raw text), we could also have such generic `RawReplaceJob`. Or what would be the advantage here?

Stefanwuu:
> I simply use `SearchOutputRawReplaceJob(output, [(" ", ""), ("▁", " ")])` for that. That is specific for our text-dict format. For other formats (e.g. just raw text), we could also have such generic `RawReplaceJob`. Or what would be the advantage here?

If I understand correctly, the job you mentioned is to convert sentencepiece units to words? But this job is the other way round to convert words to spm units(a wrapper over spm.encode). The reason we want this is that we are still creating full-word lexicons for spm for both training and recognition.

albertz:
Ah sorry, I did not look careful enough. Ok, that is certainly reasonable to have.

text/label/sentencepiece/apply.py
```diff
@@ -0,0 +1,104 @@
...
+        )
+
+
+class ApplySentencepieceToTextJob(Job):
+    """
+    Apply sentencepiece model on a text file, basically a wrapper for spm.encode
+    """
+
+    __sis_hash_exclude__ = {"gzip_output": False}
+
```
albertz, line 28: I think we usually never add this for a new job.

```suggestion
```

text/label/sentencepiece/apply.py
```diff
@@ -0,0 +1,104 @@
...
+    def __init__(
+        self,
+        text_file: tk.Path,
```
albertz, line 31:
```suggestion
        *,
        text_file: tk.Path,
```

text/label/sentencepiece/apply.py
```diff
@@ -0,0 +1,104 @@
...
+        sentencepiece_model: tk.Path,
+        map_unk: bool = False,
+        buffer_size: int = 32768,
+        gzip_output: bool = False,
+        mini_task=True,
```
albertz, line 36:
```suggestion
        mini_task: bool = True,
```

text/label/sentencepiece/apply.py
```diff
@@ -0,0 +1,104 @@
...
+
+            if self.gzip_output:
+                with util.uopen(tmp_outfile, "rt") as fin, util.uopen(self.out_sentencepiece_text, "wb") as fout:
+                    sp.call(["gzip"], stdin=fin, stdout=fout)
+            else:
+                shutil.copy(tmp_outfile, self.out_sentencepiece_text.get_path())
+
+    @classmethod
+    def hash(cls, parsed_args):
+        del parsed_args["mini_task"]
```
albertz, line 103: You should also exclude `buffer_size`.

```suggestion
        del parsed_args["mini_task"]
        del parsed_args["buffer_size"]
```

text/label/sentencepiece/apply.py
```diff
@@ -0,0 +1,104 @@
...
+        gzip_output: bool = False,
+        mini_task=True,
+    ):
+        """
+        :param text_file: words text file to convert to sentencepiece
+        :param sentencepiece_model: path to the trained sentencepiece model
+        :param map_unk: when encoding string to string, spm won't map oov symbol to <unk> but keep it as is.
+            This option forces the oov labels to be <unk> by cecking encoded indices.
+        :param gzip_output: use gzip on the output text
+        :param mini_task: if the Job should run locally, e.g. only a small (<1M lines) text should be processed
```
albertz, line 44: `buffer_size` is not documented.

text/label/sentencepiece/apply.py
```diff
@@ -0,0 +1,104 @@
...
+        with tempfile.TemporaryDirectory(prefix=gs.TMP_PREFIX) as tmp:
+            input_file = self.text_file.get_path()
+            tmp_infile = os.path.join(tmp, "in_text.txt")
+            tmp_outfile = os.path.join(tmp, "out_text.txt")
+            # normalize text format
+            with util.uopen(tmp_infile, "wt") as out:
+                sp.call(["zcat", "-f", input_file], stdout=out)
+
+            with util.uopen(tmp_infile, "rt") as fin, util.uopen(tmp_outfile, "wt") as fout:
+                buf = []
```
NeoLegends, line 79: I wonder: are the file objects themselves not already buffered (see https://docs.python.org/3/library/io.html#io.DEFAULT_BUFFER_SIZE)? Does the job get too slow without your manual intermediate buffer? Maybe just remove the manual buffer altogether?

text/label/sentencepiece/apply.py
```diff
@@ -0,0 +1,104 @@
...
+                        fout.writelines(buf)
+                        buf.clear()
+                if buf:  # write any remainder
+                    fout.writelines(buf)
+
+            if self.gzip_output:
+                with util.uopen(tmp_outfile, "rt") as fin, util.uopen(self.out_sentencepiece_text, "wb") as fout:
+                    sp.call(["gzip"], stdin=fin, stdout=fout)
+            else:
+                shutil.copy(tmp_outfile, self.out_sentencepiece_text.get_path())
```
NeoLegends, line 99: The way we create gzipped files is usually a bit different. `uopen`'s job is to disambiguate transparently between gzipped files and non-gzipped files by looking at the file extension. If the file ends in `.gz` it applies gzip, otherwise it won't.

This part can be made simpler, because the extension of `out_sentencepiece_text` in `__init__` is already set correct. Here, you can just do:
```suggestion
            with util.uopen(tmp_outfile, "rt") as fin, util.uopen(self.out_sentencepiece_text, "wt") as fout:
                for line in fin:
                    fout.write(line)
```

text/label/sentencepiece/apply.py
```diff
@@ -0,0 +1,104 @@
...
+
+        spm = sentencepiece.SentencePieceProcessor(model_file=self.sentencepiece_model.get_path())
+        unk_id = spm.unk_id()
+        with tempfile.TemporaryDirectory(prefix=gs.TMP_PREFIX) as tmp:
+            input_file = self.text_file.get_path()
+            tmp_infile = os.path.join(tmp, "in_text.txt")
+            tmp_outfile = os.path.join(tmp, "out_text.txt")
+            # normalize text format
+            with util.uopen(tmp_infile, "wt") as out:
+                sp.call(["zcat", "-f", input_file], stdout=out)
```
NeoLegends, line 76: Does this work when `input_file` is not gzip compressed? To be safe in any case you could just use another `uopen` instead of the subprocess call:

```suggestion
            with util.uopen(input_file, "rt") as in_file, open(tmp_infile, "wt") as out:
                for line in in_file:
                    out.write(line)
```

text/label/sentencepiece/apply.py
Stefanwuu: I think it's another level of buffering. Here, without an explicit buffer, you would either write all lines together(making the list object too large) or write line by line(potentially many write operations).

Stefanwuu: I didn't try without buffering though.

NeoLegends: Yes, I am expecting that the many write operations of simple line-by-line writing will be caught by the I/O buffer.

text/label/sentencepiece/apply.py
```diff
@@ -0,0 +1,104 @@
...
+
+    __sis_hash_exclude__ = {"gzip_output": False}
+
+    def __init__(
+        self,
+        text_file: tk.Path,
+        sentencepiece_model: tk.Path,
+        map_unk: bool = False,
+        buffer_size: int = 32768,
+        gzip_output: bool = False,
```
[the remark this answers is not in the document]
NeoLegends: @Stefanwuu change the default here?

text/label/sentencepiece/apply.py
NeoLegends: > potentially many write operations

My argument is that this is likely just fine and less complicated without your manual buffering. Python adds buffering by itself already to every file object (transparently, you don't see it), precisely so that many write calls do not become a problem. Due to this, you won't need your manual buffer.

Or you say "I don't care", then it's fine with me. Your manual buffering is not a problem after all.

NeoLegends (commented):
Some feedback wasn't adressed. I'm okay with this if done consciously and you say "I don't care".

text/label/sentencepiece/apply.py
```diff
@@ -0,0 +1,101 @@
+import logging
+import shutil
+import subprocess as sp
+import tempfile
+import os
+
+from sisyphus import *
```
albertz, line 7:
```suggestion
from sisyphus import Job, Task, tk
```

text/label/sentencepiece/apply.py
```diff
@@ -0,0 +1,91 @@
...
+                    pieces = spm.encode(line.rstrip("\n"), out_type=str)
+                    if self.map_unk:
+                        pieces_id = spm.encode(line.rstrip("\n"))
+                        assert len(pieces_id) == len(pieces)
+                        if unk_id in pieces_id:
+                            pieces[:] = ["<unk>" if x == unk_id else y for x, y in zip(pieces_id, pieces)]
+                    fout.write(" ".join(pieces) + "\n")
+
+            with util.uopen(tmp_outfile, "rt") as fin, util.uopen(self.out_sentencepiece_text, "wt") as fout:
+                shutil.copyfileobj(fin, fout)
```
albertz, line 86:
```suggestion
            shutil.copy(tmp_outfile, self.out_sentencepiece_text)
```

text/label/sentencepiece/apply.py
```diff
@@ -0,0 +1,91 @@
...
+        spm = sentencepiece.SentencePieceProcessor(model_file=self.sentencepiece_model.get_path())
+        unk_id = spm.unk_id()
+        with tempfile.TemporaryDirectory(prefix=gs.TMP_PREFIX) as tmp:
+            input_file = self.text_file.get_path()
+            tmp_infile = os.path.join(tmp, "in_text.txt")
+            tmp_outfile = os.path.join(tmp, "out_text.txt")
+            # normalize text format
+            with util.uopen(input_file, "rt") as in_file, open(tmp_infile, "wt") as out:
+                for line in in_file:
+                    out.write(line)
```
albertz, line 73:
```suggestion
            shutil.copy(input_file, tmp_infile)
```

text/label/sentencepiece/apply.py
```diff
@@ -0,0 +1,91 @@
...
+        else:
+            yield Task("run", rqmt=self.rqmt)
+
+    def run(self):
+        import sentencepiece
+
+        spm = sentencepiece.SentencePieceProcessor(model_file=self.sentencepiece_model.get_path())
+        unk_id = spm.unk_id()
+        with tempfile.TemporaryDirectory(prefix=gs.TMP_PREFIX) as tmp:
+            input_file = self.text_file.get_path()
```
albertz, line 67:
```suggestion
```

text/label/sentencepiece/apply.py
```diff
@@ -0,0 +1,91 @@
...
+                    out.write(line)
+
+            with util.uopen(tmp_infile, "rt") as fin, util.uopen(tmp_outfile, "wt") as fout:
+                for line in fin:
+                    pieces = spm.encode(line.rstrip("\n"), out_type=str)
+                    if self.map_unk:
+                        pieces_id = spm.encode(line.rstrip("\n"))
+                        assert len(pieces_id) == len(pieces)
+                        if unk_id in pieces_id:
+                            pieces[:] = ["<unk>" if x == unk_id else y for x, y in zip(pieces_id, pieces)]
```
albertz, line 82:
```suggestion
                            pieces = ["<unk>" if x == unk_id else y for x, y in zip(pieces_id, pieces)]
```

text/label/sentencepiece/apply.py
Stefanwuu: Actually this should not be committed(I already undo it). The goal of opening the two files is to have unified behavior between gziped and not gzipped output by taking advantage of util.uopen

text/label/sentencepiece/apply.py
Stefanwuu: Actually this should not be committed(I already undo it). The goal of opening the two files is to have unified behavior between gziped and not gzipped output by taking advantage of util.uopen

text/label/sentencepiece/apply.py
```diff
@@ -0,0 +1,87 @@
...
+            yield Task("run", mini_task=True)
+        else:
+            yield Task("run", rqmt=self.rqmt)
+
+    def run(self):
+        import sentencepiece
+
+        spm = sentencepiece.SentencePieceProcessor(model_file=self.sentencepiece_model.get_path())
+        unk_id = spm.unk_id()
+        with tempfile.TemporaryDirectory(prefix=gs.TMP_PREFIX) as tmp:
```
albertz, line 66: Note: I'm not really sure how much needed the whole logic for the tmp infile/outfile is here. I don't think that the IO stress is so big here anyway. Maybe we can remove this, and simplify further? @NeoLegends what do you think?

Stefanwuu:
> Some feedback wasn't adressed. I'm okay with this if done consciously and you say "I don't care".

Yeah, sorry, was distracted by sth else. I now address all and tested it

text/label/sentencepiece/apply.py
albertz: > Yes, temp output file is definitely uncompressed

No, it's not, with the code I suggested.

text/label/sentencepiece/apply.py
```diff
@@ -0,0 +1,91 @@
...
+                    pieces = spm.encode(line.rstrip("\n"), out_type=str)
+                    if self.map_unk:
+                        pieces_id = spm.encode(line.rstrip("\n"))
+                        assert len(pieces_id) == len(pieces)
+                        if unk_id in pieces_id:
+                            pieces = ["<unk>" if x == unk_id else y for x, y in zip(pieces_id, pieces)]
+                    fout.write(" ".join(pieces) + "\n")
+
+            with util.uopen(tmp_outfile, "rt") as fin, util.uopen(self.out_sentencepiece_text, "wt") as fout:
+                shutil.copyfileobj(fin, fout)
```
albertz, line 86:
```suggestion
            shutil.copy(tmp_outfile, self.out_sentencepiece_text.get_path())
```

text/label/sentencepiece/apply.py
albertz: You are doing `uopen` twice (both for the tmp out, and for the final out). In one place, it's not needed. But see my code suggestion.

Stefanwuu: Ah OK, sorry, I missed the other code suggestion you made. It should work this way then. Thanks!

text/label/sentencepiece/apply.py
```diff
@@ -0,0 +1,91 @@
...
+    """
+
+    def __init__(
+        self,
+        *,
+        text_file: tk.Path,
+        sentencepiece_model: tk.Path,
+        map_unk: bool = False,
+        gzip_output: bool = True,
+        mini_task: bool = True,
```
albertz, line 34: Is this a good default here? I think this might make more sense:
```suggestion
        mini_task: bool = False,
```

text/label/sentencepiece/apply.py
```diff
@@ -0,0 +1,91 @@
...
+        unk_id = spm.unk_id()
+        with tempfile.TemporaryDirectory(prefix=gs.TMP_PREFIX) as tmp:
+            input_file = self.text_file.get_path()
+            # Note: Important to overtake potential file extensions from the input/output filenames, e.g. ".gz".
+            tmp_infile = os.path.join(tmp, "in_text." + os.path.basename(self.text_file.get_path()))
+            tmp_outfile = os.path.join(tmp, "out_text." + os.path.basename(self. out_sentencepiece_text.get_path()))
+            # normalize text format(gz or txt)
+            with util.uopen(input_file, "rt") as in_file, open(tmp_infile, "wt") as out:
+                for line in in_file:
+                    out.write(line)
```
albertz, line 74:
```suggestion
            
            shutil.copy(self.text_file.get_path(), tmp_infile)
```

text/label/sentencepiece/apply.py
```diff
@@ -0,0 +1,91 @@
...
+        else:
+            yield Task("run", rqmt=self.rqmt)
+
+    def run(self):
+        import sentencepiece
+
+        spm = sentencepiece.SentencePieceProcessor(model_file=self.sentencepiece_model.get_path())
+        unk_id = spm.unk_id()
+        with tempfile.TemporaryDirectory(prefix=gs.TMP_PREFIX) as tmp:
+            input_file = self.text_file.get_path()
```
albertz, line 67:
```suggestion
```

text/label/sentencepiece/apply.py
NeoLegends: Yeah, I agree. Probably this is not that much FS load.

albertz: Ok, then let's remove this. (@Stefanwuu)

text/label/sentencepiece/apply.py
```diff
@@ -0,0 +1,82 @@
...
+
+    def run(self):
+        import sentencepiece
+
+        spm = sentencepiece.SentencePieceProcessor(model_file=self.sentencepiece_model.get_path())
+        unk_id = spm.unk_id()
+
+        with util.uopen(self.text_file.get_path(), "rt") as fin, util.uopen(
+            self.out_sentencepiece_text.get_path(), "wt"
+        ) as fout:
```
albertz, line 69:
```suggestion
        with util.uopen(self.text_file, "rt") as fin, util.uopen(self.out_sentencepiece_text, "wt") as fout:
```

text/label/sentencepiece/apply.py
```diff
@@ -0,0 +1,80 @@
...
+        )
+
+        self.mini_task = mini_task
+        self.rqmt = {"cpu": 1, "mem": 2, "time": 2}
+
+    def tasks(self):
+        if self.mini_task:
+            yield Task("run", mini_task=True)
+        else:
+            yield Task("run", rqmt=self.rqmt)
```
michelwi, line 59: we sometimes use 
```suggestion
        self.rqmt = {"cpu": 1, "mem": 2, "time": 2}

    def tasks(self):
        yield Task("run", rqmt=self.rqmt, mini_task=self.rqmt is None)
```
this way we could drop another parameter and the custom hash function

text/label/sentencepiece/apply.py
```diff
@@ -0,0 +1,80 @@
...
+        unk_id = spm.unk_id()
+
+        with util.uopen(self.text_file, "rt") as fin, util.uopen(self.out_sentencepiece_text, "wt") as fout:
+            for line in fin:
+                pieces = spm.encode(line.rstrip("\n"), out_type=str)
+                if self.map_unk:
+                    pieces_id = spm.encode(line.rstrip("\n"))
+                    assert len(pieces_id) == len(pieces)
+                    if unk_id in pieces_id:
+                        pieces = ["<unk>" if x == unk_id else y for x, y in zip(pieces_id, pieces)]
```
michelwi, line 74: should we make the unk label configurable or is `"<unk>"` universally correct?

text/label/sentencepiece/apply.py
albertz: Also fine with me. But you should declare the type of `self.rqmt: Optional[Dict[str, Any]]` or so, i.e. make it clear that this can be set to None.

text/label/sentencepiece/apply.py
albertz: Actually what does `spm.encode` return for those unknown labels? Isn't that already `"<unk>"` (or whatever you defined in the SPM)? So why is this code then needed at all?

Stefanwuu: Calling `spm.encode(line.rstrip("\n"), out_type=str)` will not return `"<unk>"` for OOV but the original OOV symbol, calling `spm.encode(line.rstrip("\n"))` will return unk ID for OOV correctly. I don't know if it's a bug or intended.  It's why I do the `map_unk` thing. 
Wrt Willi's question, `"<unk>"` will be universally correct as long as we treat `"<unk>"` as unknown in the following jobs(e.g. PipelineJob). It has nothing to do with the sentencepiece model itself, because we use `spm.unk_id()` to detect unknown.

albertz: The SPM options are documented [here](https://github.com/google/sentencepiece/blob/master/doc/options.md).
There is the `unk_piece` option, which is the unknown token for the unknown id (`unk_id` option), and which is `"<unk>"` by default, but you can configure it to anything else as well.
So, for what case is this `unk_piece` used then?

I see that there is also `SetEncodeExtraOptions`, and you can set `SetEncodeExtraOptions("unk")` or so? I think this will enable the unknown handling? I'm quite sure there is some way to properly enable this. Otherwise having the `unk_piece` doesn't really make sense?

So, I think we should use the official correct way then, instead of hacking this here.

Stefanwuu: Yes, you are correct! I tried `SetEncodeExtraOptions("unk")` and it actually enables unknown handling. The documentation is tbh horrible but great guess:)

text/label/sentencepiece/apply.py
```diff
@@ -0,0 +1,71 @@
...
+
+    def tasks(self):
+        yield Task("run", rqmt=self.rqmt, mini_task=self.rqmt is None)
+
+    def run(self):
+        import sentencepiece
+
+        spm = sentencepiece.SentencePieceProcessor(model_file=self.sentencepiece_model.get_path())
+        if self.enable_unk:
+            spm.SetEncodeExtraOptions("unk")
```
albertz, line 61: I think using the PEP8-style function names is now the preferred API.
```suggestion
            spm.set_encode_extra_options("unk")
```

text/label/sentencepiece/apply.py
```diff
@@ -0,0 +1,71 @@
...
+        """
+        self.text_file = text_file
+        self.sentencepiece_model = sentencepiece_model
+        self.enable_unk = enable_unk
+
+        self.out_sentencepiece_text = self.output_path(
+            "words_to_sentencepiece.txt.gz" if gzip_output else "words_to_sentencepiece.txt"
+        )
+
+        self.rqmt = {"cpu": 1, "mem": 2, "time": 2}
```
albertz:
```suggestion
        self.rqmt: Optional[Dict[str, Any]] = {"cpu": 1, "mem": 2, "time": 2}
```

text/label/sentencepiece/apply.py
```diff
@@ -31,25 +32,24 @@ def __init__(
         sentencepiece_model: tk.Path,
         enable_unk: bool = True,
         gzip_output: bool = True,
-        mini_task: bool = False,
+        rqmt: Optional[Dict[str, Any]] = None,
```
albertz, line 35: @michelwi Was it your suggestion to have this here as an argument? I thought the suggestion was to do `job.rqmt = None` if you want to use it as a mini task.

In any case, now here the default changed to have it as mini task again. I don't think this is a good default here.

Stefanwuu: I did the modification because I didn't understand how I could set mini_task if self.rqmt is set directly in the job, but not passed as a parameter. If it's preferred to have that way, I can change back.

albertz: > I didn't understand how I could set mini_task if self.rqmt is set directly in the job, but not passed as a parameter

Just like that:
```python
job = ApplySentencepieceToTextJob(...)
job.rqmt = None  # make it a mini task
...
```
At least I thought that is how I understood @michelwi how he intended it to be. But you can anyway do it like this.

But you can also have it as an argument here if you want. However, I think the default should be to not be a mini task.

text/label/sentencepiece/apply.py
```diff
@@ -0,0 +1,71 @@
...
+            spm.set_encode_extra_options("unk")
+
+        with util.uopen(self.text_file, "rt") as fin, util.uopen(self.out_sentencepiece_text, "wt") as fout:
+            for line in fin:
+                pieces = spm.encode(line.rstrip("\n"), out_type=str)
+                fout.write(" ".join(pieces) + "\n")
+
+    @classmethod
+    def hash(cls, parsed_args):
+        del parsed_args["mini_task"]
```
albertz, line 70:
```suggestion
        del parsed_args["rqmt"]
```

text/label/sentencepiece/apply.py
michelwi: my intention was as Albert inferred it: have no parameter in the constructor and manually set `job.rqmt=None` if need be.

This is assuming that the default you give is reasonable for the vast majority of cases where this job needs to be used. If the requirements change often (because of different dataset sizes or layouts) then we can discuss about parameterizing it.

albertz (approved):

text/label/sentencepiece/apply.py
```diff
@@ -0,0 +1,64 @@
...
+        :param gzip_output: use gzip on the output text
+        """
+        self.text_file = text_file
+        self.sentencepiece_model = sentencepiece_model
+        self.enable_unk = enable_unk
+        self.rqmt: Optional[Dict[str, Any]] = {"cpu": 1, "mem": 2, "time": 2}
+
+        self.out_sentencepiece_text = self.output_path(
+            "words_to_sentencepiece.txt.gz" if gzip_output else "words_to_sentencepiece.txt"
+        )
```
curufinwe, line 49:
```suggestion

        self.out_sentencepiece_text = self.output_path(
            "words_to_sentencepiece.txt.gz" if gzip_output else "words_to_sentencepiece.txt"
        )

        self.rqmt: Optional[Dict[str, Any]] = {"cpu": 1, "mem": 2.0, "time": 2.0}
```

text/label/sentencepiece/apply.py
```diff
@@ -0,0 +1,64 @@
...
+
+    def tasks(self):
+        yield Task("run", rqmt=self.rqmt, mini_task=self.rqmt is None)
+
+    def run(self):
+        import sentencepiece
```
curufinwe, line 55: This import is not needed. `sentencepiece` is imported on a module level and if that failed this import will also fail.

curufinwe (commented):
Two mostly style-guide comments. I would prefer them addressed, but don't want to block merging if they are deemed undesierable.

text/label/sentencepiece/apply.py
```diff
@@ -0,0 +1,64 @@
...
+
+        spm = sentencepiece.SentencePieceProcessor(model_file=self.sentencepiece_model.get_path())
```
NeoLegends, line 57: Last one nit: I feel like we should allow passing any custom options to the sentencepiece processor. Maybe you want to use e.g. sampling to generate a few different variants per sentence.

NeoLegends (commented):
Almost ready!

text/label/sentencepiece/apply.py
NeoLegends: Okay, right, sampling is maybe a bad example (although I think spm by default uses a fixed random seed, or you can fix it), but character normalization could be a relevant topic.

NeoLegends (approved):
Ok!

PR #600, opened 2025-05-09
Add `CallImport`

NeoLegends:
Like `PartialImport`, but actually calls through the callable. This is helpful for when you are trying to import callable classes that you wish to initialize during the import process. When you do this through `PartialImport`, the `PartialImport` will only fix the arguments, but not actually call your classes' `__init__`.

E.g. this simplifies importing https://github.com/rwth-i6/returnn/blob/0baf1d99bdc1afbd90aeaa080642098c2aa13f52/returnn/datasets/postprocessing.py#L396-L402

### DIFF
--- a/serialization/base.py
+++ b/serialization/base.py
@@ -220,7 +220,9 @@ class PartialImport(Import):
 
 class CallImport(PartialImport):
     """
-    Like PartialImport, but issues a full, not just a partial call to the imported callable.
+    Imports some function/class and calls it with given args.
+
+    See also :class:`PartialImport` to get a partial function/class.
     """
 
     TEMPLATE = textwrap.dedent(


DanEnergetics:
Would it make sense that, similar to the `Call` class, the assignment variable is optional? This would make it possible to pass a `CallImport` object directly as an input to another one.
Furthermore, none of the classes right now support non-keyword args. However this would be necessary for the creating the [Sequential](https://github.com/rwth-i6/returnn/blob/0baf1d99bdc1afbd90aeaa080642098c2aa13f52/returnn/datasets/postprocessing.py#L460) class. This may be beyond the scope of this PR though.

serialization/base.py
```diff
@@ -215,6 +218,20 @@ def _sis_hash(self) -> bytes:
         return sis_hash_helper({"import": super_hash, "hashed_arguments": self.hashed_arguments})
 
 
+class CallImport(PartialImport):
+    """
+    Like PartialImport, but issues a full, not just a partial call to the imported callable.
```
albertz, line 223: The doc is a bit confusing: There is no such thing as a "partial call". There are partial functions.
```suggestion
    Imports some function/class and calls it with given args.

    See also :class:`PartialImport` to get a partial function/class.
```

albertz:
Fine with the change itself.

Note though that I think we should in general go away from this type of serialization. It's just quite annoying to use. I think my serialization V2 is the approach which makes more sense, where you don't need to care at all about any of this anymore. It just always works (it works in the same way as pickling). See [this doc](https://github.com/rwth-i6/i6_experiments/blob/main/users/zeyer/serialization.rst) for an overview of all the different serialization approaches that we have.

NeoLegends:
I'm actually using these imports with the new serialization as a way to avoid importing large dependencies during manager runtime.

michelwi (approved):

albertz (approved):

NeoLegends:
But importing models from `i6_models` will.

PR #601, opened 2025-05-09
Add v2 RETURNN config serialization

NeoLegends:
This PR adds an adapted version of the v2 serialization written by @albertz. I intend for this PR to house discussion on what features to support in v2 serialization and what not to support.

The adaptations I made mainly concern themselves around being compatible with existing RETURNN configs that use the v1 serialization mechanisms like the `Import`, `PartialImport`, etc.

I also removed `Dim` serialization support to avoid having to import RETURNN from the sisyphus manager (to improve mgr loading speed). I think we should bring back this feature in some way though, because it's very necessary for RF setups.

### DIFF
--- a/serialization_v2/serialization.py
+++ b/serialization_v2/serialization.py
@@ -1,0 +1,1165 @@
+"""
+New simplified config serialization, usually for RETURNN configs.
+
+See https://github.com/rwth-i6/i6_experiments/blob/main/users/zeyer/serialization.rst
+for more details.
+
+This is conceptually similar to :class:`i6_experiments.common.utils.dump_py_code.PythonCodeDumper`
+and :func:`i6_experiments.common.setups.returnn.serialization.get_serializable_config`.
+
+See :func:`serialize_config` for the main entry point.
+
+Note: Sisyphus hashes are currently just defined by the config keys/values,
+using the `sis_hash_helper` function, without any special handling,
+except for dim tags (RETURNN :class:`Dim` objects).
+That means, e.g. functions/classes get hashed by ``(obj.__module__, obj.__qualname__)``.
+We currently don't provide a way to customize the hashing behavior
+(except of ``post_config`` which is not hashed at all).
+This could be extended in the future by allowing more custom behavior
+e.g. for module scopes.
+Also, e.g. to specify ``unhashed_package_root`` for some of the references.
+
+Note: Sisyphus Path objects are serialized directly using :func:`sisyphus.Path.get_path`.
+
+We handle those objects specially:
+- primitive types (int, float, bool, str)
+- Sisyphus Path objects
+- ~~RETURNN Dim objects~~ (removed for now to avoid importing RETURNN in the sisyphus manager)
+- dict, list, tuple, set
+- functions, classes, modules
+- functools.partial (just some nicer repr)
+- i6_core.serialization objects
+- CodeWrapper
+
+All other generic objects are handled in the same way as pickle does it
+(or also :class:`i6_experiments.common.utils.dump_py_code.PythonCodeDumper`),
+i.e. using ``__reduce__`` (etc).
+This also allows circular references.
+"""
+
+from __future__ import annotations
+
+
+__all__ = ["serialize_config", "SisPathHandling", "ReturnnConfigWithNewSerialization", "SerializedConfig"]
+
+import base64
+import builtins
+from dataclasses import dataclass
+import enum
+import functools
+import inspect
+import math
+import os
+import pickle
+import subprocess
+import sys
+import types
+from types import BuiltinFunctionType, FunctionType, MethodType, ModuleType
+from typing import Any, Collection, Dict, List, Optional, Sequence, Tuple, Union
+
+from sisyphus import Path
+from sisyphus.delayed_ops import DelayedBase
+from sisyphus.hash import sis_hash_helper
+import tree
+
+from i6_core.returnn.config import CodeWrapper, ReturnnConfig
+from i6_core.serialization import (
+    Call,
+    CallImport,
+    CodeFromFile,
+    CodeFromFunction,
+    ExplicitHash,
+    ExternalImport,
+    Import,
+    NonhashedCode,
+    PartialImport,
+    SerializerObject,
+)
+from i6_core.serialization import Collection as SerializerCollection
+
+
+def serialize_config(
+    config: Dict[str, Any],
+    post_config: Optional[Dict[str, Any]] = None,
+    *,
+    inlining: bool = True,
+    known_modules: Collection[str] = (),
+    extra_sys_paths: Sequence[str] = (),
+    sis_path_handling: SisPathHandling = None,
+) -> SerializedConfig:
+    """serialize config. see module docstring for more info."""
+    serializer = _Serializer(
+        config=config, post_config=post_config, known_modules=known_modules, sis_path_handling=sis_path_handling
+    )
+    for path in extra_sys_paths:
+        serializer.add_sys_path(path, recursive=False)
+    serializer.work_queue()
+    if inlining:
+        serializer.work_inlining()
+    return SerializedConfig(code_list=list(serializer.assignments_dict_by_idx.values()))
+
+
+class SisPathHandling(enum.Enum):
+    """
+    SisPathHandling enum.
+    """
+
+    NONE = None
+    AS_STRING = "as_string"
+    NO_DEPS = "no_deps"
+
+
+def _instanciate_delayed_copy(o: Any) -> Any:
+    """
+    Recursively traverses a structure and calls .get() on all
+    existing Delayed Operations, especially Variables in the structure
+
+    In contrast to :func:`i6_core.util.instanciate_delayed` this function does not operate inplace.
+
+    :param o: nested structure that may contain DelayedBase objects
+    :return: o with all DelayedBase objects replaced by their .get() value
+    """
+    import tree
+
+    def _instanciate_delayed_obj(o: Any) -> Any:
+        if isinstance(o, DelayedBase) and not isinstance(o, SerializerObject):
+            return o.get()
+        return o
+
+    return tree.map_structure(_instanciate_delayed_obj, o)
+
+
+def _parse_python(code: Any, name=None):
+    if code is None:
+        return ""
+    if isinstance(code, str):
+        return code
+    if isinstance(code, DelayedBase):
+        return _parse_python(code.get())
+    if isinstance(code, (tuple, list)):
+        return "\n".join(_parse_python(c) for c in code)
+    if isinstance(code, dict):
+        return "\n".join(_parse_python(v, name=k) for k, v in code.items())
+    if inspect.isfunction(code):
+        try:
+            return inspect.getsource(code)
+        except OSError:
+            # cannot get source, e.g. code is a lambda
+            assert name is not None
+            args = [
+                code.__code__.co_argcount,
+                code.__code__.co_kwonlyargcount,
+                code.__code__.co_nlocals,
+                code.__code__.co_stacksize,
+                code.__code__.co_flags,
+                code.__code__.co_code,
+                code.__code__.co_consts,
+                code.__code__.co_names,
+                code.__code__.co_varnames,
+                code.__code__.co_filename,
+                code.__code__.co_name,
+                code.__code__.co_firstlineno,
+                code.__code__.co_lnotab,
+                code.__code__.co_freevars,
+                code.__code__.co_cellvars,
+            ]
+            compiled = base64.b64encode(pickle.dumps(args)).decode("utf8")
+            return (
+                "import types; import base64; import pickle; "
+                'code = types.CodeType(*pickle.loads(base64.b64decode("%s".encode("utf8")))); '
+                '%s = types.FunctionType(code, globals(), "%s")' % (compiled, name, code.__name__)
+            )
+    if inspect.isclass(code):
+        return inspect.getsource(code)
+    raise RuntimeError("Could not serialize %s" % code)
+
+
+def _is_valid_python_identifier_name(name: str) -> bool:
+    """
+    :return: whether the name is a valid Python identifier name (including attrib name)
+    """
+
+    # Very hacky. I'm sure there is some clever regexp, but I don't find it and too lazy...
+    class _Obj:
+        pass
+
+    obj = _Obj()
+    try:
+        exec(f"obj.{name} = 'ok'", {"obj": obj})
+    except SyntaxError:
+        return False
+    assert getattr(obj, name) == "ok"
+    return True
+
+
+class ReturnnConfigWithNewSerialization(ReturnnConfig):
+    """
+    Overwrites the serialization behavior of ReturnnConfig
+    """
+
+    @staticmethod
+    def from_cfg(old_returnn_cfg: ReturnnConfig):
+        assert not old_returnn_cfg.staged_network_dict  # not supported
+        return ReturnnConfigWithNewSerialization(
+            config=old_returnn_cfg.config,
+            post_config=old_returnn_cfg.post_config,
+            python_epilog=old_returnn_cfg.python_epilog,
+            python_epilog_hash=old_returnn_cfg.python_epilog_hash,
+            python_prolog=old_returnn_cfg.python_prolog,
+            python_prolog_hash=old_returnn_cfg.python_prolog_hash,
+            sort_config=False,
+        )
+
+    def _serialize(self) -> str:
+        # This is usually run within the worker, but it shouldn't really matter.
+        assert not self.staged_network_dict  # not supported
+
+        self.check_consistency()
+
+        config = _instanciate_delayed_copy(self.config)
+        post_config = _instanciate_delayed_copy(self.post_config)
+
+        # I'm not really sure about it.
+        # Our automatic mechanism will find direct imports (e.g. i6_experiments).
+        # However, it will not find indirect imports (e.g. sisyphus),
+        # and thus the generated code might fail.
+        # So add all other paths here which we currently have.
+        # (While testing this, this was sisyphus + returnn + recipes,
+        #  but returnn is excluded below.)
+        extra_sys_paths = [p for p in sys.path if p not in _get_base_sys_path_list()]
+
+        # Handle ExternalImports
+        extra_sys_paths += [
+            item.import_path
+            for item in tree.flatten(config) + tree.flatten(post_config)
+            if isinstance(item, ExternalImport)
+        ]
+
+        python_prolog_code = _parse_python(self.python_prolog)
+        python_epilog_code = _parse_python(self.python_epilog)
+        serialized = serialize_config(
+            config,
+            post_config,
+            # Of course RETURNN knows about itself, no need to add to sys.path.
+            # Also, we don't want to force the current RETURNN here,
+            # but allow the config to be used with any other RETURNN version.
+            known_modules={"returnn"},
+            extra_sys_paths=extra_sys_paths,
+            # instanciate_delayed should already have handled it (e.g. Path),
+            # or if it has not, then we want to keep it as it is (e.g. PtCheckpoint),
+            # but without dependencies.
+            sis_path_handling=SisPathHandling.AS_STRING,
+        )
+        return "\n\n".join(
+            part
+            for part in [
+                "#!returnn/rnn.py",
+                python_prolog_code,
+                serialized.as_serialized_code(),
+                python_epilog_code,
+                "# -*- mode: python; tab-width: 4 -*-\n",
+            ]
+            if part.strip()
+        )
+
+
+@dataclass
+class SerializedConfig:
+    code_list: List[PyCode]
+
+    def as_serialization_collection(self) -> SerializerCollection:
+        """as serialization Collection"""
+        return SerializerCollection(self.code_list)
+
+    def as_serialized_code(self) -> str:
+        """as serialized code"""
+        return "".join(code.py_code for code in self.code_list)
+
+
+class _Serializer:
+    def __init__(
+        self,
+        config: Dict[str, Any],
+        post_config: Optional[Dict[str, Any]] = None,
+        known_modules: Collection[str] = (),
+        sis_path_handling: SisPathHandling = None,
+    ):
+        self.config = config.copy()
+        self.post_config = post_config.copy() if post_config else {}
+        self.assignments_dict_by_value_ref: Dict[_Ref, PyCode] = {}  # value ref -> code
+        self.assignments_dict_by_name: Dict[str, PyCode] = {}  # var name -> code
+        self.assignments_dict_by_idx: Dict[int, PyCode] = {}  # idx -> code
+        self.assignments_dict_by_value_by_type: Dict[type, Dict[Any, PyCode]] = {}  # type -> dict value -> code
+        self.reduce_cache_by_value_ref: Dict[_Ref, Tuple[Any, ...]] = {}  # value ref -> (func, args, ...)
+        self.added_sys_paths = set()
+        self.known_modules = set(known_modules)
+        for mod_name in known_modules:
+            mod = sys.modules[mod_name]
+            # Don't add those module path to sys.path again.
+            self.added_sys_paths.add(_get_module_path_from_module(mod))
+        self._next_sys_path_insert_idx = 0
+        self.sis_path_handling = sis_path_handling
+        self._cur_added_refs: List[PyCode] = []
+        self._next_assignment_idx = 0
+        # We first serialize everything without inlining anything.
+        # There we also count how often a value is used (ref_count).
+        # Then we can inline those values which are not direct config entries
+        # and which are only used once.
+        self._inlining_stage = False
+
+    def work_queue(self):
+        self._inlining_stage = False
+        queue: List[Union[_AssignQueueItem, _DeferredStateQueueItem]] = [
+            _AssignQueueItem(required_var_name=key, value=value)
+            for key, value in list(self.config.items()) + list(self.post_config.items())
+        ]
+        queue.reverse()  # we will pop from the end
+        while queue:
+            deferred_state: Optional[_DeferredStateQueueItem] = None
+            queue_item = queue[-1]
+            self._cur_added_refs.clear()
+            try:
+                if isinstance(queue_item, _AssignQueueItem):
+                    deferred_state = self._handle_next_queue_item(queue_item)
+                elif isinstance(queue_item, _DeferredStateQueueItem):
+                    self._handle_deferred_state_queue_item(queue_item)
+                else:
+                    raise TypeError(f"unexpected queue item type {type(queue_item).__name__}")
+                assert queue[-1] is queue_item
+                queue.pop(-1)
+            except _SerializationDependsOnNotYetSerializedOtherVarException as exc:
+                exc.queue_item.via_queue_item = queue_item
+                queue.append(exc.queue_item)
+                for code in self._cur_added_refs:
+                    code.ref_count -= 1
+            if deferred_state:
+                queue.append(deferred_state)
+
+    def work_inlining(self):
+        self._inlining_stage = True
+        self._next_assignment_idx = -1
+        for assign in list(self.assignments_dict_by_idx.values()):
+            assert assign.idx > self._next_assignment_idx
+            self._next_assignment_idx = assign.idx
+            if assign.py_name and not assign.has_later_state_setup:
+                new_assign = self._serialize_value_assignment(assign.value, name=assign.py_name)
+                assert isinstance(new_assign, PyCode)
+                assign.py_value_repr = new_assign.py_value_repr
+                assign.py_code = new_assign.py_code
+        self._next_assignment_idx += 1
+
+    def _handle_next_queue_item(self, queue_item: _AssignQueueItem) -> Optional[_DeferredStateQueueItem]:
+        value_ref = _Ref(queue_item.value)
+        if queue_item.required_var_name:
+            assert queue_item.required_var_name not in self.assignments_dict_by_name
+        if not queue_item.required_var_name and value_ref in self.assignments_dict_by_value_ref:
+            # No need to assign it again.
+            return None
+        name = queue_item.required_var_name
+        if not name and value_ref in _InternalReservedNamesByValueRef:
+            name = self._get_unique_suggested_name(
+                _InternalReservedNamesByValueRef[value_ref], allow_internal_reserved_name=True
+            )
+        if not name and (
+            isinstance(queue_item.value, (type, FunctionType, BuiltinFunctionType, ModuleType, Import, Call))
+            or (getattr(queue_item.value, "__module__", None) and getattr(queue_item.value, "__qualname__", None))
+            # or (isinstance(queue_item.value, Dim) and queue_item.value.name)
+        ):
+            # For those types, prefer a name based on the value, even over any other suggested name.
+            name = self._get_unique_suggested_name(self._suggest_name_from_value(queue_item.value))
+        if not name and queue_item.suggested_var_name:
+            name = self._get_unique_suggested_name(queue_item.suggested_var_name)
+        if not name:
+            name = self._get_unique_suggested_name(self._suggest_name_from_value(queue_item.value))
+        serialized = self._serialize_value_assignment(value=queue_item.value, name=name)
+        deferred_state: Optional[_DeferredStateQueueItem] = None
+        if isinstance(serialized, _PyCodeWithDeferredStateQueueItem):
+            serialized, deferred_state = serialized.code, serialized.extra
+            serialized.has_later_state_setup = True
+            deferred_state.via_queue_item = queue_item
+        assert isinstance(serialized, PyCode)
+        serialized.idx = self._next_assignment_idx
+        self._next_assignment_idx += 1
+        if queue_item.required_var_name:
+            serialized.is_direct_config_entry = True
+            if queue_item.required_var_name in self.config:
+                serialized.use_for_hash = True
+        assert serialized.py_name == name
+        assert name not in self.assignments_dict_by_name  # double check
+        self.assignments_dict_by_name[name] = serialized
+        if value_ref not in self.assignments_dict_by_value_ref:
+            self.assignments_dict_by_value_ref[value_ref] = serialized
+        value_dict = self.assignments_dict_by_value_by_type.get(type(queue_item.value))
+        if value_dict is not None:
+            if queue_item.value in value_dict:
+                if serialized.is_direct_config_entry:
+                    # Same reasoning as above for assignments_dict_by_value_ref.
+                    assert value_dict[queue_item.value].is_direct_config_entry
+            else:
+                value_dict[queue_item.value] = serialized
+        self.assignments_dict_by_idx[serialized.idx] = serialized
+        return deferred_state
+
+    def _handle_deferred_state_queue_item(self, rv: _DeferredStateQueueItem):
+        name = rv.py_name
+        value = rv.value
+        state = rv.state
+        listitems = rv.listitems
+        dictitems = rv.dictitems
+        state_setter = rv.state_setter
+        code_lines = []
+
+        if listitems is not None:
+            for i, item in enumerate(listitems):
+                item_s = self._serialize_value(item, prefix=f"{name}_listitem{i}", recursive=True)
+                code_lines.append(f"{name}.append({item_s.py_inline()})\n")
+
+        if dictitems is not None:
+            for key, v in dictitems:
+                serialized_key = self._serialize_value(key, prefix=f"{name}_key", recursive=True)
+                assert isinstance(serialized_key, PyEvalCode)
+                if (isinstance(key, str) and _is_valid_python_identifier_name(key)) or isinstance(key, (int, bool)):
+                    prefix_name = str(key)
+                else:
+                    prefix_name = "value"
+                serialized_value = self._serialize_value(v, prefix=f"{name}_{prefix_name}", recursive=True)
+                assert isinstance(serialized_value, PyEvalCode)
+                code_lines.append(f"{name}[{serialized_key.py_inline()}] = {serialized_value.py_inline()}\n")
+
+        if state is not None:
+            if state_setter is None:
+                # See pickle._Unpickler.load_build.
+                setstate = getattr(value, "__setstate__", None)
+                if setstate is not None:
+                    state_s = self._serialize_value(state, prefix=f"{name}_state", recursive=True)
+                    assert isinstance(state_s, PyEvalCode)
+                    code_lines.append(f"{name}.__setstate__({state_s.py_inline()})\n")
+                else:
+                    slotstate = None
+                    if isinstance(state, tuple) and len(state) == 2:
+                        state, slotstate = state
+                    if state:
+                        state_s = self._serialize_value(state, prefix=f"{name}_state", recursive=True)
+                        assert isinstance(state_s, PyEvalCode)
+                        code_lines.append(f"{name}.__dict__.update({state_s.py_inline()})\n")
+                    if slotstate:
+                        # not handled yet
+                        raise NotImplementedError(
+                            f"serialize {rv.py_name} = {rv.value!r} with slotstate {slotstate!r},"
+                            f" via {rv.debug_trace()}"
+                        )
+
+            else:
+                raise NotImplementedError  # not handled yet
+
+        code = PyCode(py_name=None, value=None, py_code="".join(code_lines))
+        code.idx = self._next_assignment_idx
+        self._next_assignment_idx += 1
+        self.assignments_dict_by_idx[code.idx] = code
+
+    @staticmethod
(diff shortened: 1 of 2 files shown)


serialization_v2/serialization.py
```diff
@@ -0,0 +1,1165 @@
...
+        return inspect.getsource(code)
+    raise RuntimeError("Could not serialize %s" % code)
+
+
+def _is_valid_python_identifier_name(name: str) -> bool:
+    """
+    :return: whether the name is a valid Python identifier name (including attrib name)
+    """
+
+    # Very hacky. I'm sure there is some clever regexp, but I don't find it and too lazy...
```
michelwi, line 182: https://docs.python.org/3/library/stdtypes.html#str.isidentifier ?

NeoLegends:
I brought back Dim serialization in a way where RETURNN is not imported in the manager.

serialization_v2/serialization.py
```diff
@@ -0,0 +1,1272 @@
...
+
+class _Serializer:
+    def __init__(
+        self,
+        config: Dict[str, Any],
+        post_config: Optional[Dict[str, Any]] = None,
+        known_modules: Collection[str] = (),
+        sis_path_handling: SisPathHandling = None,
+    ):
+        from returnn.tensor import Dim, batch_dim, single_step_dim
```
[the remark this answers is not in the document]
NeoLegends: You can now specify an explicit `returnn_root` that is added to `sys.path` before the serializer is instantiated, so that imports can be directed to the specified path.

serialization_v2/serialization.py
```diff
@@ -0,0 +1,1284 @@
...
+    *,
+    inlining: bool = True,
+    known_modules: Collection[str] = (),
+    extra_sys_paths: Sequence[str] = (),
+    returnn_root: Optional[Union[str, Path]] = None,
+    sis_path_handling: SisPathHandling = None,
+) -> SerializedConfig:
+    """serialize config. see module docstring for more info."""
+    if returnn_root is not None:
+        assert tk.running_in_worker(), "returnn_root only supported in sisyphus worker context"
```
albertz: Why? While we currently mostly would use it that way, I don't see that this is really a restriction that is necessary.

serialization_v2/serialization.py
```diff
@@ -0,0 +1,1284 @@
...
+
+
+def serialize_config(
+    config: Dict[str, Any],
+    post_config: Optional[Dict[str, Any]] = None,
+    *,
+    inlining: bool = True,
+    known_modules: Collection[str] = (),
+    extra_sys_paths: Sequence[str] = (),
+    returnn_root: Optional[Union[str, Path]] = None,
```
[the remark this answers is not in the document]
NeoLegends: This is for the other way around: if RETURNN is not within your `sys.path`, how do you use this module? The standard AppTek template works this way (see Haotian's comment), RETURNN is not importable from the recipes, and only used from the `ReturnnTrainingJob` via git clone job. Whether this is right or wrong is a discussion that I consider out-of-scope for this PR.

The serializer wants to import some RETURNN types/primitives like Dim so that those can be serialized properly, if they appear within the config. Although in that case you know that RETURNN was available in the recipes beforehand, so maybe I can make the imports lazy/only import them when the config to be serialized contains RETURNN objects.

serialization_v2/serialization.py
```diff
@@ -0,0 +1,1284 @@
...
+
+    def _instanciate_delayed_obj(o: Any) -> Any:
+        if isinstance(o, DelayedBase) and not isinstance(o, SerializerObject):
+            return o.get()
+        return o
+
+    return tree.map_structure(_instanciate_delayed_obj, o)
+
+
+def _parse_python(code: Any, name=None):
```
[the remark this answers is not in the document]
NeoLegends: This is no longer there

serialization_v2/serialization.py
```diff
@@ -0,0 +1,1272 @@
...
+
+
+def _is_valid_python_identifier_name(name: str) -> bool:
+    """
+    :return: whether the name is a valid Python identifier name (including attrib name)
+    """
+    return name.isidentifier() and not iskeyword(name)
+
+
+class ReturnnConfigWithNewSerialization(ReturnnConfig):
```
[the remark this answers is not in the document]
albertz: You could argue though that @Icemole is right, that this would supersedes the existing class in functionality. A V2 doesn't need to be compatible in functionality to V1.

serialization_v2/serialization.py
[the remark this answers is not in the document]
albertz: See my other comment on this. I think you can do this in a safe way (checking `obj.__class__.__module__.startswith("returnn.")` on objects). There should not be a `returnn_root` that you need to specify.

serialization_v2/serialization.py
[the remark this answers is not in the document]
albertz: To expand why I think this is doubly bad: This just adds another source for potential errors. And it adds complexity. And it is just not needed.

serialization_v2/serialization.py
```diff
@@ -0,0 +1,1243 @@
...
+        self._inlining_stage = False
+
+        # Avoid to use them, but if necessary (when inside the config), they can be used.
+        # The config keys always have precedence.
+        self._internal_reserved_names = {
+            "sys": sys,
+            # TODO: are these entries needed for correctness?
+            # "batch_dim": batch_dim,
+            # "single_step_dim": single_step_dim,
+            # "Dim": Dim,
```
albertz: I think these are not needed. I would however maybe adapt the code for suggesting the variable name to try to keep the original name here.

serialization_v2/serialization.py
```diff
@@ -0,0 +1,1243 @@
...
+                raise NotImplementedError  # not handled yet
+
+        code = PyCode(py_name=None, value=None, py_code="".join(code_lines))
+        code.idx = self._next_assignment_idx
+        self._next_assignment_idx += 1
+        self.assignments_dict_by_idx[code.idx] = code
+
+    @staticmethod
+    def _isinstance_returnn_dim(obj: Any) -> bool:
+        return obj.__class__.__name__ == "Dim" == obj.__class__.__module__.startswith("returnn.")
```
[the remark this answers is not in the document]
NeoLegends: Ah, yes. I also noticed the bug/typo by now due to failing tests. I like the suggestion! Applied.

serialization_v2/serialization.py
```diff
@@ -353,9 +353,11 @@ def _fill_internal_reserved_names_dict_with_known_globals(self, obj: Any):
 
         if _Serializer._isinstance_returnn_obj(obj):
             # obj is from returnn, so we're free to import its known globals now
-            from returnn.tensor import Dim, batch_dim, single_step_dim
+            from returnn.tensor import Dim, Tensor, batch_dim, single_step_dim
 
-            known_globals.update({"batch_dim": batch_dim, "single_step_dim": single_step_dim, "Dim": Dim})
+            known_globals.update(
+                {"batch_dim": batch_dim, "single_step_dim": single_step_dim, "Dim": Dim, "Tensor": Tensor}
+            )
```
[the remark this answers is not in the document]
albertz: That why I said I would adapt the code for the variable name, that you get these cases here. Maybe in general that classes, global funcs and other global objects always keep their original name.

NeoLegends:
<details>
<summary>Example config of reasonable complexity serialized with the latest changes:</summary>

```python
#!returnn/rnn.py

import sys

sys.path.insert(0, "/home/mgunz/setups/2025-03-07--ws-asr/recipe/apptek_asr/..")
import sys

sys.path.insert(
    0, "/home/mgunz/setups/2025-03-07--ws-asr/recipe/apptek_asr/../i6_models"
)
get_model = __import__("functools").partial(
    __import__(
        "apptek_asr.lib.pytorch.networks.conformer_aed",
        fromlist=["ConformerTransformerAedCtcModelV1"],
    ).ConformerTransformerAedCtcModelV1,
    **{
        "sampling_rate": 16000,
        "model_dim": 512,
        "out_dim": 10241,
        "num_heads": 8,
        "num_enc_layers": 12,
        "num_dec_layers": 6,
        "aux_loss_layers": (4, 8),
        "bos_idx": 0,
        "eos_idx": 1,
        "specaug_start": (10480, 31440, 52400),
        "use_rf_init": True,
        "share_embedding": True,
        "logits_bias": False,
        "aux_logits_bias": True,
    }
)


import sys

sys.path.insert(0, "/usr/local/bin")
task = "train"
target = "classes"
learning_rate_file = "learning_rates"
use_horovod = True
behavior_version = 21
file_cache_opts = {"cache_directory": "/ssd/$USER/returnn/file_cache"}
torch_dataloader_opts = {"num_workers": 1}
grad_scaler = None
torch_amp = "bfloat16"
torch_distributed = {"param_sync_step": 100, "reduce_type": "param"}
learning_rate_control = "constant"
batch_size = 3600000
max_seqs = 200
max_seq_length = {"classes": 250, "data": 720000}
from apptek_asr.lib.pytorch.training.optimizer import (
    conformer_aed_weight_decay_blacklist_v2,
)

optimizer = {
    "class": "adamw",
    "epsilon": 1e-16,
    "weight_decay": 0.01,
    "weight_decay_custom_include_check": conformer_aed_weight_decay_blacklist_v2,
}
gradient_clip_global_norm = 5.0
extern_data = {
    "data": {"dim": 1},
    "classes": {
        "dim": 10241,
        "sparse": True,
        "vocab": {
            "class": "SentencePieces",
            "model_file": "/nas/models/asr/artefacts/subword_units/sentencepiece/ES/2025-04-spm_10240-telephony-research-task/10240-nmt_nfkc_cf-sep.spm",
            "alpha": 0.7,
            "enable_sampling": True,
        },
    },
}
from returnn.util.file_cache import CachedFile
from returnn.tensor import Dim
from copyreg import __newobj__
from returnn.datasets.postprocessing import Sequential

eval_datasets_value_dataset_map_seq_stream = __newobj__(Sequential)
ExtractAudioFeaturesV1 = __import__(
    "apptek_asr.lib.data_postprocessing", fromlist=["ExtractAudioFeaturesV1"]
).ExtractAudioFeaturesV1(
    data_key="data",
    data_is_in_codec_form=False,
    sample_rate=16000,
    feat_extract_options={
        "features": "raw",
        "peak_normalization": True,
        "preemphasis": None,
        "pre_process": None,
    },
)
eval_datasets_value_dataset_map_seq_stream_state = {"funcs": (ExtractAudioFeaturesV1,)}
eval_datasets_value_dataset_map_seq_stream.__dict__.update(
    eval_datasets_value_dataset_map_seq_stream_state
)
T_dim_1 = Dim(None, name="T")
F_dim = Dim(1, name="F")
T_dim_2 = Dim(None, name="T")
C_dim = Dim(10241, name="C")
eval_datasets_value_1_dataset_map_seq_stream = __newobj__(Sequential)
eval_datasets_value_1_dataset_map_seq_stream_state = {
    "funcs": (ExtractAudioFeaturesV1,)
}
eval_datasets_value_1_dataset_map_seq_stream.__dict__.update(
    eval_datasets_value_1_dataset_map_seq_stream_state
)
eval_datasets = {
    "corpus.ES.f8kHz.es-filtered.Fisher-batch.2.dev.v2": {
        "class": "MultiProcDataset",
        "buffer_size": 800,
        "dataset": {
            "class": "PostprocessingDataset",
            "dataset": {
                "class": "MetaDataset",
                "datasets": {
                    "features": {
                        "class": "HDFDataset",
                        "files": [
                            CachedFile(
                                "/nas/data/speech/ES/8kHz/Fisher/corpus/batch.2.dev.v2/hdf-raw_wav.8kHz.split-1/Fisher-batch.2.dev.v2.hdf.1"
                            )
                        ],
                        "partition_epoch": 1,
                        "seq_ordering": "sorted_reverse",
                    },
                    "targets": {
                        "class": "PostprocessingDataset",
                        "dataset": {
                            "class": "HDFDataset",
                            "files": [
                                CachedFile(
                                    "/home/mgunz/setups/2025-03-07--ws-asr/work/apptek_asr/hdf/create/BlissToTextHdfJob.nMV2kJd36eqY/output/corpus.hdf.001"
                                )
                            ],
                            "partition_epoch": 1,
                        },
                        "map_outputs": {
                            "classes": {
                                "dims": [Dim(None, name="T")],
                                "sparse_dim": Dim(10241, name="classes"),
                                "vocab": {
                                    "class": "SentencePieces",
                                    "model_file": "/nas/models/asr/artefacts/subword_units/sentencepiece/ES/2025-04-spm_10240-telephony-research-task/10240-nmt_nfkc_cf-sep.spm",
                                    "alpha": 0.7,
                                    "enable_sampling": True,
                                },
                                "dtype": "int32",
                            }
                        },
                        "map_seq": __import__(
                            "apptek_asr.lib.data_postprocessing",
                            fromlist=["DecodeTextAndApplyVocabularyV1"],
                        ).DecodeTextAndApplyVocabularyV1(
                            data_key="data",
                            out_data_key="classes",
                            vocab_options={
                                "class": "SentencePieces",
                                "model_file": "/nas/models/asr/artefacts/subword_units/sentencepiece/ES/2025-04-spm_10240-telephony-research-task/10240-nmt_nfkc_cf-sep.spm",
                                "alpha": 0.7,
                                "enable_sampling": True,
                            },
                        ),
                    },
                },
                "data_map": {
                    "data": ("features", "data"),
                    "classes": ("targets", "classes"),
                },
                "seq_order_control_dataset": "features",
            },
            "map_seq_stream": eval_datasets_value_dataset_map_seq_stream,
            "map_outputs": {
                "data": {"dims": [T_dim_1, F_dim], "dtype": "float32"},
                "classes": {"dims": [T_dim_2], "sparse_dim": C_dim, "dtype": "int32"},
            },
        },
        "num_workers": 1,
        "sharding_method": "dedicated",
    },
    "corpus.ES_US.f8kHz.es-filtered.AAA-batch.1.v1": {
        "class": "MultiProcDataset",
        "buffer_size": 800,
        "dataset": {
            "class": "PostprocessingDataset",
            "dataset": {
                "class": "MetaDataset",
                "datasets": {
                    "features": {
                        "class": "HDFDataset",
                        "files": [
                            CachedFile(
                                "/nas/data/speech/ES_US/8kHz/AAA/corpus/batch.1.v1/20250414-hdf-raw_wav.16kHz.split-1/AAA-batch.1.v1.hdf.1"
                            )
                        ],
                        "partition_epoch": 1,
                        "seq_ordering": "sorted_reverse",
                    },
                    "targets": {
                        "class": "PostprocessingDataset",
                        "dataset": {
                            "class": "HDFDataset",
                            "files": [
                                CachedFile(
                                    "/home/mgunz/setups/2025-03-07--ws-asr/work/apptek_asr/hdf/create/BlissToTextHdfJob.SNlQA1HZoJne/output/corpus.hdf.001"
                                )
                            ],
                            "partition_epoch": 1,
                        },
                        "map_outputs": {
                            "classes": {
                                "dims": [Dim(None, name="T")],
                                "sparse_dim": Dim(10241, name="classes"),
                                "vocab": {
                                    "class": "SentencePieces",
                                    "model_file": "/nas/models/asr/artefacts/subword_units/sentencepiece/ES/2025-04-spm_10240-telephony-research-task/10240-nmt_nfkc_cf-sep.spm",
                                    "alpha": 0.7,
                                    "enable_sampling": True,
                                },
                                "dtype": "int32",
                            }
                        },
                        "map_seq": __import__(
                            "apptek_asr.lib.data_postprocessing",
                            fromlist=["DecodeTextAndApplyVocabularyV1"],
                        ).DecodeTextAndApplyVocabularyV1(
                            data_key="data",
                            out_data_key="classes",
                            vocab_options={
                                "class": "SentencePieces",
                                "model_file": "/nas/models/asr/artefacts/subword_units/sentencepiece/ES/2025-04-spm_10240-telephony-research-task/10240-nmt_nfkc_cf-sep.spm",
                                "alpha": 0.7,
                                "enable_sampling": True,
                            },
                        ),
                    },
                },
                "data_map": {
                    "data": ("features", "data"),
                    "classes": ("targets", "classes"),
                },
                "seq_order_control_dataset": "features",
            },
            "map_seq_stream": eval_datasets_value_1_dataset_map_seq_stream,
            "map_outputs": {
                "data": {"dims": [T_dim_1, F_dim], "dtype": "float32"},
                "classes": {"dims": [T_dim_2], "sparse_dim": C_dim, "dtype": "int32"},
            },
        },
        "num_workers": 1,
        "sharding_method": "dedicated",
    },
}
import functools

sys.path.insert(1, "/home/mgunz/setups/2025-03-07--ws-asr/recipe")
from apptek_asr.meta.returnn_experiments.subword_asr_v1 import (
    _get_subep_split_hdf_oggzip_dataset_v2,
)

train = {
    "class": "DistributeFilesDataset",
    "buffer_size": 800,
    "files": [
        (
            "/nas/data/speech/ES/8kHz/Fisher/corpus/batch.1.train.v2/20250414-hdf-raw_wav.16kHz.split-17/Fisher-batch.1.train.v2.hdf.1",
            "/home/mgunz/setups/2025-03-07--ws-asr/work/apptek_asr/hdf/create/BlissToTextHdfJob.C0qNk9mzWQQ6/output/corpus.hdf.001",
        ),
        (
            "/nas/data/speech/ES/8kHz/Fisher/corpus/batch.1.train.v2/20250414-hdf-raw_wav.16kHz.split-17/Fisher-batch.1.train.v2.hdf.2",
            "/home/mgunz/setups/2025-03-07--ws-asr/work/apptek_asr/hdf/create/BlissToTextHdfJob.C0qNk9mzWQQ6/output/corpus.hdf.002",
        ),
        # ...
    ],
    "get_sub_epoch_dataset": functools.partial(
        _get_subep_split_hdf_oggzip_dataset_v2,
        buf_size=200,
        label_decoding_postprocessors=None,
        num_workers=4,
        partition_epoch=1,
        postprocessing_functions=[
            ExtractAudioFeaturesV1,
            __import__(
                "apptek_asr.lib.data_postprocessing", fromlist=["SpeedPerturbationV1"]
            ).SpeedPerturbationV1(
                data_key="data",
                sample_rate=16000,
                factors={0.7: 1, 0.8: 1, 0.9: 1, 1.0: 1, 1.1: 1},
            ),
            __import__(
                "apptek_asr.lib.data_postprocessing", fromlist=["SeqConcatV1"]
            ).SeqConcatV1(
                classes_key="classes",
                classes_sep=3,
                data_key="data",
                max_num_seqs=3,
                max_seq_len=480000,
                num_features_per_class=None,
                num_seqs_dist=("geometric", 0.3),
                seq_tag_sep=":",
            ),
        ],
        postprocessor_map_outputs={
            "data": {"dims": [T_dim_1, F_dim], "dtype": "float32"},
            "classes": {"dims": [T_dim_2], "sparse_dim": C_dim, "dtype": "int32"},
        },
        ordering_postprocessing_function=__import__(
            "returnn.datasets.postprocessing", fromlist=["LaplaceOrdering"]
        ).LaplaceOrdering(num_seqs_per_bin=1000, length_key="data"),
        source_dataset_ordering="random",
        vocab_size=10241,
        vocab_config={
            "class": "SentencePieces",
            "model_file": "/nas/models/asr/artefacts/subword_units/sentencepiece/ES/2025-04-spm_10240-telephony-research-task/10240-nmt_nfkc_cf-sep.spm",
            "alpha": 0.7,
            "enable_sampling": True,
        },
    ),
    "partition_epoch": 6,
    "seq_ordering": "random",
}
device = "gpu"
log = ["./returnn.log"]
log_verbosity = 5
num_epochs = 600
save_interval = 1
cleanup_old_models = {
    "keep": [
        100,
        125,
        150,
        175,
        200,
        225,
        250,
        275,
        300,
        325,
        350,
        375,
        400,
        425,
        450,
        475,
        500,
        525,
        550,
        575,
        590,
        591,
        592,
        593,
        594,
        595,
        596,
        597,
        598,
        599,
        600,
    ]
}
debug_print_layer_output_template = True
log_batch_size = True
multiprocessing = True
backend = "torch"
use_lovely_tensors = True
log_grad_norm = True
use_train_proc_manager = True
torch_log_memory_usage = True
watch_memory = True
model = "/home/mgunz/setups/2025-03-07--ws-asr/work/i6_core/returnn/training/ReturnnTrainingJob.TaZhQFRcTP65/output/models/epoch"


dynamic_learning_rate = __import__("functools").partial(
    __import__(
        "apptek_asr.lib.pytorch.training.learning_rates",
        fromlist=["dyn_lr_piecewise_linear"],
    ).dyn_lr_piecewise_linear,
    **{
        "piecewise_epochs": [0.0, 270.0, 540.0, 600.0],
        "piecewise_values": [1e-05, 0.001, 1e-05, 1e-06],
    }
)

train_step = __import__("functools").partial(
    __import__(
        "apptek_asr.lib.pytorch.training.aed_ctc_train_step",
        fromlist=["aed_train_step"],
    ).aed_train_step,
    **{
        "aed_loss_scale": 1.0,
        "aux_loss_scales": (1.0, 1.0),
        "label_smoothing": 0.1,
        "label_smoothing_start_epoch": 0,
    }
)


# -*- mode: python; tab-width: 4 -*-
```

</details>

NeoLegends:
@albertz Look at the example config, and notice how `returnn.datasets.postprocessing.Sequential` is serialized. That currently goes through the pickle route. Do you think/have any idea how to make serialization for these simple, stateless classes prettier in a generic way?

I added some special-casing to serialize `CachedFile` better (because that really blows up the verbosity of your config depending on how you use it), and I can sure add some for `Sequential` as well, but at some point a more generic solution would be better. Can also be done in a follow-up though.

returnn/config.py
```diff
@@ -36,6 +37,54 @@ def __repr__(self):
         return self.code
 
 
+def unparse_python(code: Any, name: Optional[str] = None, hash_full_python_code: bool = False) -> str:
```
albertz: Why do you have that here?

albertz: Ah sorry, I see that we are now in `returnn/config.py`.

But then my question is, why did you move this code? Why did you touch `__parse_python` at all? This is not really relevant for this PR? Or is it? How?

serialization_v2/serialization.py
```diff
@@ -0,0 +1,1280 @@
...
+    return name.isidentifier() and not iskeyword(name)
+
+
+class ReturnnConfigWithNewSerialization(ReturnnConfig):
+    """
+    Overwrites the serialization behavior of ReturnnConfig.
+    """
+
+    @staticmethod
+    def from_cfg(old_returnn_cfg: ReturnnConfig):
```
albertz: I think this cannot be done in a safe way.

serialization_v2/serialization.py
```diff
@@ -0,0 +1,1280 @@
...
+    SerializerObject,
+)
+from i6_core.serialization import Collection as SerializerCollection
+
+if TYPE_CHECKING:
+    from returnn.tensor import Dim
+    from returnn.util.file_cache import CachedFile
+
+
+def serialize_config(
```
albertz: I wonder a bit whether we should make this private, i.e.:

```suggestion
def _serialize_config(
```

Because I never actually made use of this directly. It was always via `ReturnnConfigWithNewSerialization`.

(If we make this private, maybe we should also:
* move `ReturnnConfigWithNewSerialization` to the top here
* make `SerializedConfig` private
* adapt the module docstring

)

albertz:
There are some other things I noticed from your example. I will make separate comments for each (so that I can mark each as resolved individually).

```python
sys.path.insert(0, "/home/mgunz/setups/2025-03-07--ws-asr/recipe/apptek_asr/..")
```
The path here is not normalized. Should we maybe normalize it to `"/home/mgunz/setups/2025-03-07--ws-asr/recipe"`?

albertz:
```python
import sys
...
import sys
```
It seems that `import sys` is done again and again. That seems like a bug. That should not happen.

serialization_v2/serialization.py
```diff
@@ -0,0 +1,1280 @@
...
+        if mod_name in self.known_modules:
+            return
+        mod = sys.modules[mod_name]
+        if not hasattr(mod, "__file__"):
+            return  # assume builtin module or so
+        mod_path = _get_module_path_from_module(mod)
+        self.add_sys_path(mod_path)
+        self.known_modules.add(mod_name)
+
+    def add_sys_path(self, path: str, *, recursive: bool = True):
```
albertz:
```suggestion
    def add_sys_path(self, path: str, *, recursive: bool = True, code_comment: str = ""):
```
And then below add that `code_comment` as a prefix to the `PyCode` `py_code` string.

albertz: (This is an addition I made in the meantime after you made the PR.)

albertz:
I noticed, you have basically 4 times a dim with the name "T". Did you do that intentionally?
```python
T_dim_1 = Dim(None, name="T")
T_dim_2 = Dim(None, name="T")
```
and also two times `Dim(None, name="T")` deep inside some dict.

serialization_v2/serialization.py
```diff
@@ -0,0 +1,1280 @@
...
+        if isinstance(value, functools.partial):
+            return self._serialize_functools_partial(value, name)
+        if isinstance(value, Call):
+            return self._serialize_call(value, name)
+        if isinstance(value, CallImport):
+            return self._serialize_call_import(value, name)
+        if isinstance(value, PartialImport):
+            return self._serialize_partial_import(value, name)
+        if isinstance(value, Import):
+            return self._serialize_import(value, name)
```
albertz: Despite the discussion whether you should or should not use those here for `ReturnnConfigWithNewSerialization`, I wonder, why do you need to have this code here at all? Will you even ever get here? Because you have called `instanciate_delayed_copy` already on the config, which should have resolved all of that, or not?

serialization_v2/serialization.py
```diff
@@ -0,0 +1,1272 @@
...
+                serialized.as_serialized_code(),
+                python_epilog_code,
+                "# -*- mode: python; tab-width: 4 -*-\n",
+            ]
+            if part.strip()
+        )
+
+
+@dataclass
+class SerializedConfig:
```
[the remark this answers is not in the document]
albertz: I think my idea was at some point that I could reuse the existing `ReturnnConfig` and put this whole thing there into the python epilog or so. So then one would not need `ReturnnConfigWithNewSerialization`. But then, I added `ReturnnConfigWithNewSerialization`, and I found this to be much easier and cleaner to use.

So yes, we can maybe remove the whole thing, if you also don't think this is useful / necessary.

`serialize_config` could just return `list(serializer.assignments_dict_by_idx.values())` directly.

serialization_v2/serialization.py
```diff
@@ -0,0 +1,1280 @@
...
+        sis_path_handling: SisPathHandling = None,
+    ):
+        self.config = config.copy()
+        self.post_config = post_config.copy() if post_config else {}
+        self.assignments_dict_by_value_ref: Dict[_Ref, PyCode] = {}  # value ref -> code
+        self.assignments_dict_by_name: Dict[str, PyCode] = {}  # var name -> code
+        self.assignments_dict_by_idx: Dict[int, PyCode] = {}  # idx -> code
+        self.assignments_dict_by_value_by_type: Dict[type, Dict[Any, PyCode]] = {
+            # TODO: is this entry needed for correctness?
+            # Dim: {},
```
albertz: I think this is needed, because the `Dim` serialization is actually simplified, and drops some of the deeper structures, including the `same_as` stuff, and then two dim objects which are equal but not the same instance would not be equal anymore after the serialization.

Maybe you can add such code below:
```python
if "returnn" in sys.modules:
    from returnn.tensor import Dim

    self.assignments_dict_by_value_by_type[Dim] = {}
```

albertz:
Btw, another thing that I added, that was somewhat useful:
```python

_in_serialize_config_flag = False


def in_serialize_config() -> bool:
    """
    Whether we are currently inside a serialize_config call.
    """
    return _in_serialize_config_flag


@contextmanager
def _set_in_serialize_config_ctx(value: bool = True):
    global _in_serialize_config_flag
    old_value = _in_serialize_config_flag
    _in_serialize_config_flag = value
    try:
        yield
    finally:
        _in_serialize_config_flag = old_value
```

And in `serialize_config`, I used:
```python
...
with _set_in_serialize_config_ctx():
    serializer.work_queue()
    if inlining:
        serializer.work_inlining()
...
```

See my `i6_experiments.users.zeyer.returnn.delayed_dim.DelayedReduceDim` as an example for how I use this. This `DelayedReduceDim` is an alternative to the `_DelayedDim` that I see sometimes being used, but with one important difference: `_DelayedDim` (and other `DelayedBase` objects) are resolved via `instanciate_delayed_copy`. This happens usually in the `create_files` Sisyphus job task. `DelayedReduceDim` (and other uses of `in_serialize_config`) can implement the `__reduce__` logic in such a way that the actual `Dim` and its opts are only created really when RETURNN is run (in the `run` Sisyphus job task). You get the same effect also when you use `CallImport` and co in the Python prolog/epilog.

It would be good if we can also add this here.

serialization_v2/serialization.py
```diff
@@ -0,0 +1,1280 @@
...
+        if isinstance(value, Call):
+            return self._serialize_call(value, name)
+        if isinstance(value, CallImport):
+            return self._serialize_call_import(value, name)
+        if isinstance(value, PartialImport):
+            return self._serialize_partial_import(value, name)
+        if isinstance(value, Import):
+            return self._serialize_import(value, name)
+        if isinstance(value, (Checkpoint, PtCheckpoint)):
+            return self._serialize_checkpoint(value)
```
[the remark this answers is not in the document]
NeoLegends: Generally I expect `ReturnnConfig(config={"load", returnn_training_job.out_checkpoints[420]})` to work. And I think before this this caused some serialization issues.

But maybe it was a different bug/incorrectly specified path handling/delayed value instantiation. Let me check if this is still necessary.

serialization_v2/serialization.py
NeoLegends: See https://github.com/rwth-i6/i6_core/pull/601#issuecomment-3540972389

albertz:
> > The path here is not normalized. Should we maybe normalize it to "/home/mgunz/setups/2025-03-07--ws-asr/recipe"?
> 
> Again, I think this is an AppTek thing. It just injects `{path}/..` and does not normalize that concatenation

Yes, but we could normalize it here. So that was my question. Should we maybe normalize it? (Or you mean because this comes via python epilog, this cannot be done here? I was mostly thinking about the `sys.path` modifications that we do here.)

NeoLegends:
Yes this comes from python_prolog/python_epilog.

albertz:
> avoiding imports from `torch` in the mgr was to allow patterns like:
> 
> ```
> config = ReturnnConfig({
>   "dynamic_learning_rate": PartialImport(code_object_path="i6_experiments.my.lr.function", hashed_arguments={...})
> })
> ```
> 
> and then the serializer will assign the imported value to the config key that you specified.

Did you actually test this? Because, as said, currently I think it would run `instanciate_delayed_copy` before the serialization. And the `instanciate_delayed_copy` will replace the object just by a string. So the serialization logic will never see the `PartialImport` object? So this will just not work?

serialization_v2/serialization.py
albertz: I thought mostly about using Python epilog / prolog when making this comment. As you want to keep this, and you accept that this is not safe, then also this function here could stay.

But I would say, only if you really use it. Do you?

albertz:
> I generally agree now that the instantiate_delayed should go away, or rather that the functionality should be moved into the serializer. But then again this then introduces another break of the "behaves just like pickle" rule. Those delayed objects kind of break this rule anyway though.

But I argued that this functionality should *not* be part of the serializer. The serializer should just exactly behave like pickle.

NeoLegends:
Sorry I edited my comment now. Since they behave like a stand-in for a value that is lazily computed, I wonder what the value is in serializing the "stand-in" part if the value they stand for is "ready". I think we should simply serialize their value like pickle, but not the container (the delayed object) itself.

serialization_v2/serialization.py
albertz: To answer myself: This is because the `instanciate_delayed_copy` was also modified to be inconsistent to `instanciate_delayed` in that it would *not* handle `SerializerObject`.

albertz:
> This is an actual example.

No? `postprocessor` is not a valid RETURNN option. RETURNN will ignore this.

NeoLegends:
I read RETURNN sources and from them I see that if we have `load = PtCheckpoint(...)` then I expect it to break. E.g. the torch-engine calls `checkpoint.endswith(...)` and this function does not (obviously) exist on the checkpoint class.

Are you sure that serializing the checkpoints without going through their repr will work correctly? `instanciate_delayed` does not handle them I think.

NeoLegends:
> which again goes against the pickle principle.

Maybe this is not a problem after all since we _do_ resolve delayed objects before invoking the serializer, so does it really matter if the serializer does this as well?

NeoLegends:
> Yes. As said, this is the example I gave before for PtCheckpoint. And this is just one example. You can think of many similar examples.

Ah, yes, right. So that means we actually cannot have both some values be serialized properly and use delayed objects to instanciate them only when needed.

I am wondering if all those use cases only come up because we use these ill-defined wrapper objects. Maybe `(Pt)Checkpoint` should instead be refactored to work with any `PathLike` (which includes `str`) and not just `sisyphus.Path`, or maybe it should never become part of the config?

You say right now "keep the wrapper objects in the config, RETURNN can handle it", but is that a good default choice? Can it really handle these objects in all/most of the cases?  My experience tells me that we generally don't use the wrapper objects as "interface object" to other tools, but always use the raw underlying `str` representation of the paths. So people might expect something else. E.g. when writing RASR-configs we also special-case the wrapper objects and print the path as str 🤔.

And so the "flipped" solution to this problem is to special case _all_ the wrapper objects that we have (`Delayed*`, Path, Checkpoints) and never serialize them in their `i6_core`-/python-form, but always as `str`, and then possibly provide escape hatches the other way if people really want to preserve the pythonic/class representation.

serialization_v2/serialization.py
```diff
@@ -0,0 +1,1146 @@
...
+    def _serialize(self) -> str:
+        import tree
+
+        # This is usually run within the worker, but it shouldn't really matter.
+        assert not self.staged_network_dict  # not supported
+
+        self.check_consistency()
+
+        config = instanciate_delayed(self.config)
+        post_config = instanciate_delayed(self.post_config)
```
albertz:
```suggestion
        config = instanciate_delayed_copy(self.config)
        post_config = instanciate_delayed_copy(self.post_config)
```

serialization_v2/serialization.py
```diff
@@ -212,8 +169,8 @@ def _serialize(self) -> str:
 
         self.check_consistency()
 
-        config = _instanciate_delayed_copy(self.config)
-        post_config = _instanciate_delayed_copy(self.post_config)
+        config = instanciate_delayed(self.config)
+        post_config = instanciate_delayed(self.post_config)
```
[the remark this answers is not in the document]
NeoLegends: Why would it matter? We're running this in `create_files`, right?

NeoLegends:
> I lost track a bit: Do all V1 cases work properly now in V2? Or was there sth which does not work? What exactly does not work?

I think yes, since that was all prolog/epilog + `CodeWrapper`.

albertz:
> I just found out the TF checkpoint wrapper class does not have an `__fspath__` method, i.e. it is not `PathLike`. If you simply serialize that and want RETURNN to load it, it will break (currently).

I think this does not really matter for this PR? Then just let's fix the TF checkpoint wrapper.

serialization_v2/serialization.py
```diff
@@ -0,0 +1,1182 @@
...
+        if self.via_queue_item:
+            return (
+                f"{self.required_var_name or self.suggested_var_name} <{type(self.value).__name__}>"
+                f" -> {self.via_queue_item.debug_trace()}"
+            )
+        return f"config {self.required_var_name or self.suggested_var_name} = {self.value!r}"
+
+
+@dataclass
+class PyCode(SerializerObject):
```
albertz: Maybe this should be private?

```suggestion
class _PyCode(SerializerObject):
```

serialization_v2/serialization.py
```diff
@@ -0,0 +1,1135 @@
...
+
+        import sys  # assign `sys`
+    """
+
+    py_name: Optional[str]
+    value: Any
+    py_code: str
+    py_value_repr: Optional[_PyEvalCode] = None
+    is_direct_config_entry: bool = False
+    use_for_hash: bool = False
```
albertz: This `use_for_hash` is maybe not needed anymore?

```suggestion
```

serialization_v2/serialization.py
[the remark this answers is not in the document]
NeoLegends: I mean yeah, I removed most of it by now. There was a part in the module docstring about the hashing, some handling of whether some part of the PyCode should go into the hash or not and some hash tests. This is now all not there anymore.

NeoLegends:
> The path here is not normalized. Should we maybe normalize it to "/home/mgunz/setups/2025-03-07--ws-asr/recipe"?

This path came from an apptek artefact that injects the corresponding code into `python_prolog`, it did not come from the `serialization_v2` code.

I think in the V2 serialization we will never inject unnormalized paths into the config (because we don't do directory traversal there), and I'm hesitant to add logic to modify these manually injected paths. Actually I think it is difficult to do without actually parsing the prolog/epilog.

So I say no: don't do path normalization.

serialization_v2/serialization.py
albertz: Well, I mean, I did not suggest to remove it. I just wanted to understand the use case. But ok, if you think it's better to remove, then let's keep it removed for now.

serialization/serialization_v2.py
```diff
@@ -0,0 +1,1033 @@
...
+
+Allows serializing arbitrary Python objects inside the config
+including functions/classes/modules/objects,
+with automatic import handling.
+
+See :class:`i6_core.returnn.ReturnnConfigV2` for the :class:`ReturnnTrainingJob` (etc.) integration.
+
+See :func:`serialize_config` for a lower level entry point to config dict serialization.
+
+See https://github.com/rwth-i6/i6_experiments/blob/main/users/zeyer/serialization.rst
```
albertz:
```suggestion
See https://github.com/rwth-i6/i6_core/blob/main/serialization/README.rst
```

albertz (approved):
See my last small comments. I think otherwise all is fine for merging now.

returnn/config.py
```diff
@@ -366,6 +323,93 @@ def _sis_hash(self):
...
+        self.check_consistency()
+
+        config = instanciate_delayed_copy(self.config)
+        post_config = instanciate_delayed_copy(self.post_config)
+
+        # I'm not really sure about it.
+        # Our automatic mechanism will find direct imports (e.g. i6_experiments).
+        # However, it will not find indirect imports (e.g. sisyphus),
+        # and thus the generated code might fail.
+        # So add all other paths here which we currently have.
```
michelwi: Are we now sure? -> remove/rewrite comment otherwise, lets try to make sure

returnn/config.py
```diff
@@ -366,6 +323,93 @@ def _sis_hash(self):
...
+
+        # Handle ExternalImports
+        extra_sys_paths += [
+            item.import_path
+            for item in tree.flatten(self.python_prolog) + tree.flatten(self.python_epilog)
+            if isinstance(item, ExternalImport)
+        ]
+
+        python_prolog_code = unparse_python(self.python_prolog, hash_full_python_code=True)
+        python_epilog_code = unparse_python(self.python_epilog, hash_full_python_code=True)
```
michelwi: why is `hash_full_python_code` hardcoded here? It seems the only thing it does is prevent us from using `DelayedBase` objects in prolog/epilog without actually affecting the hash

```suggestion
        python_prolog_code = unparse_python(self.python_prolog, hash_full_python_code=self.hash_full_python_code)
        python_epilog_code = unparse_python(self.python_epilog, hash_full_python_code=self.hash_full_python_code)
```

of course, if the parameter does not make sense in the new implementation, then I would rather add an
```
assert hash_full_python_code == True
```
to the `__init__` of `ReturnnConfigV2`.

But then I am wondering that hashing and serialization should be independent of each other.. so what is the case here?

EDIT: I read below
> There is a :class:`SerializerObject` base class, deriving from Sisyphus :class:`DelayedBase`.

so actually the `# Handle ExternalImports` part does not make sense if we disallow them in the prolog/epilog.

serialization/serialization_v2.py
```diff
@@ -0,0 +1,1032 @@
...
+
+    def _serialize_reduce(self, value: Any, name: str) -> Union[_PyEvalCode, _PyCodeWithDeferredStateQueueItem]:
+        # Generic fallback using __reduce__ or __reduce_ex__.
+        # This is very much following the original pickle logic (slightly simplified).
+        value_ref = _Ref(value)
+        if value_ref in self.reduce_cache_by_value_ref:
+            rv = self.reduce_cache_by_value_ref[value_ref]
+        else:
+            reduce = getattr(value, "__reduce_ex__", None)
+            reduce_proto = 4  # not sure...
```
michelwi: > ` # not sure...`

why not? can we make sure?

albertz: Yea, sounds good to change the comment like that.

NeoLegends:
🎉

PR #603, opened 2025-05-14
Recording alignment confidence filtering job

Icemole:
We have a filtering job for alignments whose average confidence score is greater than a threshold specified by the user.

This PR proposes the `FilterSegmentsByRecordingAlignmentConfidenceJob`, which:
1. Obtains the average confidence score of each recording (simply defined as the average of the segment alignment scores that compose each recording).
2. Filters out a recording (that is, filters out all segments from the recording) if the recording confidence score is greater than the threshold specified by the user.

### DIFF
--- a/corpus/filter.py
+++ b/corpus/filter.py
@@ -2,6 +2,7 @@ __all__ = [
     "FilterSegmentsByListJob",
     "FilterSegmentsByRegexJob",
     "FilterSegmentsByAlignmentConfidenceJob",
+    "FilterSegmentsByRecordingAlignmentConfidenceJob",
     "FilterCorpusBySegmentsJob",
     "FilterCorpusRemoveUnknownWordSegmentsJob",
     "FilterCorpusBySegmentDurationJob",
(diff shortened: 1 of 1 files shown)


corpus/filter.py
```diff
@@ -209,6 +211,111 @@ def run(self):
                 segment_file.write(segment + "\n")
 
 
+class FilterSegmentsByRecordingAlignmentConfidenceJob(Job):
```
michelwi, line 214: I think quite a lot of code could be shared with FilterSegmentsByAlignmentConfidenceJob

Icemole: Yes. There are several options here:

1. In `FilterSegmentsByAlignmentConfidenceJob` I've considered adding a boolean flag `filter_whole_recordings` or so, with default value `False` (same behavior as right now), but if `True` it applies the filtering in `FilterSegmentsByRecordingAlignmentConfidenceJob`. However, I felt like it was too distinct from the original purpose of the job.
2. Another option could be abstracting the code into a common function to be used by both jobs.

I prefer 1. Let me know what you think.

Icemole: Implementing 1.

Icemole: Done. I'll try to test the new functionality soon. Please let me know what you think.

Icemole: Note: https://github.com/rwth-i6/i6_core/pull/603/commits/e4a31ba231749c485a4b41c7ee6ff91dee47c847 is the commit that changes from one implementation to another. Both can be easily compared :)

sarahberanek (approved):

sarahberanek:
I was thinking if `FilterRecordingsByAlignmentConfidenceJob` is a better fit then `FilterSegmentsByRecordingAlignmentConfidenceJob` ... but then the output is a segment list 🤔 ?

Icemole:
> I was thinking if `FilterRecordingsByAlignmentConfidenceJob` is a better fit then `FilterSegmentsByRecordingAlignmentConfidenceJob` ... but then the output is a segment list 🤔 ?

Yeah, that's a good point. Both names are valid in my view. We could use `FilterRecordingsByAlignmentConfidenceJob` for a name which is a bit shorter (and we could also output the recording list, if that bothers anyone :slightly_smiling_face:), but I don't have a strong preference towards either.

sarahberanek (approved):

PR #604, opened 2025-05-14
DownloadAndPrepareHuggingFaceDatasetJob, trust_remote_code, some doc ext

albertz:


datasets/huggingface.py
```diff
@@ -81,6 +90,7 @@ def run(self):
                 cache_dir=tmp_dir,
                 split=self.split,
                 token=self.token,
+                **({"trust_remote_code": self.trust_remote_code} if self.trust_remote_code is not None else {}),
```
michelwi, line 93: why do we make it so complicated instead of passing the value directly?

According to [the documentation](https://huggingface.co/docs/datasets/v3.6.0/en/package_reference/loading_methods#datasets.load_dataset.trust_remote_code) it seems to me that even `None` can be given, although I would be fine to pass `False` by default (Based in my semantic interpretation of the parameter).
Or are we afraid that someone might be using a version of `datasets` prior to 2.16. where the parameter did not exist?

albertz: Yes, it's for backward compat.

albertz: This job was introduced April 2022. datasets 2.16 is from Dec 2023. datasets 2.20 is from June 2024.

michelwi (approved):

NeoLegends (approved):

PR #605, opened 2025-05-19
`ComputeTimeStampErrorJob`: allow multiple caches as input, add output file with individual TSEs per segment pair

Icemole:
1. Added option to input multiple caches instead of a single one. The TSE will be accumulated for each cache file provided.
2. Added output file for sorted (descending) list of highest TSEs.

### DIFF
--- a/mm/alignment.py
+++ b/mm/alignment.py
@@ -644,102 +644,105 @@ class ComputeTimeStampErrorJob(Job):
         start_differences = Counter()
         end_differences = Counter()
         differences = Counter()
+        tse_dict: Dict[str, Tuple[str, float]] = {}  # ref_seg_name: (hyp_seg_name, avg_tse)
 
-        hyp_alignments = rasr_cache.open_file_archive(self.hyp_alignment_cache.get())
-        hyp_alignments.setAllophones(self.hyp_allophone_file.get())
-        if isinstance(hyp_alignments, rasr_cache.FileArchiveBundle):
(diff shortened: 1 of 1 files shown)


mm/alignment.py
```diff
@@ -747,6 +755,15 @@ def run(self) -> None:
...
         self.out_tse_frames.set(statistics.mean(abs(diff) for diff in differences.elements()))
+        with util.uopen(self.out_highest_tse_differences_file.get_path(), "wt") as f:
+            for ref_seq_tag, (hyp_seq_tag, avg_tse) in sorted(
+                tse_dict.items(), key=lambda k_v: k_v[1][1], reverse=True
+            ):
+                if ref_seq_tag == hyp_seq_tag:
+                    # No need to be redundant in the output format.
+                    f.write(f"{ref_seq_tag}\t{avg_tse}\n")
+                else:
+                    f.write(f"{ref_seq_tag}\t{hyp_seq_tag}\t{avg_tse}\n")
```
michelwi, line 766: I would not make this decision on a per-segment level. The chances of one tag being different while others are the same might be low, but it would completely destroy the parseabillity of the output.. in fact, for better parsing I would just stick to one format anyway even if it is redundand.

Icemole: That's a very fair point.

michelwi (approved):
It would be great if you could be more precise in your final commit message than just "Improvements" ;)

kuacakuaca (approved):

PR #606, opened 2025-05-20
Add job to write dict contents to a csv file

Icemole:
The `WriteToTextFileJob` wasn't having the format (generic csv) that I wanted, so I decided to write another job for that.

Note: there's a [`TextDictToTextLinesJob`](https://github.com/rwth-i6/i6_core/blob/main/text/convert.py#L12), but it doesn't seem to handle actual dictionaries but rather some RETURNN format.

### DIFF
--- a/text/processing.py
+++ b/text/processing.py
@@ -17,7 +17,7 @@ import shutil
 import subprocess
 from collections.abc import Iterable
 import tempfile
-from typing import List, Optional, Union
+from typing import Dict, List, Optional, Union
 
 from sisyphus import Job, Task, Path, global_settings as gs, toolkit as tk
 from sisyphus.delayed_ops import DelayedBase
@@ -340,7 +340,10 @@ class WriteToCsvFileJob(WriteToTextFileJob):
     __sis_hash_exclude__ = {}  # It was filled in the base class, but it's not needed anymore since this is a new job.
 
     def __init__(
-        self, content: Union[str, dict, Iterable, DelayedBase], out_name: str = "file.txt", delimiter: str = "\t"
+        self,
+        content: Dict[str, Union[str, List[str]]],
+        out_name: str = "file.txt",
+        delimiter: str = "\t",
     ):
         """
         :param content: input which will be written into a text file
@@ -361,8 +364,10 @@ class WriteToCsvFileJob(WriteToTextFileJob):
         content = util.instanciate_delayed(self.content)
         if isinstance(content, dict):
             for key, val in content.items():
-                csv_writer.writerow((key, val))
-
+                if isinstance(val, list):
+                    csv_writer.writerow((key, *val))
+                else:
+                    csv_writer.writerow((key, val))
         else:
             raise NotImplementedError("Content of unknown type different from (str, dict, Iterable).")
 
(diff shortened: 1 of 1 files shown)


text/processing.py
```diff
@@ -296,22 +308,62 @@ def __init__(self, content: Union[str, dict, Iterable, DelayedBase], out_name: s
...
+        """
+        Writes the input contents (from `self.content`) into the file provided as parameter as a csv file.
+
+        :param file_handler: Open file to write the contents of `self.content` to.
+        """
+        csv_writer = csv.writer(file_handler, delimiter=self.delimiter)
+        content = util.instanciate_delayed(self.content)
+        if isinstance(content, dict):
+            for key, val in content.items():
+                csv_writer.writerow((key, val))
```
sarahberanek, line 363: what happens if value is a list?

text/processing.py
```diff
@@ -296,22 +308,62 @@ def __init__(self, content: Union[str, dict, Iterable, DelayedBase], out_name: s
...
+    """
+    Write a given content into a csv file, one entry per line.
+
+    This job only supports dictionaries as input type. Each key/value pair is written as `<key><delimiter><value>`.
+    """
+
+    __sis_hash_exclude__ = {}  # It was filled in the base class, but it's not needed anymore since this is a new job.
+
+    def __init__(
+        self, content: Union[str, dict, Iterable, DelayedBase], out_name: str = "file.txt", delimiter: str = "\t"
```
kuacakuaca, line 342: the type should be updated.

text/processing.py
Icemole: It can't be a list, the job only supports dictionary for now (see above).

We can discuss whether we want to support more data, and if so what would be the way to add it.

If we're supporting lists: maybe something like this?

```
for line in content:
    csv_writer.write(f"{self.delimiter.join(line)}\n")
```

Icemole: The job supports lists as values now.

text/processing.py
```diff
@@ -339,7 +340,10 @@ class WriteToCsvFileJob(WriteToTextFileJob):
     __sis_hash_exclude__ = {}  # It was filled in the base class, but it's not needed anymore since this is a new job.
 
     def __init__(
-        self, content: Union[str, dict, Iterable, DelayedBase], out_name: str = "file.txt", delimiter: str = "\t"
+        self,
+        content: Dict[str, Union[str, List[str]]],
```
albertz, line 344:
```suggestion
        content: Dict[str, Union[str, List[str]]],
        *,
```

Icemole:
One could argue whether we're programming `WriteToTextFileJob` with a custom delimiter. Of course, that's not the main issue, since the user can easily do this with a simple `delimiter.join(...)` and converting the values to be written by the base `WriteToTextFileJob` into a list.

This job mainly has the purpose of adapting to any `csv.reader` call. While this could be partially fulfilled by the `WriteToTextFileJob`, the issue is that the base job doesn't escape any conflictive input, and that must be done by the user instead.

---

Another question that might arise is the arbitrary length of the list value in `Dict[str, List[str]]`, because then the columns could be of an arbitrary length. I'd leave this responsibility to the user.

michelwi (approved):

PR #607, opened 2025-05-20
Add `BlissToAudioHDFJob`

NeoLegends:
See #593.

I think @Icemole also wants to use this for raw wave HDFs on another large corpus.


returnn/hdf.py
```diff
@@ -346,6 +360,288 @@ def run(self):
...
+                )
+            except Exception as exc:
+                if i < max_tries - 1:
+                    _logging.warning(f"Failed to process audio {audio_file}: {exc}")
+                else:
+                    _logging.warning(f"Failed to process audio {audio_file} after {max_tries} attempts. Skipping.")
+        return []
+
+    @staticmethod
+    def __process_seq(
```
albertz, line 557: Double underscore `__` has a reserved special meaning in Python. You should not use it.

returnn/hdf.py
```diff
@@ -346,6 +360,288 @@ def run(self):
...
+                    output_dtype=output_dtype,
+                    rounding_strategy=rounding_strategy,
+                    round_factor=round_factor,
+                    target_sampling_rate=target_sampling_rate,
+                )
+            except Exception as exc:
+                if i < max_tries - 1:
+                    _logging.warning(f"Failed to process audio {audio_file}: {exc}")
+                else:
+                    _logging.warning(f"Failed to process audio {audio_file} after {max_tries} attempts. Skipping.")
```
albertz, line 553: This also doesn't seem to be a safe solution, to just skip this? E.g. if `ffmpeg` or so is not correctly installed, it would fail for every seq, and then you get an empty corpus, without error?

Or if the behavior is non-deterministic (I assume so if you try 3 times; why 3 times?), then it means the result of this job here will be non-deterministic, if you just skip it then. This is not good.

returnn/hdf.py
```diff
@@ -346,6 +360,288 @@ def run(self):
...
+                + channel_mix_args
+                + [
+                    "-af",
+                    "aresample=resampler=soxr",
+                    "-ar",
+                    str(target_sampling_rate),
+                    audio_file_out,
+                ],
+                stdout=sp.DEVNULL,
+                stderr=sp.DEVNULL,
```
albertz, line 606: Why? Wouldn't it be good to see errors? I guess you don't want to spam your log? Maybe just add `-hide_banner -loglevel error` to the args instead?

returnn/hdf.py
```diff
@@ -346,6 +360,288 @@ def run(self):
...
+                    segments,
+                    compress_format=compress_format,
+                    compress_factor=compress_factor,
+                    multi_channel_strategy=multi_channel_strategy,
+                    output_dtype=output_dtype,
+                    rounding_strategy=rounding_strategy,
+                    round_factor=round_factor,
+                    target_sampling_rate=target_sampling_rate,
+                )
+            except Exception as exc:
```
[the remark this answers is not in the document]
NeoLegends: I think I had some random FS hiccups, but I removed this for now and will re-run and then only handle the specific cases.

returnn/hdf.py
```diff
@@ -346,6 +360,288 @@ def run(self):
...
+                        raise NotImplementedError(f"RoundingScheme {rounding_strategy} not implemented.")
+
+                    audio.seek(start)
+                    data = audio.read(duration, always_2d=False, dtype=output_dtype)
+                    assert data.ndim == 1
+
+                    if compress_format is not None:
+                        assert compress_factor is not None
+                        buf = io.BytesIO()
+                        sf.write(
```
albertz, line 594: I wonder a bit about the logic: You first use ffmpeg to convert to wav, and then `soundfile` to convert it to `compress_format`? Why not use ffmpeg directly to convert it to `compress_format`?

returnn/hdf.py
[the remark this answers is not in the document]
NeoLegends: So I think the problem that this job is facing is that when the corpora are so large, there just happen to be some corrupted files in the data that you can do nothing about but skip.

returnn/hdf.py
```diff
@@ -399,6 +399,7 @@ def __init__(
         rounding: BlissToPcmHDFJob.RoundingScheme = BlissToPcmHDFJob.RoundingScheme.rasr_compatible,
         round_factor: int = 1,
         target_sampling_rate: int = 16000,
+        skip_on_error: bool = False,
```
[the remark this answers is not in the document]
NeoLegends: Hm, so maybe this option should be there but then only apply to known exceptions?

returnn/hdf.py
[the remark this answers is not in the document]
albertz: Hm looks bad. But in any case, I would not add such workaround to skip errors. The job should run deterministically. Now it seems if you maybe install a new libflac/libsndfile version later, you would maybe get a different set of segments in the HDF? That should be avoided.

So either you find a working safe workaround here, or we should not use soundfile here for writing flac files. Or you can just make some `assert compression_format != "flac"` because FLAC is broken, linking to the issues. Do you actually use FLAC or sth else? I thought you use ogg?

returnn/hdf.py
```diff
@@ -346,6 +360,274 @@ def run(self):
...
+                    (full_name, segment.start, segment.end)
+                    for segment in rec.segments
+                    if (full_name := segment.fullname()) in all_whitelists
+                ]
+            )
+            > 0
+        )
+        # use imap instead of imap_unordered to have reproducible results, even though this
+        # causes a slight speed penalty
+        for results in pool.imap(apply_func, recs, chunksize=8):
```
NeoLegends, line 511: @albertz Since the users need to use the split files to recover the file ordering anyway, WDYT about switching this to `imap_unordered` for more speed? This will result in the HDFs having their segments ordered nondeterministically though, so two instances of the same job may not produce exactly the same HDFs, just "equivalent" HDFs.

albertz: It should be deterministic.

returnn/hdf.py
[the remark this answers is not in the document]
albertz: I think this option should not be there. Unless you have a reliable way to check for the exception, without even using soundfile/ffmpeg, just based on the raw file content, which works always, no matter what version is used, which is 100% guaranteed to be deterministic, no matter the soundfile/ffmpeg versions. But I doubt that this is possible.

returnn/hdf.py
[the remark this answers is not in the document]
NeoLegends: Yeah, I do use ogg, it still happens. So this error report is only remotely related.

returnn/hdf.py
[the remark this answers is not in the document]
NeoLegends: I wonder how this can be approached then. The only other way I see would be filtering the data/segment lists from the segments that error. But then you only know which segments fail to process when you do try to process them. It seems overkill to have another job just "try" the processing out just to produce a segment list of recordings that do not fail. Then you can also just create the HDFs and record all failing recordings in a separate output variable.

returnn/hdf.py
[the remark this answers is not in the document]
albertz: Have you tried with newer libsndfile / libvorbis / libogg (whatever else is involved)?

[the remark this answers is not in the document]
albertz: Otherwise, to me the conclusion would be, we just cannot use `soundfile`.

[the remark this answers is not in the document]
albertz: This `psf_fseek` error happens during reading or writing?

You are reading a wav file here. I wonder how this can get wrong?

But I anyway wonder how this function can fail. Maybe some memory corruption in libsndfile. Which would be very bad, and which means we cannot really rely on it (anything can happen then; it sometimes works, sometimes does not, sometimes would just put random noise, or silence, or whatever).

For just reading a wav file, and also doing this cutting of the segment, we can simply use the `wave` package.

returnn/hdf.py
albertz: Ok I see. This should be documented in the code.

Note (as written also in the other thread), for the slicing, you don't need `soundfile`, you an use the `wave` package. So if you really can't do reliable slicing with FFmpeg (but I somehow doubt that also...), a solution could be:

* FFmpeg to wav
* Slice with `wave` package, save again as wav
* Optional: FFmpeg to target format

returnn/hdf.py
```diff
@@ -346,6 +360,274 @@ def run(self):
...
+            # We send every segment of the recording grouped together with the audio to
+            # avoid having to read the audio file more than once.
+            #
+            # We only pass down the required audio/segment metadata instead of the whole
+            # `i6_core.lib.corpus.Segment` object to avoid excessive pickling overhead,
+            # stalling the worker processes (if we don't).
+            (rec.audio, segments)
+            for rec in c.all_recordings()
+            if len(
+                segments := [
```
Icemole, line 508: The walrus operator was introduced in python 3.8, so we should probably state this somewhere around the code, right? Are there other operations in i6_core as a whole that assume that the code is for python 3.8+?

returnn/hdf.py
```diff
@@ -346,6 +360,274 @@ def run(self):
...
+            ):
+                if segment_name not in segments_whitelist:
+                    continue
+                _logging.info(f"Writing {segment_name} to {hdf_writer.filename}.")
+                hdf_writer.insert_batch(
+                    inputs=data.reshape(1, -1, 1),
+                    seq_len=[data.shape[0]],
+                    seq_tag=[segment_name],
+                )
+        for hdf_writer in hdf_writers:
```
Icemole, line 530: As a suggestion, instead of manually closing the files, you can do:
```
with contextlib.ExitStack() as stack:
    hdf_writers: List[TextIO] = [
        SimpleHDFWriter(filename=out_hdf.get_path(), dim=1) for out_hdf in out_hdfs
    ]
    ...
```
And when you exit the context stack, the files are automatically closed.

The downside is another level of indentation. Maybe you could create the `hdf_writers`  further down to alleviate that. It's up to you.

returnn/hdf.py
```diff
@@ -346,6 +360,274 @@ def run(self):
...
+        multi_channel_strategy: BlissToPcmHDFJob.BaseStrategy,
+        output_dtype: str,
+        rounding_strategy: BlissToPcmHDFJob.RoundingScheme,
+        round_factor: int,
+        target_sampling_rate: int,
+    ) -> List[Tuple[str, np.ndarray]]:
+        audio_file, segments = recording
+        _logging.info(f"Processing {audio_file} with {len(segments)} segments.")
+
+        assert isinstance(multi_channel_strategy, (BlissToPcmHDFJob.Mixdown, BlissToPcmHDFJob.PickNth)), (
```
Icemole, line 565: You're already catching this in `__init__`, I don't think there's a need to catch it here as well.

If this is because the method is static, see also my comment above regarding whether static methods are needed.

returnn/hdf.py
```diff
@@ -346,6 +360,274 @@ def run(self):
...
+                    "-ar",
+                    str(target_sampling_rate),
+                    audio_file_out,
+                ],
+                stdout=sp.DEVNULL,
+            )
+            with sf.SoundFile(audio_file_out) as audio:
+                for segment_name, start, end in segments:
+                    if rounding_strategy == BlissToPcmHDFJob.RoundingScheme.start_and_duration:
+                        start = int(start * audio.samplerate / round_factor) * round_factor
```
Icemole, line 607: At this point I'd think about extrapolating these into functions in `lib/`. They seem to be too relevant for them to be reimplemented at each job. Maybe not at this PR if you don't want, but let's keep that in mind.

returnn/hdf.py
Icemole: Right now my experience with the base `BlissToPcmHDFJob` is that it's running at an (optimistic) xRTF of 5.0, since it takes a bit more than 2 hours to process a split composed of 10 hours of audio data.

If @NeoLegends managed to have an xRTF of 5000, that's a win in my book...

... as long as the cascaded conversion is correct and it doesn't induce any sampling errors... I don't know how the code would behave if resampling from some format to wav was required, maybe we'd lose a few frames in the way. @NeoLegends have you tried this?

returnn/hdf.py
```diff
@@ -346,6 +360,274 @@ def run(self):
...
+                _logging.info(f"Writing {segment_name} to {hdf_writer.filename}.")
+                hdf_writer.insert_batch(
+                    inputs=data.reshape(1, -1, 1),
+                    seq_len=[data.shape[0]],
+                    seq_tag=[segment_name],
+                )
+        for hdf_writer in hdf_writers:
+            hdf_writer.close()
+
+    @staticmethod
```
Icemole, line 533: What's the advantage of having these methods set as static? They won't be used from outside of the class, right?

If you have instance methods here, you have the advantage of knowing that some things like the `multi_channel_strategy` are properly set, you don't have to use `partial` and you don't have to pass so many arguments around. In fact, as I see in [this line](https://github.com/rwth-i6/i6_core/pull/607/files#diff-1b4a30ead91f576041505969c0f243f61369eb50d60fbfc702f24359c71822e6R480), you're passing everything that already belongs to `self`, so I think you could simplify the code if you make these instance methods (assuming I haven't skipped some reasoning, which I might have).

returnn/hdf.py
```diff
@@ -346,6 +360,274 @@ def run(self):
...
+            segment_whitelists.append(segments_whitelist)
+
+        assert len(segment_whitelists) == len(out_hdfs)
+        all_whitelists = set.union(*segment_whitelists)
+
+        c = corpus.Corpus()
+        c.load(self.bliss_corpus.get_path())
+
+        pool = multiprocessing.Pool(max(self.rqmt["cpu"] - 1, 1))
+        apply_func = partial(
```
Icemole, line 487: See my comment below regarding static methods.

TL;DR: if `_process_seq_handle_errors` was an instance method, you could save this call to `partial` and manually specifying the arguments.

returnn/hdf.py
albertz: I agree. If there are so far no requirements for Python >=3.8, we should not introduce this now. Rather, we should open a separate issue about it, discuss it, then somewhere explicitly specify the Python version we require, and then we can follow this. (But this PR here does not need to wait for that. You can just rewrite the code without walrus operator for now.)

returnn/hdf.py
albertz: You forgot sth like:
```python
for w in hdf_writers:
    stack.callback(w.close)
```
But I don't really see how this makes the code much easier. It just looks more convoluted and indirect (harder to understand) instead.

returnn/hdf.py
albertz: It's using multiprocessing here. Maybe otherwise you have serialization issues? But not sure. @NeoLegends did you actually try?

albertz: If you tried, and there are problems, are these problems not solvable? If not, you should add at least some comment here.

returnn/hdf.py
albertz: Ah, actually, we do require Python 3.8 in i6_core. See https://github.com/rwth-i6/i6_core/pull/409. (But it could be argued that this is not well documented, which is a separate issue.)

returnn/hdf.py
Icemole: No, it's not directly related to it, but I thought your worries about the logic were a higher waiting time for the job to finish.

returnn/hdf.py
```diff
@@ -346,6 +366,278 @@ def run(self):
...
+                        raise NotImplementedError(f"RoundingScheme {rounding_strategy} not implemented.")
+
+                    audio.seek(start)
+                    data = audio.read(duration, always_2d=False, dtype=output_dtype)
+                    assert data.ndim == 1
+
+                    if compress_format is not None:
+                        assert compress_factor is not None
+                        buf = io.BytesIO()
+                        sf.write(
```
Icemole, line 622: Right now my experience with the base `BlissToPcmHDFJob` is that it's running at an (optimistic) xRTF of 5.0, since it takes a bit more than 2 hours to process a split composed of 10 hours of audio data.

If @NeoLegends managed to have an xRTF of 5000, that's a win in my book...

... as long as the cascaded conversion is correct and it doesn't induce any sampling errors... I don't know how the code would behave if resampling from some format to wav was required, maybe we'd lose a few frames in the way. @NeoLegends have you tried this?

returnn/hdf.py
```diff
@@ -346,6 +360,274 @@ def run(self):
...
+        self.skip_on_error = skip_on_error
+
+        self.out_hdfs = [self.output_path(f"{i + 1:0d}.hdf") for i in range(len(splits))]
+
+        num_cores = min(len(splits) + 1, 16)
+        mem_per_core = 1
+        self.rqmt = {"cpu": num_cores, "mem": num_cores * mem_per_core + 2, "time": 48}
+
+    def tasks(self):
+        yield Task("run", rqmt=self.rqmt, args=range(self.concurrent))
```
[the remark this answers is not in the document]
Icemole: > entirely doable

Let me give more details: as I see it, you'd have to wrap [this line](https://github.com/rwth-i6/i6_core/pull/607/files#diff-1b4a30ead91f576041505969c0f243f61369eb50d60fbfc702f24359c71822e6R525) with some kind of `if` statement that checks whether there's already a sequence with the segment name defined in the writer, something like the following:
```
if not hdf_writer.has_batch([segment_name]):
    hdf_writer.insert_batch(
        inputs=data.reshape(1, -1, 1),
        seq_len=[data.shape[0]],
        seq_tag=[segment_name],
    )
```

Update: I've been reviewing the [`SimpleHDFWriter`](https://github.com/rwth-i6/returnn/blob/master/returnn/datasets/hdf.py#L1070) code, and there doesn't seem to be a check that the segment name was inserted. Besides, `self._seq_tags` seems to be a list.

Maybe you can do the checks yourself by passing around some `inserted_segment_names: Set[str] = set()`.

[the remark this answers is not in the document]
Icemole: I see more issues with the `resumable` approach. To keep the state of the HDF files on disk, we should close them when an exception is triggered. But then:
1. You can't be sure that you're closing the HDF when another worker is writing to it as well.
3. If you do the checks yourself that a sequence has been inserted into the HDF, then that whole context is lost, and you have to start from scratch.

For a solution to (2), you could create the initial set as `set(SimpleHDFWriter._seq_tags)`.

A solution to (1) is more complicated. I think we'd need some kind of function that runs by only one worker after a sisyphus worker sees an exception and all workers have synchronized (so that they all know that the job is going to fail). I guess that's not implemented yet.

returnn/hdf.py
```diff
@@ -346,6 +366,278 @@ def run(self):
...
+
+    def run(self, index: int):
+        out_hdfs = list(util.chunks(self.out_hdfs, self.concurrent))[index]
+        splits = list(util.chunks(self.splits, self.concurrent))[index]
+        assert len(out_hdfs) == len(splits)
+        _logging.info(f"Writing audio to {out_hdfs} from {splits}.")
+
+        SimpleHDFWriter = get_returnn_simple_hdf_writer(self.returnn_root.get_path())
+        hdf_writers = [SimpleHDFWriter(filename=out_hdf.get_path(), dim=1) for out_hdf in out_hdfs]
```
Icemole, line 473: Couldn't you have race condition issues here when writing with different workers into a single file? Does anyone know if `SimpleHDFWriter` manages writes atomically?

returnn/hdf.py
Icemole: @albertz you're right, it's because of serialization issues: see https://stackoverflow.com/a/21345308. `pickle` can't dump non-static methods, we've had this for a while in sisyphus as well. We should use a third party library called `pathos`, which has a `multiprocessing` module that uses `dill` inside, which is able to serialize "almost anything". Then it would be supported, but I guess we don't want to pick that route.

There are other homemade solutions in the thread for Stack Overflow post for managing a set of workers, but these seem complicated as well. So I definitely understand why @NeoLegends wanted to set these functions as static.

returnn/hdf.py
albertz: No, I worried about unnecessary complexity (having FFmpeg + soundfile instead of just FFmpeg). And I worried about unnecessary additional dependencies (soundfile could maybe be avoided). It just looked weird to me, and I did not understand why soundfile was needed here. @NeoLegends answers that. But then I gave another suggestion to achieve the same without needing soundfile.

returnn/hdf.py
albertz: Why do you make the comment here at this point in the code? Is this related to the use of soundfile here?

returnn/hdf.py
albertz: > `pickle` can't dump non-static methods

That's wrong. Of course it can. This works:
```python
import pickle

class Foo:
    def __init__(self, v):
        self.value = v

    def func(self):
        return self.value
    
x = Foo(42)

f = pickle.loads(pickle.dumps(x.func))
assert f() == 42
```

The StackOverflow issue you link is about something else. It's about that these are *local* functions, which are not accessible from the global module namespace. That does not work. You just have to make them globally accessible. Same for all involved classes (`Foo` in the example). But that's all.

returnn/hdf.py
albertz: But the workers don't write into it?

returnn/hdf.py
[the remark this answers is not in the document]
albertz: There is no problem with having it resumable. Just let it start over from the beginning. The HDFs are anyway not in place, because it writes first to some temp file, and only in `SimpleHDFWriter.close`, it will move it over to the real destination.

returnn/hdf.py
```diff
@@ -322,6 +334,8 @@ def run(self):
                 data = audio.read(duration, always_2d=True, dtype=self.output_dtype)
                 if isinstance(self.multi_channel_strategy, self.PickNth):
                     data = data[:, self.multi_channel_strategy.channel]
+                elif isinstance(self.multi_channel_strategy, self.Mixdown):
+                    data = np.sum(data, axis=-1)
```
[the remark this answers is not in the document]
albertz: While there are certainly multiple ways of doing this, I think it's fine to just leave the given approach here (or any other one, eg if you say that `np.mean` makes in general more sense?), and not have multiple options. Or maybe just let this be handled to FFmpeg. You can just pass `-ac 1` to FFmpeg. (Actually not sure what FFMpeg does then without any further options, but I guess it does sth reasonable.)

returnn/hdf.py
NeoLegends: Ah, I think I just feared pickling issues and went with a static right away. I can check if I can make this instance methods.

returnn/hdf.py
NeoLegends: Yes, I preferred to double-check this.

returnn/hdf.py
Icemole: Well, sort of. I guess my comment would be better off in the `ffmpeg` call, but it's actually related to the combination of `ffmpeg` and `soundfile` to extract features. I want to make sure that the `ffmpeg` process doesn't remove any relevant samples by converting an audio format into another, and then `soundfile` does a wrong interpretation of the features, thus leading to an off-by-one error when training.

We've had off-by-one errors in the past, and they're very annoying to deal with because we're removing segments from the training that are completely ok because of a software issue, so I'd like to ensure that this doesn't happen here as well.

albertz: Ah ok. I somehow interpret your comment as: "It's nice to have this huge speedup (better xRTF). But of course only if it is correct." Which is of course nice, but the comments here are about some specific suggestions to change sth in the code, or at least to discuss this, so I somehow failed to see what your specific suggestion/comment here is about. I'm still not exactly sure. Of course we should make sure that it is correct. Do you think it's not? Or is it? This code here? Or sth else? Or are you just asking @NeoLegends whether he verified this?

returnn/hdf.py
Icemole: I guess we lose nothing by trying. @NeoLegends can you try to set this up if possible? Thanks!

returnn/hdf.py
Icemole: Well, when `close()` is triggered, then the file is moved + copied into the respective location, as shown [here](https://github.com/rwth-i6/returnn/blob/master/returnn/datasets/hdf.py#L1413).

However, I failed to realize that each HDF file is called `close()` once as per the `chunks()` call at the beginning of the `run` task. So there will be no issues with multiple workers writing to the same file on disk, and therefore this won't be an issue.

returnn/hdf.py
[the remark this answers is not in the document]
Icemole: > if you say that np.mean makes in general more sense?

I'm not sure, that's why I proposed the function to be customizable, maybe with a default. In my mind, summing artificially increases the amplitude of the waveform, while taking the average might also artificially decrease the amplitude of the waveform. Either way, I guess that the Fourier transform wouldn't care about this, except maybe for the wave's power, which is also output by the Fourier transform, if I remember correctly. So this doesn't seem to be a really big deal, and we can leave it as `sum` (unless anyone has some other point regarding this).

> You can just pass `-ac 1` to FFmpeg.

Yes, that's already done, see [here](https://github.com/rwth-i6/i6_core/pull/607/files#diff-1b4a30ead91f576041505969c0f243f61369eb50d60fbfc702f24359c71822e6R569).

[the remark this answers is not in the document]
albertz: > Yes, that's already done, see [here](https://github.com/rwth-i6/i6_core/pull/607/files#diff-1b4a30ead91f576041505969c0f243f61369eb50d60fbfc702f24359c71822e6R569).

But this is `BlissToAudioHDFJob`. Here we talk about `BlissToPcmHDFJob`. I actually confused this as well. Here in `BlissToPcmHDFJob`, there is no FFmpeg at all currently, right?

But then I would not add this code at all, unless we are 100% sure that it is consistent to the behavior of `BlissToAudioHDFJob`. Maybe/probably `BlissToAudioHDFJob` FFmpeg `-ac 1` does sth different than this code here.

I would just throw an exception here, "currently not supported" or sth like that, with a comment pointing to BlissToAudioHDFJob.

returnn/hdf.py
albertz: But the workers don't even get the HDF writers at all? I don't really understand your comment about `close()`.

> So there will be no issues with multiple workers writing to the same file on disk

But workers never at all access the HDFs. That happens all here in the parent process, in the `run` func.

returnn/hdf.py
```diff
@@ -346,6 +366,278 @@ def run(self):
...
+        self.rounding = rounding
+        assert round_factor > 0
+        self.round_factor = round_factor
+        self.returnn_root = returnn_root
+        self.target_sampling_rate = target_sampling_rate
+        self.skip_on_error = skip_on_error
+
+        self.out_hdfs = [self.output_path(f"{i + 1:0d}.hdf") for i in range(len(splits))]
+
+        num_cores = min(len(splits) + 1, 16)
```
michelwi, line 459: Why does the `num_cores` scale with the `len(splits)`?
It seems to me that if at all it should rather depend on `len(splits)/concurrent` but then making that ratio larger would just make the job take more time ~and we should instead always just use as many cores as ffmpeg can comfortably handle.~ 
EDIT: I saw below this is used by multiprocessing in the python process and have a separate thread for that now.

NeoLegends:
So I just attempted to use FFmpeg for compression as well, calling it as subprocess, passing in audio to its stdin and reading from stdout. Everything is done purely over pipes, no disk I/O.

It's much slower than using soundfile for compression. For some reason the CPU usage is much lower, which is strange for a task that should be CPU bound. I also tried increasing the pipe buffer size (to 1MB), but this doesn't help either.

I believe the workflow should therefore be:
1. FFmpeg: input -> wav
2. wave: wav -> samples
3. optional: soundfile: samples -> compression

albertz:
I doubt that FFMpeg is slower than libsndfile. That cannot be. Sth must be wrong.

Maybe it's the overhead due to the pipes?

In any case, I don't really think you have any other option. From our previous discussion, it seems like soundfile is not an option at all (due to non-deterministic behavior, random crashes).

NeoLegends:
I think for compression soundfile is fine.

returnn/hdf.py
```diff
@@ -346,6 +366,278 @@ def run(self):
...
+                segments_whitelist = {line.strip() for line in f if len(line.strip()) > 0}
+            segment_whitelists.append(segments_whitelist)
+
+        assert len(segment_whitelists) == len(out_hdfs)
+        all_whitelists = set.union(*segment_whitelists)
+
+        c = corpus.Corpus()
+        c.load(self.bliss_corpus.get_path())
+
+        pool = multiprocessing.Pool(max(self.rqmt["cpu"] - 1, 1))
```
michelwi, line 486: I am wondering why we have two levels of parallelism: 
* run `concurrent` many sisyphus tasks on possibly different nodes
* run `self.rqmt["cpu"]-1` many workers for multiprocessing

could we achieve similar speed by just using one form of parallelism?

albertz:
> Still strange to me that giving so much buffer to them doesn't speed things up.

Why should the buffer help here? Maybe it's actually an issue that the data does not arrive fast enough. If CPU usage is lower than it should be, then this clearly shows the bottleneck is somehow in your pipes, or the way you handle them.

> I think for compression soundfile is fine

Well, I asked before (you did not answer yet), the errors/crashes you potentially get (psf_fseek error) happens during reading or writing? You mentioned ogg in that context, so I thought it is about writing/compression. But now you say the opposite?

If you say your proposed way works without any issues/crashes and is deterministic, then ok, my biggest criticism is resolved. But I still think having only a single dep (FFmpeg) would be somehow nicer than to use both FFmpeg + soundfile. And also FFmpeg has much more options for the compression and about possible formats.

What is your FFmpeg compression command actually?

returnn/hdf.py
michelwi: I think @Icemole is referring to a solution like
```
with ExitStack() as stack:
    files = [stack.enter_context(open(fname)) for fname in filenames]
    # All opened files will automatically be closed at the end of
    # the with statement, even if attempts to open files later
    # in the list raise an exception
```
which would only work/be more elegant, if the `SimpleHDFWriter` was a contextmanager... _looks at implementation_ oh it is. then one can do
```
with contextlib.ExitStack() as stack:
    hdf_writers: List[TextIO] = [
        stack.enter_context(SimpleHDFWriter(filename=out_hdf.get_path(), dim=1)) for out_hdf in out_hdfs
    ]
    ...
```
and closing would be done implicitly upon leaving the context

albertz: Yes. But I don't really see how this makes the code much easier. It just looks more convoluted and indirect (harder to understand) instead.

returnn/hdf.py
NeoLegends: There are two levels here because with every unit of `concurrent` I/O becomes less efficient, as we can end up reading a single audio file multiple times.

Within one concurrency unit the code avoids reading the audio sources more than once and is therefore as I/O efficient as can be. As the corpora grow, however, the single job instance (with multiprocessing) becomes too little and then you can eventually start to tolerate some I/O inefficiency.

albertz:
> I don't think there's great harm in using [soundfile] for compression here as well [instead of FFmpeg]

Well, as said, two things:

* I don't trust soundfile, after you were mentioning about these errors. When there happen random crashes/errors while reading a wav file, something very serious is wrong (you cannot really mess up reading a wav file; it must be some of the hidden structures, maybe some memory corruption?), which very likely will also affect writing. Maybe it just does not lead to a crash but just corrupt data (random noise in it, silence in it, missing some parts, whatever...). And that is likely non-deterministic, maybe also rare, and if there is no crash/error, then you won't even notice it. Did you test this actually? I.e. this job multiple times, and get identical output? Or you can show me the actual bug in the libsndfile/soundfile code and show that this will not affect compression/writing. If you cannot show this, I think it is reasonable to expect that there are problems as well.

* FFmpeg has much more options, and supports more formats.

returnn/hdf.py
NeoLegends: It in fact works as instance method.

NeoLegends:
So, this now uses only FFmpeg, and I'm doing a test run to see if there are any failures now.

returnn/hdf.py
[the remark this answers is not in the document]
albertz: > Not that this is very relevant for the discussion here.

Well, it is to some degree. My argument is, `Mixdown` should behave the same for both `BlissToAudioHDFJob` and `BlissToPcmHDFJob`. With BlissToAudioHDFJob, you use FFmpeg `-ac 1`, here you use `np.sum`. Do you know that this has exactly the same effect? If not, then you should not use `Mixdown` here. If you don't know, then I also would not implement it that way without verifying that. Just throw an exception "this is unsupported" or so.

returnn/hdf.py
```diff
@@ -610,31 +620,42 @@ def _process_segment(segment: Tuple[str, float, float]):
...
+                            "-",
+                        ],
+                        bufsize=2**20,
+                        check=True,
+                        capture_output=True,
+                        input=data_bytes,
+                        timeout=60,
+                    )
+                    ffmpeg_out = ffmpeg_proc.stdout
+                except sp.ChildProcessError as e:
```
albertz, line 653: Same question here. Why this? Why not directly output it?

NeoLegends: See other comment.

returnn/hdf.py
```diff
@@ -551,29 +551,39 @@ def _process_seq(self, recording: Tuple[str, Sequence[Tuple[str, float, float]]]
...
+                    str(self.target_sampling_rate),
+                    "-f",
+                    "wav",
+                    "-",
+                ],
+                check=True,
+                capture_output=True,
+            )
+            ffmpeg_out = ffmpeg_proc.stdout
+        except sp.CalledProcessError as e:
```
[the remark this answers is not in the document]
albertz: Just don't set `capture_output`? Instead just set `stdout=PIPE`.

returnn/hdf.py
```diff
@@ -346,6 +364,297 @@ def run(self):
...
+                        str(c_level),
+                        "-f",
+                        c_format,
+                        "-",
+                    ],
+                    bufsize=2**20,
+                    check=True,
+                    input=data_bytes,
+                    stdout=sp.PIPE,
+                    timeout=max(30, 5 * (len(data_bytes) / self.target_sampling_rate)),
```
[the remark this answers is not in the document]
albertz:
```suggestion
```

returnn/hdf.py
```diff
@@ -346,6 +364,297 @@ def run(self):
...
+            The value of `level` depends on the codec and container used.
+            For e.g. libvorbis, compression level ranges from -1 to 10 (inclusive) where -1 is the
+            highest compression and 10 the lowest.
+        :param multi_channel_strategy: defines what should happen to multi-channel audio files.
+            Currently implemented are:
+            Mixdown(): Mix down all channels to one channel, default.
+            PickNth(n): Takes audio from n-th channel.
+        :param num_workers: Number of workers used for parallel recording processing.
+            It can be increased to e.g. match the number of CPU cores on a big cluster machine,
+            and the job will stay I/O efficient.
```
albertz, line 437: I would maybe reorder this that `num_workers` comes after `concurrent`.

returnn/hdf.py
```diff
@@ -346,6 +364,297 @@ def run(self):
...
+                        "-f",
+                        "s16le",
+                        "-i",
+                        "-",
+                        "-c:a",
+                        c_codec,
+                        "-qscale:a",
+                        str(c_level),
+                        "-f",
+                        c_format,
```
albertz, line 634: I think this is too specific, and too less generic, and too complicated.
This `qscale` is codec specific. And often, you would not use this.
E.g., for mp3, I saw that you would maybe use `-c:a libmp3lame -b:a 128k` instead (no `-qscale:a`, also no `-f`). Or for Ogg/Vorbis, `-c:a libvorbis -b:a 64k`.
Or maybe AAC, with VBR: `-c:a libfdk_aac -vbr 3`.

So, instead of this here, I would just put some `*ffmpeg_output_args`.
```suggestion
                        *self.ffmpeg_output_args,
```

returnn/hdf.py
```diff
@@ -346,6 +364,297 @@ def run(self):
...
+        self.rounding = rounding
+        assert round_factor > 0
+        self.round_factor = round_factor
+        self.returnn_root = returnn_root
+        self.target_sampling_rate = target_sampling_rate
+
+        self.out_hdfs = [self.output_path(f"{i + 1:0d}.hdf") for i in range(len(splits))]
+
+        self.rqmt = {
+            "cpu": 1 + (num_workers // 2),
```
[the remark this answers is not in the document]
NeoLegends: Because the job is mainly I/O limited, not compute limited. So it doesn't make sense to request as many CPU cores as there are workers, they mostly just wait for the disk.

returnn/hdf.py
```diff
@@ -346,6 +364,297 @@ def run(self):
...
+                        "-",
+                        "-c:a",
+                        c_codec,
+                        "-qscale:a",
+                        str(c_level),
+                        "-f",
+                        c_format,
+                        "-",
+                    ],
+                    bufsize=2**20,
```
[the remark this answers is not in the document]
NeoLegends: I wanted to make sure the compressor process is not limited by pipe I/O during its work (remember: it's quite slow, not maxing its CPU core), which is why I set the large buffer here. I guess since this doesn't make a difference I can remove the large buffer here again.

returnn/hdf.py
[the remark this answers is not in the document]
NeoLegends: But then it's still a bit arbitrary to have `// 2`, as other numbers may be just as realistic there. Maybe I'll just decouple these two numbers and let the user specify them manually.

returnn/hdf.py
NeoLegends: Oh, yeah. Good point. Implemented.

returnn/hdf.py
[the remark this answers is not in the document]
NeoLegends: No, this has changed now.

returnn/hdf.py
```diff
@@ -346,6 +364,296 @@ def run(self):
...
+            Within job concurrency is handled by the multiprocessing library, using `num_workers`
+            as parallelism factor.
+            Note that within-job concurrency is more I/O efficient than between-job concurrency,
+            so prefer increasing `num_workers` over increasing `concurrent`, when possible.
+        :param cpu_rqmt: How many CPUs to assign to the job.
+            The job is mainly I/O limited, so it's okay to assign fewer CPUs than the number of worker processes.
+        :param mem_rqmt: How much memory to assign to the job.
+        :param num_workers: Num subprocs (multiprocessing.Pool size) used for parallel recording processing.
+            It can be increased to e.g. match the number of CPU cores on a big cluster machine,
+            and the job will stay I/O efficient.
```
albertz, line 450: Now you have both `cpu_rqmt` and `num_workers`? Do you really need that? My suggestion was to either have `cpu_rqmt` or to use `num_workers` (but definitely not both). Whatever of that you choose, the code should reflect that exactly. I.e. if you specify `num_workers`, that should be what the pool size will be exactly in the end, but `rqmt["cpu"]` can be derived from `num_workers`. If you specify `cpu_rqmt`, then that would be exactly the `rqmt["cpu"]`, and you can infer `num_workers` somehow from it. Both options would be fine with me. But you somehow mixed everything up (it was neither the num workers nor the cpu rqmt).

returnn/hdf.py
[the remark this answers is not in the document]
albertz: My comment ("calling the variable num_workers is wrong") was referred to my earlier comment ("why //2") which was referred to the code you had before you changed it.

returnn/hdf.py
NeoLegends: I exposed both values because I think it is difficult to have a value that properly represents the relation between the number of worker procs and the CPU requirement that is correct for all use cases.

albertz: But before, you also only had one option.

I think there should be a reasonable default for the other value when you specify one of it. E.g. when you only specify `num_workers`, it should set a reasonable cpu rqmt based on it (and not the hardcoded value you have now).

And then, also I don't think you need `cpu_rqmt` as an option here. For other things, for the rare case that you really want to overwrite this, you can simply do:
```python
job = BlissToAudioHDFJob(...)
job.rqmt["cpu"] = ...
```

@michelwi made a very similar argument [here](https://github.com/rwth-i6/i6_core/pull/599#discussion_r2090792950).

I think now it's just unnecessary complicated (having two options) and also not reasonable that cpu rqmt are independent of num_workers by default.

returnn/hdf.py
```diff
@@ -346,6 +364,296 @@ def run(self):
...
+        audio_file, segments = recording
+
+        assert isinstance(self.multi_channel_strategy, (BlissToPcmHDFJob.Mixdown, BlissToPcmHDFJob.PickNth)), (
+            "unknown multi_channel_strategy"
+        )
+        channel_mix_args = (
+            ["-ac", "1"]
+            if isinstance(self.multi_channel_strategy, BlissToPcmHDFJob.Mixdown)
+            else ["-map_channel", f"0.{self.multi_channel_strategy.channel}"]
+        )
```
albertz, line 550: Btw, I wonder a bit: You expect that the user of `BlissToAudioHDFJob` always wants to get a single audio channel out? But at the same time, you allow to write any custom audio format into the HDF. Doesn't it make sense to also allow for multiple channels?

I can see two further reasonable cases here:
* `multi_channel_strategy is None` -> `channel_mix_args = []`, i.e. do nothing, leave it as it is
* `multi_channel_strategy` is sth like `CustomFfmpegChannelArgs` (or so, maybe better name) -> `channel_mix_args = multi_channel_strategy.channel_mix_args`, i.e. just specify any custom args

returnn/hdf.py
```diff
@@ -346,6 +364,296 @@ def run(self):
...
+                )
+        for hdf_writer in hdf_writers:
+            hdf_writer.close()
+
+    def _process_seq(self, recording: Tuple[str, Sequence[Tuple[str, float, float]]]) -> List[Tuple[str, np.ndarray]]:
+        audio_file, segments = recording
+
+        assert isinstance(self.multi_channel_strategy, (BlissToPcmHDFJob.Mixdown, BlissToPcmHDFJob.PickNth)), (
+            "unknown multi_channel_strategy"
+        )
```
michelwi, line 545:
```suggestion
        if not isinstance(self.multi_channel_strategy, (BlissToAudioHDFJob.Mixdown, BlissToPcmHDFJob.PickNth)):
            raise NotImplementedError(f"multi_channel_strategy {self.multi_channel_strategy} is not supported by BlissToAudioHDFJob")
```
But better move this into the constructor so that we fail already at graph creation time and not only at runtime

returnn/hdf.py
```diff
@@ -346,6 +364,296 @@ def run(self):
...
+                "aresample=resampler=soxr",
+                "-ar",
+                str(self.target_sampling_rate),
+                "-f",
+                "wav",
+                "-",
+            ],
+            check=True,
+            stdout=sp.PIPE,
+        )
```
michelwi, line 583: remove training comma to let black collapse the lists:
```suggestion
        ffmpeg_proc = sp.run(
            ["ffmpeg", "-hide_banner", "-loglevel", "error", "-y", "-i", audio_file, "-c:a", "pcm_s16le"]
            + channel_mix_args
            + ["-af", "aresample=resampler=soxr", "-ar", str(self.target_sampling_rate), "-f", "wav", "-"],
            check=True,
            stdout=sp.PIPE,
        )
```

returnn/hdf.py
NeoLegends: There already is an assertion in `__init__`, but I improved the messages for now.

returnn/hdf.py
NeoLegends: Yeah, this is mainly because the other job doesn't allow for multiple channels either at the moment. I agree, we can in any case be more relaxed about this in the future :+1:

returnn/hdf.py
```diff
@@ -346,6 +364,276 @@ def run(self):
...
+
+    @dataclass(frozen=True)
+    class Mixdown(BlissToPcmHDFJob.BaseStrategy):
+        """Multi channel strategy that instructs FFmpeg to mix down all channels to one channel."""
+
+    def __init__(
+        self,
+        bliss_corpus: tk.Path,
+        splits: Sequence[tk.Path],
+        *,
```
albertz, line 399:
```suggestion
        *,
        splits: Sequence[tk.Path],
```

returnn/hdf.py
```diff
@@ -346,6 +364,276 @@ def run(self):
...
+
+        assert concurrent > 0
+        self.concurrent = concurrent
+        assert num_workers > 0
+        self.num_workers = num_workers
+
+        self.out_hdfs = [self.output_path(f"{i + 1:0d}.hdf") for i in range(len(splits))]
+
+        self.rqmt = {
+            "cpu": 1 + (num_workers // 3),  # 3 workers per CPU core + 1 for the main process
```
albertz, line 477: Before, you had:
```suggestion
            "cpu": 1 + (num_workers // 2),  # 2 workers per CPU core + 1 for the main process
```
Why now `//3` instead?

albertz (approved):
Despite my last smaller comments, I think this is fine now.

returnn/hdf.py
NeoLegends: Observing the job I noticed //3 reflects the actual CPU usage better than //2.

Icemole (changes requested):

returnn/hdf.py
```diff
@@ -346,6 +364,281 @@ def run(self):
...
+    Can optionally write compressed audio data to the HDF.
+
+    See:
+        - https://github.com/rwth-i6/i6_core/pull/593 for discussion,
+        - https://github.com/rwth-i6/i6_core/pull/593#issuecomment-2883024538 for why
+          this job is faster than `BlissToPcmHDFJob`.
+    """
+
+    @dataclass(frozen=True)
+    class Mixdown(BlissToPcmHDFJob.BaseStrategy):
```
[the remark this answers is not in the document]
albertz: You are asking why `Mixdown` is defined here? -> The base class does not support this. This is only supported here in the new class.

Or you are asking why we did not implement the mixdown method also in `BlissToPcmHDFJob`? -> Because it's not obvious how to do it (in a way that is consistent to  `BlissToAudioHDFJob`).

returnn/hdf.py
```diff
@@ -346,6 +364,281 @@ def run(self):
...
+                        "-i",
+                        "-",
+                        *self.ffmpeg_output_args,
+                        "-",
+                    ],
+                    check=True,
+                    input=data_bytes,
+                    stdout=sp.PIPE,
+                )
+                data = np.frombuffer(ffmpeg_proc.stdout, dtype=np.uint8)
```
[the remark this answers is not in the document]
NeoLegends: Yes, uint8 is the correct choice, as written in the parameter comment for `ffmpeg_output_args`.

returnn/hdf.py
```diff
@@ -346,6 +364,281 @@ def run(self):
...
+        *,
+        ffmpeg_output_args: Optional[Sequence[str]] = None,
+        multi_channel_strategy: Optional[BlissToPcmHDFJob.BaseStrategy] = None,
+        output_dtype: Optional[str] = "int16",
+        returnn_root: Optional[tk.Path] = None,
+        rounding: BlissToPcmHDFJob.RoundingScheme = BlissToPcmHDFJob.RoundingScheme.rasr_compatible,
+        round_factor: int = 1,
+        target_sampling_rate: int = 16000,
+        concurrent: int = 1,
+        num_workers: int = 21,
```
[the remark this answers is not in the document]
NeoLegends: I was first rejecting your proposal, but actually it doesn't hurt much to reduce the default here.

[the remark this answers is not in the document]
NeoLegends: Reducing to 9 workers, which means the job will grab 4 CPU cores by default.

Icemole (approved):

PR #608, opened 2025-06-05
`Filter*ByAlignmentConfidenceJob`: add flag to disregard alignments that didn't reach a final state

Icemole:
This PR adds a parameter in the `_parse_alignment_logs` method of the `FilterSegmentsByAlignmentConfidenceJob` which represents whether to take or remove the alignments that haven't finished.

In `FilterRecordingsByAlignmentConfidenceJob` this parameter is set to a hardcoded `True` because the alignment caches don't contain these alignments. Then, considering them only pollutes any statistics that we can extract from the recording as a whole. This is explained in the code.

### DIFF
--- a/corpus/filter.py
+++ b/corpus/filter.py
@@ -128,6 +128,8 @@ class FilterSegmentsByRegexJob(Job):
 
 
 class FilterSegmentsByAlignmentConfidenceJob(Job):
+    __sis_hash_exclude__ = {"remove_dnf_alignments": False}
+
     def __init__(
         self,
         alignment_logs: Dict[int, Path],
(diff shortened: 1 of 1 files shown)


Icemole:
I've tested the job. Besides the obvious differences in segments, there's an unwanted difference from the XML dump regarding self-closing XML tags:
```
378,379c378,379
<             <frames port="alignments" number="1"/>
<             <frames port="features" number="253"/>
---
>             <frames port="alignments" number="1" />
>             <frames port="features" number="253" />
```
I think this is purely aesthetical and I wouldn't give it too much thought, but please let me know if this bothers you.

Icemole:
If you want I can also expose the new flag in the `FilterSegmentsByAlignmentConfidenceJob` as well, so that we don't have these pesky outliers at the end of each distribution.

michelwi (approved):

kuacakuaca (approved):

PR #609, opened 2025-06-19
A small fix for text read with uopen

Stefanwuu:
Without explicitly setting mode to rt, the default mode for gzip.open is rb, which can't read gzipped text file correctly.

### DIFF
--- a/lexicon/conversion.py
+++ b/lexicon/conversion.py
@@ -207,7 +207,7 @@ class LexiconFromTextFileJob(Job):
 
         phonemes = set()
         seen_lemma = {}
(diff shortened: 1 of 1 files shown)


albertz (approved):

albertz:
Yes.

Btw, maybe we should even make the `mode` argument mandatory for `uopen`, and require that it contains `t` or `b`. Because the whole purpose for `uopen` is to support both txt and gzip, but if you don't specify `t` or `b`, it's always to be inconsistent. (@NeoLegends)

michelwi (approved):
Approve, and I agree to @albertz suggestion.

PR #610, opened 2025-06-20
`Filter*ByAlignmentConfidenceJob`: better alignment plots

Icemole:
The alignment plots were previously useless when dealing with values outside of the disposed range, such as when aligning with NNs (negative alignment scores). Now the X axis range entirely depends on the distribution.

Also added a custom plot for the recording alignment scores, since the Y axis label was wrong (it said "segments" instead of "recordings").

Notes:
1. The previous plots would be irreproducible. However, for me this is not an issue due to the "cosmetic" nature of the change.
2. If the job is not triggered with `remove_dnf_alignments=True` (which is the recommended value anyway, and only set to `False` for retrocompatibility purposes), the graph might look bad (e.g. X axis extending from 0 to 1e+36, one big bar at [0, 200] and then nothing else). However, since I use NN-based alignments, it does already look bad to me for most of my filterings (just one histogram bar at 0 and then an empty graph), so I believe this would already be an improvement. To circumvent this, the user should set `remove_dnf_alignments=True`.
3. To fix (2), we can think of setting `remove_dnf_alignments=True` by default always. The alignments that did not finish aren't in the alignment caches anyway, so these being included here only pollutes the plot. What do you think?

### DIFF
--- a/corpus/filter.py
+++ b/corpus/filter.py
@@ -386,6 +386,26 @@ class FilterRecordingsByAlignmentConfidenceJob(FilterSegmentsByAlignmentConfiden
 
         return filtered_segments
 
+    def _plot(self, recording_dict: Dict[str, List[Tuple[str, float]]]):
+        """
+        Plots the average recording alignment scores.
+
+        :param recording_dict: Dictionary of recording full names to list of (segment full name, alignment score).
+        """
+        import matplotlib
+        import matplotlib.pyplot as plt
+
+        matplotlib.use("Agg")
+
+        score_np = self._get_alignment_scores_array(recording_dict)
+
+        # Before filtering.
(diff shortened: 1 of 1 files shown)


michelwi:
Do we not have a separate Job to plot alignment scores that is much more versatile and configurable?
https://github.com/rwth-i6/i6_core/blob/b19149e43dbc1c779eb3cee763fb0bebada15380/mm/alignment.py#L429

So instead of changing the suboptimal plotting function of the filter jobs to something slightly-less-but-still-not-so-good , I would just disregard the plots they produce and use the dedicated plotting job..?

Maybe instead if this PR, add a parameter `filter_list` to the `PlotAlignmentJob` and use the `self.out_kept_recordings`

PR #613, opened 2025-06-30
Add `CombineRasrCachesJob` and `CombineRasrLogsJob`

Icemole:
Job that does what it advertises: obtains "original" and "updated" RASR caches (whatever kind of caches), and runs the RASR archiver binary such that a "final" cache is output, in which the content from the "original" caches is overwritten by the corresponding content from the "updated" caches.

~~The job might optionally receive "original" and "updated" log files. If so, the job also outputs "final" logs, in which the content inside `<segment>` tags from the "original" logs is overwritten by the corresponding content from the "updated" logs.~~ Update 2025/07/01: I've created another job `CombineRasrLogsJob` in this same PR for the purpose of updating the RASR logs.

Note that the order in which we specify the caches is important. Intuitively, the RASR archiver reads the files from the end to the beginning, and keeps whatever it has read latest. Therefore, the order in which the files are specified should be `final <- updated <- original`, which is reflected in the archiver call in the code.

I've used this job for alignment fixes and it works as expected:
- Sequences from "original" that are not in "updated" are left untouched.
- Sequences from "original" that are in "updated" obtain the contents from "updated".

Credits: the archiver's ability to merge caches was known during a conversation with @michelwi (thanks!). More information and details can also be found in [the RASR archiver's wiki entry](https://www-i6.informatik.rwth-aachen.de/rwth-asr/manual/index.php/Archiver).


rasr/util.py
```diff
@@ -184,3 +188,127 @@ def run(self):
...
+                "Original and updated dictionaries don't have a 1-to-1 correspondence:\n"
+                f"Original - updated: {set_original_logs_keys.difference(set_updated_logs_keys)}\n"
+                f"Updated - original: {set_updated_logs_keys.difference(set_original_logs_keys)}"
+            )
+            self.out_final_logs = {i: self.output_path(f"log.{i}.gz") for i in range(1, len(self.original_logs) + 1)}
+
+        self.rqmt = {"cpu": 1, "mem": 1.0, "time": 1.0}
+
+    def tasks(self):
+        yield Task("run", resume="run", rqmt=self.rqmt, args=range(1, len(self.original_caches) + 1))
```
NeoLegends, line 278: We specify the args here as range from 1 to length of the original caches, should those perhaps simply be defined by `self.original_caches.keys()`? I suppose since this is mostly used with RASR-infrastructure, where many things are passed as dictionaries with 1-based integer keys it's fine if we don't do it that way, but it seems a bit more correct to me if we use the dict keys instead of a range here.

NeoLegends (approved):
Just one minor comment, otherwise LGTM!

rasr/util.py
```diff
@@ -184,3 +188,127 @@ def run(self):
...
+
+    def run(self, task_id: int):
+        sp.check_call(
+            [
+                self.rasr_archiver_exe,
+                "--mode",
+                "combine",
+                self.out_final_single_caches[task_id].get_path(),
+                self.updated_caches[task_id].get_path(),
+                self.original_caches[task_id].get_path(),
```
NeoLegends, line 290: Wait, the order here does not match the order in the docstring of the class.

NeoLegends (changes requested):
I believe there is an ordering mismatch in the args.

rasr/util.py
Icemole: No, you're right, it does not. This seems to be intended by RASR. The specific ordering should be `final`, `updated`, and `original`. This is the inverse of the corresponding (hypothetical) flow file, which should have links `original -> updated` and `updated -> final`.

The `final` caches should definitely be set at the beginning, as per the RASR wiki that I reference in the job's docstring. What remains to be seen to theoretically prove the correct functionality of the code is the ordering between `updated` and `original`. Sadly, as far as I've read, the wiki doesn't properly specify this.

However, I've empirically confirmed that the last cache set here is read first, and the ordering is reverse to what we would think as "normal". More specifically, I used first the `final` / `original` / `updated` structure, thinking that it would work, but it didn't update the caches. The current structure did update the caches. And it makes sense to have a flow from "beginning" to "end".

Of course, any confirmation from a RASR expert would be awesome!

rasr/util.py
```diff
@@ -184,3 +188,127 @@ def run(self):
...
+    For that, the RASR archiver binary is used in `combine` mode, such that:
+    1. The original cache information is read on a segment-wise basis.
+    2. The updated cache information is also read, and overwrites the former original cache information.
+    3. The result is dumped into the final cache.
+
+    The call in `run` is equivalent to:
+    ```
+    archiver                  \
+        --mode combine        \
+        --allow-overwrite yes \
```
albertz, line 207: You don't use this option?

rasr/util.py
Icemole: That makes sense! I've changed the code. Let's see if other people disagree.

rasr/util.py
Icemole: You're right. I copy-pasted the example from the wiki, which uses `copy` mode, and adapted it. However, it seems that `combine` doesn't need to overwrite, because it already overwrites things itself.

I'll fix the docstring.

rasr/util.py
```diff
@@ -184,3 +188,126 @@ def run(self):
...
+
+class UpdateRasrCachesJob(Job):
+    """
+    Updates the information from a given set of alignment caches ("original")
+    with the information on another set of alignment caches ("updated").
+
+    For that, the RASR archiver binary is used in `combine` mode, such that:
+    1. The original cache information is read on a segment-wise basis.
+    2. The updated cache information is also read, and overwrites the former original cache information.
+    3. The result is dumped into the final cache.
```
albertz, line 201: I wonder about the name of the job, and the documentation. I don't fully understand what this does.

What exactly is in the cache file? What exactly is taken from the original, and what from the updated cache information? I guess not everything from the original is used, otherwise we simply have result = updated? So sth is used from original. But what?

Or is the cache file a generic container of segment name -> dict(any key -> value), and this will update all the dicts for every segment? Does this assume the original and updated caches cover the same segment names?

In that case, the name of the job is maybe ok (corresponds to `dict.update`).

All of this should be documented.

Icemole: > 1. What exactly is in the cache file?

Well, I don't think we can know for sure. The definition from the [RASR wiki](https://www-i6.informatik.rwth-aachen.de/rwth-asr/manual/index.php/Architecture#Cache) is: _"The usual storage format is an archive (sometimes also called cache), which is a binary format that supports compression."_

You can confirm the information given in the wiki by glancing over the [RASR cache implementation](https://github.com/rwth-i6/rasr/blob/master/src/Flow/Cache.cc) (which I'll link in the code). You'll see that it just opens the file and reads/writes from/onto it.

As I understand it, the RASR cache is generic enough so that it can hold anything inside (it's just defined as "a binary [file]") However, in practice it usually follows a map structure with defined key-value pairs, as you said. I don't think it would make sense in practical terms for it not to follow a key-value structure, since for ASR training purposes we need to be able to lookup segment information in a short amount of time. The way our pipeline uses them, both feature and alignment caches follow this key-value structure, in which the key is the segment's full name, and the value is the information about the segment (features, alignments...).

As you said, the key-value part is important the the functionality of the code, and I'll add this to the documentation. However, overwrites of more generic caches that don't follow the key-value pairs depend on what the RASR archiver decides to do with the RASR caches. I can't tell you what happens when you don't have such a key-value structure. I haven't tried this because it's not interesting to me (and I don't think it's interesting to anyone running RASR because of the usual way in which we use the RASR caches for ASR training).

> 2. What exactly is taken from the original, and what from the updated cache information?

All sequences are taken from the original cache. Then, the sequences found in the updated cache have their value overwritten with the information in the updated cache. It's indeed very similar to a `dict.update` operation, or if you want more explicit pseudocode:

```
final_cache[segment_full_name] = updated_cache[segment_full_name] if segment_full_name in updated_cache else original_cache[segment_full_name]
```

I hope this clarifies the functionality of the job. I'll try to add as much information about this as possible to the docstring.

albertz: > ```
> final_cache[segment_full_name] = updated_cache[segment_full_name] if segment_full_name in updated_cache else original_cache[segment_full_name]
> ```

Oh ok, but this is not what I thought. I thought it is:
```python
for segment_full_name in original_cache:
    assert segment_full_name in updated_cache
    final_cache[segment_full_name] = original_cache[segment_full_name].copy()
    final_cache[segment_full_name].update(updated_cache[segment_full_name])
```

Icemole: > Oh ok, but this is not what I thought.

I see the source of confusion. Well, I don't know the internal details too much to refine the implementation details that much. I interpret that the caches are similar to `Dict[str, Any]`, and the functionality is equivalent to
```
final_cache = copy.deepcopy(original_cache)
final_cache.update(updated_cache)
```
but I don't know for 100% sure. I guess the difference with respect to both implementations is that my value `Any` is interpreted by you as `Dict[...]`. Assuming that both caches follow the same structure (which I would say is a valid assumption to make in general, since both caches are usually generated in a similar way), then I guess your code is equivalent to mine, only that I abstract further by triggering an update on the external dict, and you go through the internal dicts.

> So is "combine" maybe a better name?

That makes sense. I've changed the job's name (and created a second one).

Icemole: > I don't know the internal details

Maybe a RASR expert can chime in :slightly_smiling_face: calling @hannah220 @curufinwe (but this is not too important, I would say).

Icemole: Agreed! Thanks for the feedback! Resolving conversation.

rasr/util.py
```diff
@@ -184,3 +189,160 @@ def run(self):
...
+    """
+
+    def __init__(
+        self, original_caches: Dict[int, tk.Path], updated_caches: Dict[int, tk.Path], rasr_archiver_exe: tk.Path
+    ):
+        """
+        :param original_caches: Caches that must be overwritten.
+            The items that are not available in :param:`updated_caches` will remain as found in these caches.
+
+            The dictionary keys provided here must have a 1-to-1 correspondence with those of :param:`updated_caches`.
```
albertz, line 247: "The dictionary keys provided here must have a 1-to-1 correspondence with those of :param:`updated_caches`." - I don't really understand what this is supposed to mean. What "1-to-1 correspondence"? You mean original cache should have the same keys as updated cache? But that's not true. Or you mean that in both cases, the keys should be segment names? But this is a very weird way to say this. I would maybe just remove this sentence. (Same also for `updated_caches`.)

Icemole: What I meant was that the dictionary of ints to original caches must have the same keys as the dictionary of ints to updated caches. Do you think there's a better way of stating this?

Icemole: Note that if we don't make them coincide, then we should properly catch this in the functionality of the job. This isn't too complicated to code at first thought, but I believe that the user should usually have the same number of original/updated caches.

Keep in mind that the job is for updating (combining) the content of some cache with the content of another cache. As a consequence, if the contents of the updated caches are not a subset (in terms of segment full names, not in terms of complete information) of the original caches, then we're potentially duplicating information. This is the normal functionality from the archiver, and I can make it clearer in the docstring.

albertz: Ah, I missed that part. This was also confusing to me: Why do you have that at all, such keys? I would have expected just `original_cache: tk.Path, updated_cache: tk.Path`.

albertz: I see. This is maybe what you should document here.

But I'm again confused by this. You said original and updated would potentially contain different segments (otherwise the whole job does not make sense, as we discussed before). So then `original_cache` and `updated_cache` would likely also contain different splits, and definitely the split indices would not correspond to each other?

albertz: Hm ok, this sounds very specific. I wonder how generic this is, i.e. in how many other situations you could reasonably make use of this `CombineRasrCachesJob`. But I don't say you should think about making this more generic now.

The "1-to-1 correspondence" is just not clear. Write instead sth like "we require `original_cache.keys() == updated_caches.keys()`", and that "each `original_cache[task_id]` is combined with `updated_caches[task_id]`". I think that's enough.

albertz (approved):

PR #614, opened 2025-07-08
Add OOV log for FilterCorpusRemoveUnknownWordSegmentsJob

Stefanwuu:
I think it's a nice add-on to let users check which OOVs they filtered out and which frequencies the OOVs have. To implement it I just made the original unknown_filter stateful, which should be harmless.

### DIFF
--- a/corpus/filter.py
+++ b/corpus/filter.py
@@ -491,9 +491,9 @@ class FilterCorpusRemoveUnknownWordSegmentsJob(Job):
         self.corpus = bliss_corpus
         self.lexicon = bliss_lexicon
         self.case_sensitive = case_sensitive
-        assert (
-            all_unknown is None or segment_oov_tolerance is None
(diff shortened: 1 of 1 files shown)


corpus/filter.py
```diff
@@ -523,28 +529,49 @@ def maybe_to_lower(s):
...
+                orth = segment.orth
+                if not orth:
+                    return True
+
+                words = [maybe_to_lower(o) for o in orth.strip().split(" ")]
+                oov_words = [word for word in words if word not in vocabulary]
+
+                if log_oov_list:
+                    for oov in oov_words:
+                        self.oov_counts[oov] += 1
```
michelwi, line 562: This sounds like a task for `collections.Counter`. Would you consider using that over defaultdict?

Stefanwuu: I didn't know about this, and it's indeed a better fit here. Thanks for the advice!

albertz (approved):

michelwi (approved):

PR #616, opened 2025-08-05
Recog v2 updates

hannah220:
This is an update needed to run the new recognizer implemented by `SearchAlgorithmV2` in https://github.com/rwth-i6/rasr/blob/0e159e5ebe3ba12b712940fff6f18f8ce69bbe62/src/Search/SearchV2.hh#L52. 

For V2, `label-scorer` must be additionally set in RASR config (rasr/config.py). 
And `BuildGlobalCacheJob` must be expanded to handle search types other than `advanced-tree-search`.

### DIFF
--- a/rasr/config.py
+++ b/rasr/config.py
@@ -252,12 +252,9 @@ def build_config_from_mapping(crp, mapping, include_log_config=True, parallelize
     :return: config, post_config
     :rtype: (RasrConfig, RasrConfig)
     """
-    import pdb
-
     config = RasrConfig()
     post_config = RasrConfig()
 
-    pdb.set_trace()
     if include_log_config:
         config._update(crp.log_config)
         post_config._update(crp.log_post_config)
--- a/rasr/config.py
+++ b/rasr/config.py
@@ -252,14 +252,17 @@ def build_config_from_mapping(crp, mapping, include_log_config=True, parallelize
     :return: config, post_config
     :rtype: (RasrConfig, RasrConfig)
     """
+    import pdb
+
     config = RasrConfig()
     post_config = RasrConfig()
 
+    pdb.set_trace()
     if include_log_config:
         config._update(crp.log_config)
         post_config._update(crp.log_post_config)
 
-    for mkey in ["corpus", "lexicon", "acoustic_model", "language_model", "recognizer"]:
+    for mkey in ["corpus", "lexicon", "acoustic_model", "language_model", "recognizer", "label_scorer"]:
         if mkey not in mapping:
             continue
         keys = mapping[mkey]
--- a/recognition/advanced_tree_search.py
+++ b/recognition/advanced_tree_search.py
@@ -883,6 +883,7 @@ class BuildGlobalCacheJob(rasr.RasrCommand, Job):
         config, post_config = rasr.build_config_from_mapping(crp, mapping_dict)
 
         config.speech_recognizer.recognition_mode = "init-only"
+        config.speech_recognizer.search_type = search_type
         config.speech_recognizer.reocgnizer_type = recognizer_type
         config.speech_recognizer.global_cache.file = "global.cache"
         config.speech_recognizer.global_cache.read_only = False
--- a/recognition/advanced_tree_search.py
+++ b/recognition/advanced_tree_search.py
@@ -818,9 +818,16 @@ class BuildGlobalCacheJob(rasr.RasrCommand, Job):
     Standalone job to create the global-cache for advanced-tree-search
     """
 
-    __sis_hash_exclude__ = {"search_type": "advanced-tree-search"}
+    __sis_hash_exclude__ = {"search_type": "advanced-tree-search", "recognizer_type": "recognizer"}
 
-    def __init__(self, crp, search_type="advanced-tree-search", extra_config=None, extra_post_config=None):
+    def __init__(
+        self,
+        crp,
+        search_type="advanced-tree-search",
+        recognizer_type="recognizer",
+        extra_config=None,
+        extra_post_config=None,
+    ):
         """
         :param rasr.CommonRasrParameters crp: common RASR params (required: lexicon, acoustic_model, language_model, recognizer)
         :param rasr.Configuration extra_config: overlay config that influences the Job's hash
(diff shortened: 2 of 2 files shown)


recognition/advanced_tree_search.py
```diff
@@ -855,19 +864,20 @@ def cleanup_before_run(self, cmd, retry, *args):
...
-                "recognizer": "speech-recognizer.recognizer",
-            },
-        )
+    def create_config(cls, crp, search_type, recognizer_type, extra_config, extra_post_config, **kwargs):
+        mapping_dict = {
+            "lexicon": "speech-recognizer.model-combination.lexicon",
+            "acoustic_model": "speech-recognizer.model-combination.acoustic-model",
+            "language_model": "speech-recognizer.model-combination.lm",
+            "recognizer": "speech-recognizer.recognizer",
+            "label_scorer": "speech-recognizer.model-combination.label-scorer",
```
michelwi, line 873: will unconditionally adding `"label_scorer"` here not make it be included in every config, breaking all our hashes in the process?

even better: the pipeline crashes with 
```
AttributeError: label_scorer_config
```
here if label_scorer is not present in the crp.


Maybe a better solution would be 
```
mapping_dict = {...}
if recognizer_type == "recognizer-v2":
  mapping_dict["label_scorer"] =  "speech-recognizer.model-combination.label-scorer"
```

recognition/advanced_tree_search.py
```diff
@@ -817,7 +817,16 @@ class BuildGlobalCacheJob(rasr.RasrCommand, Job):
     Standalone job to create the global-cache for advanced-tree-search
     """
 
-    def __init__(self, crp, extra_config=None, extra_post_config=None):
+    __sis_hash_exclude__ = {"search_type": "advanced-tree-search", "recognizer_type": "recognizer"}
```
michelwi, line 820: no need for `__sis_hash_exclude__` for this Job, since it implements its own hash method that makes the hash only depend on the generated config and not the input arguments that are passed to the job.

recognition/advanced_tree_search.py
```diff
@@ -855,19 +864,20 @@ def cleanup_before_run(self, cmd, retry, *args):
...
+            "recognizer": "speech-recognizer.recognizer",
+            "label_scorer": "speech-recognizer.model-combination.label-scorer",
+        }
+
+        config, post_config = rasr.build_config_from_mapping(crp, mapping_dict)
 
         config.speech_recognizer.recognition_mode = "init-only"
-        config.speech_recognizer.search_type = "advanced-tree-search"
+        config.speech_recognizer.search_type = search_type
+        config.speech_recognizer.reocgnizer_type = recognizer_type
```
michelwi, line 880: Do not change config for old setups:
```suggestion
        if recognizer_type != "recognizer":
            config.speech_recognizer.reocgnizer_type = recognizer_type
```

michelwi (changes requested):

recognition/advanced_tree_search.py
```diff
@@ -855,19 +862,23 @@ def cleanup_before_run(self, cmd, retry, *args):
...
+        if recognizer_type == "recognizer-v2":
+            mapping_dict["label_scorer"] = "speech-recognizer.model-combination.label-scorer"
+
+        config, post_config = rasr.build_config_from_mapping(crp, mapping_dict)
 
         config.speech_recognizer.recognition_mode = "init-only"
-        config.speech_recognizer.search_type = "advanced-tree-search"
+        config.speech_recognizer.search_type = search_type
+        if recognizer_type != "recognizer":
+            config.speech_recognizer.reocgnizer_type = recognizer_type
```
curufinwe, line 881: This does not work. The speech-recognizer doesn't know anything about recognizer-v2. Right now it uses this method to create the recognizer: https://github.com/rwth-i6/rasr/blob/master/src/Search/Module.cc#L77 (called from https://github.com/rwth-i6/rasr/blob/master/src/Tools/SpeechRecognizer/SpeechRecognizer.cc#L114 )

Also note that the new recognizer uses a parameter named "type" and not "search-type" to determine the exact algorithm. This also has to change.

rasr/config.py
```diff
@@ -259,7 +259,7 @@ def build_config_from_mapping(crp, mapping, include_log_config=True, parallelize
         config._update(crp.log_config)
         post_config._update(crp.log_post_config)
 
-    for mkey in ["corpus", "lexicon", "acoustic_model", "language_model", "recognizer"]:
+    for mkey in ["corpus", "lexicon", "acoustic_model", "language_model", "recognizer", "label_scorer"]:
```
curufinwe, line 262: If you add this key here, please also initialize it with None in https://github.com/rwth-i6/i6_core/blob/main/rasr/crp.py#L13

recognition/advanced_tree_search.py
hannah220: Here is the i6 RASR PR: https://github.com/rwth-i6/rasr/pull/142

hannah220: And `search_type` is actually not used in the speech-recognizer. 
But deleting this will result in hash break so I'll leave this..

curufinwe:
I figured out why the pipeline is failing. The reason is that the `FilterSegmentsByAlignmentConfidenceJob` and `FilterRecordingsByAlignmentConfidenceJob`
https://github.com/rwth-i6/i6_core/blob/main/corpus/filter.py#L130 take a crp parameter into their hashes. We have added two more members here in this PR and that changes the crp object.

My suggestion would be to fix the job to compute the hashes properly or refactor it to get an optional concurrent parameter and not the whole crp and fix the setups and accept the hash breakage.

Only 3 setups are affected on the AppTek side:
* setups/am/TA_IN/f16kHz/20220423-wwang-gmm/config/align.py
* setups/am/UK_UA/f16kHz/20220429-wmichel-gmm/config/align.py
* setups/am/UK_UA/f16kHz/20220316-wmichel-gmm/config/align.py

@hannah220 @michelwi what do you think about this?

michelwi:
> What would be the correct order to proceed now?

I would suggest to create a separate PR with the changes of https://github.com/rwth-i6/i6_core/commit/32bdd3b258c7cc0fa303da9d7e122790cfa1e130, which is merged first. Then we merge https://bitbucket.org/omnifluent/apptek_asr/branch/update-hash-with-crp-argument and then this PR can proceed without problems and we get to check if any additional problems are detected by our pipeline.

hannah220:
This https://github.com/rwth-i6/i6_core/pull/620 contains change in hash func for `FilterSegmentsByAlignmentConfidenceJob` and `FilterRecordingsByAlignmentConfidenceJob`.

PR #622, opened 2025-09-17
make intermediate g2p models available, make rqmt more generous

JackTemaki:
When g2p training takes longer, it would be nice to use some intermediate model. Also, for Arabic it went completely out of the assigned resources, thus I wanted to loosen them a bit.


JackTemaki:
@michelwi was ist da mit den hashes los?

PR #629, opened 2025-10-20
`ReturnnForwardJob[V2]`: accept `DelayedBase` as model

Icemole:
Counter-PR to https://github.com/rwth-i6/i6_core/pull/628.

Thanks to @patrick-wilken, I was able to trigger `DelayedGetItem` on a dictionary and retrieve the delayed value itself. This seems to work wonderfully in most jobs as a `DelayedBase` variable, which most jobs support, except for `ReturnnForwardJob[V2]`. The base model receives it nicely, but there's an internal check that the model's absolute path should exist on disk.

This PR enables the check to also pass with `DelayedBase` objects, such as `DelayedGetItem`.

**One thing I'm wondering is: why would the job require a check that the path to the model exists? If it didn't exist, would the job simply not have been triggered?** Maybe we can remove the check altogether. What do you think?

---

PS: I also took the chance to improve the `parameter_tuning.py` file by adding an `__all__` list and reordering the imports for clarity (python, third party, specific to `i6_core`). If you don't like this or feel it's too off-topic, please let me know and I'll abandon these changes.

### DIFF
--- a/returnn/forward.py
+++ b/returnn/forward.py
@@ -10,7 +10,7 @@ import os
 import shutil
 import subprocess
 import tempfile
-from typing import List, Optional, Union, get_args
+from typing import List, Optional, Union
 
 from sisyphus import Job, Task, gs, setup_path, tk
 from sisyphus.delayed_ops import DelayedBase


returnn/forward.py
```diff
@@ -390,11 +397,27 @@ def hash(cls, kwargs):
...
+def _get_model_path(model: Checkpoint) -> str:
+    """
+    :param model: Base model checkpoint represented as a python/sisyphus object.
+    :return: Path on disk to the model checkpoint.
+    """
+
+    base_checkpoint_classes = get_args(Checkpoint)
+    assert isinstance(model, base_checkpoint_classes), (
+        f"Expected model to be of type {base_checkpoint_classes} but found model of type {type(model)}."
+    )
```
albertz: Remove this. This is not necessary. You will do the same check already a few lines below, in a better way.
```suggestion
```

returnn/forward.py
```diff
@@ -390,11 +397,27 @@ def hash(cls, kwargs):
...
-    raise TypeError(f"Unknown model checkpoint type: {type(model)}")
+        model = model.index_path
+    elif isinstance(model, PtCheckpoint):
+        model = model.path
+
+    assert isinstance(model, AbstractPath), (
+        "Expected model after unwrapping to be of type AbstractPath "
+        f"(like sisyphus.job_path.Path or sisyphus.job_path.Variable), but found model of type {type(model)}."
+    )
+    return model.get_path()
```
albertz: Why do you change the API of `_get_model_path`? This change seems somewhat unnecessary?
```suggestion
    return model
```

returnn/forward.py
```diff
@@ -30,6 +33,8 @@ class ReturnnForwardJob(Job):
     Also allows to run an "eval" task pass, which is similar to "forward" but treats all layers as in training mode,
     which can be used to e.g. do cheating experiments.
 
+    As of 2025-10, the job also accepts the :class:`sisyphus.delayed_ops.DelayedGetItem` sisyphus class.
+
```
albertz:
```suggestion
```

returnn/forward.py
Icemole: The thing is that by doing it the first time here, we're telling the user that we don't expect the *input* type, as in `model_checkpoint=42`.

However, in the check below we're checking that the *wrapped* datatype is a proper `AbstractPath` (usually `tk.Path`).

I feel like both checks do different things. If you still feel strongly against this change, I'll add it as you prefer, since it's a small detail.

returnn/forward.py
```diff
@@ -390,11 +397,27 @@ def hash(cls, kwargs):
...
+        model = model.get()
+
     if isinstance(model, TfCheckpoint):
-        return model.index_path
-    if isinstance(model, PtCheckpoint):
-        return model.path
-    raise TypeError(f"Unknown model checkpoint type: {type(model)}")
+        model = model.index_path
+    elif isinstance(model, PtCheckpoint):
+        model = model.path
```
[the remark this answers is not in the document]
Icemole: See https://github.com/rwth-i6/i6_core/pull/629/files#r2445435877.

returnn/forward.py
```diff
@@ -390,11 +397,27 @@ def hash(cls, kwargs):
...
+    :return: Path on disk to the model checkpoint.
+    """
+
+    base_checkpoint_classes = get_args(Checkpoint)
+    assert isinstance(model, base_checkpoint_classes), (
+        f"Expected model to be of type {base_checkpoint_classes} but found model of type {type(model)}."
+    )
+    if isinstance(model, DelayedBase) and not isinstance(model, AbstractPath):
+        # For instance, sisyphus.delayed_ops.DelayedGetItem could wrap around a tk.Path or PtCheckpoint.
+        model = model.get()
```
michelwi: The apptek pipeline fails because you shouldn't be calling `get()` in the manager.. 

I don't think there is a way to check the existence for general DelayedBase objects. 

>why would the job require a check that the path to the model exists? 

Maybe we could just drop the check?

Icemole: ... Neither can I call `get()` in a sisyphus `Task`? What differentiates `create_files` from `run`?

> Maybe we could just drop the check?

I strongly think so too. What would be the use case for this check? As I said in the PR description, if the path does not exist, then the corresponding sisyphus variable won't be ready, so the job will never be triggered. Maybe someone with more experience in sisyphus can correct me.

albertz: > The apptek pipeline fails because you shouldn't be calling `get()` in the manager..

I don't understand. This is never called in the manager here?

albertz: > > why would the job require a check that the path to the model exists?
> 
> Maybe we could just drop the check?

The check does not hurt. Leave it.

returnn/forward.py
albertz: I don't understand what you mean. Maybe you missed my other code change suggestion? I'm not talking about the check on `AbstractPath`. I mean the `else` branch.

returnn/forward.py
```diff
@@ -4,24 +4,27 @@
...
-from i6_core.returnn.training import Checkpoint as TfCheckpoint, PtCheckpoint
+from i6_core.returnn.training import Checkpoint as TfCheckpoint
+from i6_core.returnn.training import PtCheckpoint
 import i6_core.util as util
 
 
 Path = setup_path(__package__)
 
-Checkpoint = Union[TfCheckpoint, PtCheckpoint, tk.Path]
+Checkpoint = Union[TfCheckpoint, PtCheckpoint, AbstractPath, DelayedBase]
```
michelwi:
```suggestion
Checkpoint = Union[TfCheckpoint, PtCheckpoint, DelayedBase]
```

`AbstractPath` is already a subclass of `DelayedBase`.

albertz: Yes but I think it's good to highlight this anyway. Giving it a `Path` is probably way more common than any other type of `DelayedBase`.

Btw, I would use `Path` instead of `AbstractPath` though.

returnn/forward.py
```diff
@@ -391,10 +393,29 @@ def hash(cls, kwargs):
...
-        return model.index_path
-    if isinstance(model, PtCheckpoint):
-        return model.path
-    raise TypeError(f"Unknown model checkpoint type: {type(model)}")
+        model = model.index_path
+    elif isinstance(model, PtCheckpoint):
+        model = model.path
+    elif isinstance(model, tk.Path):
+        # Already in tk.Path, which is what we need to return.
+        pass
```
albertz: Add an `else` branch here, raising a `TypeError`.

michelwi (approved):

PR #632, opened 2025-11-13
Add job to select random lines of text from file in memory-efficient way

NeoLegends:
This job/use-case keeps plopping up (for me in SPM training), let's make it into an i6_core job.

### DIFF
--- a/text/processing.py
+++ b/text/processing.py
@@ -506,11 +506,10 @@ class TakeNRandomLinesJob(Job):
                 f"which is fewer than the requested {self.num_lines} lines."
             )
 
+        np.random.seed(self.seed)
+        indices = np.random.choice(num_non_empty_lines, size=self.num_lines, replace=False)
+        indices_set = set(int(v) for v in indices)
         with util.uopen(self.text, "rt") as in_file, util.uopen(self.out, "wt") as out_file:
             non_empty_lines = (line for line in in_file if line.strip())
-
-            np.random.seed(self.seed)
-            indices = np.random.choice(num_non_empty_lines, size=self.num_lines, replace=False)
-            indices_set = set(int(v) for v in indices)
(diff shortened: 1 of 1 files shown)


Atticus1806:
I think in our Pipelines we usually do it by first applying `ShuffleAndSplitSegmentsJob` with a split of 1.0/0.0 and then applying `HeadJob`.
While this is not the most elegant solution I guess, the functionality "exists".

That being said I don't mind adding a new job, but I just wanted to bring that to attention / discussion.

albertz (approved):

text/processing.py
```diff
@@ -453,3 +454,45 @@ def run(self):
...
+
+        num_to_take = min(self.num_lines, num_lines_in_file)
+        take_map = [True for _ in range(num_to_take)] + [False for _ in range(max(num_lines_in_file - num_to_take, 0))]
+        random.seed(self.seed)
+        random.shuffle(take_map)
+
+        with util.uopen(self.text, "rt") as in_file, util.uopen(self.out, "wt") as out_file:
+            non_empty_lines = (line for line in in_file if line.strip())
+            lines_to_write = (line for line, take_it in zip(non_empty_lines, take_map) if take_it)
+            out_file.writelines(lines_to_write)
```
curufinwe:
```suggestion
            for line, take_it in zip(filter(lambda line: line.strip(), in_file), take_map):
                if take_it:
                    out_file.write(line)
```

Atticus1806 (approved):
Fine by me, would wait for @curufinwe to resolve his comment before merge.

text/processing.py
albertz: I also prefer the variant by @NeoLegends. I'm not so much a fan of `filter`. I think list comprehensions / generator expressions are usually easier to read.

PR #633, opened 2025-11-21
Add library for reading/writing new ASR result format

curufinwe:


tests/lib/sample.jsonl
```diff
@@ -0,0 +1,8 @@
+{"type": "rec", "name": "rec1"}
+{"type": "seg", "name": "seg1", "start": 0.0, "end": 2.0, "n_best_index": 0}
+{"type": "tt", "text": "hello", "start": 0.0, "end": 1.0}
+{"type": "tt", "text": "world", "start": 1.0, "end": 2.0}
+{"type": "seg", "name": "seg2", "start": 2.0, "end": 4.0, "n_best_index": 0}
+{"type": "tt", "text": "bar", "start": 3.0, "end": 4.0}
+{"type": "seg", "name": "seg2", "start": 2.0, "end": 4.0, "n_best_index": 1}
+{"type": "tt", "text": "foo", "start": 2.0, "end": 3.0}
```
NeoLegends: missing newline

.github/workflows/hashes.yml
```diff
@@ -21,9 +21,9 @@ jobs:
       with:
         repository: "rwth-i6/sisyphus"
         path: "sisyphus"
-    - uses: actions/setup-python@v4
+    - uses: actions/setup-python@v6
       with:
-        python-version: 3.8
+        python-version: 3.10.12
```
NeoLegends: Q:
1. What is required from 3.10 that requires this version bump?
2. Not that it's necessary for this PR (if you don't import this module), but should 3.10 then become the new minimally required version for `i6_core`? I vaguely remember the time we bumped the python version on apptek_asr that i6_core wasn't ready for a bump yet/there are people using it with older pythons than 3.10. I think this could still be the case.

NeoLegends (commented):
A few more smaller things came to my mind.

lib/asr_results.py
```diff
@@ -0,0 +1,237 @@
...
+
+    Reads the file line by line, parsing each line as a JSON object and validating it
+    against the ASRResultsStream union type (Recording, Segment, or TimedTranscript).
+
+    :param path: Path to the input file.
+    :yield: Parsed ASR result objects (Recording, Segment, or TimedTranscript).
+    """
+    record_adapter = TypeAdapter(ASRResultsStream)
+
+    path = Path(path)
```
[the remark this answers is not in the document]
curufinwe: This is a remnant from the time before I used uopen.

lib/asr_results.py
```diff
@@ -0,0 +1,237 @@
...
+    with uopen(path, "rt", encoding="utf-8") as f:
+        for line in f:
+            line = line.strip()
+            if not line:
```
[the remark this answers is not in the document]
curufinwe: You can do this, without then using them later in the pipeline. Json doesn't have comments and I would stick to this here.

tests/lib/sample.jsonl
curufinwe: Why is this a problem?

.github/workflows/hashes.yml
curufinwe: I talked to the people at i6 who probably have the oldest setups (@Marvin84 @mmz33 @JackTemaki ) and they are all OK with bumping the required version to 3.10.

PR #650, opened 2026-03-27
Adding corpus segment track based hdf5 dump strategy

guest-tz:


returnn/hdf.py
```diff
@@ -311,7 +337,7 @@ def run(self):
         out_hdf = SimpleHDFWriter(filename=self.out_hdf, dim=1)
 
         for recording in c.all_recordings():
-            audio = None
+            audio = audio_file = None
```
curufinwe:
```suggestion
            audio = None
            audio_file = None
```

curufinwe (approved):
Only one small style comment

returnn/hdf.py
```diff
@@ -232,13 +232,38 @@ class BaseStrategy:
...
+            return data
+
     @dataclass(frozen=True)
     class PickNth(BaseStrategy):
         channel: int
 
         def __eq__(self, other):
             return super().__eq__(other) and other.channel == self.channel
 
+        def process(self, data, segment: corpus.Segment):
```
michelwi:
```suggestion
        def process(
            self,
            data: np.ndarray,
            segment: corpus.Segment,
        ):
```
nit: typing + black

returnn/hdf.py
```diff
@@ -232,13 +232,38 @@ class BaseStrategy:
...
+            assert data.shape[-1] > self.channel, "Audio has too few channels."
+            data = data[:, self.channel : self.channel + 1]
```
michelwi, line 258: the previous implementation was 
```
 data = data[:, self.multi_channel_strategy.channel]
```
i.e. results in different shape.

Did you check the new shape works as intended, especially with the resampling block?
(request changes/verification here)

michelwi (changes requested):

returnn/hdf.py
Icemole:
```suggestion
        def process(self, data: np.ndarray, segment: corpus.Segment):
```
Would black enforce a multiline function definition?

Edit: I see that this is the case above as well. 🤷‍♀️

returnn/hdf.py
Icemole: > results in different shape.

I, for one, was suprised by this. For reference:

```
>>> import numpy as np
>>> a = np.array([[1, 2, 3], [4, 5, 6]])
>>> a[:, 1 : 2]
array([[2],
       [5]])
>>> a[:, 1 : 2].shape
(2, 1)
>>> a[:, 1]
array([2, 5])
>>> a[:, 1].shape
(2,)
```

This PR is 2 weeks old so I would assume that the functionality works as intended, but still, I'm surprised that the functionality works as intended. I subscribe to @michelwi's comment.

returnn/hdf.py
michelwi: I mean it could be a case of black with default 80 linewidths adding newline + comma and then black with 120 linewidths seeing the comma and going 🤷

returnn/hdf.py
michelwi: >  I would assume that the functionality works as intended

later there is anyway a reshape call, when the data is written to hdf. I am only concerned about the resample if-branch, which might not have been activated in @guest-tz cases.

guest-tz: previously it was not consistent, ndim was 1 after picking, without picking ndim was 2.
Now there is also a check to force ndim ==  2, and single channel.
Overall, at the end it does not matter, i just did not like it, see line:
[line](https://github.com/guest-tz/i6_core/blob/e108c4447db9be58c5abc290320d67bbb94467c8/returnn/hdf.py#L388)

michelwi: > especially with the resampling block?

According to [documentation](https://librosa.org/doc/main/generated/librosa.resample.html) `librosa.resample` supports arbitrary shapes and only works on the specified axis. We specify `axis=0` and all cases have the time-axis at position 0, so it should be correct and working.

I agree with your comment about consistency.

michelwi (approved):

michelwi:
I manually ran the AppTek hashes pipeline (#23414) and it was green.

PR #651, opened 2026-04-15
Update CommonRasrParameters and build_config_from_mapping to support multiple label scorers

hannah220:
In SearchAlgorithmV2, each label scorer config must be set under individual block `model-combination.label-scorer-{i}`.

### DIFF
--- a/rasr/config.py
+++ b/rasr/config.py
@@ -269,9 +269,9 @@ def build_config_from_mapping(crp, mapping, include_log_config=True, parallelize
             if mkey == "label_scorers":
                 label_scorer_configs = getattr(crp, mkey)
                 num_label_scorers = len(label_scorer_configs)
-                label_scorer_top_level_key = key.split(".")[:len(key.split("."))-1]
+                label_scorer_top_level_key = key.split(".")[: len(key.split(".")) - 1]
                 config[".".join(label_scorer_top_level_key + ["num-scorers"])] = num_label_scorers
-                
+
                 for i, (c, pc) in enumerate(label_scorer_configs, start=1):
                     if c is not None:
                         config[f"{key}-{i}"] = c
@@ -281,7 +281,7 @@ def build_config_from_mapping(crp, mapping, include_log_config=True, parallelize
                 c = getattr(crp, "%s_config" % mkey)
                 if c is not None:
                     config[key] = c
-    
+
                 c = getattr(crp, "%s_post_config" % mkey)
                 if c is not None:
                     post_config[key] = c
--- a/rasr/config.py
+++ b/rasr/config.py
@@ -267,16 +267,19 @@ def build_config_from_mapping(crp, mapping, include_log_config=True, parallelize
             keys = (keys,)
         for key in keys:
             if mkey == "label_scorers":
-                label_scorer_configs = getattr(crp, mkey)
-                num_label_scorers = len(label_scorer_configs)
-                label_scorer_top_level_key = key.split(".")[: len(key.split(".")) - 1]
-                config[".".join(label_scorer_top_level_key + ["num-label-scorers"])] = num_label_scorers
-
-                for i, (c, pc) in enumerate(label_scorer_configs, start=1):
-                    if c is not None:
-                        config[f"{key}-{i}"] = c
-                    if pc is not None:
-                        post_config[f"{key}-{i}"] = pc
+                # ONLY continue when crp contains label_scorers
+                # most cases (recognizer-v1), crp does not have any label_scorers
+                label_scorer_configs = getattr(crp, mkey, None)
+                if label_scorer_configs is not None:
+                    num_label_scorers = len(label_scorer_configs)
+                    label_scorer_top_level_key = key.split(".")[: len(key.split(".")) - 1]
+                    config[".".join(label_scorer_top_level_key + ["num-label-scorers"])] = num_label_scorers
+    
+                    for i, (c, pc) in enumerate(label_scorer_configs, start=1):
+                        if c is not None:
+                            config[f"{key}-{i}"] = c
+                        if pc is not None:
+                            post_config[f"{key}-{i}"] = pc
             else:
                 c = getattr(crp, "%s_config" % mkey)
                 if c is not None:
--- a/rasr/config.py
+++ b/rasr/config.py
@@ -267,12 +267,21 @@ def build_config_from_mapping(crp, mapping, include_log_config=True, parallelize
             keys = (keys,)
         for key in keys:
             if mkey == "label_scorers":
-                label_scorer_configs = getattr(crp, mkey, None)
+                label_scorer_configs = getattr(crp, mkey)
                 num_label_scorers = len(label_scorer_configs)
-                if num_label_scorers > 0:
-                    label_scorer_top_level_key = key.split(".")[: len(key.split(".")) - 1]
-                    config[".".join(label_scorer_top_level_key + ["num-label-scorers"])] = num_label_scorers
-    
+                if num_label_scorers == 0:
+                    continue
+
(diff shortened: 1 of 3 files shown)


rasr/config.py
```diff
@@ -266,13 +266,27 @@ def build_config_from_mapping(crp, mapping, include_log_config=True, parallelize
...
-            c = getattr(crp, "%s_config" % mkey)
-            if c is not None:
-                config[key] = c
-
-            c = getattr(crp, "%s_post_config" % mkey)
-            if c is not None:
-                post_config[key] = c
+            num_config = 1
+            if mkey == "label_scorer":
+                num_label_scorers = getattr(crp, "num_label_scorers")
```
michelwi: this does not seem to work.

From the failed Pipeline:
```
[2026-04-15 15:58:30,645] INFO: 		  File "/opt/atlassian/pipelines/agent/build/i6_core/recognition/advanced_tree_search.py", line 934, in hash
[2026-04-15 15:58:30,645] INFO: 		    config, post_config = cls.create_config(**kwargs)
[2026-04-15 15:58:30,645] INFO: 		                          ^^^^^^^^^^^^^^^^^^^^^^^^^^^
[2026-04-15 15:58:30,645] INFO: 		  File "/opt/atlassian/pipelines/agent/build/i6_core/recognition/advanced_tree_search.py", line 917, in create_config
[2026-04-15 15:58:30,645] INFO: 		    config, post_config = rasr.build_config_from_mapping(crp, mapping_dict)
[2026-04-15 15:58:30,645] INFO: 		                          ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
[2026-04-15 15:58:30,645] INFO: 		  File "/opt/atlassian/pipelines/agent/build/i6_core/rasr/config.py", line 271, in build_config_from_mapping
[2026-04-15 15:58:30,645] INFO: 		    num_label_scorers = getattr(crp, "num_label_scorers")
[2026-04-15 15:58:30,645] INFO: 		                        ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
[2026-04-15 15:58:30,645] INFO: 		  File "/opt/atlassian/pipelines/agent/build/i6_core/rasr/crp.py", line 60, in __getattr__
[2026-04-15 15:58:30,645] INFO: 		    raise AttributeError(name)
[2026-04-15 15:58:30,645] INFO: 		AttributeError: num_label_scorers
```

Maybe need to add a default value of 1?

Otherwise my additional advice here would be that `getattr` is not needed with a static key and one could also write 

```suggestion
                num_label_scorers = crp.num_label_scorers
```

curufinwe (changes requested):
requested changes discussed offline

rasr/crp.py
```diff
@@ -26,7 +26,7 @@ def __init__(self, base=None):
             self.language_model_post_config = None
             self.recognizer_config = None
             self.recognizer_post_config = None
-            self.label_scorer_config = None
+            self.label_scorers = None
             self.label_scorer_post_config = None
```
michelwi: From the code change above, I assume that the `label_scorers` should now be a `list[tuple[RasrConfig, RasrConfig]]`, is that right? 

Then first of all - I don't like that this is different from everything else so far. Also for the implementation in rasr it would have probably been better to have a modular CombineLabelScorer instead of having the log-linear combination hardcoded in the interface.

Secondly, then probably the former label_scorer_post_config can go away.
```suggestion
            self.label_scorers = None
```

Thirdly, maybe a better interface would be to have `[]` (empty list) as the placeholder for "no label scorers" so that the implementation simplifies (e.g. we can always loop over self.label_scorers)

Fourthly, maybe at this point some typing info would be great so that we know how to use label_scorers

Fifthly (that sounds odd.. I never counted so high xD): The pipeline broke, because hashes of the setups that previously used the old way to write the config are now different.

hannah220: First: the reason that label scorers are listed directly after `model-combination` like `*.recognizer-v2.model-combination.label-scorer-{i}` instead of using CombineLabelScorer is because the search algorithm should have access to individual label scorer to do intermediate pruning (prune hypotheses after scoring with each label scorer). 
CombineLabelScorer has label scorers internally and combines all the scores internally so intermediate pruning does not work well from the search side. 

Second: You're right, I'll remove that. 

Third: I only want to set `num-label-scorers` when there is at least one label scorer, so I think I cannot get rid of the if sentence. 

Forth: Also yes. 

Fifth: Right, so I'll have to update the hashes on apptek side once this is merged?

rasr/config.py
```diff
@@ -259,20 +259,42 @@ def build_config_from_mapping(crp, mapping, include_log_config=True, parallelize
...
             continue
         keys = mapping[mkey]
         if type(keys) == str:
             keys = (keys,)
         for key in keys:
-            c = getattr(crp, "%s_config" % mkey)
-            if c is not None:
-                config[key] = c
+            if mkey == "label_scorers":
+                label_scorer_configs = getattr(crp, mkey)
```
michelwi:
```suggestion
                label_scorer_configs = crp.label_scorers
```
this is now `getattr` with static key with extra steps.

or you could inline `crp.label_scorers` everywhere.

rasr/config.py
```diff
@@ -259,20 +259,42 @@ def build_config_from_mapping(crp, mapping, include_log_config=True, parallelize
...
+                num_label_scorers = len(label_scorer_configs)
+                if num_label_scorers == 0:
+                    continue
+
+                label_scorer_top_level_key = key.split(".")[: len(key.split(".")) - 1]
+                config[".".join(label_scorer_top_level_key + ["num-label-scorers"])] = num_label_scorers
```
michelwi: This is only needed in the case of having more than one label scorer (as 1 is the default). It does not hurt to always set it - except for breaking those two setups that have used it already.
I guess if you moved these two lines below in the else-branch, it would fix the hashes

hannah220:
I think hash still fails because of `TelephonyStreamingQuantizedOnnxStreamingFinetuned1BConformerCTCMbwTelArpaLm202510V1Factory` that uses a single label scorer (not a list). 
To be fixed by https://bitbucket.org/omnifluent/apptek_asr/pull-requests/2185/diff#Lartefacts/prod/speech_server/grpc_recognizer/EN/repo.pyF1226.

michelwi:
> I think hash still fails because [...]

One problem is that this branch is too old and is missing the `LexiconToPhonemeListJob`, pls merge/rebase this branch to current main.
The other problem is related to the way how the AppTekSpeechServerRecognitionJob builds the crp for the BuildGlobalCacheJob, which needs to be adapted to this change. cf [apptek_asr #2201](https://bitbucket.org/omnifluent/apptek_asr/pull-requests/2201)

curufinwe (approved):

rasr/config.py
```diff
@@ -259,20 +259,39 @@ def build_config_from_mapping(crp, mapping, include_log_config=True, parallelize
...
+
+                if num_label_scorers == 1:
+                    c, pc = label_scorer_configs[0]
+                    if c is not None:
+                        config[key] = c
+                    if pc is not None:
+                        post_config[key] = pc
+                elif num_label_scorers > 1:
+                    label_scorer_top_level_key = key.split(".")[: len(key.split(".")) - 1]
+                    config[".".join(label_scorer_top_level_key + ["num-label-scorers"])] = num_label_scorers
```
curufinwe:
```suggestion
                    label_scorer_top_level_key = key.rsplit(".", maxsplit=1)[0]
                    config[label_scorer_top_level_key + ".num-label-scorers"] = num_label_scorers
```

michelwi (approved):
AppTek Pipeline #24363 is green 👍

PR #652, opened 2026-04-16
Add LexiconToPhonemeListJob

hannah220:
Add a job to extract phonemes from bliss lexicon. 
This list can then be used as "lexicon" input file for lexicon free search in search v2, which corresponds to output vocabulary  (subword tokens) of the model.

### DIFF
--- a/lexicon/conversion.py
+++ b/lexicon/conversion.py
@@ -43,7 +43,7 @@ class LexiconToWordListJob(Job):
 
 class LexiconToPhonemeListJob(Job):
     def __init__(self, bliss_lexicon: Path):
-        self.set_vis_name("Lexicon to Word List")
+        self.set_vis_name("Lexicon to Phoneme List")
 
         self.bliss_lexicon = bliss_lexicon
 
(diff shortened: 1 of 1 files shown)


lexicon/conversion.py
```diff
@@ -41,6 +41,26 @@ def run(self):
...
+    def tasks(self):
+        yield Task("run", mini_task=True)
+
+    def run(self):
+        with uopen(self.bliss_lexicon.get_path(), "r") as lexicon_file, open(
+            self.out_phoneme_list.get_path(), "wt"
+        ) as phoneme_file:
+            lexicon = ET.fromstring(lexicon_file.read())
+            for p in lexicon.findall("./phoneme-inventory/phoneme/symbol"):
+                phoneme_file.write(f"{p.text}\n")
```
michelwi, line 61: We have a [lib](https://github.com/rwth-i6/i6_core/blob/main/lib/lexicon.py) to parse lexica, but this one works as well..

michelwi (approved):

lexicon/conversion.py
hannah220: done

curufinwe (approved):

PR #653, opened 2026-05-12
Add CMakeJob

SimBe195:
This PR adds a compilation job for projects with a cmake build system. It runs a configure -> build -> install sequence and marks the install directory as its output. I tested it with RASR compilation which works well.

### DIFF
--- a/tools/compile.py
+++ b/tools/compile.py
@@ -108,3 +108,87 @@ class MakeJob(Job):
         d = kwargs.copy()
         d.pop("num_processes")
         return super().hash(d)
+
+
+class CMakeJob(Job):
+    """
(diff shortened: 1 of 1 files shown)


tools/compile.py
```diff
@@ -108,3 +108,87 @@ def hash(cls, kwargs):
...
+
+class CMakeJob(Job):
+    """
+    Builds a CMake project using a configure, build and install sequence.
+    """
+
+    def __init__(
+        self,
+        source_folder: tk.Path,
+        cmake_opts: Optional[List[str]] = None,
```
michelwi, line 121: 1. I would specify in the variabel name that those go to the 'configure' call.
2. do we also need custom options for the 'compile' and 'install' calls?

PR #654, opened 2026-05-26
TransformAndMapHuggingFaceDatasetJob and ExtractTextFromHuggingFaceDatasetJob: support nodes without internet access.

robin-p-schmitt:
These jobs fail if there is no internet access, even if the required data is present in `HF_HOME`. This fixes this by using `snapshot_download` first with the `local_files_only=True` option.

### DIFF
--- a/datasets/huggingface.py
+++ b/datasets/huggingface.py
@@ -15,6 +15,27 @@ if TYPE_CHECKING:
     TransformFuncT = Union[Callable[[DatasetDict], DatasetDict], Callable[[Dataset], Dataset]]
 
 
+def load_hf_dataset(path, **opts):
+    """
+    If `HF_HUB_OFFLINE` is set to true, only use the local files in `HF_HOME` to load the dataset.
+    Otherwise, `load_dataset` tries to load metadata from the internet.
+
+    This can be used for nodes which don't have internet access (e.g. JUPITER in Jülich).
+    """
+    import os
+    from datasets import load_dataset
+    from huggingface_hub import snapshot_download
+
+    if os.environ.get("HF_HUB_OFFLINE") == "1":
+        path = snapshot_download(
+            repo_id=path,
+            repo_type="dataset",
+            local_files_only=True,
+        )
+
+    return load_dataset(path, **opts)
+
+
 class DownloadAndPrepareHuggingFaceDatasetJob(Job):
     """
     https://huggingface.co/docs/datasets/
@@ -242,7 +263,7 @@ class TransformAndMapHuggingFaceDatasetJob(Job):
                 " or via job.set_env"
             )
 
-            ds = load_dataset(dataset_path, **load_dataset_opts)
+            ds = load_hf_dataset(dataset_path, **load_dataset_opts)
             assert isinstance(ds, (Dataset, DatasetDict))
 
         if self.transform:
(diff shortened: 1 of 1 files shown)


datasets/huggingface.py
```diff
@@ -15,6 +15,27 @@
...
+    Otherwise, `load_dataset` tries to load metadata from the internet.
+
+    This can be used for nodes which don't have internet access (e.g. JUPITER in Jülich).
+    """
+    import os
+    from datasets import load_dataset
+    from huggingface_hub import snapshot_download
+
+    if os.environ.get("HF_HUB_OFFLINE") == "1":
+        path = snapshot_download(
```
albertz, line 32: So you say `snapshot_download` does not have this problem that it accesses the internet?

robin-p-schmitt: Yes, I checked this on one of the Jupiter nodes without internet and it work. But if you set HF_HUB_OFFLINE=1, then it anyway won't access the internet

datasets/huggingface.py
```diff
@@ -15,6 +15,27 @@
...
+    If `HF_HUB_OFFLINE` is set to true, only use the local files in `HF_HOME` to load the dataset.
+    Otherwise, `load_dataset` tries to load metadata from the internet.
+
+    This can be used for nodes which don't have internet access (e.g. JUPITER in Jülich).
+    """
+    import os
+    from datasets import load_dataset
+    from huggingface_hub import snapshot_download
+
+    if os.environ.get("HF_HUB_OFFLINE") == "1":
```
michelwi: I asked the AI about the differences in offline mode and it said your function should be good. However it gave this warning. ~Pls verify`... no, wait I should verify the AI slob. https://huggingface.co/docs/huggingface_hub/en/package_reference/environment_variables#boolean-values there, AI is right about it, but the documentation has a better function: [is_offline_mode](https://huggingface.co/docs/huggingface_hub/v1.16.0/en/package_reference/utilities#huggingface_hub.is_offline_mode)
```
One subtle issue in your wrapper: it only checks exactly this:

os.environ.get("HF_HUB_OFFLINE") == "1"

So your wrapper recognizes HF_HUB_OFFLINE=1, but not HF_HUB_OFFLINE=true, ON, or YES, even though Hugging Face itself recognizes those as true.

A more robust version would be:

def is_hf_offline():
    return os.environ.get("HF_HUB_OFFLINE", "").upper() in {"1", "ON", "YES", "TRUE"}

or, if available:

from huggingface_hub.constants import HF_HUB_OFFLINE

Then use that boolean instead of manually parsing the env var.
```


```suggestion
    if is_offline_mode():
```

datasets/huggingface.py
michelwi: So you mean this change only works with `HF_HUB_OFFLINE=1` and if you set it, then the old coder would work as well? So we don't need it?

albertz (approved):

datasets/huggingface.py
albertz: I understand it as if you set `HF_HUB_OFFLINE=1`, then `snapshot_download` works, but `load_dataset` does not. Which is somewhat weird, but I just accepted it as a weirdness of HF datasets.

Maybe adding such a comment to the code would be helpful though.

robin-p-schmitt: I just checked again. `snapshot_download` only locates the files of the dataset, if the dataset exists, while `load_dataset` does more things including a check for metadata from the internet.

michelwi (approved):

PR #656, opened 2026-05-27
LibriSpeechCreateBlissCorpusJob, fix speaker_name

albertz:
``Recording.speaker_name`` is typed ``Optional[str]`` and ``Corpus.dump`` runs it through ``saxutils.escape`` (since #625), which rejects non-str.
LibriSpeech speaker ids are numeric strings in SPEAKERS.TXT; we keep them as ``int`` internally for stable sort order in ``self._speakers``, but stringify when handing them to the corpus model.

### DIFF
--- a/datasets/librispeech.py
+++ b/datasets/librispeech.py
@@ -138,7 +138,7 @@ class LibriSpeechCreateBlissCorpusJob(Job):
             name = "{0}-{1}-{2:04d}".format(transcript["speaker_id"], transcript["chapter"], transcript["segment"])
             recording = corpus.Recording()
             recording.name = name
-            recording.speaker_name = transcript["speaker_id"]
+            recording.speaker_name = str(transcript["speaker_id"])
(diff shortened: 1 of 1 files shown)


mmz33 (approved):

michelwi:
Does this change the file that is generated (in a meaningful way)? If so, we usually add a parameter to control this behavior for reproducabillity.

albertz:
> Does this change the file that is generated (in a meaningful way)? If so, we usually add a parameter to control this behavior for reproducabillity.

Without this change, the job currently cannot create a file at all. It just crashes.

albertz:
Check #625. I think that code was never tested. `speaker_name` must be a str, otherwise `saxutils.escape` crashes.

michelwi:
mh.. right. Before #625 the integer value would have become a string anyway, so I think your changes here will preserve the same output that was created back than.

michelwi (approved):

PR #660, opened 2026-06-28
 ExtractTextFromHuggingFaceDatasetJob: add revision arg

polycrit:
Add an optional revision argument to `ExtractTextFromHuggingFaceDatasetJob`, forwarded to `load_hf_dataset`/`load_dataset`, so callers can pin a specific dataset commit instead of resolving the moving default branch.

The job hash stays stable for callers that do not set revision (the `hash` classmethod drops it when `None`), mirroring the per-argument hash handling in `TransformAndMapHuggingFaceDatasetJob`.

### DIFF
--- a/datasets/huggingface.py
+++ b/datasets/huggingface.py
@@ -348,22 +348,32 @@ class ExtractTextFromHuggingFaceDatasetJob(Job):
         *,
         split: Optional[str] = "train",
         column_name: str = "text",
+        revision: Optional[str] = None,
     ):
         """
         :param path: for :func:`datasets.load_dataset`
         :param name: for :func:`datasets.load_dataset`
         :param split: for :func:`datasets.load_dataset`
         :param column_name: name of the text column to extract
+        :param revision: dataset version (git revision) of the dataset
         """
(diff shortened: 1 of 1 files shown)


datasets/huggingface.py
```diff
@@ -348,23 +348,33 @@ def __init__(
...
         self.split = split
         self.column_name = column_name
+        self.revision = revision
 
         self.rqmt = {"cpu": 4, "mem": 8, "time": 10}
 
         self.out_text = self.output_path("text.txt.gz")
 
+    @classmethod
+    def hash(cls, kwargs):
```
michelwi: Instead of overriding the hash method I would recommend to set
```
__sis_hash_exclude__ = {"revision": None}
```
cf https://github.com/rwth-i6/sisyphus/blob/ddcd0281834841873562fecfc8f60c1828504cf5/sisyphus/job.py#L174

michelwi (changes requested):

michelwi (approved):

albertz (approved):

PR #661, opened 2026-07-16
Make `prettify` optional for `ReorderPhonemeInventoryJob`

mmz33:
This PR makes `prettify` optional since it can be very slow for big lexicon files. I had a job which took 30 sec when setting `prettify` to False instead of hours...

### DIFF
--- a/lexicon/modification.py
+++ b/lexicon/modification.py
@@ -281,9 +281,12 @@ class ReorderPhonemeInventoryJob(Job):
     The job will raise an error if the phoneme set does not match between both lexica.
     """
 
-    def __init__(self, lex_to_modify: tk.Path, ref_lex: tk.Path):
+    __sis_hash_exclude__ = {"prettify": True}
+
+    def __init__(self, lex_to_modify: tk.Path, ref_lex: tk.Path, prettify: bool = True):
         self.lex_to_modify = lex_to_modify
         self.ref_lex = ref_lex
+        self.prettify = prettify
 
         self.out_lexicon = self.output_path("lex.xml.gz")
 
@@ -310,4 +313,4 @@ class ReorderPhonemeInventoryJob(Job):
 
         lex_to_modify.phonemes = ref_lex.phonemes
 
-        write_xml(self.out_lexicon.get_path(), lex_to_modify.to_xml())
+        write_xml(self.out_lexicon.get_path(), lex_to_modify.to_xml(), prettify=self.prettify)


.github/workflows/tests.yml
```diff
@@ -28,6 +28,7 @@ jobs:
       - uses: actions/checkout@v6
         with:
           repository: "rwth-i6/returnn"
+          ref: "master"
```
michelwi: why do we need to add `ref` here and not for sisyphus?

mmz33: The changes are needed to avoid the failing default-branch API lookup so yeah not very sure why the check passed for `tests.yml` for sisyphus. In `hashes.yml` we explicitly set the branch to `master` for sisyphus so i think we should also do the same for `tests.yml`. maybe it is using the newer checkout version (`checkout@v6`) which handled it differently, not sure.

michelwi: so maybe update all checkouts to the latest versions instead?

> yeah not very sure why the check

maybe ask your friend? xP
I told you I'll me weary of AI created code changes that "the author" does not understand

mmz33: I do not understand your comment. I understand the problem even if an AI solved it. The tests also passed so the comment you asked does not break or add anything and it is not very important for this PR. This discussion is even unrelated to this PR and should be done in a separate PR.

michelwi: > This discussion is even unrelated to this PR and should be done in a separate PR.

yes. We can move the pipeline fix in a separate PR and discuss there

> I understand the problem

Well I don't understand the problem. Especially not that some repos need an explicit ref and others do not. And now I am concerned that other repos might need an explicit ref as well soon. Or maybe the problem was something different and providing an explicit ref just made the code follow a different code path and "the problem" is in fact not solved

> even if an AI solved it. The tests also passed

"an AI solved it" == I used a tool I don't understand to arrive at an outcome I don't understand
As the tests are not comprehensive, a "pass" does not guarantee there is no problem. (e.g. the AI could change all tests to be no-ops and then they would immediately pass)
Granted, understanding also does not guarantee there is no problem, but I'd feel better if someone understood it.

michelwi: I have looked into the problem and it seems that the cause of the problem was a (temporary) server outage on githubs side which broke the automatic resolution of the default branch. 
Maybe the servers were available sometimes and sometimes not so that some requests were answered sucessfully and others werent (which explains why it works despite leaving out some branch refs).

To verify, I have rerun all the tests on your initial commit and this is now completely green, i.e. no further changes to the pipeline are necessary at the moment.

I therefore suggest:
* in this PR we remove all updates to the github actions and then it should be fine
* we can have a second PR adding refs t oal lthe branches and updating the checkout workflow version, as v2 is rather old and the log shows various deprecation warnings. The most recent version as of now is v7 https://github.com/actions/checkout

 (this keeps unrelated changes separated and easier to find)

mmz33: Ok thanks for checking. I did the change.

michelwi (approved):

albertz (approved):

PR #663, opened 2026-09-13
ExtractOovWordsFromTextJob

albertz:


corpus/stats.py
```diff
@@ -76,6 +81,55 @@ def change_casing(text_str):
...
 
 
+class ExtractOovWordsFromTextJob(Job):
+    """
+    Extracts the out of vocabulary words of a plain text corpus
+    (one utterance per line, words separated by whitespace) w.r.t. a lexicon.
+    Same semantics as :class:`ExtractOovWordsFromCorpusJob`, without a Bliss corpus.
+    """
+
+    def __init__(self, text_file: tk.Path, *, bliss_lexicon: tk.Path, casing: str = "none"):
```
Gerstenberger, line 91: I am wondering in general:
For `casing`, we have three different options which are different literals. In Python, I see people use either strings or enums usually, i.e. one could also achieve this using `enum.StrEnum` or so. Both have advantages/disadvantages, for example you would have to import the enum-type too and so on.

In the end, I guess this is a preference. Not sure how enum-type arguments are handled in sisyphus/if there are problems.

Maybe then it would be good to type this using `typing.Literal`? I guess this is not crucial and only relevant for some type checkers.

Gerstenberger (approved):

corpus/stats.py
michelwi: I am ok to keep this consistent to ExtractOovWordsFromCorpusJob

michelwi (approved):

